Issue #120, opened 2021-02-17
Add enums to markdowngen


cmungall:
Currently the markdown generator ignores enums

It should generate a .md file for every enum. The content of the markdown will reflect the enum itself.

This should also be linked when an enum is used in a range

For testing, I recommend using the schema here https://github.com/cmungall/ontology_associations

hrshdhgd:
@cmungall , I committed a first stab of code to generate markdown from YAML corresponding to enums. I used the ['association.yaml'](https://github.com/cmungall/ontology_associations/blob/main/src/schema/association.yaml) file as an input. This is what I receive as an output corresponding to ``` gp2term_relation_enum```.

```
# Enum: Gp2termRelationEnum

URI: [ontology_association:Gp2termRelationEnum](https://w3id.org/ontology_association/Gp2termRelationEnum)
```

I did notice there is more information in the enum object for the property 'Permissible Values' that looks like this:

```
permissible_values=
    {'protein_coding_gene': PermissibleValue(text='protein_coding_gene',
                            description=None,
                            meaning='SO:0001217',
                            alt_descriptions={},
                            deprecated=None,
                            todos=[],
                            notes=[],
                            comments=[],
                            examples=[],
                            in_subset=[],
                            from_schema=None,
                            imported_from=None,
                            see_also=[],
                            deprecated_element_has_exact_replacement=None,
                            deprecated_element_has_possible_replacement=None,
                            is_a=None,
                            mixins=[],
                            extensions=[],
                            annotations=[]),
    'ncRNA_gene': PermissibleValue(text='ncRNA_gene',
                            description=None,
                            meaning='SO:0001263',
                            alt_descriptions={},
                            deprecated=None,
                            todos=[], notes=[],
                            comments=[],
                            examples=[],
                            in_subset=[],
                            from_schema=None,
                            imported_from=None,
                            see_also=[],
                            deprecated_element_has_exact_replacement=None,
                            deprecated_element_has_possible_replacement=None,
                            is_a=None,
                            mixins=[],
                            extensions=[],
                            annotations=[]),
    'mRNA': PermissibleValue(text='mRNA',
                            description=None,
                            meaning='SO:0000234',
                            alt_descriptions={},
                            deprecated=None,
                            todos=[],
                            notes=[],
                            comments=[],
                            examples=[],
                            in_subset=[],
                            from_schema=None,
                            imported_from=None,
                            see_also=[],
                            deprecated_element_has_exact_replacement=None,
[...]

PR #174, opened 2021-04-22
Extend markdowngen to include enums as ranges, #120 

hrshdhgd:
Diplayes properties within enums that have no values as well.


cmungall:
I think this looks good!

cmungall (approved):

hsolbrig (approved):

PR #207, opened 2021-05-13
Linkml runtime

hsolbrig:
This needs testing before we official put it in


cmungall:
@hsolbrig what is the status of this?

hsolbrig:
I think there was one issue -- something about someone not being able to reference the linkml-model.

My take is that we should go ahead and merge this and the linkml-model changes and be done with it.  Need to see what is going on with the unit tests above before I do

PR #210, opened 2021-05-26
Correct description of domain of linkml

bpow:
"Subject" was listed twice-- replace the second occurrence with "Object"


cmungall (approved):
Thank you!

PR #212, opened 2021-05-30
Turned off title case in schema name

gaurav:
By default, LinkML displays schema names in Markdown in title case. This is not ideal in cases where the schema name is an acronym, e.g. `CCDH` is turned into `Ccdh`. This PR changes the display of schema names to be exactly as entered in the `schema.name` field.


cmungall (approved):
Thank you!! This was really bugging me, I have many acronym linkml projects!!

PR #213, opened 2021-05-30
Make predicate_cardinality rendering identical to range_cardinality rendering

gaurav:
@mbrush requested this on the CCDH model documentation, so I thought I'd send the idea upstream: I'm not sure if predicate_cardinality rendering should always be identical to range_cardinality rendering, but maybe this should be configurable? Also, in the future, 'REQ' and 'OPT' are probably going to be harder to internationalize than '1..1' and '0..1'.

### DIFF
--- a/linkml/generators/markdowngen.py
+++ b/linkml/generators/markdowngen.py
@@ -412,7 +412,7 @@ class MarkdownGenerator(Generator):
         if slot.multivalued:
             card_str = '1..*' if slot.required else '0..*'
         else:
-            card_str = 'REQ' if slot.required else 'OPT'
+            card_str = '1..1' if slot.required else '0..1'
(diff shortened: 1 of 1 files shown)


cmungall (commented):
I mildly prefer human readable terms like "required" or "optional" (which should ideally be hyperlinked and not computerese like `REQ`), but I don't have strong opinions. @hsolbrig and others please post your thoughts here

mbrush:
I prefer 0..1,  0..m, 1, as these are more expressive.  Indicated that a field can be multi-valued is often important to convey.  Configurable would be great of course - so implementations can chose how they want to convey this info.

bfurner:
i also prefer the numeric representation: 0..1, 0..*, 1..1, 1..*, etc.

cmungall (approved):

Issue #201, opened 2021-05-07
problem generating enum in json schema


smrgeoinfo:
I'm trying to have an enumerated list of terms as the value range for a property (slot) in my JSON schema. I must be missing something in my yaml schema description.

here's what's in in the yaml:

```
enums:
  materialtype:
    description: categories for kinds of material that constitute the sample 
    permissible_values:
      anyanthropogenicmaterial:
        text: anyanthropogenicmaterial
      anthropogenicmetal:
        text: anthropogenicmetal
      otheranthropogenicmaterial:
        text: otheranthropogenicmaterial
```
...
```
slots:		
  hasMaterialCategory:
    range: materialtype
    domain: PhysicalSampleRecord
    description: specify the kind of material that constitutes the sample
    multivalued: true		
```
....

The generated JSON schema:
		
```
            "hasMaterialCategory": {
               "description": "specify the kind of material that constitutes the sample",
               "items": {
                  "type": "string"
               },
               "type": "array"
            },
```
			
			
I expected 
... 
```
"items": {
                  "type": "string"
				  "enum": [
					"anyanthropogenicmaterial",
					"anthropogenicmetal",
					"otheranthropogenicmaterial"
				  ]
               },
```

cmungall:
Your expectations are correct, the json-schema generator should emit this, as per https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values

Please note also there is an issue with the python generator right now, if you use the generated python model to create models and then emit json, it will emit an object and not a string. See #157 - we should hopefully fix this soon

gaurav:
As I just mentioned in the LinkML meeting, I'm going to be working on coming up with a quick-and-dirty solution to this problem this weekend in [the cancerDHC fork of LinkML](https://github.com/cancerDHC/linkml/). If I get a working solution, I'll follow up here to look into the possibility of converting my fix into a PR for LinkML.

cmungall:
Just noticed this is not resolved. Can we get your fix in soon @gaurav . @wdduncan I'm surprised this hasn't bitten us yet?

gaurav:
I _think_ this is fixed by PR #239 -- the example tests currently generate:

https://github.com/linkml/linkml/blob/555d044d84128d37b009ebe6e09305a8d04baac7/tests/test_biolink_model/output/biolink-model.schema.json#L8676-L8685

and:

[...]

PR #239, opened 2021-06-18
Add enums to JSON schema

gaurav:
This is a first stab at adding enumerations to JSON Schema generation (#201). It uses the `text` field of the [`permissible_values` slot](https://linkml.github.io/linkml-model/docs/permissible_values.html) to build a list of [enumerated values](https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values) in JSON Schema. Enumerations are generated separately in the file and then referenced via an inlined slot.

I'm not very familiar with all the options of [`pv_formula`](https://linkml.github.io/linkml-model/docs/pv_formula.html), so this code only works with the `permissible_values` slot at the moment. Let me know if there are other cases we need to support here, or if this is good enough for now.

One fix I had to make to fix a bug in linkml-runtime is in b616df8: to fix the Python enum generation bug (https://github.com/linkml/linkml/issues/254), I added a workaround in the Python code being generated here. Once this has been fixed in linkml-runtime (presumably in PR https://github.com/linkml/linkml-runtime/pull/17), this workaround -- and the related test files -- can be removed. If that PR isn't very controversial, it might be worth it to wait for that to be incorporated and published, in which case some code can be removed from this PR before merging it.

This PR modifies the kitchen sink tests and some expected Python and JSON Schema files to bring them in line with the code changes made in this PR. It also adds "chardet" as a prerequisite as the tests don't run without them -- I think it's a prerequisite for pyshexc, but not declared as such in its requirements, so we need to install this ourselves.


cmungall:
do you want help writing the tests @gaurav?

if you want to wait for 
#247
there is a general purpose jsonschema test that operates over the kitchen_sink schema, which is intended to capture all possible linkml constructs

gaurav:
I added some basic tests in 58278d8 -- hopefully that's better than nothing! Let me know what you think.

(I tried to merge PR #247 into this, but I ended up running into a bunch of test errors, so I thought it would be easier to just make an issue-specific test. Feel free to suggest improvements/request that I merge it into #247 anyway.)

gaurav:
Ack, no, the kitchen test is still failing. I've filed that in a separate issue, which might be better tackled by whoever wrote the enumeration Python data class generator: https://github.com/linkml/linkml/issues/254.

cmungall:
Sorry about that! Harold and I merged the kitchen sink tests even though
there was still some issue. We intended to fix after. Let's talk about it
tomorrow

hsolbrig:
@gaurav  -- I'll take a look at this now.   Don't worry about creating a clean test case...

cmungall:
just checking where we are with this...

gaurav:
@cmungall I've got the tests passing, so I think this is now ready for review! I've put in a workaround for https://github.com/linkml/linkml/issues/254 in b616df8, but if we can fix that directly in linkml-runtime (presumably in PR https://github.com/linkml/linkml-runtime/pull/17), we can remove it from this PR. Or we can merge this PR now and remove the workaround later. Either way works for me!

cmungall (approved):

gaurav:
@cmungall I removed the workaround for linkml/linkml-runtime#17 since that has now been merged into linkml-release in 1d79ec9. I think this PR is ready to be merged now!

gaurav:
@cmungall Let me know if you need anything else from me before merging this in!

cmungall (approved):

PR #240, opened 2021-06-19
Replace `simple_name` with `name` in the Slot header in the Markdown generator

gaurav:
In the Markdown generator, the Slot page currently uses the "simple name" (i.e. the CURIE without the prefix) of the slot in the page header. When a slot is defined as part of a class (such as e.g. [Specimen.derived_product](https://cancerdhc.github.io/ccdhmodel/v1.0.1/specimen__derived_product/)), this is displayed as `Slot: specimen__derived_product`, which is not very readable. In the CCDH project, we have replaced this with the `name` field of the slot, giving us the much more readable "Slot: derived_product".

Alternatively, we could include the defining class in the Slot page header, possibly with a link to the entity (i.e. "[Specimen](https://cancerdhc.github.io/ccdhmodel/v1.0.1/Specimen/).derived_product"). I'm not sure which one makes more sense given how LinkML works with slot definitions.

### DIFF
--- a/linkml/generators/markdowngen.py
+++ b/linkml/generators/markdowngen.py
@@ -250,7 +250,7 @@ class MarkdownGenerator(Generator):
             obj_type = 'Enum'
         else:
             obj_type = 'Class'
-        self.header(1, f"{obj_type}: {simple_name}" + (f" _(deprecated)_" if obj.deprecated else ""))
+        self.header(1, f"{obj_type}: {name}" + (f" _(deprecated)_" if obj.deprecated else ""))
         self.para(be(obj.description))
         print(f'URI: [{curie}]({uri})')
         print()


cmungall (approved):

cmungall:
Hi @gaurav thanks for the change, I approved

But it would be good to get @hsolbrig's thoughts on this.

There are some deeper questions here about slot identity that we should discuss. It looks like in your case, you are using attributes, which induce a slot name `c__s` (double underscore). I have sometimes avoided attributes for these aesthetic reasons.

But note also that `slot_usage` will induce a `c_s`. For example, I might have a fairly generic slot `name` that has a range of string, a common description, etc. But I may want one regex `pattern` to apply for humans, and another for pets. The metamodel creates these different slots under the hood as they have different semantics (minimally).

The question is, how should this be reflected in the end-user oriented markdown?

 1. collapse into a single `s` page, and put any info from `c_s` on the page for `c`, under where `s` is listed
 2. have separate pages for each induced slot, but improve their aesthetics in the spirit of this PR
     - for attribute-induced I think it is sufficient just to use name as you do
     - for usage-induced, we may end up with different pages all titled `s`, so maybe we want to say something like "s in the context of c"?

There is some discussion of this here:

https://github.com/linkml/linkml/discussions/144

gaurav:
> There are some deeper questions here about slot identity that we should discuss. It looks like in your case, you are using attributes, which induce a slot name `c__s` (double underscore). I have sometimes avoided attributes for these aesthetic reasons.

Actually, I think CCDH needs this name mangling approach: in our model, there is no reason to assume that `Specimen.id` should have the same definition or constraints as `ResearchSubject.id`, so the current behavior -- to treat them as entirely different properties -- is correct to me. So maybe it would just be better to include a link to the containing entity in the slot page title?

If we did need a property to be defined identically across multiple classes, we would probably do one of two things:
1. Create a mixin (e.g. `Identifiable` or `HasID`) with an `id` attribute that could be inherited to every class that needed it. We actually used to have an `id` attribute in our top-level `Entity` class at one point during development, but that was removed before the v1.0 release.
2. Check to see if an attribute is only defined in a single class. If it is, we can promote it to a top-level slot. If not, we would leave it as a class attribute, but produce a warning suggesting data modelers work on pulling it out into a mixin or eliminate the duplicate property in some other way.

cmungall:
I think in the CCDH case it's still fine though. You would go to the page for "id" to see the very general properties shared by all IDs. This page could also show you how id is used in all different slot_usage concepts. And the individual pages for each class would show id in the context of that class.

gaurav:
Yup, we could do that! I still worry that this *might* get confusing if we have two identically named fields in two different classes, but since the class-specific definitions will be available on the class pages, it should all be clear. I'll look into that if and when it comes up.

PR #259, opened 2021-07-05
Issue 202

hsolbrig:
A medium term fix for issue_202.   Add a table of python to JSON types in the json schema generator which should carry us through for a while.

We do need to consider a general patter for mapping if we're going to start dealing with lots of languages


hsolbrig:
I give up.  I'm going to merge with errors and we'll figure them out later

Issue #260, opened 2021-07-07
jsonschemagen should include format restrictions for dates 


cmungall:
The kitchen sink test emits json like:

```
     "EmploymentEvent": {
         "additionalProperties": false,
         "description": "",
         "properties": {
            "employed_at": {
               "type": "string"
            },
            "ended_at_time": {
               "type": "string"
            },
            "is_current": {
               "type": "boolean"
            },
            "started_at_time": {
               "type": "string"
            }
         },
```

"format": "date" should be added if. the linkml type is date

see https://opis.io/json-schema/2.x/formats.html

cmungall:
as per https://json-schema.org/understanding-json-schema/reference/string.html#format

> JSON Schema implementations are not required to implement this part of the specification, and many of them do not."
  
it appears the python json schema implementation does not implement this, so we skip the test for this for now

cmungall:
Fixed by #261

PR #261, opened 2021-07-07
Additional tests and fixing #260

cmungall:


cmungall:
@hsolbrig - initially my PR conflicted with yours. I like your lookup table approach. I extended this as necessary to deal with #260 -- I opted to map between the linkml base type and a tuple of (type, format), as is required in json-schema for things like date fields.

cmungall:
merging after discussion with @hsolbrig

PR #264, opened 2021-07-07
allowing ontologyIRI to be different from schema.id, fixes #246 Added owlgen tests

### LINKED ISSUE #246: owlgen: what should be OntologyIRI be for OWL derived from a LinkML model
@balhoff was asking what the URI should be for the OWL of biolink

currently we reuse the URI of the schema, e.g

```turtle
<https://w3id.org/biolink/biolink-model> a owl:Ontology ;
    rdfs:label "Biolink-Model" ;
    dcterms:license "https://creativecommons.org/publicdomain/zero/1.0/" ;
```

Note
https://w3id.org/biolink/biolink-model

doesn't resolve (but this is more of a biolink issue than linkml - @sierra-moxon )

I think we need a separate URL for the OWL artefact from the main schema (and possibly for shex, JSON-Schema, etc)

 1. 🚀  use file suffix - e.g https://w3id.org/biolink/biolink-model.owl.ttl 
 2. 🎉  different path - e.g. https://w3id.org/biolink/biolink-model/ontology

I think the first is most straightforward and is consistent with OBO. Vote using the appropriate symbol.

cmungall:


cmungall:
merging after discussion with @hsolbrig

PR #275, opened 2021-07-16
Issue 106: add --not-closed/--closed to cli to control global additional properties

### LINKED ISSUE #106: jsonschema generation should generate additionalProperties: false for enclosing properties
we correctly generate

for each class

but we should also do this on the enclosing object

wdduncan:
Fixes #106 

Add functionality to specify at the global level whether the schema allows additional properties.  
The default is `not-closed`; i.e., the schema does allow additional properties. Used `--closed` in the cli to set `additionalProperties: false`.

### DIFF
--- a/linkml/generators/jsonschemagen.py
+++ b/linkml/generators/jsonschemagen.py
@@ -1,13 +1,29 @@
 import os
-from typing import Union, TextIO, Optional
+from typing import Union, TextIO, Optional, Dict, Tuple
 
 import click
 from jsonasobj2 import JsonObj, as_json
-
-from linkml.utils.generator import Generator, shared_arguments
 from linkml_runtime.linkml_model.meta import SchemaDefinition, ClassDefinition, SlotDefinition
 from linkml_runtime.utils.formatutils import camelcase, be, underscore
 
+from linkml.utils.generator import Generator, shared_arguments
+
+# Map from underlying python data type to json equivalent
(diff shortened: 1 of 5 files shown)


linkml/generators/jsonschemagen.py
```diff
@@ -73,14 +58,27 @@ def end_class(self, cls: ClassDefinition) -> None:
...
-        elif slot.range in self.schema.classes and slot.inlined:
+        if slot.range in self.schema.classes and slot.inlined:
             rng = f"#/definitions/{camelcase(slot.range)}"
+        elif slot.range in self.schema.types:
+            rng = self.schema.types[slot.range].base
         else:
+            # note we assume string for non-lined complex objects
             rng = "string"
 
+        # translate to json-schema builtins
```
cmungall, line 69: wait, this seems to be introducing old code

cmungall:
Running the tests as is fails to make roottest3.jsonld due to the clashing nae

if I rename the 2nd test test_tree_root_closed then I get the expected difference:

```
 $ diff tests/test_scripts/output/genjsonschema/rootttest3.jsonld  tests/test_scripts/output/genjsonschema/rootttest4.jsonld 
4c4
<    "additionalProperties": false,
---
>    "additionalProperties": true,
```

it would be good to explicitly test this but let's not worry about that for now

PR #283, opened 2021-07-20
abstract common add_mapping code into parent generator

### LINKED ISSUE #163: RDF Generator problems
see: https://github.com/biolink/biolink-model/issues/667

### LINKED ISSUE #278: Abstract add_mappings code into common generator superclass
`def add_mappings`  is duplicated across different generators with drift between the code, causing a variety of subtle issues relating to loss of prefix information

See my comment here:

https://github.com/linkml/linkml/issues/192#issuecomment-883537162

The fix is to simply move add_mappings from jsonldcontext gen into the parent, and to remove it from pythongen and prefixmapgen

I hypothesize this will fix #163

cmungall:
Fixes #278

Also:

Fixes #163
Fixes biolink/biolink-model#667


linkml/generators/owlgen.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, schema: Union[str, TextIO, SchemaDefinition], ontology_uri_su
             SchemaLoader(METAMODEL_YAML_URI, base_dir=META_BASE_URI, importmap=kwargs.get('importmap', None),
                          mergeimports=self.merge_imports)
         self.metamodel.resolve()
+        self.emit_prefixes: Set[str] = set()
```
cmungall, line 42: @hsolbrig doesn't seem satisfactory having to do this in every __init__ method - move up to parent and set as a default?

hsolbrig:
It appears that lines 572 through 575 in generator.py are causing the above error.  Why were these 4 lines introduced:
```
        if isinstance(defn, ClassDefinition):
             mappings.append(defn.class_uri)
        if isinstance(defn, SlotDefinition):
            mappings.append(defn.slot_uri)
```

Issue #285, opened 2021-07-23
Schema/model version not included in artifacts generated via generators


deepakunni3:
Schema/model version, via `version` slot, is not serialized into artifacts like Python dataclasses, OWL, JSONSchema, etc.
The only artifact that seems to have the version is the RDF TTL.

I think there is value in including both metamodel version as well as model version.

deepakunni3:
If this has not yet been worked on then I can take a stab at it :)

sierra-moxon:
closing as completed; thanks @deepakunni3 !

PR #288, opened 2021-07-24
Ensure version and metamodel version are part of all relevant serializations

deepakunni3:
This PR addresses https://github.com/linkml/linkml/issues/285.

### DIFF
--- a/linkml/generators/sqlddlgen.py
+++ b/linkml/generators/sqlddlgen.py
@@ -193,6 +193,7 @@ class SQLDDLGenerator(Generator):
         self.rename_foreign_keys = rename_foreign_keys
         self.direct_mapping = direct_mapping
         self.sqlschema = SQLSchema()
+        self.generate_header()
 
     def _is_hidden(self, cls: ClassDefinition) -> bool:
         if cls.mixin or cls.abstract:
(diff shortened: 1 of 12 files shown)


deepakunni3:
@cmungall @hsolbrig Would also like to hear your thoughts if this is the right approach.

deepakunni3:
So the tests are failing because of "output mismatch". 

If the approach above makes sense then I can fix the tests by updating the outputs that its being compared against.

linkml/generators/sqlddlgen.py
```diff
@@ -204,6 +205,11 @@ def _class_name_to_table(self, cn: ClassDefinitionName) -> str:
         """
         return underscore(cn)
 
+    def generate_header(self):
```
cmungall, line 208: are you sure '#' as a comment delimiter is respected by all SQL engines?

Currently in this module we only write DDL via sqlalchemy -- this should take care of emitting comments with the correct syntax

cmungall (changes requested):

linkml/generators/sqlddlgen.py
deepakunni3: Thank you for pointing that out @cmungall ! 

I made the change such that we are using the more traditional SQL comment syntax (`/* COMMENT */`)

deepakunni3:
This PR also brings to attention on `generator_version`? I think there is value to keeping track of the version of the generator that was used to autogenerate various artifacts.

But is the generator version updated on changes to the individual generators? Is that part of the LinkML development SOP?

cmungall (approved):

PR #293, opened 2021-08-06
Added PROV example (prov association subset)

cmungall:


wdduncan:
A perhaps overly critical observation on my part is that:
```
Association:
    class_uri: prov:Association
    slots:
    - agent
    - hadRole
  ```
doesn’t make it clear that it is the agent that has the role, not the association.  
I think this structure would be more clear:
```
Association:
    class_uri: prov:Association
    slots:
    - agent: { hadRole: ...}
 ```
The value `agent` would in this case be a `person`.  
```
Association:
    class_uri: prov:Association
    slots:
    - agent
    slot_usage:
       agent:
          range: Person
 ```
The `hadRole` would be a slot added to person.
```
Person:
    class_uri: prov:Person
    is_a: Agent
   slots: hadRole
```
Also, we already have `person value` in the schema. Shouldn't we reuse that?

PR #294, opened 2021-08-09
Fix issue 273: Generation of SQLAlchemy classes fails when handling enums

deepakunni3:
This PR fixes the bug discovered when trying to generate SQLAlchemy classes and DDL for postgresql dialect.

This was because of two things:
- the executor used to serialize SQL statements was causing issues
- the name of enum was not being passed from linkml model to the `sqlalchemy.sql.sqltypes.Enum` object (which is required for Postgres - https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.Enum.params.name)


cmungall:
awesome! thanks so much!

cmungall (approved):

PR #310, opened 2021-08-27
draft of javagen fixes 175

cmungall:


kevinschaper:
@cmungall I think this is ready for one last look. 

I don't love that the java specific command line flags are showing up for all of the generators (the documentation really highlights that), but maybe that's something that we can fix in a later PR?

cmungall:
> I don't love that the java specific command line flags are showing up for all of the generators (the documentation really highlights that), but maybe that's something that we can fix in a later PR?

Hmm, I don't seem to get this:


```bash
$ pipenv run gen-python --help
Courtesy Notice: Pipenv found itself running within a virtual environment, so it will automatically use that environment, instead of creating its own for any project. You can set PIPENV_IGNORE_VIRTUALENVS=1 to force pipenv to ignore that environment and create its own instead. You can set PIPENV_VERBOSITY=-1 to suppress this warning.
Usage: gen-python [OPTIONS] YAMLFILE

  Generate python classes to represent a LinkML model

Options:
  --head / --no-head              Emit metadata heading
  --genmeta / --no-genmeta        Generating metamodel. Only use this for
                                  generating meta.py

  --classvars / --no-classvars    Generate CLASSVAR info
  --slots / --no-slots            Generate Slot information
  -f, --format [py]               Output format (default=py)
  --metadata / --no-metadata      Include metadata in output
                                  (default=--metadata)

  --useuris / --metauris          Include metadata in output
                                  (default=--useuris)

  -im, --importmap FILENAME       Import mapping file
  --log_level [CRITICAL|ERROR|WARNING|INFO|DEBUG]
                                  Logging level (default=WARNING)
  --mergeimports / --no-mergeimports
                                  Merge imports into source file
                                  (default=mergeimports)

  --help                          Show this message and exit.
```

cmungall:
hmm, not sure how we ended up with diffs of 65 files... maybe something to do with the merge to master.. we need developer docs on rebase etc!

linkml/utils/generator.py
```diff
@@ -625,6 +625,15 @@ def decorator(f: Command) -> Command:
                    default=g.valid_formats[0]))
         f.params.append(
             Option(("--metadata/--no-metadata", ), default=True, help="Include metadata in output (default=--metadata)"))
+        f.params.append(
```
cmungall, line 628: oh I see what you mean.... but these don't need to go here. Just annotate the click options here

```python
@shared_arguments(JavaGenerator)
@click.command()
def cli(yamlfile, head=True, genmeta=False, classvars=True, slots=True, **args):
    print(JavaGenerator(yamlfile, emit_metadata=head, genmeta=genmeta, gen_classvars=classvars, gen_slots=slots,  **args).serialize(emit_metadata=head, **args))
```

kevinschaper: That makes much more sense!

kevinschaper:
Sorry for the messy commit, I tend to lean towards merges rather than rebase after a frustrating experience with re-rebasing a complicated conflict, but I can definitely stick to rebasing here in the future. 

The high volume of conflicts & commits that I'm getting seem to be from the test artifacts. I thought since they were being regenerated and are committed, that I should commit them, but it seems like probably that's not the best idea.

cmungall:
We definitely need to document the test framework!

cmungall:
Replaced by #343

PR #329, opened 2021-09-07
sphinx docs

cmungall:
- trying to follow diataxis.fr. Using django as a kind of exemplar
 - following the template @matentzn and @hrshdhgd used for sssom-py and ontorunner

The source files go in `sphinx`


matentzn:
Nice one!

PR #332, opened 2021-09-08
add name to setup.py to enable dependents graph in github fixes #331

### LINKED ISSUE #331: enable "dependents" graph for linkml and associated repos
Currently, the "dependents" graph for linkml is empty, despite many repositories that depend on it. 

Screen Shot 2021-09-08 at 1 39 24 PM

It might be nice to add "name=linkml" to the setup.py in linkml (similar change to other repositories in the linkml space) so that the dependents graph can be generated.

Low - it's an enhancement but not crucial for work
Short-term - 2-4 weeks

[](https://github.community/t/bug-project-not-showing-network-dependents/3161/38) gave me the answer using setup.py, I'm not sure yet how to do it with pipfile.lock, though it does seem like a supported formated.

sierra-moxon:


sujaypatil96:
Just came across this PR, while I was committing to another open PR. I'm curious as to why the same contents that exist in [setup.cfg](https://github.com/linkml/linkml/blob/main/setup.cfg) need to be present in [setup.py](https://github.com/linkml/linkml/blob/main/setup.py) as well?

sierra-moxon:
Hi @sujaypatil96! :) It looks like setup.cfg is not one of the github [supported package ecosystems](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph#dependents-included).  So for things like linkml that are actually dependencies for many other projects, we don't actually get a 'dependents' graph in github which is kind of a bummer.  I'd also like to have just one or the other.  A next step would be to figure out what benefits the setup.cfg with the setup.py (with pbr) give us that outweigh the benefits of a plain setup.py file.

cmungall (approved):
guess we should maybe make setup.py the standard? WWCD?

sujaypatil96:
Thank you for clarifying that @sierra-moxon. :) I see the problem now.

Also, I think the main reason why people prefer the `setup.cfg` over the `setup.py` is for performance reasons? It's more efficient to parse a `.cfg` file rather than to run the logic in a `.py` file to extract package information. But if it's posing problems like this one, where dependency graphs aren't getting generated, probably best to stick with `setup.py`, I'm not sure how significant the improvement in the package build process is.

cmungall (approved):
go ahead and merge, but I think we should try and reduce duplication, and use one or the other?

cmungall (approved):

Issue #237, opened 2021-06-18
Need contributing.md


hsolbrig:
Need to document how testing is done

cmungall:
* [ ] document how to extend linkml-model and how to bring this into linkml-runtime (there is a video of @hsolbrig and I doing this)
 * [ ] document test framework
      - when do we check things into output/ folders
      - naming conventions for test_issues (I use linkml_test_NNN to distinguish from numbering system from biolinkml)
      - document the strategy of comparing outputs and how to read the false positives

cmungall:
@sujaypatil96 added: https://github.com/linkml/linkml/blob/main/CONTRIBUTING.md

Issue #73, opened 2020-09-04
Generator versions are not maintained


hsolbrig:
Generator version numbers are not being updated.  As an example, the json schema generator version is still at "0.0.2".  We need to either come up with a semi-automated way to maintain these or just remove this characteristic

cmungall:
I think we should remove

joeflack4:
I don't mind removing either. It is a nice-to-have, so I think it should be kept in some backlog, but there's probably a lot of other stuff that should be prioritized before this.

PR #366, opened 2021-09-23
markdowngen: Various improvements

joeflack4:
markdowngen: Various improvements
  - Deprecated elements: Now rendered with 'strikethrough' formatting


linkml/generators/markdowngen.py
```diff
@@ -296,7 +296,15 @@ def element_header(self, obj: Element, name: str, curie: str, uri: str) -> None:
             obj_type = 'Subset'
         else:
             obj_type = 'Class'
-        self.header(1, f"{obj_type}: {name}" + (f" _(deprecated)_" if obj.deprecated else ""))
+
+        # TODO: for some reason obj.deprecated dosn't appear to be True even
```
joeflack4, line 300: For some reason, even though I added `deprecated` as a `slot` in [`kitchen_sink.yaml`](https://github.com/linkml/linkml/blob/markdown-gen-updates/tests/test_generators/input/kitchen_sink.yaml#L65), `obj.deprecated` was returning `None` instead of `True`, which I did not expect.

I'm assuming this is either because:
a. Adding 'deprecated' as a slot is not how we mark a class as deprecated, or 
b. There is some error which is causing the 'deprecated' slot this to not be detected.

If (b), I assume the issue is in [`linkml_runtime/linkml_model/meta`](https://github.com/linkml/linkml-runtime/blob/main/linkml_runtime/linkml_model/meta.py), but I'm not sure exactly where or how to fix it yet. It looks like the fix would need to be in several different classes (which may mean that the way inheritance and mixins are setup in these classes are not optimally). 

So for now, as a hack, I've detected 'deprecated' another way. But I'd like to hear back concerning this. If it is indeed a problem in `meta.py`, I can try to tackle, or make a new issue for it.

linkml/generators/markdowngen.py
```diff
@@ -296,7 +296,15 @@ def element_header(self, obj: Element, name: str, curie: str, uri: str) -> None:
...
+        # ...though 'deprecated' in obj.slots. When that's fixed, can update
+        # ...this clause to be simpler. - joeflack4 2021/09/23
+        if (hasattr(obj, 'slots') and 'deprecated' in obj.slots) or obj.deprecated:
+            self.header(1, f"{obj_type}: ~~{name}~~ _(deprecated)_")
```
joeflack4, line 304: Since we have strikethrough now, should we get rid of `_(deprecated)_`, or keep it? I'm kind of agnostic; it might help for extra clarity.
Screen Shot 2021-09-23 at 8 03 14 PM

tests/test_generators/input/kitchen_sink.yaml
```diff
@@ -60,6 +60,10 @@ classes:
       aliases:
         multivalued: true
 
+  God:
```
joeflack4, line 63: I was being rather cheeky =P

tests/test_generators/input/kitchen_sink.yaml
```diff
@@ -240,6 +244,17 @@ slots:
   city:
   has birth event:
     range: BirthEvent
+  deprecated:
```
joeflack4, line 247: I'm not sure if I filled all these fields out correctly, or if they're even useful. Maybe it doesn't really matter in this case.

tests/test_generators/input/kitchen_sink.yaml
```diff
@@ -240,6 +244,17 @@ slots:
...
+    name: deprecated
+    definition_uri: https://w3id.org/linkml/deprecated
+    description: Description of why and when this element will no longer be used
+    from_schema: https://w3id.org/linkml/meta
+#    domain: God
```
joeflack4, line 252: I left this out arbitrarily, but I can uncomment.

linkml/generators/markdowngen.py
```diff
@@ -16,7 +16,7 @@
 
 class MarkdownGenerator(Generator):
     generatorname = os.path.basename(__file__)
-    generatorversion = "0.1.1"
+    generatorversion = "0.2.1"
```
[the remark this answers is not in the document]
cmungall: hmm, I am not sure we have consistently applied versioning to the generators. We can discuss tomorrow!

linkml/generators/markdowngen.py
cmungall: So the range of deprecated is a string, containing the deprecation reason. This is admittedly very odd! We can discussion tomorrow, but I suspect this may be the heart of your issue

linkml/generators/markdowngen.py
cmungall: I say keep it

tests/test_generators/input/kitchen_sink.yaml
cmungall: :-)

linkml/generators/markdowngen.py
joeflack4: Hmm, I see. Well, I do have the `range` defined, though it does just say "string":
```yml
  deprecated:
    name: deprecated
    definition_uri: https://w3id.org/linkml/deprecated
    description: Description of why and when this element will no longer be used
    from_schema: https://w3id.org/linkml/meta
#    domain: God
    range: string
    slot_uri: linkml:deprecated
    owner: deprecated
    domain_of:
    - common_metadata
```

will discuss later

tests/test_generators/input/kitchen_sink.yaml
```diff
@@ -60,6 +60,10 @@ classes:
       aliases:
         multivalued: true
 
+  God:
+    slots:
+      - deprecated
```
cmungall, line 65: I'm not sure what you are trying to do here. I know it's a fun test example but sometimes this can lead to confusion

You have a class God, which means that you are saying there are many God instances - say: Thor, Ganesh, Dave, ...

The class God has slot "deprecated" (which is confusingly the same name as a slot from the metamodel), so you can declare individual instances as deprecated...

Let's keep this simple. Let's only introduce one change to kitchensink, under classes:

```yaml
   Patient:
     is_a: Person
     deprecated: deprecated because we want to treat this as a role rather than a subclass
```

cmungall (changes requested):

tests/test_generators/input/kitchen_sink.yaml
joeflack4: Hey Chris. Ah, I see. I think this is just a misunderstanding of LinkML. I thought adding `deprecated` as a slot was the way to do it, but I guess it's defined in the metamodel. If deprecated is not a slot, I guess it is a "keyword"? As for God, it's just a toy example; Anything else that makes sense is good.

I'll make this change.

I think this is why I was having an issue where the generator wasn't detecting it as deprecated; because I'd put it as a slot, which is not the correct way to do this.

cmungall: formally deprecated is a slot:

[linkml:deprecated](https://w3id.org/linkml/deprecated)

it's just that it's a slot in the metamodel not in your domain model. The meta stuff gets a bit confusing and the docs may be inconsistent in how these kinds of things are described!

linkml/generators/markdowngen.py
joeflack4: I think this was caused, as we discovered, because I declared deprecated under 'slots', rather than as a root attribute of the class.

cmungall:
how is this coming along @joeflack4 ?

joeflack4:
@cmungall Thanks for pinging me. I made the updates and reversions you suggested. This should be ready to merge now.

I changed the title of this to just be about the 'deprecated' change, since that's all that's here. With Dazhi leaving, a lot of stuff has fallen onto my plate, so I think that my progress with the MarkdownGen issue will be slow.

One other question. I had to force add the new output file `Patient.md` for the new Patient class we added to `kitchen_sink.yaml`. I see that `output/` is in the `.gitignore`, which is why I had to force add it. Is there a reason that we are committing our test output files? Or is this a mistake and do they need to be removed?

cmungall (approved):

cmungall:
> Also, I left the version as 0.2.0. Feel free to change that to 0.1.0 or 0.1.1 if you like.

Fine to leave as is, but we need to make this clearer overall.

See #73

> assuming the merge-check build failures I'm getting are not related to this work. 

Hmm, can you investigate further? This is odd

> One other question. I had to force add the new output file Patient.md for the new Patient class we added to kitchen_sink.yaml. I see that output/ is in the .gitignore, which is why I had to force add it. Is there a reason that we are committing our test output files? Or is this a mistake and do they need to be removed?

You are asking all the good difficult questions.

I was hoping @jiaola might have been able to document this a bit more before he left

We have this ticket #237 

For this

Let's discuss this on Friday

cmungall:
here is the problem:

https://github.com/linkml/linkml/runs/3808748176?check_suite_focus=true#step:6:1960

kitchen_sink.yaml is used by various other tests, and this is messing up the SQL DDL test

sorry to be annoying but can you change the example and do something completely unrelated to existing classes, e.g.

```
  fake_class:
    deprecated: this is not a real class, we are using it to test deprecation
```

joeflack4:
Hey Chris, thanks for finding the source of the issue. I made that change. Looks like the checks are failing again >,<. About to give that a check.

As for gitignore and test outputs, I added some thoughts to #237.

Issue #312, opened 2021-08-31
YAMLGenerator injects source_file_date in wrong format


cmungall:
minimal.yaml:

```yaml
id: foo
prefixes:
  foo: http://x.org/
default_prefix: foo
default_range: string
```

produces:

```yaml
name: foo
id: foo
prefixes:
  foo:
    prefix_prefix: foo
    prefix_reference: http://x.org/
default_prefix: foo
default_range: string
metamodel_version: 1.7.0
source_file: minimal.yaml
source_file_date: Mon Aug 30 22:25:12 2021
source_file_size: 81
generation_date: 2021-08-30 22:26
```

When we try and parse this:

```bash
  File "/Users/cjm/.local/share/virtualenvs/linkml-6dchTQmP/lib/python3.9/site-packages/linkml_runtime/utils/metamodelcore.py", line 277, in __new__
    raise ValueError(f"{value} is not a valid datetime")
ValueError: Mon Aug 30 22:25:12 2021 is not a valid datetime
```

culprit:
https://github.com/linkml/linkml/blob/edf57962211d37004f0bc390f95083ad36dee5af/linkml/utils/rawloader.py#L89

sujaypatil96:
@cmungall: Could you clarify what commands were run to produce the outputs in the second and third code blocks?

cmungall:
`gen-yaml`

Issue #89, opened 2021-03-25
xsd:dateTime values in ttl are invalid


balhoff:
Triples like these using `xsd:dateTime` are using the wrong date format and cause parsing errors in some RDF tools:

```turtle
<https://w3id.org/biolink/vocab/Biolink-Model>  meta:generation_date "2020-11-12 22:05"^^xsd:dateTime .
```

dateTime should be `YYYY-MM-DDThh:mm:ss`

cmungall:
See also: #312

PR #400, opened 2021-10-07
fix date format of `source_file_date`

sujaypatil96:
This PR seeks to address issue #312, i.e., to fix the current `time.ctime()` date time format so that it displays date time in the `YYYY-MM-DDThh:mm:ss` format, as noted by @balhoff in issue #89.

Here is the output from `minimal.yaml`:

```
name: foo
id: foo
prefixes:
  foo:
    prefix_prefix: foo
    prefix_reference: http://x.org/
default_prefix: foo
default_range: string
metamodel_version: 1.7.0
source_file: minimal.yaml
source_file_date: 2021-10-05 15:09:57
source_file_size: 81
generation_date: 2021-10-07 15:43
```

### DIFF
--- a/linkml/utils/rawloader.py
+++ b/linkml/utils/rawloader.py
@@ -13,7 +13,7 @@ from linkml_runtime.linkml_model.meta import (
 )
 from linkml_runtime.loaders import yaml_loader
 from linkml_runtime.utils.yamlutils import YAMLMark, YAMLRoot
-from linkml_runtime.linkml_model.types import Datetime
+from linkml_runtime.utils.metamodelcore import XSDDateTime
 
 from linkml.utils.mergeutils import set_from_schema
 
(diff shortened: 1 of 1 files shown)


sujaypatil96:
Thanks for the review @balhoff. Apologies, but what is the `-06:00` in `2002-05-30T09:30:10-06:00` supposed to represent?

Is it okay if the output looks like this:

```
name: foo
id: foo
prefixes:
  foo:
    prefix_prefix: foo
    prefix_reference: http://x.org/
default_prefix: foo
default_range: string
metamodel_version: 1.7.0
source_file: minimal.yaml
source_file_date: '2021-10-05T15:09:57'
source_file_size: 81
generation_date: 2021-10-07 18:44
```

Note the file date being wrapped in single quotes.

PR #401, opened 2021-10-08
Add testing for alternative python versions

### LINKED ISSUE #404: Add support for python 3.10
As downstream dependencies are facing failing tests in 3.10, we should support 3.10 upstream here. 
See PR #401 

Or declare that we do not yet support 3.10 in our README, etc.

cthoyt:
Closes #404 

Since this project claims to work on py37+, I added explicit tests for py37, py39, and the newly released py310. These tests are applied in both the main.yml and pr-test.yml configurations.

I found in https://github.com/mapping-commons/sssom-py/pull/167/checks?check_run_id=3837024754 that it breaks in py310, so this is expected to fail and lead to other bug fixes for the time being

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8]
+        python-version: [ "3.7", "3.8", "3.9", "3.10" ]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


dalito:
#764 added tests for other Python versions (3.7...3.10).

sierra-moxon:
I think we already have this (probably as a result of this ticket, thanks all! :) ) in the github action "main.yaml". which runs on every PR.  I do like the convention of pr-test.yaml being the file that runs on each PR instead of 'main' -- but I think we could address that in another PR (I would go in and fix this PR's conflicts for everyone, but because it is a fork, I think its best to just open another PR).

Issue #393, opened 2021-10-04
jsonschemagen should use tree_root to determine top_class


cmungall:
if top-class not specified on CLI, use tree_root property to determine roo

sujaypatil96:
Do we need to write a handler wrapping [line 51](https://github.com/linkml/linkml/blob/main/linkml/generators/jsonschemagen.py#L51)? I'm guessing it's **not** as simple as:

```python
self.topCls = top_class if top_class else self.tree_root
```

If not, how would be construct the logic to determine what the value of `self.topCls` should be?

cmungall:
I think it is that simple! Let's try it!

sujaypatil96:
Is this what we want: https://github.com/linkml/linkml/blob/spatil/jsonschemagen-tree-root/linkml/generators/jsonschemagen.py#L174-L176?

I tried logging the output of `cls.tree_root` on a couple of example schemas, `personinfo.yaml` and `mixs.yaml` but the output was `None` on all occasions.

PR #406, opened 2021-10-13
use tree root to determine top class

sujaypatil96:
This PR seeks to address issue #393. It adds conditional logic to assign `cls.tree_root` to `self.topCls` if there isn't any value supplied for `--top-class` optional argument.

@cmungall: Should it be `cls.tree_root` or `cls.name` that should be assigned to `self.topCls`?


sujaypatil96:
Oh no worries Chris, the experience helped increase my familiarity with the codebase. I'll go ahead and close this PR now.

Issue #76, opened 2020-10-12
Pipfile.lock needs updating


hsolbrig:
We also need to update the Pipfile.lock file in the main workflow

PR #415, opened 2021-10-18
Updating pipenv specific files

sujaypatil96:
This PR seeks address issue #76. The following points are the highlight of the PR:

- Update instructions so that `pipenv` is used only by package developer or contributors to install the package locally for testing purposes
- Regular users should stick to using `pip` to install the package from PyPI as detailed in the [docs](https://linkml.io/linkml/intro/install.html#install-linkml)
- Create `requirements.txt` and `requirements-dev.txt`. These files have been generated using `pipenv` and not `pip`


sujaypatil96:
I can see in the [requirements.txt](https://github.com/linkml/linkml/blob/main/requirements.txt) and [requirements-dev.txt](https://github.com/linkml/linkml/blob/main/requirements-dev.txt) files that where it says in the comments that the requirements files have been automatically generated using `pipenv`.

The commands that freeze out the requirements using `pipenv` are as follows:

```
$ pipenv lock --requirements > requirements.txt
```

```
$ pipenv lock --requirements --dev-only > requirements.txt
```

The requirements files that are in this PR have been generated using the above commands. If we don't want to use them, I'm happy to remove them.

Issue #378, opened 2021-09-29
Add `--version` flag


cmungall:
As per https://clig.dev/ we should have a `--version` flag on generator and other commands. This should display the linkml version - perhaps also runtime.

not sure how to implement this:

```python
>>> import linkml
RDFLib Version: 5.0.0
>>> linkml.__version__
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'linkml' has no attribute '__version__'
```

cmungall:
We need to solve this problem generally:

- the metamodel docs should show the version of linkml-model
- generated schemas should show the version of the schema in their own docs
- all tools should be able to report on the version of that tool, and also the version of the metamodel

The general blocker here is that we use github releases to handle all aspects of releases, including assigning version tags. This means the versions in all out pyproject.yaml files is 0.0.0, making introspection hard.

dalito:
For automatic versioning with poetry, you may use https://github.com/mtkennerly/poetry-dynamic-versioning - It solved poetry`s automatic versioning problem for me. The version can be derived from a git-tag and temporarily injected to files you want. Can this not solve the 0.0.0 problem?

cmungall:
oooh this may be just what we need!! thank you @dalito!

PR #434, opened 2021-10-23
Add version flag

rosdyana:
Solve issue #378


sujaypatil96:
Thank you for clarifying that @rosdyana. I had a couple of other comments though.

setup.py
```diff
@@ -2,6 +2,10 @@
 import sys
 from setuptools import setup
 from warnings import warn
+import versioneer
+
+version = versioneer.get_version()
```
sujaypatil96, line 7: You're getting the version here, but not passing it to the `setup()`. From the `pandas` repo that you referenced, here is the [line](https://github.com/pandas-dev/pandas/blob/master/setup.py#L649) I'm referring to.

setup.py
```diff
@@ -2,6 +2,10 @@
...
+cmdclass = versioneer.get_cmdclass()
```
sujaypatil96, line 8: Similar to `version`, you're not passing `cmdclass` to `setup()` either. Here is the [line](https://github.com/pandas-dev/pandas/blob/master/setup.py#L651) from pandas.

sujaypatil96 (changes requested):
Would you be able to resolve these two comments @rosdyana? I think we should be good to merge it right after.

rosdyana:
@sujaypatil96 Sure, I just pushed a new commit to resolve that issue, thanks!

sujaypatil96 (approved):
Thank you @rosdyana. I think this PR is good to be merged.🚀

sujaypatil96:
Converting this PR to draft.

Reason: Passing explicit `version` and `cmdclass` options to `setup()` in setup.py without [pbr](https://github.com/openstack/pbr) will cause the application to break. So we need to explore alternative options of implementing this. See this SO [post](https://stackoverflow.com/questions/55921981/how-can-i-use-pbr-version-from-source).

sierra-moxon:
Hi @rosdyana - we just migrated LinkML to use the poetry virtual environment and package management framework.  I think I need to close this PR as it changes files that were removed during this migration.  With that change, have we addressed your versioning concerns?  If not, please do submit another PR and we can address it ASAP.

PR #445, opened 2021-10-27
Added show_default=True at every click event with default

### LINKED ISSUE #441: Add show_default=True on all click options that have a default
Right now if I run

```bash
gen-python --help
```

I see:

```
  --mergeimports / --no-mergeimports
                                  Merge imports into source file
                                  (default=mergeimports)
```

but this isn't universally the case, see #78. All options with defaults should have display_defaults=True

https://stackoverflow.com/questions/39359420/type-and-default-input-value-of-a-click-option-in-help-option

joseporiolayats:
Added show_default=True to all files with the click.option decorator and default value specified (first PR) Closes #441


linkml/validators/sparqlvalidator.py
```diff
@@ -126,7 +126,7 @@ def cli(input, output=None, input_format=None, output_format=None,
             raise Exception(f'Must pass one of --endpoint-url OR --input')
         input_format = _get_format(input, input_format)
         results = validator.validate_file(input, format=input_format)
-    output_format = _get_format(output, output_format, default='json')
+    output_format = _get_format(output, output_format, default='json', show_default=True)
```
sujaypatil96, line 129: This is not a click option, so we don't need `show_default=True` here.

sujaypatil96 (changes requested):

sujaypatil96:
@joseporiolayats: Thank you for the PR. Would you be able to help with the change requested above? I think PR should be ready to be merged after that.

cmungall:
@sujaypatil96 go ahead and make a PR on top of @joseporiolayats's and we can merge that

sujaypatil96:
I think it's ready to be merged.🚀

linkml/generators/golrgen.py
```diff
@@ -84,7 +84,7 @@ def visit_class_slot(self, cls: ClassDefinition, aliased_slot_name: str, slot: S
 
 @shared_arguments(GolrSchemaGenerator)
 @click.command()
-@click.option("--dir", "-d", default='golr-views', help="Output directory")
+@click.option("--dir", "-d", default='golr-views', show_default=True, help="Output directory")
```
cmungall, line 87: is this meaningful? There is no default defined

linkml/generators/csvgen.py
```diff
@@ -55,7 +55,7 @@ def visit_class(self, cls: ClassDefinition) -> bool:
 
 @shared_arguments(CsvGenerator)
 @click.command()
-@click.option("--root", "-r", default=None, multiple=True, help="Class(es) to transform")
+@click.option("--root", "-r", default=None, show_default=True, multiple=True, help="Class(es) to transform")
```
cmungall, line 58: is it useful to do this is default is none?

cmungall (approved):

linkml/generators/csvgen.py
sujaypatil96: Oh, that's a good point. Probably not. I think we should remove `show_default=True` when `default=None`. Also I don't think we need to explicitly specify `default=None`, I think the option value is `None` implicitly.

linkml/generators/golrgen.py
sujaypatil96: Oh there is, isn't there? The default value, in this case for the directory name has been set to `golr-views`.

linkml/generators/csvgen.py
cmungall: oh you are quite right, I misread!!

linkml/generators/golrgen.py
cmungall: oh you are quite right,I misread!

Issue #447, opened 2021-11-02
Add tsv option for gen-prefix-map


cmungall:
This is a trivial but useful extension to genprefixmap.py

Add a format option `tsv` or `csv`

This will be rftab-compliant (see https://github.com/ontodev/rdftab.rs#prefixes). See https://github.com/cmungall/semantic-sql/blob/main/prefixes/prefixes.csv for an example

Make sure that this option is documented in sphinx, see #446

PR #451, opened 2021-11-03
add `--format tsv` option to `gen-prefix-map` generator

sujaypatil96:
Per my understanding of issue #447, I have added support to `prefixmapgen.py` as in this following PR.

Execute the following command to see it in action:

```
gen-prefix-map examples/PersonSchema/personinfo.yaml --format tsv
```

See the sample output from the above command below:

```
CODE    http://example.org/code/
GEO     http://example.org/geoloc/
GSSO    http://purl.obolibrary.org/obo/GSSO_
P       http://example.org/P/
ROR     http://example.org/ror/
famrel  https://example.org/FamilialRelations#
linkml  https://w3id.org/linkml/
personinfo      https://w3id.org/linkml/examples/personinfo/
prov    http://www.w3.org/ns/prov#
rdf     http://www.w3.org/1999/02/22-rdf-syntax-ns#
rdfs    http://www.w3.org/2000/01/rdf-schema#
schema  http://schema.org/
skos    http://example.org/UNKNOWN/skos/
xsd     http://www.w3.org/2001/XMLSchema#
```

You can also redirect the tsv output into a `.tsv` file in the following way:

```
gen-prefix-map examples/PersonSchema/personinfo.yaml --format tsv > examples/personinfo.tsv
```

CC: @cmungall


cmungall:
can you also add a test?

sujaypatil96:
@cmungall: yes, I'll add a test as well.

Also I realized that `prefixmapgen.py` didn't have support to redirect output using the `--output` option, so I added in in the above [commit](https://github.com/linkml/linkml/pull/451/commits/39b3bad2a404d2ad6b0dc73d9652296ea23c8e5b).

sujaypatil96:
I'm not sure how we would write a test for this @cmungall?

Since the value of `--format` is being captured in `end_schema()` rather than in the `PrefixGenerator.__init__()` constructor, there's no way we can invoke that particular use case in the `test_prefixgen()` test case [here](https://github.com/linkml/linkml/blob/main/tests/test_prefixes/test_prefixes.py#L48)?

cmungall:
there must be a way to pass in format as an argument - either in the constructor, or in the serialize method?

sujaypatil96:
@cmungall: I've added a test for the new `--format tsv` option for `genprefixgen.py` in `test_prefixes.py`.

linkml/generators/prefixmapgen.py
```diff
@@ -64,11 +70,32 @@ def end_schema(self, base: Optional[str] = None, output: Optional[str] = None, *
             context[k] = v
         for k, v in self.slot_class_maps.items():
             context[k] = v
-        if output:
-            with open(output, 'w') as outf:
-                outf.write(as_json(context))
+
+        if self.output_path:
```
cmungall, line 74: The generator class doesn't need to worry about output path.

The job of any one generator is to produce the output, everything else is generic

sujaypatil96:
Looks like all the checks went through Chris, this PR is probably ready to be merged.

linkml/generators/prefixmapgen.py
```diff
@@ -64,11 +66,32 @@ def end_schema(self, base: Optional[str] = None, output: Optional[str] = None, *
             context[k] = v
         for k, v in self.slot_class_maps.items():
             context[k] = v
+
         if output:
-            with open(output, 'w') as outf:
-                outf.write(as_json(context))
+            output_ext = output.split(".")[-1]
```
cmungall, line 71: We should probably warn if this over-rides the users explicit choice

linkml/generators/prefixmapgen.py
```diff
@@ -64,11 +66,32 @@ def end_schema(self, base: Optional[str] = None, output: Optional[str] = None, *
...
+                    outf.write(as_json(context))
         else:
-            print(as_json(context))
+            if self.format == "tsv":
+                mapping: Dict = {}  # prefix to IRI mapping
+                for prefix in sorted(self.emit_prefixes):
+                    mapping[prefix] = self.namespaces[prefix]
+                
+                for key, value in mapping.items():
+                    print(key, value, sep='\t')
```
cmungall, line 92: is this right?

cmungall: ah I see, it's just a 2 column tsv so this is OK - but why not use this same simple form when writing to a file?

linkml/generators/prefixmapgen.py
sujaypatil96: Hm, so this isn't necessarily overriding anything. It just parses out the file extension from the filename that is supplied via the `--output` option.

linkml/generators/prefixmapgen.py
sujaypatil96: I wanted to use `csv.writer()` to make sure the data is being correctly formatted and written to a tsv file rather than stringing the `key` and `value` together with a `\t` separator.

PR #456, opened 2021-11-05
First pass at pydanticgen docs

kevinschaper:
also removed output directory and package parameters that didn't have any effect on pydanticgen.

### DIFF
--- a/sphinx/generators/pydantic.rst
+++ b/sphinx/generators/pydantic.rst
@@ -1,4 +1,4 @@
-Java
+Pydantic
 ======
 
 Overview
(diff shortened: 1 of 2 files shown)


sphinx/generators/pydantic.rst
```diff
@@ -0,0 +1,50 @@
...
+Code
+^^^^
+
+
+.. autoclass:: PydanticGenerator
+    :members: serialize
+
+Additional Notes
+----------------
+LinkML contains two Python generators. The Pydantic dataclass generator is specifically useful for FastAPI, but is newer and less full featured than the standard `Python generator <python.html>`_.
```
cmungall, line 31: I think the idiom is you `<python>` but not an rst expert!

cmungall (approved):

Issue #446, opened 2021-11-02
Add a page for gen-prefix-map


cmungall:
We need a page in sphinx/generators for the gen-prefix-map command

We have this page:

https://linkml.io/linkml/generators/jsonld-context.html

However, json-ld 1.1 is quite complex and many consumers want a simple yaml prefix mapping, see https://github.com/biolink/biolink-model/issues/710

sujaypatil96:
Running `gen-prefix-map` on `examples/PersonSchema/personinfo.yaml` in the following way:

```
gen-prefix-map examples/PersonSchema/personinfo.yaml
```

gives us the following output:

```
{
   "CODE": "http://example.org/code/",
   "GEO": "http://example.org/geoloc/",
   "GSSO": "http://purl.obolibrary.org/obo/GSSO_",
   "P": "http://example.org/P/",
   "ROR": "http://example.org/ror/",
   "famrel": "https://example.org/FamilialRelations#",
   "linkml": "https://w3id.org/linkml/",
   "personinfo": "https://w3id.org/linkml/examples/personinfo/",
   "prov": "http://www.w3.org/ns/prov#",
   "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
   "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
   "schema": "http://schema.org/",
   "skos": "http://example.org/UNKNOWN/skos/",
   "xsd": "http://www.w3.org/2001/XMLSchema#",
   "Address": {
      "@id": "schema:PostalAddress"
   },
   "Organization": {
      "@id": "schema:Organization"
   },
   "Person": {
      "@id": "schema:Person"
   }
}
```

I wanted to understand why only `Address`, `Organization` and `Person` classes were showing up?

PR #457, opened 2021-11-08
add sphinx page for `gen-prefix-map`

sujaypatil96:
This PR does the following:

- add sphinx page for the `gen-prefix-map` command as required in issue #446 
- remove `personinfo.yaml` as the example output from `gen-prefix-map` and add JSON and TSV output files instead


sujaypatil96:
Updated the code in the wrong branch, so closing this PR.

Issue #463, opened 2021-11-11
In some cases generation of post_init inlining method prematurely makes a key from some slots


turbomam:
__UPDATE from CJM__ I created a minimal example, scroll down to https://github.com/linkml/linkml/issues/463#issuecomment-977487300

I don't know if this is a LinkML error or something specific to the CRDC-H model, so I am [cross posting](https://github.com/cancerDHC/ccdhmodel/issues/154) from the repo where the CRDC-H model lives. Note the worked example lives in yet another repo!

TL;DR: I created a YAML datafile by using the CRDC-H Python data classes. When I try to load that (or use linkml-convert), I get the unhashable type: 'CodeableConcept' error

----

including the care crossposted link for link discovery: https://github.com/cancerDHC/ccdhmodel/issues/154

gaurav:
- Example file that (correctly) generates a data file: https://github.com/cancerDHC/example-data/blob/061556713077371d25b8b252e142dc08c03ae7a9/gen_diag_with_stage_obs_set.py
  - I'm not sure which version of `crdch_model` Mark was using, but I think you can replicate the error with https://pypi.org/project/crdch-model/1.1.5/
- Example data file that causes an error when reloading it using the YAML loader: https://github.com/cancerDHC/example-data/blob/061556713077371d25b8b252e142dc08c03ae7a9/gen_diag_with_stage_obs_set.yaml

Is there somewhere in the linkml or linkml-runtime repos where something like this could be written up as a test case?

cmungall:
ideally this could be distilled into a smaller test case. I'm having a hard time figuring out how to make this actionable. I'm also having a hard time finding the relevant schema on github since it doesn't seem to be tagged with 1.1.5... and it's a large schema

cmungall:
OK, I made a more minimal test file:

```yaml
stage:
- method_type: []
  observations:
  - observation_type:
      coding:
      - code: C25605
        system: http://ncithesaurus.nci.nih.gov/
        label: Overall
    value_codeable_concept:
      coding:
      - code: C96258
        system: http://ncithesaurus.nci.nih.gov/
        label: FIGO Stage IIIC
```

script that tests this:

```python
from linkml_runtime.loaders import yaml_loader
from crdch_model import Diagnosis
yaml_loader.load('test.yaml', target_class=Diagnosis)
```


I can replicate the error this way (but still ideally we would have a minimal reproducible test)

I think this is telling:

```python
[...]

PR #488, opened 2021-11-24
minimal test demonstrating post_init issue for #463

cmungall:
This is a minimal test for #463 

Currently this fails, as the python generator creates code:

```python
@dataclass
class Container(YAMLRoot):

    class_class_uri: ClassVar[URIRef] = EX.Container
    class_class_curie: ClassVar[str] = "ex:Container"
    class_name: ClassVar[str] = "Container"
    class_model_uri: ClassVar[URIRef] = EX.Container

    contains: Optional[Union[Union[dict, Contained], List[Union[dict, Contained]]]] = empty_list()

        self._normalize_inlined_as_list(slot_name="contains", slot_type=Contained, key_name="type", keyed=False)

```


cmungall:
the test class shouldn't be called  IssueJSONSchemaInlinedAsDictCase -- doesn't related to jsonschema per se

cmungall (commented):
I can't approve since I did the initial PR, but I approve the fix, thanks Harold!

PR #498, opened 2021-12-03
new DocumentGenerator, using SchemaView

cmungall:


joeflack4:
Just adding a note here that we discussed during the call today the utility of having a "tab group"; i.e. we can click through tabs to see different formats, e.g. yaml, shex, python, etc. I'm pondering how that would best be implemented, since markdown doesn't have that. Maybe generate the markdown, then do markdown->html, and then edit the HTML file to add the tab group.

cmungall:
> Maybe generate the markdown, then do markdown->html, and then edit the HTML file to add the tab group.

or just use completely different jinja2 templates - it's only a minimal amount of redundancy

cmungall:
See also https://stackoverflow.com/questions/37606292/how-to-create-tabbed-code-blocks-in-mkdocs-or-sphinx

cmungall:
I'm going to hand this PR over to you now @joeflack4

cmungall:
We can also embed HTML directly in markdown. May all get difficult to
manage if there are js dependencies though...

PR #507, opened 2021-12-10
Remove the pandas dependency

hsolbrig:
Also add python 3.10 to the tox unit tests

Issue #506, opened 2021-12-10
linkml shouldn't import pandas


hsolbrig:
**Describe the bug**
Pandas is a huge module and linkml does not need it.  It should not be a dependency.

cmungall:
where are we with this @hrshdhgd ?

hrshdhgd:
I believe this is taken care of ,although there are still `pandas` references in `toml` , `requirements.txt` and `pipfile`s

hrshdhgd:
<img width="293" alt="Screen Shot 2022-05-27 at 1 03 19 PM" src="https://user-images.githubusercontent.com/5247075/170766141-b179fb1b-d483-40e8-b53c-def356ca04ae.png">

hrshdhgd:
[PR](https://github.com/linkml/linkml/pull/804) to remove remnants.

PR #518, opened 2021-12-14
removed pandas from sssomgen operations

### LINKED ISSUE #516: Create a generic framework for managing schemas as google sheets
We have a few different projects with their own sheet2linkml python frameworks (ccdh, nmdc, ...). Other projects have expressed interest in modeling schemas as sheets.

Is a generic framework possible

I think having everyone conform to the exact same columns is too restrictive

instead we can be inspired by https://robot.obolibrary.org/template and have a 2nd row that provides a mapping between the local metamodel names and the linkml model

E.g. ccdh has a sheet for alcohol observations:

Status | Entity | Attribute | Description | Cardinality | Type | Example Values | Comments | Source Mapping | Indirect  Source Mapping | Profile Binding | Value Set Binding | Value Binding | Dev Issues | Dev Log | caDSR Mapping | ADM Mapping | FHIR Mapping | BRIDG Mapping |   |   |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
include | AlcoholExposureObservation |   | A structured object that describes a single data item about an individual's exposure to alcohol, as generated through a point-in-time observation, measurement, or interpretation. |   |   |   |   |   |   |   |   |   |   | Mappings:- FHIR:Observation- BRIDG:PerformedObservation + PerformedObservationResult |   |   |   |   |   |   |  
include | AlcoholExposureObservation | id | The 'logical' identifier of the entity within the system of record.  The simple value of this attribute stands for an identifier of this data object within the system, it can be used as a reference from other objects within the same system (i.e. primary and foreign keys), and it should be unique per type of object. The same data object copied to a different system will likely have a different "id" in the new system since "id" values are system specific and do not represent persistent business identifiers. Business identifiers are assigned outside the information system and are captured in the "identifier" field. The "id" field is more likely to be a serially or randomly generated value that is assigned to the data object as it is created in a system. | 0..0 | string |   | FHIR naming conventions are followed here, where 'id' is used for logical identifiers, and 'identifier' for business identifiers.   The identified entity may have a different 'id' in a different system, and have many external business 'identifiers' associated with it. If an "id" value can also be considered to be a business identifier, and should be transmissible to other systems, it should be duplicated in the "identifier" attribute along with the additional metadata needed to give context to the value of "id". An entity in this model that does not have an "id" field is considered to be an embeddable entity that can not stand on its own. |   |   |   |   |   |   |   |   |   |   |   |   |   |  
include | AlcoholExposureObservation | category | The general category of observation described | 0..0 | CodeableConcept |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |  
include | AlcoholExposureObservation | observation_type | The specific type of alcohol exposure observation made - i.e. the feature or characteristic that was observed, measured, estimated, etc.. | 1..1 | CodeableConcept | alcohol_days_per_week \| alcohol_drinks_per_day \| alcohol_history \| alcohol_intensity | This field holds the 'key' in the core key-value pair comprised of the observation_type field and the relevant value(s) field. |   |   |   | AlcoholExposureObservation.observation_type |   |   |   |   |   |   |   |   |   |  

There is an implicit mapping here:

| CRCD-H | LinkML |
| --- | --- |
| Entity | Class |
| Attribute | attributes |
| Description | description |
| Cardinality | required & multivalued |
| Type | Range |

etc

We have either row 2 of the sheet contain the mapping, or do this is a separate sheet or sssom file or even its own linkml model

The value of each cell in the mapping row would be a yaml string implicitly bounded by `{...}`

It could include values like

 * slot.name
 * class.name
 * enum.name
 * permissible_value

E.g. if a column has `slot.name` in its mapping row then the value of subsequent rows in that column would be a slot name

These could also be parameterized, e.g.

 * `type: {C: ClassDefinition, S: SlotDefinition}`

Allowing the local schema to use simple codes for linkml elements

This schema could be used to have sheets define classes, slots, enums, types, or even a mixture

 * if class.name and slot.name present, it's a slot_usage or slot
 * if just slot.name, it's a slot
 * if just class.name, it's a class
 * and so on for even schemas, enums
 * optionally a type field, e.g `type: {C: ClassDefinition, S: SlotDefinition}`

There could be special purpose values in the mapping row, e.g. `cardinality` would map UML syntax to required + multivalued tuples

so the example for CCDH could be

              Status: include
              Entity: ClassDefinition.name
           Attribute: SlotDefinition.name
         Description: description
         Cardinality: cardinality
                Type: range
      Example Values: 
            Comments: comments: {sep: '|'}
        FHIR Mapping: exact_mappings: {prefix: FHIR}

### LINKED ISSUE #517: Remove pandas dependency from sssomgen
In response to #506 and #507

hrshdhgd:
Addresses #506 and is supplement to #507. Fixes #517

### DIFF
--- a/linkml/generators/sssomgen.py
+++ b/linkml/generators/sssomgen.py
@@ -12,7 +12,6 @@ from linkml_runtime.linkml_model.meta import (
     EnumDefinition,
     Definition,
 )
-import pandas as pd
 
 DEFAULT_OUTPUT_FILENAME = "sssom.tsv"
 
@@ -49,8 +48,6 @@ class SSSOMGenerator(Generator):
         "exact_mappings": "skos:exactMatch",
     }
 
-    final_df = pd.DataFrame(columns=msdf_columns)
-
     def __init__(
         self,
         schema: Union[str, TextIO, SchemaDefinition],
@@ -60,6 +57,7 @@ class SSSOMGenerator(Generator):
     ) -> None:
 
         self.sourcefile = schema
+        self.table_as_list = []
         if output:
             self.output_file = output
         else:
@@ -69,10 +67,14 @@ class SSSOMGenerator(Generator):
             format = self.valid_formats[0]
         super().__init__(schema, **kwargs)
 
-    def make_msdf(self, row_as_dict: dict) -> None:
-        tmp_df = pd.DataFrame(row_as_dict, index=[0])
-        tmp_df = tmp_df.reindex(columns=self.msdf_columns, fill_value="")
-        self.final_df = pd.concat([self.final_df, tmp_df], ignore_index=True)
+    def make_msdf_list(self, row_as_dict: dict) -> None:
+        list_of_row = []
+        for col_name in self.msdf_columns:
+            if col_name in row_as_dict.keys():
+                list_of_row.append(row_as_dict[col_name])
+            else:
+                list_of_row.append("")
+        self.table_as_list.append(list_of_row)
 
     def definition_extract_info(self, obj: Definition):
         subject_source = obj.from_schema
@@ -108,7 +110,8 @@ class SSSOMGenerator(Generator):
                         row_dict["subject_source"] = subject_source
                         if match_type:
                             row_dict["match_type"] = match_type
-                        self.make_msdf(row_dict)
+
+                        self.make_msdf_list(row_dict)
 
         elif type(obj) is EnumDefinition:
             if obj.permissible_values:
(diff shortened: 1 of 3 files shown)


cmungall (approved):

hrshdhgd:
I believe this is taken care of ,although there are still `pandas` references in `toml` , `requirements.txt` and `pipfile`s

PR #521, opened 2021-12-17
Adding tool developer guide

cmungall:


cmungall:
@ddooley - if you have time to look this over that would be great!

ddooley:
Ok I will check it out.  Completely packed day though today.

sphinx/developers/tool-developer-guide.rst
```diff
@@ -0,0 +1,441 @@
...
+
+There are a few broad approaches here:
+
+- 1. Use the direct output of your language's YAML/JSON parser, and
+  implement logic ad-hoc
+- 2. Develop a full blown analog of the python SchemaView class for
+  your language
+- 3. Write python code to transform LinkML YAML into a native form
+  geared for your application
+
```
ddooley, line 145: I think there is a fourth option, which we're doing in DataHarmonizer: Use LinkML python SchemaView and other calls to create a compiled data structure which can be read as YAML or JSON, which is then read-in by your platform and language of choice (e.g. javascript in a browser).

cmungall:
That's actually what I intended by option 3 -- "native form" == "compiled
data structure" - maybe I need to make that clearer!

Issue #104, opened 2021-01-25
Using reserved words such as 'null' generates undecipherable stack traces


cmungall:
if I use the keyword `null` as a slot name:

```yaml
id: http://w3id.org/test
description: test
imports:
- biolinkml:types
prefixes:
  biolinkml: https://w3id.org/biolink/biolinkml/
  test: https://w3id.org/mixs/
default_prefix: test
slots:
  null: {}
  
classes:
  core:
    slots:
    - null
```

and try and use any generator I get stack traces that don't identify the root cause. this can be hard to debug in large schemas

```bash
$ gen-py-classes ~/tmp/bad.yaml 
/Users/cjm/repos/mixs-demo/venv/lib/python3.7/site-packages/biolinkml/__init__.py:158: UserWarning: Some URL processing will fail with python 3.7.5 or earlier.  Current version: sys.version_info(major=3, minor=7, micro=1, releaselevel='final', serial=0)
  warn(f"Some URL processing will fail with python 3.7.5 or earlier.  Current version: {sys.version_info}")
Traceback (most recent call last):
  File "/Users/cjm/repos/mixs-demo/venv/bin/gen-py-classes", line 11, in <module>
    sys.exit(cli())
  File "/Users/cjm/repos/mixs-demo/venv/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/cjm/repos/mixs-demo/venv/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/cjm/repos/mixs-demo/venv/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/cjm/repos/mixs-demo/venv/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/cjm/repos/mixs-demo/venv/lib/python3.7/site-packages/biolinkml/generators/pythongen.py", line 881, in cli
    print(PythonGenerator(yamlfile, emit_metadata=head, gen_meta=genmeta, gen_classvars=classvars, gen_slots=slots,  **args).serialize(emit_metadata=head, **args))
  File "/Users/cjm/repos/mixs-demo/venv/lib/python3.7/site-packages/biolinkml/generators/pythongen.py", line 35, in __init__
    super().__init__(schema, format, **kwargs)
[...]

PR #525, opened 2021-12-18
Ensure valid enums from Pydanticgen 

kevinschaper:
I'm looking to match biolink-model-pydantic's behavior of using 'value_{index}' in place of enumerations that don't translate nicely to python identifiers. 

Inspired by old fashioned web MVC, I'm attempting to move in the direction of structuring the data structure passed out of pydanticgen and into the template in a way that matches the needs of the templating. 

For this PR, that means that I need to pass a label and value for each member of an enumeration. My initial commit is passing as a dictionary, but I need to convert to data classes (pydantic data classes? why not!), which has the benefit of better documenting what is available to a template. 

Ideally I can also figure out how to test the enum generation method in its own standalone test, but I had a little trouble with that.


kevinschaper:
For the biolink strand enum, biolink-model-pydantic gives 

```
class StrandEnum(str, Enum):
    """
    strand
    """
    value_0 = "+"
    value_1 = "-"
    value_2 = "."
    value_3 = "?"
```

and because the underscore method gets called, this is giving:
```
class StrandEnum(str, Enum):
    
    value_0 = "+"
    _ = "-"
    value_2 = "."
    value_3 = "?"
```

which is a little weird, for PredicateQualifierEnum, bmp gives
```
@unique
class PredicateQualifierEnum(str, Enum):
    """
    constrained list of qualifying terms that soften or expand the definition of the predicate used. can be used to
    constrain or qualify any predicate (any child of related_to).
    """
    predicted = "predicted"
    possibly = "possibly"
    hypothesized = "hypothesized"
    validated = "validated"
    value_4 = "supported by real-world evidence"
    value_5 = "supported by clinical evidence"
```

and this code is giving:
```
class PredicateQualifierEnum(str, Enum):
    
    predicted = "predicted"
    possibly = "possibly"
    hypothesized = "hypothesized"
    validated = "validated"
    supported_by_real_world_evidence = "supported by real-world evidence"
    supported_by_clinical_evidence = "supported by clinical evidence"
```

Which I think shows the benefit of the underscore method. I can probably just fix the pathological minus to underscore thing by not running underscore on one character enums.

kevinschaper:
I was able to make something that felt close enough to a unit test by just defining a very small schema. I decided not to try to fix the "-" enum being translated to "_", it seemed maybe a bit of an over-fit. It would be equally strange for "--" to be translated to "__", but I also didn't feel like adding a lot of extra code, when the real goal is simply to make an enum that is an identifier and the actual name maybe isn't so important. At least not for now!

cmungall:
code all looks great

policy-wise, I'm wondering about the decision to use an incremental index. insert a new PV anywhere at the end of the list would break code.

what about using something analogous to URL encoding; perhaps spelling out, e.g. '+1' => 'PLUS_1'

I am sure there must be a standard for this.. maybe this is relevant:

https://stat.ethz.ch/R-manual/R-devel/library/base/html/make.names.html

kevinschaper:
Great point! 

I couldn't find a python equivalent of make.names, but I think I can use unicodedata.name()

kevinschaper:
@cmungall Would it make more sense to make the unicodedata.name replacements as a part of the underscore method in linkml_runtime instead?

cmungall:
Yes, I think that makes sense. Replace any non alphanum/underscore.

Do we still want to universally prefix with `value_`. It seems if I take care to make my enums computer friendly I want to use them directly and not have some unpythonesque `value_foo`

so perhaps:

 1. For first character `c`:
     - use directly if alpha
     - if 0-9 or underscore, then prepend with `value_`, followed by that character
     - otherwise `name(c)`
 2. For subsequent characters
    - use directly if alphanumeric or underscore
    - otherwise `name(c)`

This has the disadvantage of some "surprise" for mixed enums, but it keeps the default case simple and clean

btw, we can be using similar rules for any kind of language conversion. LinkML allows classes or slots to start with numbers (e.g. `5 prime`), and I believe any kind of  character, though I wouldn't recommend (e.g. `5'`). See also #104 

others should weight in here!

kevinschaper:
Oh, there's no `value_` getting prepended now, but I am prepending an underscore for anything that starts with a number. Here's the test examples I have so far (but I think I could reasonable add a lot more):

```
enums:
  TestEnum:
    permissible_values:
      123:
      +:
      This & that, plus maybe a 🎩:
      Ohio:
```
getting mapped to:
```
        assert enum['values']['_123'] == '123'
        assert enum['values']['PLUS_SIGN'] == '+'
        assert enum['values']['This_AMPERSAND_that_plus_maybe_a_TOP_HAT'] == 'This & that, plus maybe a 🎩'
        assert enum['values']['Ohio'] == 'Ohio'
```

I can swap that `_123` to `value_123` though, that might be more sensible given that underscore has a meaning.

cmungall:
That may be too extreme, as it could end up changing people's existing
jsonschema (for example)

I think we keep underscore as is but start planning a general purpose name
warper that can be used and configured for different generators. This will
take a bit of thought and planning

cmungall:
love it!!!

no strong opinions about the leading number. Maybe number_123?

cmungall (approved):
merge when ready!

PR #591, opened 2022-02-04
Pydanticgen: fixes to description quoting & class ordering 

kevinschaper:
I also brought the PydanticConfig  over from biolink_model_pydantic and set the dataclass annotation.

For now, the class ordering is just sitting locally within pydanticgen, but I put in a note about moving it to SchemaView. 

It looks like we've made it past the initial syntax error stage, so it may be possible to use with FastAPI.

### DIFF
--- a/linkml/generators/pydanticgen.py
+++ b/linkml/generators/pydanticgen.py
@@ -62,17 +62,16 @@ class {{ c.name }}(
     {{ c.description }}
     \"\"\"
     {%- endif %}
-    {% for attr in c.attributes.values() -%}
+    {% for attr in c.attributes.values() if c.attributes -%}
     {{attr.name}}: {{ attr.annotations['python_range'].value }} = Field(None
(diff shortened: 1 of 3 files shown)


kevinschaper:
I'm stuck here on `linkml:Any` 

I haven't found a way to create a python class with no validation at all, I've tried an alternate PydanticConfig with validation options set to false, but it still won't allow the metadata assignment from the example. 

I can generate `AnyObject = Any` and explicitly avoid generating a class, but then it the `Person.schema_json` method in test_pydanticgen fails, because it's not a pydantic object. 

It works fine if I just skip generating AnyObject completely and just use Any directly in the attribute, but maybe that would violate the expectation of defining a `linkml:Any` class in the schema? 
```
metadata: Optional[Any] = Field(None, description="""Example of a slot that has an unconstrained range""")
```

cmungall (approved):

Issue #592, opened 2022-02-05
pip dependency conflict?


RichardBruskiewich:
```
ERROR: Cannot install -r requirements.txt (line 33), -r requirements.txt (line 36), click>=8.0.0 and linkml==1.1.13 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested click>=8.0.0
    black 22.1.0 depends on click>=8.0.0
    linkml 1.1.13 depends on click~=7.0
...
```
KGX pulls in **black** with (`click>=8.0.0)` but also sets `Click~=7.0`. Can **linkml** also rather use click>=8.0.0 ?

Also affects code upstream, like **kgx** (see https://github.com/biolink/kgx/issues/371)

dalito:
The [Pipfile](https://github.com/linkml/linkml/blob/f32b173ff2163b3dd64cacb1c9499ee0a5ca0a76/Pipfile#L7) is not yet updated. It still contains:
```
click = "~=7.0"
```

RichardBruskiewich:
> The [Pipfile](https://github.com/linkml/linkml/blob/f32b173ff2163b3dd64cacb1c9499ee0a5ca0a76/Pipfile#L7) is not yet updated. It still contains:
> 
> ```
> click = "~=7.0"
> ```

Sorry.. my bad. Patched and committed.

cmungall:
can this be closed?

PR #593, opened 2022-02-05
click version set >= 7.0.0

RichardBruskiewich:
Resolves issue #592


cmungall (changes requested):
For libraries, I recommend being relaxed about versions

pinning to 7x was too strict
pinning yo 8x also too strict (unless we really need these features)

I recommend instead allowing 7 or greater

RichardBruskiewich:
> For libraries, I recommend being relaxed about versions
> 
> pinning to 7x was too strict pinning yo 8x also too strict (unless we really need these features)
> 
> I recommend instead allowing 7 or greater

Hi @cmungall (cc: @hsolbrig, @sierra-moxon),

The practical problem actually arises with [the `black` package used by LinkML, KGX, etc. which _**enforces**_ the dependency on  >8.0.0](https://github.com/psf/black/blob/main/setup.py#L100).  

Pinning our version to >=7.0.0 still won't (always) work  (would only satisfy if 8.0.0 is pulled in by LinkML and KGX).

cmungall:
I'm not following. Why won't >=7.0.0 work?

If a downstream requires 8.0.0 then 8.0.0 will be used

cmungall (approved):

Issue #613, opened 2022-02-17
Change tests so that they follow procedure documented in CONTRIBUTING.md


sujaypatil96:
Some tests in the linkml repos, specifically ones in [test_issues](https://github.com/linkml/linkml/tree/main/tests/test_issues) do not follow the guidelines detailed in the CONTRIBUTING.md doc, in that are incorrectly named, i.e., without the `test_` prefix in the filenames, which cause them not to be run by the Actions workflows.

Urjaswi-2655:
@sujaypatil96 @krishna-saravan  Only tests which are named with numbers of issue are too be corrected right? If that's the condition then there are only first three tests with wrong name i.e. test no. 163,313 and 315.

sujaypatil96:
@Urjaswi-2655 @krishna-saravan yes, that's right. Test issues 163, 313, and 315 need to be prefixed with the `test_` keyword in the filename, per the guidelines in the [CONTRIBUTING.md](https://github.com/linkml/linkml/blob/main/CONTRIBUTING.md) doc.

Some sanity checks:
* Run the test suite with `python -m unittest discover` to make sure the tests are going through
* You can log some output from these tests to check if they are going through

or

* Run `python -m unittest tests/test_issues/*`
* Check if the number of tests that were executed is equal to the number of tests in the `test_issues` folder

Once you've completed the above steps on a branch, you can submit a PR which we will review and merge. If all looks good, you can apply the same strategy to other linkml repos as well.

To make a PR, I suggest creating a fork of the linkml repo, and working on a branch called `issue-613-fix`.

PR #614, opened 2022-02-18
changed prefixs in test_issues

krishna-saravan:
#613 issue | changed the issues according to contributing.md guidelines


cmungall:
Thank you @krishna-saravan!

As you can see the build is failing. However, I think you did the right thing. I think your fix has revealed the fact that we have had a failing test all this time.

Apologies for the fact the github actions output is so hard to read: we have an open issue to clean this up

It looks like this is failing:

test_linkml_test_issue_163.py

In fact this test is for a long-standing issue https://github.com/linkml/linkml/issues/163 (@hsolbrig any chance we can finally fix this?)

I will leave it to @sujaypatil96 to coordinate how to proceed. One path you might want to consider is placing a `@skip` annotation of the failing test, just so we can get this merged (we never merge failing tests). If we do this, then we should add a note in the comments for #163 that we need to unskip the test. However I would not expect someone new to the project to fix 163 as it gets into the guts of linkml a lot!

sujaypatil96:
Thank you for the change @krishna-saravan.

Could you change the files accordingly:

* `tests/test_issues/test_linkml_issue_313.py`
* `tests/test_issues/test_linkml_issue_315.py`
* `tests/test_issues/test_linkml_issue_163.py`

This goes against our guidelines in the CONTRIBUTING.md doc, but for historical reasons and to avoid conflicts with tests coming from biolink, we should do this.

Like @cmungall suggested, you can skip test_linkml_issue_163.py by adding a `@unittest.skip()` construct, similar to this [test](https://github.com/linkml/linkml/blob/main/tests/test_generators/test_sqltransform.py#L25). We can enable it once, issue #163 has been resolved.

sujaypatil96 (commented):
Thank you for the change @krishna-saravan.

Could you change the files accordingly:

tests/test_issues/test_linkml_issue_313.py
tests/test_issues/test_linkml_issue_315.py
tests/test_issues/test_linkml_issue_163.py
This goes against our guidelines in the CONTRIBUTING.md doc, but for historical reasons and to avoid conflicts with tests coming from biolink, we should do this.

Like @cmungall suggested, you can skip [test](https://github.com/linkml/linkml/blob/main/tests/test_generators/test_sqltransform.py#L25)_linkml_issue_163.py by adding a @unittest.skip() construct, similar to this test. We can enable it once, issue https://github.com/linkml/linkml/issues/163 has been resolved.

cmungall:
https://github.com/linkml/linkml/runs/5280664896?check_suite_focus=true#step:6:2206

```
  File "/home/runner/work/linkml/linkml/tests/test_issues/test_linkml_issue_163.py", line 43, in IssueRDFNamespaceTestCase
    @unittest.skip()
TypeError: skip() missing 1 required positional argument: 'reason'
```

See https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures

PR #618, opened 2022-02-24
fix 404 in https://linkml.io/linkml/generators/jsonld-context.html

StroemPhi:
I have no idea if this is the right file to fix this in, but I just wanted to let you know, that those links in the docs are broken otherwise.


cmungall:
it is the right file, thank you!

We should probably use the w3ids for all metamodel element links, but for now let's merge this and fix the broken links

note: once merged it may take a bit to percolate to the live docs as this is still a manual push step

Issue #642, opened 2022-03-09
Create a generator that makes an 'unfolded' json with all induced slots as attributes


cmungall:
This is the composition of two pieces of functionality:

 1. materialize induced slots as attributes
 2. trivial json serialization

Note we already have logic for 1 here:

https://github.com/linkml/linkml/blob/cad58a0fef9995a218b09a1c0fa814f350f7d58e/linkml/generators/docgen.py#L369-L388

PR #643, opened 2022-03-10
add JSON generator to LinkML generator library

sujaypatil96:
This PR seeks to address issue #642.

To test the generator, run:

```
gen-json tests/test_generators/input/personinfo.yaml --materialize-attributes --output ~/path/to/output.json
```


linkml/generators/jsongen.py
```diff
@@ -0,0 +1,91 @@
+import logging
+import os
+import json
+from typing import Dict, List, TextIO, Union
+
+import click
+
+from linkml_runtime.dumpers import json_dumper
+
```
cmungall, line 9: is this standard style, to have these spaces?

linkml/generators/jsongen.py
```diff
@@ -0,0 +1,91 @@
...
+from linkml_runtime.utils.schemaview import SchemaView
+
+from linkml.utils.generator import Generator, shared_arguments
+
+
+logger = logging.getLogger(__name__)
+
+
+class JsonGenerator(Generator):
+    generatorname = os.path.basename(__file__)
```
cmungall, line 21: Add a docstring:

This generator provides a direct conversion of a LinkML schema into json, optionally merging imports and unrolling induced slots into attributes

linkml/generators/jsongen.py
```diff
@@ -0,0 +1,91 @@
...
+    def __init__(
+        self,
+        schema: Union[str, TextIO, SchemaDefinition],
+        class_name: str,
+        **kwargs,
+    ):
+        self.schemaview = SchemaView(schema)
+        self.class_name = class_name
+
+    def _get_class_defn(self, class_name: str) -> ClassDefinition:
```
cmungall, line 34: minor stylistic point: it's probably not worth having a method for this, just call schemaview directly. But YMMV

linkml/generators/jsongen.py
```diff
@@ -0,0 +1,91 @@
...
+                )
+                logger.info(f"The JSON file has been written to {output}")
+                return output
+        else:
+            return json_dict
+
+
+@shared_arguments(JsonGenerator)
+@click.option(
+    "--class_name", "-c", required=True, help="Class to retrieve all induced slots on"
```
cmungall, line 81: sorry, I mis-specified. It's not so useful doing this on a per-class basis

This should be a boolean option `--materialize-attributes/--no-materialize-attributes`, if set then induced slots are materialized as attributes for *all* classes.

cmungall (approved):
great start! most of my comments are stylistic, apart from the main one about the correct option to use

cmungall:
Also now I think about this more, how about calling this `gen-linkml`?

it sounds a bit counterintuitive - why would you generate linkml from linkml? but this really is a direct transform leaving structure intact and changing surface syntax, applying optional transformations. It would also be useful to have a format option for yaml, because sometimes it's useful to create yaml that is unfolded

linkml/generators/jsongen.py
sujaypatil96: Hm, perhaps this will reduce redundancy when we add more options to `gen-linkml`, like the YAML format option?

linkml/generators/jsongen.py
sujaypatil96: My bad, I was trying to separate out imports from standard libraries, third party libraries and custom libraries.

PR #647, opened 2022-03-11
add `--output` option to `gen-linkml`

sujaypatil96:
Allow users to specify path to which the output of `gen-linkml` should be outputted to.

The advantage of this can be seen when we want to read in the expected test result for comparison in a test case.

### DIFF
--- a/linkml/utils/helpers.py
+++ b/linkml/utils/helpers.py
@@ -2,7 +2,6 @@ def remove_duplicates(lst):
     """Remove duplicate tuples from a list of tuples."""
     return [t for t in (set(tuple(i) for i in lst))]
 
-def write_to_file(file_path, data, mode="w"):
-    """Write string data to a file."""
-    with open(file_path, mode) as f:
+def write_to_file(file_path, data, mode="w", encoding="utf-8"):
+    with open(file_path, mode, encoding=encoding) as f:
         f.write(data)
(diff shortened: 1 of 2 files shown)


linkml/utils/helpers.py
```diff
@@ -1,3 +1,8 @@
 def remove_duplicates(lst):
     """Remove duplicate tuples from a list of tuples."""
     return [t for t in (set(tuple(i) for i in lst))]
+
+def write_to_file(file_path, data, mode="w"):
```
cmungall, line 5: seems a bit overkill to have a helper class - or is the idea this is where we abstract over ensuring everything is utf-8?

sujaypatil96: Oh, it's not a class? I thought `linkml.utils.helpers` module was where we put in all our helper methods, that we can use within our core generators, validators, transformers, etc.?

cmungall: sorry, I meant to say helper method

dalito: I would expect that such helper method would be easily overlooked - as easy as it is to forget adding `encoding="utf-8"` like below :-) . Unfortunately the filesystem encoding cannot be changed. So discovering this in tests is difficult unless you want to activate testing on windows in github actions.

cmungall (approved):
thanks, and much cleaner!

PR #651, opened 2022-03-16
taking down the error level to an info.

sierra-moxon:


cmungall (approved):

cmungall:
We can merge this but we need to spend more time investigating what is going on here...

Issue #652, opened 2022-03-17
Take into account descendants of a range class when generating JSON Schema


deepakunni3:
This request is pertaining to the jsonschemagen.py and the use-case can be best described as follows:

If I have the following schema,

```yaml
id: https://w3id.org/Example-Schema
name: Example-Schema
description: >-
  An Example Schema
version: 0.0.0
imports:
  - linkml:types

prefixes:
  linkml: https://w3id.org/linkml/
  example: https://w3id.org/example/

default_prefix: example

classes:
  c1:
    slots:
        - prop1

  c2:
    is_a: c1
    slots:
        - prop2

  c3:
    is_a: c1
    slots:
      - prop3

  named thing:
    slots:
      - prop4


slots:
  prop1:
    required: true

  prop2:
    range: string
    required: true

  prop3:
    range: string
    required: true

  prop4:
    range: c1
```

Then the jsonschema generated by the jsonschema-gen looks as follows:

```json
{
   "$defs": {
      "C1": {
         "additionalProperties": false,
         "description": "",
         "properties": {
            "prop1": {
               "type": "string"
            }
         },
         "required": [
            "prop1"
         ],
         "title": "C1",
         "type": "object"
      },
      "C2": {
         "additionalProperties": false,
         "description": "",
         "properties": {
            "prop1": {
               "type": "string"
            },
            "prop2": {
               "type": "string"
            }
         },
         "required": [
            "prop1",
            "prop2"
         ],
         "title": "C2",
         "type": "object"
      },
      "C3": {
         "additionalProperties": false,
         "description": "",
         "properties": {
            "prop1": {
               "type": "string"
            },
            "prop3": {
               "type": "string"
            }
         },
         "required": [
            "prop1",
            "prop3"
         ],
         "title": "C3",
         "type": "object"
      },
      "NamedThing": {
         "additionalProperties": false,
         "description": "",
         "properties": {
            "prop4": {
               "$ref": "#/$defs/C1"
            }
         },
         "required": [],
         "title": "NamedThing",
         "type": "object"
      }
   },
   "$id": "https://w3id.org/Example-Schema",
   "$schema": "http://json-schema.org/draft-07/schema#",
   "additionalProperties": true,
   "metamodel_version": "1.7.0",
   "properties": {},
   "required": [],
   "title": "Example-Schema",
   "type": "object",
   "version": "0.0.0"
}
```

[...]

PR #653, opened 2022-03-17
Add option to include range class descendants in generated jsonschema

deepakunni3:
This PR adds the functionality as described in issue #652 where an optional flag `--include-range-class-constraints` for jsonschema-gen will add all descendants of the range class when defining the type of the property in jsonschema.

@cmungall would be curious to hear your thoughts on this. Not sure if this is a niche use case or if an approach like this is not ideal for the JSON Schema realm.

### DIFF
--- a/linkml/generators/jsonschemagen.py
+++ b/linkml/generators/jsonschemagen.py
@@ -1,12 +1,13 @@
 import logging
 import os
 from copy import deepcopy
-from typing import Union, TextIO, Optional, Dict, Tuple
+from typing import List, Union, TextIO, Optional, Dict, Tuple
 
 import click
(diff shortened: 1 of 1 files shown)


cmungall:
Can you add the example you provided in #652 as a test? or we can do that for you after the fact in a separate PR

I think this all makes sense, thank you!!

The longer term strategy here is to have a jsonschema-independent way of unrolling a class hierarchy into a non-OO linkml schema, making use of linkml constructs like any_of. See https://linkml.io/linkml/schemas/experimental.html. However, any_of doesn't yet get translated to jsonschema yet. It may make sense to first convert jsonschemagen to use schemaview.

But this is a great step to getting there!

Issue #155, opened 2021-04-04
Create a LinkML editor


cmungall:
I would love to have a flexible schema-driven editor framework

The idea is not to have a LinkML editor. The idea is to have a LinkML-schema driven editor, to be able to edit instance data for any LinkML schema. And since a LinkML schema is described with LinkML we can dogfood it and use the generic editor as a LinkML editor.

The core editor would be plain form-based. I have been tempted to write this using justpy, but it should probably have a traditional Vue-FastAPI type architecture.

On top of the basic features there would be plugins, including things such as hierarchical views. Again this would be flexible and hierarchies could be build from any slot, so I can either (a) have hierarchical editing of an instance graph using my domain model (b) view and modify the hierarchy in a linkml schema.

As a test case, we could convert phenopackets to linkml (using the to-be-written proto2linkml converter) and then instantly make a phenopacket editor.

cmungall:
out of scope for now

cmungall:
But note that https://github.com/cidgoh/DataHarmonizer can be used for this

PR #669, opened 2022-03-24
update Pipfile on repo use pyyaml>=5.4

sujaypatil96:
This PR is analogous to PR [#155](https://github.com/linkml/linkml-runtime/pull/155) on linkml-runtime. All instances of `yaml.load()` without an explicit `loader` argument have been changed to `safe_load()`.


Pipfile
```diff
@@ -13,7 +13,7 @@ prefixcommons = ">=0.1.7"
 rdflib = ">=6.0.0"
 requests = ">=2.22"
 pyjsg = ">=0.11.6"
-pyyaml = "<5.4"
+pyyaml = ">=5.4"
```
cmungall, line 16: do we want to force this? do we know it won't work pre 5.4?

PR #676, opened 2022-04-05
Add support for populating "designates type" fields in Pydanticgen

### LINKED ISSUE #672: Use designates_type field when generating datamodels to auto-populate the field
Some schemas use https://w3id.org/linkml/designates_type specify that a field is used to hold the (linkml) class instantiated, e.g.

```yaml
  category:
    is_a: type
    domain: entity
    range: category type
    designates_type: true
```

this could be leveraged in e.g. pydanticgen to populate a default for that field

- if the range is uriorcurie, use the curie of the class, e.g. biolink:Gene
- if string, use the localid

kevinschaper:
I'm starting this out as a draft to see how it does with tests. 

Fixes #672

### DIFF
--- a/linkml/generators/pydanticgen.py
+++ b/linkml/generators/pydanticgen.py
@@ -67,8 +67,8 @@ class {{ c.name }}
     {%- endif %}
     {% for attr in c.attributes.values() if c.attributes -%}
     {{attr.name}}: {{ attr.annotations['python_range'].value }} = Field(
-    {%- if slot_values[c.name][attr.name] -%} 
-        [\"{{ slot_values[c.name][attr.name] }}\"]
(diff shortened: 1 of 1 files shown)


linkml/generators/pydanticgen.py
```diff
@@ -169,6 +175,20 @@ def sort_classes(self, clist: List[ClassDefinition]) -> List[ClassDefinition]:
                 )
         return slist
 
+    def get_slot_values(self) -> Dict[str, Dict[str, str]]:
```
cmungall, line 178: I might rename something like get_predefined_slot_values

cmungall (approved):

PR #678, opened 2022-04-06
Extend Pydantic classes from BaseModel

kevinschaper:
This is partially just a roll-back when when they were extending from BaseModel before, but I added the PydanticConfig within any class that directly extends from BaseModel - following the example here: https://pydantic-docs.helpmanual.io/usage/model_config/


cmungall (approved):

cmungall:
any update on this @kevinschaper?

kevinschaper:
I'll try this change out again this week and see what I run into this time in a more up to date branch. It looked like a collision with sqlalchemy, but I know that we got past another weird interaction there, so hopefully it will more straightforward on a second attempt.

PR #685, opened 2022-04-11
Supply empty lists on multivalued slots in pydanticgen

kevinschaper:
This is just a subtle tweak to bring in a convenience feature of biolink_model_pydantic. I also realized that I built in the assumption that designates_type slots would be lists, so this PR fixes that.


cmungall:
do you understand why this is failing @kevinschaper - is it related to your change? do you want to try a null PR to check?

kevinschaper:
Good call @cmungall, I made https://github.com/linkml/linkml/pull/689 to check, and it does seem like there's some sort of test problem on main

PR #688, opened 2022-04-13
add utils method that materializes slot patterns based on structured patterns

sujaypatil96:
Add `materialize_patterns()` method that materializes slot patterns based on structured patterns in the settings.

CC: @cmungall @turbomam


linkml/utils/pattern.py
```diff
@@ -0,0 +1,34 @@
...
+        format_spec["{" + k + "}"] = setting.setting_value
+
+    for _, slot_defn in schema_view.all_slots().items():
+        if slot_defn.structured_pattern:
+            struct_pat = slot_defn.structured_pattern
+
+            # compute pattern from structured patterns
+            # and format_spec dictionary
+            slot_defn.pattern = " ".join(
+                format_spec.get(pat, pat) for pat in struct_pat.syntax.split()
```
ddooley, line 33: I'm not sure how split() is working here, but has there been discussion about what should be allowed in struct_pat.syntax?  I think we should allow for "{number}{unit}" i.e. allow for no spaces - many systems might have this content. Means also that slot_defn.pattern parsing will have to preserve spaces existing between structured_pattern "{a} {b} {c}{d}{e}" etc. values.

ddooley: A quickie way to do this is (for interpolated strings only):

slot_defn.pattern = struct_pat.syntax.format(**settings_dict)

so you can probably avoid both loops above.

sujaypatil96: @ddooley: Thank you for the review.

@cmungall had the same comment as you, i.e., what if the serialization isn't space separated. We need to have provisions for handling `{foo}{bar}`.

However the problem with implementing it as:

```
slot_defn.pattern = struct_pat.syntax.format(**format_spec)
```

is that the dictionary keys have a `.` dot in them, and there's no way to escape dots apparently: https://stackoverflow.com/questions/23909449/how-to-escape-dot-in-python-str-format

sujaypatil96:
@cmungall: perhaps this is a good time to make a design decision and use some other separator, say, `:` colon instead of `.` dot in the settings dictionary?

cmungall (changes requested):

ddooley:
If we want the purity of x.y.z type names, which I really like, then this could accomplish that:

```
# Online Python compiler (interpreter) to run Python online.
# Write Python 3 code in this online editor and run it.
from types import SimpleNamespace

def recurseAttr(attrArray, value):  
    if len(attrArray) == 0:
        return value

    return SimpleNamespace(**{attrArray[0]: recurseAttr(attrArray[1:], value) } )

mydict = {'foo':1,
'bar':'2', 
'dom.da': 3, 
'di.domo.da': 4
}

# hardcoded test examples of what's going on
# setattr(mydict['dom'], 'da', 3)
# mydict['dom'] =  SimpleNamespace(**{'da': 3})
# mydict['di'] = SimpleNamespace(**{'domo': SimpleNamespace(**{'da':4}) })

mydict2 = mydict.copy()
for key in mydict:
    if '.' in key:
        dotted = key.split('.')
        mydict2[dotted[0]] = recurseAttr(dotted[1:], mydict[key])
    

pattern = '{foo}.{bar} {dom.da}{di.domo.da}'.format(**mydict2)
print(pattern)
```

sujaypatil96:
@ddooley: Thank you for the review and above snippet.

This is what a typical _settings_ dictionary with a _dot_ looks like. Instead of a `SimpleNamespace`, it is a `string`.
```
settings:
  float: "\\d+[\\.\\d+]"
  unit: "\\S+"
  unit.length: "(centimeter|meter|inch)"
  unit.weight: "(kg|g|lbs|stone)"
  email: "\\S+@\\S+{\\.\\w}+"
```

I'm trying to modify your code to fit this example.

sujaypatil96:
Thank you so much for the snippet above @ddooley. I tested it too, and it works perfectly. I've integrated your code in this PR. There's a test case accompanying the `materialize_patterns()` method in the `tests/test_utils/test_pattern.py` file.

Me adding your code on the PR, didn't feel ethical since it'll show up as my contribution, please do let me know if you would like me to revert the commit, so you can push the snippet to this branch?

ddooley (changes requested):

linkml/utils/pattern.py
```diff
@@ -0,0 +1,86 @@
...
+            struct_pat = slot_defn.structured_pattern
+
+            pattern = struct_pat.syntax
+
+            # compute pattern from structured patterns
+            # and format_spec dictionary
+
+            # regular expression capturing the various use cases
+            # for the optionally dot separated, curly braces bound, pattern syntax
+            var_name = re.compile("{([a-z0-9_-]+([\.-_ ][a-z0-9]+)*)}", re.IGNORECASE)
```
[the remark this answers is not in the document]
sujaypatil96: Great catch Damion. I've addressed this. Thank you.

Issue #692, opened 2022-04-14
Description metaproperty is not being exported with owl-gen


janik-martin:
**Describe the bug**
Description text isn't being exported with OWL generator. When exporting our schema with 'title' and 'description' metadata filled in, only 'title' gets exported into the resulting ontology (as dcterms:title), description is dropped.

**To Reproduce**
1. Create simple schema, having class and some slots with 'title' and 'description' meta
2. Export it via owl-gen e.g. to Turtle format
3. View the exported ontology, e.g. in text editor.
4. dcterms:title is present for both classes and properties, description is missing.

**Expected behavior**
Description gets exported, probably as dcterms:description or rdfs:comment (or both?)

**Additional context**
dcterms:description would perfectly match the metadata attribute name and the parallel to title -> dcterms:title. However rdfs:comment is used by automatic documentation tools like Widoco (https://dgarijo.github.io/Widoco/doc/bestPractices/index-en.html#def) and is broadly used in other ontologies like Schema.org etc...

PR #693, opened 2022-04-14
Unit test and fix for issue #692

janik-martin:
I have added the missing description export as rdfs:comment because of its broad usage in other well known ontologies.

### DIFF
--- a/linkml/generators/owlgen.py
+++ b/linkml/generators/owlgen.py
@@ -107,6 +107,8 @@ class OwlSchemaGenerator(Generator):
                 self.graph.add((uri, SKOS.altLabel, Literal(s)))
         if e.title is not None:
             self.graph.add((uri, DCTERMS.title, Literal(e.title)))
+        if e.description is not None:
+            self.graph.add((uri, RDFS.comment, Literal(e.description)))
         if e.mappings is not None:
             for m in e.mappings:
                 m_uri = self.namespaces.uri_for(m)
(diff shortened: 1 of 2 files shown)


cmungall (approved):

cmungall:
Thanks!

Having it map to rdfs:comment is quite sensible because as you say, it is used in many ontologies.

It does create a slight issue, as in the metamodel we have:

 * https://w3id.org/linkml/description uses skos:description
 * https://w3id.org/linkml/comments uses rdfs:comment

I think ultimately we want to allow for flexibility as everyone does ontology metadata differently (even in small communities like OBO we have time aligning - we just had a 3 hour term metadata workshop this week and did not resolve even some basic issues!).

I suggest that to include this PR in a release we include a command line option `--metadata-profile` which at first can either be not specified or 'rdfs'. If nothing is specified then we default to the mapping that is in the metamodel. If rdfs profile is used then definition maps to rdfs:comment.

We can later add more profiles (e.g. OBO, schemaorg) and add the ability to specify arbitrary mappings on an element by element level. But I want to be able to merge a PR that is immediately useful to you without worrying about introducing behavior we can't guarantee compatible behavior moving forward.

You have already been generous in contributing a PR, if you are willing to take a shot that would be fantastic, but if not, that's OK, we should be able to get to this ourselves very soon.

cmungall:
I will merge and make further changes

Issue #465, opened 2021-11-12
Switch from pipenv to poetry


cmungall:
* should be done across all main repos
     - linkml, runtime, model
     - don't bother with template for now
 * make sure actions work
 * I suggest making PRs on all 3 repos first, checking they work, then merge en masse
 * Ensure we have any necessary instructions in CONTRIBUTING.md for main IDEs

sujaypatil96:
Start with transition to poetry with the linkml-runtime repo.

sujaypatil96:
Reopening this issue since we are yet to _poetrify_ the core linkml repo.

PR #708, opened 2022-04-19
Auto-deploy `sphinx` docs using gh-actions.

### LINKED ISSUE #597: Move generated docs from main branch to gh-pages branch
Like the title says (was "Why is the generated html-documentation in the repo?"). It is very rare to see this. I was wondering why you do it.

hrshdhgd:
Fixes #597


cmungall:
Thanks! If I were to merge this now we would have a period of downtime with the docs?

I suggest 

 1. Do an initial commit of the generated docs to gh-pages
 2. Configure GH to point to these
 3. Merge this
 4. Do a quick PR on docs to make sure autodeploy is working

hrshdhgd:
I did one better -- Used `pipenv` to auto-deploy `sphinx` docs. Once the poetry conversion happens as stated in https://github.com/linkml/linkml/issues/465 , I will switch out the code and the transition should be smoother.

hrshdhgd:
I think it worked:  https://linkml.io/linkml/

Issue #711, opened 2022-04-22
owl-gen Custom annotations export


janik-martin:
I'm creating a new feature request related to discussion in  https://github.com/linkml/linkml/discussions/709

Export custom (local) annotations into owl ontology.

cmungall:
Let's call this one done, but feel free to add new issues if you need new custom behavior!

PR #712, opened 2022-04-22
functionality and unit test for #711

janik-martin:
Hi, regarding our annotations export discussion,  I tried to implement the basic functionality this way. What do you think of it?


cmungall:
apologies for the confusing reporting of errors

here is the salient line:
https://github.com/linkml/linkml/runs/6130359034?check_suite_focus=true#step:6:4509

janik-martin:
Sorry for the delay, I've had to switch to something else for a while. I fixed the small issue (just a leftover in sample data..) I hope now it'll go well.

cmungall:
Thanks!

It looks like you also committed the output files - no harm done, we've all done this accidentally. And I see we seem to have lost our CONTRIBUTING.md that had instructions on this, mmm

janik-martin:
Hmm, not sure how I did it.. in the commit history I only see my files and then some GitHub Action "automated adding outputs.." which is probably that? I would be glad if you point me to a resource where I can see how to avoid it.
Thanks

Issue #687, opened 2022-04-13
Support identifiers on class slots in pydanticgen


kevinschaper:
Currently from the biolink model GeneToDiseaseAssociation pydanticgen is producing a subject slot that looks like this: 

```
subject: GeneOrGeneProduct = Field(None, description="""gene in which variation is correlated with the disease, may be protective or causative or associative, or as a model""")
```

We're currently using strings for the the subject ID, which is supported with biolink_model_pydantic like this:
```
    subject: Union[str, GeneOrGeneProduct] = None
```

For completeness, here is how pythongen handles it, which wouldn't work for our usage: 
```
    subject: Union[dict, GeneOrGeneProduct] = None
```

After digging down with the debugger, I see that biolink_model_pydantic is supporting this using class_identifier_path() from generator.py to allow for the ID or the class

kevinschaper:
I think I was actually wrong about this, since the field is `inline: false` I don't actually think the class should be assignable, just a str, which is what's happening.

PR #714, opened 2022-04-22
Implement mixin inheritance in pydanticgen

kevinschaper:
At last week's LinkML meeting we talked through issue #687, and rather than taking biolink-model-pydantic's approach of always allowing either the class or the ID scalar, we thought it made more sense to implement mixin inheritance. 

The challenging bit was getting listing of classes to have a valid method resolution order in the class definition ([this is a nice explanation](https://www.educative.io/edpresso/what-is-mro-in-python)). 

This actually wasn't implemented yet in biolink-model-pydantic, which had a hardcoded fix to handle the one spot where the biolink model didn't just happen to get it right, but obviously we don't want to put it on the model to worry about the order in which classes are listed. 

Pydanticgen already sorts classes in working order for inheritance, so sorting the subset of parent classes in the same order and then reversing gives a valid MRO.

### DIFF
--- a/linkml/generators/pydanticgen.py
+++ b/linkml/generators/pydanticgen.py
@@ -114,13 +114,18 @@ class PydanticGenerator(OOCodeGenerator):
                  template_file: str = None,
                  allow_extra = False,
                  format: str = valid_formats[0],
-                 genmeta: bool=False, gen_classvars: bool=True, gen_slots: bool=True, **kwargs) -> None:
+                 genmeta: bool=False,
+                 gen_classvars: bool=True,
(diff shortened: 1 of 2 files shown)


kevinschaper:
Here is the test failure:

```
sqlalchemy.exc.InvalidRequestError: Class <class 'test.Person'> has multiple mapped bases: [<class 'test.HasNewsEvents'>, <class 'test.HasAliases'>, <class 'test.NamedThing'>]
```

I'm guessing that SQLAlchemy has some special needs how to handle multiple inheritance.

PR #725, opened 2022-04-27
Find mixin identifier ranges using descendants in Pydanticgen

kevinschaper:
The current Pydanticgen behavior is to inline a class when the slot is defined as inlined, or the range class lacks an identifier slot. For my (Biolink-model) usage, none of my mixins have identifier slots, leading to inconsistent python classes which expect classes only for mixins but take strings when a class is defined. 

This PR looks at the slot identifiers of the descendants of the mixin to generate slot ranges.

### DIFF
--- a/linkml/generators/pydanticgen.py
+++ b/linkml/generators/pydanticgen.py
@@ -229,6 +229,34 @@ class PydanticGenerator(OOCodeGenerator):
         else:
             return f"Union[{'.'.join(id_ranges)}]"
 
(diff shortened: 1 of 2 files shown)


kevinschaper:
I'm not sure, but this might be failing because kitchen_sink doesn't define an identifier slot anywhere, so it's expecting a full company object rather than just a string. I'm unsure whether I should code to the test or update the test / test data?

cmungall (approved):

Issue #726, opened 2022-04-28
gen-json-schema does not convert some slot definition attributes to "const"


gcampion-cdisc:
**Describe the bug**
During generation, some slot definition attributes should be converted to "const" in the output json-schema. One example is "equals_string", but there are others.

**To Reproduce**
1. create a linkml with the following:

```
...
slots:
  MySlot:
      equals_string: My string value
...
```

2. Generate a json-schema using gen-json-schema

**Expected behavior**

A json-schema is generated with the following:

```
...
"properties": {
    "MySlot": {
        "const": "My string value"
    }
}
...
```
(Currently, a "const" is not created.)

cmungall:
Thanks for the report!

@pkalita-lbl fixed this, and the fix is incorporated into https://github.com/linkml/linkml/releases/tag/v1.2.10, now on pypi

PR #730, opened 2022-04-29
json-schema consts #726

cmungall:


pkalita-lbl:
I added handling for `equals_string` and `equals_number` slot attributes in the jsonschema generator. I think this is ready for review now.

cmungall (commented):
Thanks!

PR #746, opened 2022-05-03
proposal: add 'ifabsent' as a slot available for use SlotDefinition

### LINKED ISSUE #633: define a default value for a slot: instantiate this default value in instances of the class.
@LarsVogt

equals_string is helpful, but only on validation.  Is there a way to specify a default value for a slot that is instantiated in all (some?) of the serializations of the model?

sierra-moxon:
(must be added after linkml-model change here: https://github.com/linkml/linkml-model/pull/95)

### DIFF
--- a/tests/test_issues/test_linkml_issue_660.py
+++ b/tests/test_issues/test_linkml_issue_660.py
@@ -34,7 +34,7 @@ classes:
       s3:
         equals_number: 32
       s5: 
-        if_absent: foo
+        ifabsent: foo
   D:
     slots:
       - s1
@@ -53,10 +53,10 @@ slots:
     equals_number: 7
     range: integer
   s5: 
-    description: test override for if_absent
+    description: test override for ifabsent
     range: string
   s6:
-    if_absent: foo
+    ifabsent: foo
     range: string
 """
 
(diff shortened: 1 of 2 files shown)


tests/test_issues/test_linkml_issue_660.py
```diff
@@ -0,0 +1,89 @@
...
+    description: test override for equals_number
+    range: integer
+  s4:
+    equals_number: 7
+    range: integer
+  s5: 
+    description: test override for if_absent
+    range: string
+  s6:
+    if_absent: foo
```
cmungall, line 59: https://w3id.org/linkml/ifabsent

      function that provides a default value for the slot.  Possible values for this slot are defined in
      linkml_runtime.utils.ifabsent_functions.default_library:
        * [Tt]rue -- boolean True
        * [Ff]alse -- boolean False
        * int(value) -- integer value
        * str(value) -- string value
        * default_range -- schema default range
        * bnode -- blank node identifier
        * slot_uri -- URI for the slot
        * class_curie -- CURIE for the containing class
        * class_uri -- URI for the containing class

cmungall (changes requested):

sierra-moxon:
I don't think this will pass until we have a release of the linkml-model repo.

PR #748, opened 2022-05-03
stub out values_from

sierra-moxon:
I am not sure the difference in intent between values_from and enumerations.

### DIFF
--- a/docs/schemas/models.md
+++ b/docs/schemas/models.md
@@ -287,6 +287,8 @@ enums:
         meaning: GSSO:000385
 ```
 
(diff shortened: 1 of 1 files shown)


cmungall:
values_from is one of those ones that was used in the original biolink-model but not very well defined for future use

biolink is the only one to use and seems to be a mix, sometimes the canonical id prefix casing, sometimes lowercase? It seems to be ambiguous as to whether it's (a) id_prefixes on a per slot basis (b) a way of applying an enumeration to a CURIE

I think let's hold off on this for now and resync when Harold has progressed further

Issue #753, opened 2022-05-06
`linkml-ws` followup: `make setup`: `poetry: command not found`


joeflack4:
**Describe the bug**
I'm following the steps [in this video](https://asciinema.org/a/GHydP2cs0NIyUdhjvwvSFvB91) to set up a new LinkML project using `linkml-ws`.

**To Reproduce**
1. Create an empty directory in which to create project.
2. (~optional) Make a virtual environment
3. `pip install linkml`
4. Try to create a new LinkML workspace using: `linmkl-ws new -D "<WORKSPACE_NAME>" <WORKSPACE_NAME>`
5. `cd <WORKSPACE_NAME>`
6. `make setup`

**Expected behavior**
Project should be set up.

**Actual behavior**
```
make setup
poetry install
bash: poetry: command not found
make: *** [install] Error 127
```

**Things I tried**
1. Making sure LinkML is installed.
```
pip freeze | grep linkml
biolinkml==1.7.6
linkml==1.2.10
linkml-dataops==0.1.0
linkml-runtime==1.2.9
linmkl-ws new -D "tccm"
zsh: command not found: linmkl-ws
```

**Possible solutions**
1. Making sure that the following is included as an argument to `setup()` in `setup.py`:
```
      install_requires=[
          'poetry',
      ],
```

joeflack4:
I mistyped `linkml-ws` as `linmkl-ws` lol

sujaypatil96:
@joeflack4 you can see the repo no longer uses setuptools and pipenv for package management. We could have documented this better, but you can install the poetry binaries on your machine, and not as a python package, by following the instructions here: https://python-poetry.org/docs/

joeflack4:
I am aware now, thanks Sujay. I discussed briefly w/ Chris during a LinkML meeting. He understands the UX argument I'm making. I think probably some automation can be done for the edge case where a random user on the internet wants to pick up LinkML, but isn't using poetry. But it is probably not worth the dev time anyway to handle that case, so I really don't mind if this PR is closed. I mainly wanted to bring attention to the issue. Hopefully I can contribute in a more meaningful way though some time in the future.

sierra-moxon:
Hi @joeflack4  - thanks!  You could also checkout the linkml-initializer, which is the start of a web-based project creation strategy.  Closing for now, but please do try it out and give @pkalita-lbl any feedback about how it could be improved.

Issue #736, opened 2022-04-30
Github Actions still runs `pipenv` instead of `poetry` to run checks


hrshdhgd:
As the title states, the workflow needs an update.

- [ ] [Install dependencies](https://github.com/linkml/linkml/blob/c103113712c9b4a6e47104ad889d098c27320ffa/.github/workflows/main.yaml#L23)
- [ ] [Update requirements](https://github.com/linkml/linkml/blob/c103113712c9b4a6e47104ad889d098c27320ffa/.github/workflows/main.yaml#L44)
- [ ] [Check in test outputs](https://github.com/linkml/linkml/blob/c103113712c9b4a6e47104ad889d098c27320ffa/.github/workflows/main.yaml#L59)

for starters and maybe other places.

dalito:
You can probaly copy from linkml-runtime which has already been renovated.

PR #754, opened 2022-05-06
Added `poetry` install requirement

joeflack4:
Fix for: linkml-ws followup: make setup: poetry: command not found #753

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,5 +11,6 @@ NAME = 'linkml'
 setup(
     name=NAME,
     setup_requires=['pbr'],
+    install_requires=['poetry'],
     pbr=True,
 )


joeflack4:
There was a converation on slack: https://monarchinitiative.slack.com/archives/C01LDTVURH9/p1651864784659989

@turbomam pointed out the [poetry docs](https://python-poetry.org/docs/#installing-with-pip) on installation. However the recommended `pipx` installation did not work for me, whether or not I was in a virtual environment:

```
pipx install poetry
/Users/joeflack4/.local/pipx/venvs/poetry/bin/python: No module named pip
```

turbomam:
> There was a converation on slack: https://monarchinitiative.slack.com/archives/C01LDTVURH9/p1651864784659989
> 
> @turbomam pointed out the [poetry docs](https://python-poetry.org/docs/#installing-with-pip) on installation. However the recommended `pipx` installation did not work for me, whether or not I was in a virtual environment:
> 
> ```
> pipx install poetry
> /Users/joeflack4/.local/pipx/venvs/poetry/bin/python: No module named pip
> ```
> 
> Normal `pip` appears to have worked, or so it seems.

I haven't used `pipx` so I can't recommend it. I recommend following the curl-based instruction at https://python-poetry.org/docs/#osx--linux--bashonwindows-install-instructions (for everybody except windows powershell users)

dalito:
MAybe you haven't seen it: linkml will switch to poetry so setup.py will be gone soon (#736). ...and pipx is indeed great.

joeflack4:
Indeed, this topic is about poetry. I was wondering about that myself, whether or not `setup.py` is used with poetry, or replaced by something else.

dalito:
poetry uses `pyproject.toml` and does not read `setup.py`.

joeflack4:
@sierra-moxon showed me that the documentation for https://github.com/linkml/linkml-project-template shows that poetry should be installed first. I was just following @cmungall 's video which did not have this step.

gaurav:
I think the command would be `pipx poetry`, not `pipx install poetry` — IIRC, pipx runs a program from Pypi after installing it into a local cache.

joeflack4:
@gaurav I tried that, but for some reason `pipx` didn't work for me: https://github.com/linkml/linkml/pull/754#issuecomment-1119949906

Just curious. @dalito mentioned that Poetry would replace pip soon, however, I think this means that (a) we will be using Poetry as the package manager for LinkML as opposed to Pip. However, I don't think this means (b) that we will be removing the ability for newer versions of LinkML to be installed via Pip. Am I correct?

If I'm indeed correct that we want to continue maintaining the ability for users to install LinkML from PyPi via `pip`, then I don't see much of a drawback to having LinkML auto-install `poetry`. I think this would be good, because I think most new LinkML users in the future will not have heard of Poetry, and it would be better UX for them to install LinkML and just have all of its features work without having to worry about additional system dependencies. However, if Poetry ultimately ends up replacing `pip`, then this will not be an issue at that time.

Not really a big deal; just figured I would try to lay out that argument a little more clearly.

dalito:
The newest pip does no longer need setup.py it can work with only pyprojecct.toml. In general, python packages created with poetry can be pip installed. So using pip for installation will be still possible. linkml-runtime was already "converted to poerty" ad you are still using it. What is happening is a switch from pipenv to poetry but not a switch from pip to poetry.

joeflack4:
> The newest pip does no longer need setup.py. It can work with only pyprojecct.toml.
oh that's cool, didn't know that

cmungall:
should we close this?

joeflack4:
@cmungall That's fine Chris. Going to close now. I brought attention to the thought, which is the important part. I don't think it's worth dev time to think about this level of fine tuning of UX, now that I understand more.

Issue #759, opened 2022-05-09
Improve LinkML documentation


nlharris:
This ticket is mostly a placeholder for me. I will work on improving the documentation and looping others in to help as appropriate.

Links:

- https://linkml.io/ -- main splash page
- https://linkml.io/linkml -- core tools docs
- https://github.com/linkml/schema-automator - schema inference
- https://linkml.io/linkml-registry/registry/

Peter: I think the [introduction](https://github.com/linkml/.github/blob/main/profile/README.md) is confusing:

> LinkML is a flexible modeling language that allows you to author schemas in YAML that describe the structure of your data. LinkML provides a framework for working with and validating data in a variety of formats (JSON, RDF, TSV) provides generators for compiling LinkML schemas to other frameworks.
 
It would be better to say
 
> LinkML is a both a flexible modeling language that allows you to author schemas in YAML that describe the structure of your data as well as a framework for working with and validating data in a variety of formats (JSON, RDF, TSV) provides generators for compiling LinkML schemas to other frameworks.
 
It would be even better not to use the same name for the schema and the tool!

Chris: So technically the schema that describes linkml is linkml-metamodel, the repos are:

https://github.com/linkml/linkml. ## various tools and scripts
https://github.com/linkml/linkml-model. ## the metamodel / self-describing schema

it would be a bit churny to change names now but I agree it's potentially confusing. Maybe if the linkml repo just better describes itself as being the tools...

Jules: Yes, I'd also like to see a lot more information in the README.md about the purpose and content of the repo, especially if it might not be the most accurately named one. Having an embedded link to detailed documentation is good, but I feel there still ought to be some detail in the README given it's the first thing anyone reads.

As you suggested, it would be good if https://github.com/linkml/linkml mentions that the repo contains various tools and scripts and briefly mention what the most important ones are and how they are used.

nlharris:
BTW, I don't have the power to self-assign this ticket or to add a label.

nlharris:
Chris suggested "we should flesh out https://linkml.io/linkml/developers/organization.html and include a link to it"

sierra-moxon:
@nlharris - I checked your permissions on both the organization and this repo in particular, and from what I can see you have write permission here, and full dev privileges on linkml. 🤔

nlharris:
[...]

PR #761, opened 2022-05-09
trying to make it clearer what's in this repo

nlharris:
as per #759


cmungall:
In future we should flesh out https://linkml.io/linkml/developers/organization.html and include a link to it

PR #764, opened 2022-05-10
update Github Actions to work with poetry

sujaypatil96:
The same Github Actions from the exemplar linkml-runtime framework have been ported over to work with linkml core, replacing the Actions that previously used pipenv.


sujaypatil96:
@dalito: any thoughts as to why the Windows tests are not going through here?

dalito:
Great to add Windows to the CI tests. I am not too surprised about the failures: linkML had still some minor issues in contrast to linkML-runtime. I will have a look.

sujaypatil96:
Oh perfect, thank you, I appreciate it.

cmungall:
Let’s break this down and handle things incrementally

first task is to complete the transition to poetry. We are in an odd in between state. Let’s fix asap. If there were hidden windows issues with the tests that were masked because we didn’t do windows tests then leave windows tests out of this pr. Let’s get the transition completed and then immediately make a new pr that adds the windows tests and solve those

dalito:
Ok. I will wait untill this is merged and look at the windows failures afterwards.

PR #771, opened 2022-05-12
with advice from Jules Jacobsen we fix up the proto generator to matc…

sierra-moxon:
…h the proto3 spec

### DIFF
--- a/linkml/generators/protogen.py
+++ b/linkml/generators/protogen.py
@@ -24,6 +24,8 @@ class ProtoGenerator(Generator):
         self.generate_header()
 
     def generate_header(self):
+        print(f' syntax="proto3";')
+        print(f' package')
         print(f"// metamodel_version: {self.schema.metamodel_version}")
         if self.schema.version:
             print(f"// version: {self.schema.version}")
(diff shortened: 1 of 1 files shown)


julesjacobsen:
```proto
syntax = "proto3";

package org.phenopackets.schema.v2.core;

import "phenopackets/schema/v2/core/base.proto";
import "phenopackets/schema/v2/core/measurement.proto";
import "phenopackets/schema/v2/core/phenotypic_feature.proto";

option java_multiple_files = true;
option java_package = "org.phenopackets.schema.v2.core";

// A Biosample refers to a unit of biological material from which the substrate
// molecules (e.g. genomic DNA, RNA, proteins) for molecular analyses (e.g.
// sequencing, array hybridisation, mass-spectrometry) are extracted. Examples
// would be a tissue biopsy, a single cell from a culture for single cell genome
// sequencing or a protein fraction from a gradient centrifugation.
// Several instances (e.g. technical replicates) or types of experiments (e.g.
// genomic array as well as RNA-seq experiments) may refer to the same Biosample.
// FHIR mapping: Specimen (http://www.hl7.org/fhir/specimen.html).
message Biosample {
}
```

sierra-moxon:
tested the "no merge imports" and "merge imports = false" flags and we get many proto files as expected, but the code to import the modularized files into the "package" part of the proto header still needs to be written.  But this PR is doing a little better than before so we can take it in and do more work on this in another PR.

Issue #770, opened 2022-05-12
add jackson/"record" annotation as a new generator for java classes


sierra-moxon:
per meeting with Jules
https://www.geeksforgeeks.org/what-are-java-records-and-how-to-use-them-alongside-constructors-and-methods/#:~:text=Java%20records%20were%20introduced%20with,DTOs%20(Data%20Transfer%20Objects).

use "record" annotation
https://dev.to/brunooliveira/practical-java-16-using-jackson-to-serialize-records-4og4
note: we would need to use Java 17 and greater for these generated classes.

julesjacobsen:
[linkml-java-record-demo.zip](https://github.com/linkml/linkml/files/8691660/linkml-java-record-demo.zip)

```java
public interface PropertyValue {

    String pred();

    String val();

    List<String> xrefs();

    Meta meta();
}

public record SynonymPropertyValue(String pred, String val, List<String> xrefs, Meta meta, String synonymType, boolean isExact, Scope scope) implements PropertyValue {
}

**Comment by kevinschaper** [comment]
Here's where I'm at right now (using Biolink Disease as an example), I haven't tried compiling, because I need to sort out the default values for the builder 

```
package org.biolink;

import java.util.List;

public record Disease(
    List<OrganismTaxon> inTaxon, 
    List<String> providedBy, 
    String id, 
    String iri, 
    List<String> category, 
    String type, 
    String name, 
    String description, 
    String source, 
    List<Attribute> hasAttribute) {

    public static Builder builder() {
        return new Builder();
    }

    public static class Builder {

        private List<OrganismTaxon> inTaxon = ""; //TODO: need to generate appropriate default/empty values for each type
        private List<String> providedBy = ""; //TODO: need to generate appropriate default/empty values for each type
        private String id = ""; //TODO: need to generate appropriate default/empty values for each type
        private String iri = ""; //TODO: need to generate appropriate default/empty values for each type
        private List<String> category = ""; //TODO: need to generate appropriate default/empty values for each type
        private String type = ""; //TODO: need to generate appropriate default/empty values for each type
        private String name = ""; //TODO: need to generate appropriate default/empty values for each type
        private String description = ""; //TODO: need to generate appropriate default/empty values for each type
        private String source = ""; //TODO: need to generate appropriate default/empty values for each type
        private List<Attribute> hasAttribute = ""; //TODO: need to generate appropriate default/empty values for each type


        public Builder inTaxon(List<OrganismTaxon> inTaxon) {
[...]

PR #776, opened 2022-05-13
Initial javagen Record support

kevinschaper:
This adds a template for generating Java 17 Records, and support in the OODocument for passing all fields to the template

Working to at least partially address issue #770


cmungall (approved):

linkml/generators/javagen.py
```diff
@@ -42,9 +42,9 @@
 
 TYPEMAP = {
     "str": "String",
-    "int": "Integer",
-    "float": "Float",
-    "Bool": "Boolean",
+    "int": "int",
+    "float": "float",
```
cmungall, line 46: it looks like you are mapping from python types - I think you should map from the uri of the type or the name of the type, this gives you the distinction eg between float and double. See #777

cmungall (approved):

linkml/generators/javagen.py
cmungall: Here is the problem:

https://github.com/linkml/linkml/blob/29a64763d113a7ee6305d73765c432661858241c/linkml/generators/javagen.py#L70-L71

We map from t.base, which has already lost information.

```yaml
  float:
    uri: xsd:float
    base: float
    description: A real number that conforms to the xsd:float specification

  double:
    uri: xsd:double
    base: float
    description: A real number that conforms to the xsd:double specification
```

instead map from t.uri

I also recommend using sv.induced_type which should propagate uris from parent types

cmungall: See also https://github.com/linkml/linkml/issues/794

PR #781, opened 2022-05-17
add the link to the range of a slot

### LINKED ISSUE #778: add link to the range of a slot in Jinja based the schema documentation
At the moment the range class of a slot is not linked to in the schema documentation. I guess, it has to do with:
https://github.com/linkml/linkml/blob/1bbf442f5c0dab5b6a4eb3309ef25b95c74d0892/linkml/generators/docgen/slot.md.jinja2#L25

where it should actually be `{{gen.link(element.range)}}`. If I'm correct, I could attempt a PR.

StroemPhi:
adresses #778

This PR alters the default Jinja template for the slot documentation by adding a link to the documentation of the class specified in the range of the slot.

### DIFF
--- a/linkml/generators/docgen/slot.md.jinja2
+++ b/linkml/generators/docgen/slot.md.jinja2
@@ -22,7 +22,7 @@ URI: [{{ gen.uri(element) }}]({{ gen.uri(element) }})
 
 ## Properties
 
- * Range: {{element.range}}
+ * Range: {{gen.link(element.range)}}
 
 {% if schemaview.usage_index().get(element.name) %}
 | used by | used in | type | used |


StroemPhi:
I just realized that this is related to https://github.com/linkml/linkml-project-template/issues/15, and that my commit won't work if the range class is one of the built in datatypes. 
For this there should be some conditional code that checks first if the range class is a built in data type and link to that if true. From what I can see this should be done in https://github.com/linkml/linkml/blob/main/linkml/generators/docgen.py#L251_L252. But I'm an autodidact coding noob, so it would probably take me a while to propose something useful and not ugly. Any pointers in the right direction are welcome.

cmungall:
Thanks for the PR!

As you note the problem you bring up is not with your PR itself but rather with gendoc not doing a good job of separating internal and external links.

I made a separate issue for this:

* #782.  

I don't think we should block your PR on solving this more generally. However, we could change the behavior of gendoc.link to first check if this is one of the files that is generated, and switch to using the URI if not.

cmungall (approved):

cmungall:
I know this is draft, but in fact separate PRs have rendered any reasons to delay moot, so this is ready to go!

StroemPhi:
@cmungall I'm a bit confused as this PR has been merged. Is this still a relevant issue? And if yes, should I better open a new issue then?

Issue #791, opened 2022-05-18
Add toggle option to gen-doc to accommodate term ids in generated URLs


sujaypatil96:
Use case in MIxS:

Currently gen-doc uses the _name_ label to encode URLs. For MIxS we want to route to slots via the URL using the MIxS term ID. So as part of this issue, we can add some functionality to gen-doc that can allow users to toggle between using _name_ and _term IDs_.

A cursory review of the docgen modules, I would think the course of action would be to tweak this block of code: https://github.com/linkml/linkml/blob/main/linkml/generators/docgen.py#L121-L128

Specifically, instead of this:

```
n = self.name(s)
```

We need to find a way to read in the term ID, which I assume will be in _id_ part of the slot? And pass that to the `_write()` method so that it creates MD files with the filenames using term ID instead of slot names.

Edit: It looks like there's no method to retrieve the term ids, like name, so we're going to have to custom define one.

CC: @cmungall

PR #792, opened 2022-05-18
Add `use_slot_uris` argument which you can use in gen-doc jinja templates

sujaypatil96:
This PR seeks to resolve issue #791.

The way in which you can enforce the usage of term_ids in URLs is by setting `term_id=True`.


cmungall (changes requested):
Logic is good, but "term" isn't a term in linkml, instead call the parameter `use_slot_uris`

cmungall:
This also leaves open to use_class_uris in future (not a MIxS reqt yet?)

sujaypatil96:
Hm, yes, that's true. It does let us extrapolate the same functionality to be applicable to classes by exposing a use_class_uris argument. But it isn't a MIxS requirement just yet.

linkml/generators/docgen.py
```diff
@@ -121,7 +124,7 @@ def serialize(self, directory: str = None) -> None:
         for sn, s in sv.all_slots().items():
             if self._is_external(s):
                 continue
-            n = self.name(s)
+            n = self.name(s, use_slot_uris=self.use_slot_uris)
```
cmungall, line 127: the logic here is odd. use_slot_uris is already an instance attribute on the Generator

cmungall (approved):

PR #804, opened 2022-05-27
removed pandas and ran `poetry update` fixes #506

### LINKED ISSUE #506: linkml shouldn't import pandas

hrshdhgd:


sierra-moxon:
thanks Harshad :) - I've been thinking about doing this as well, its such a big library.

Issue #816, opened 2022-06-01
linkml-sqldb dump doesn't want to instantiate datetimes from TSV from Postgres into SQLite


turbomam:
The offending type in Postgres is `timestamp(6)`, like `2022-01-13T20:03:48.999191`

I scrubbed some of the inserted values in the very last line of the stack trace.

```
Traceback (most recent call last):
  File "/Users/MAM/Library/Caches/pypoetry/virtualenvs/linkml-project-template-d6UeGt0V-py3.9/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1719, in _execute_context
    context = constructor(
  File "/Users/MAM/Library/Caches/pypoetry/virtualenvs/linkml-project-template-d6UeGt0V-py3.9/lib/python3.9/site-packages/sqlalchemy/engine/default.py", line 1070, in _init_compiled
    param = [
  File "/Users/MAM/Library/Caches/pypoetry/virtualenvs/linkml-project-template-d6UeGt0V-py3.9/lib/python3.9/site-packages/sqlalchemy/engine/default.py", line 1071, in <listcomp>
    processors[key](compiled_params[key])
  File "/Users/MAM/Library/Caches/pypoetry/virtualenvs/linkml-project-template-d6UeGt0V-py3.9/lib/python3.9/site-packages/sqlalchemy/dialects/sqlite/base.py", line 1003, in process
    raise TypeError(
TypeError: SQLite DateTime type only accepts Python datetime and date objects as input.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/MAM/Library/Caches/pypoetry/virtualenvs/linkml-project-template-d6UeGt0V-py3.9/bin/linkml-sqldb", line 8, in <module>
    sys.exit(main())
  File "/Users/MAM/Library/Caches/pypoetry/virtualenvs/linkml-project-template-d6UeGt0V-py3.9/lib/python3.9/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/Users/MAM/Library/Caches/pypoetry/virtualenvs/linkml-project-template-d6UeGt0V-py3.9/lib/python3.9/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/MAM/Library/Caches/pypoetry/virtualenvs/linkml-project-template-d6UeGt0V-py3.9/lib/python3.9/site-packages/click/core.py", line 1657, in invoke
[...]

PR #819, opened 2022-06-02
Test for data, schema and Makefile for linkml-sqldb dump of datetimes

turbomam:
for #816 

I'm pushing some data, a schema and a Makefile that illustrate my problem with `linkml-sqldb dump` and dates

[`make all`](l/linkml/blob/issue-816-postgres-dates/tests/test_issues/input/issue_816/birthday_schema.Makefile) will convert a TSV with `datetime`s into YAML but `linkml-sqldb dump` doesn't work (unless you change the range of birthdate to `string`, or something like that.)


cmungall:
Closing in favor of #860

Issue #815, opened 2022-06-01
linkml-sqldb dump from TSV can't handle strings longer than 131072 characters


turbomam:
In the Celniker FELIX database (for the Synbio Ontology project), most of the nucleic acid sequences are shorter than 50k nt. But there is one 202,036 nt long sequence and I would like to add that to a SQLite database with LinkML validation.

```
Traceback (most recent call last):
  File "/Users/MAM/Library/Caches/pypoetry/virtualenvs/linkml-project-template-d6UeGt0V-py3.9/bin/linkml-sqldb", line 8, in <module>
    sys.exit(main())
  File "/Users/MAM/Library/Caches/pypoetry/virtualenvs/linkml-project-template-d6UeGt0V-py3.9/lib/python3.9/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/Users/MAM/Library/Caches/pypoetry/virtualenvs/linkml-project-template-d6UeGt0V-py3.9/lib/python3.9/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/MAM/Library/Caches/pypoetry/virtualenvs/linkml-project-template-d6UeGt0V-py3.9/lib/python3.9/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/MAM/Library/Caches/pypoetry/virtualenvs/linkml-project-template-d6UeGt0V-py3.9/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/MAM/Library/Caches/pypoetry/virtualenvs/linkml-project-template-d6UeGt0V-py3.9/lib/python3.9/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/Users/MAM/Library/Caches/pypoetry/virtualenvs/linkml-project-template-d6UeGt0V-py3.9/lib/python3.9/site-packages/linkml/utils/sqlutils.py", line 325, in dump
    obj = loader.load(source=input,  target_class=py_target_class, **inargs)
  File "/Users/MAM/Library/Caches/pypoetry/virtualenvs/linkml-project-template-d6UeGt0V-py3.9/lib/python3.9/site-packages/linkml_runtime/loaders/csv_loader.py", line 43, in load
    objs = unflatten_from_csv(source, config=config, **kwargs)
[...]

PR #822, opened 2022-06-02
Test for LinkML/Python handling of long strings in TSV

turbomam:
[schema, data and Makefile](https://github.com/linkml/linkml/tree/issue-815-databasing-long-strings/tests/test_issues/input/issue_815) for showing that `linkml-sqldb dump` fails with TSV input containing strings longer than 131072 characters

as @cmungall observed, loading the same data from YAML does work

from #815


cmungall:
@turbomam can this be closed?

I think this is replaced by #859 which uses a more conventional testing strategy

turbomam:
sure

PR #838, opened 2022-06-15
Extend Pydantic from BaseModel

kevinschaper:
Rather than merge or rebase, I wanted to start in a fresh branch. The difference from the old one is that brought the PydanticConfig in in the form of kwargs when extending from BaseModel, and one of the tests is now raising a ValidationError rather than a TypeError


kevinschaper:
Here's the error it's hitting now: 

```p1 = mod.Person(id='P1', name='a b', age_in_years=22)
[2682](https://github.com/linkml/linkml/runs/6910057666?check_suite_focus=true#step:9:2683)
  File "<string>", line 2, in __init__
[2683](https://github.com/linkml/linkml/runs/6910057666?check_suite_focus=true#step:9:2684)
  File "/home/runner/work/linkml/linkml/.venv/lib/python3.8/site-packages/sqlalchemy/orm/instrumentation.py", line 451, in _new_state_if_none
[2684](https://github.com/linkml/linkml/runs/6910057666?check_suite_focus=true#step:9:2685)
    self._state_setter(instance, state)
[2685](https://github.com/linkml/linkml/runs/6910057666?check_suite_focus=true#step:9:2686)
  File "<string>", line 1, in set
[2686](https://github.com/linkml/linkml/runs/6910057666?check_suite_focus=true#step:9:2687)
  File "pydantic/main.py", line 401, in pydantic.main.BaseModel.__setattr__
[2687](https://github.com/linkml/linkml/runs/6910057666?check_suite_focus=true#step:9:2688)
AttributeError: __fields_set__```

The [one reference](https://stackoverflow.com/questions/60158357/why-do-i-get-attributeerror-fields-set-when-subclassing-a-pydantic-basemo) I can see to this `__fields_set__` error is from a person who extended from BaseModel *and* used the pydantic.dataclass, but as far as I can tell that isn't happening here.

**Comment by kevinschaper** [comment]
For some reason, if I add the pydantic config kwargs directly to my BaseModel with `__slots__ = '__weakref__', I get a complaint, but if I add the pydantic config either to the classes themselves, or add another class in the middle of the inheritance, like this, it's fine:

```
class IntermediateBaseModel(PydanticBaseModel):
    __slots__ = '__weakref__'
    
class BaseModel(IntermediateBaseModel,
                validate_assignment = True, 
                validate_all = True, 
                underscore_attrs_are_private = True, 
                extra = {% if allow_extra %}'allow'{% else %}'forbid'{% endif %}, 
                arbitrary_types_allowed = True):
    pass                    
```

I'm not sure if it's less ugly, but I might try putting the weakref fix into the classes themselves rather than do this

File: tests/test_generators/output/kitchen_sink_pydantic.py [hunk 1]
```diff
@@ -2,21 +2,21 @@
...
-
-    validate_assignment = True
-    validate_all = True
-    underscore_attrs_are_private = True
-    extra = 'forbid'
-    arbitrary_types_allowed = True  # TODO re-evaluate this
+# class IntermediateBaseModel(PydanticBaseModel):
+#    __slots__ = '__weakref__'
+    
+class BaseModel(PydanticBaseModel,
```
cmungall, line 13: Normally I would expect AdjectiveFoo is_a Foo, not Foo is_a AdjectiveFoo

Also: won't think cause issues if we have multiple yaml files - each one makes a fresh pydantic model each with a base class BaseModel. Formally these are separated and users can do from foo import BaseModel as FooBaseModel, but this is a bit awkward

Or maybe there is no need to import this and it's something that just exists as a convenience?

cmungall: an alternative is the pydantic analog of YAMLRoot in linkml-runtime, this could be an option, but I like the ability to make things purely standalone

kevinschaper: It's awkward because I seem to need these two shim layers. Renaming the import from BaseModel to PydanticBasemodel is just copied and pasted from the solution I found, but because we're generating all of the classes, it's not really necessary to preserve the name BaseModel. I could try bundling up the configuration kwargs to make that a little less ugly, and get down to just the IntermediateBaseModel, which could be maybe renamed to BaseModelShim?

I feel like the BaseModel probably isn't the only piece of inheritance that would break if generating a schema from multiple yaml files (we would need to do imports for any inheritance across yaml files, right?), maybe it would make more sense to have the generator take a collection of yaml files?

cmungall (approved):

kevinschaper:
@cmungall I updated the class names to at least be a little more descriptive.

PR #853, opened 2022-06-29
require primary id field to be not null

sierra-moxon:


cmungall (changes requested):
should it not be

Nullable IF slot is none OR (NOT slot.required AND NOT is_pk)?

Also wondering what the implications are of forcing this - maybe we can have a test for this?

Some background:

SQL Alchemy requires every table to have a primary key (simple or compound), hence the original logic in this code: if there is no identifier then make the tuple of all fields the identifier.

AFAICR the SQL spec does not expressly forbid NULLs in compound keys -- but different DBMSs handle this case differently?

PR #863, opened 2022-07-11
Add missing link to tutorial10

dalito:


dalito:
Tutorial10 is the same as 09.

PR #879, opened 2022-07-22
Add note on how to clean up after running tests

dalito:
Partly solves #759


cmungall (approved):
Much appreciated! And we really need a good longer term fix for this!

Issue #898, opened 2022-08-09
Use code formatter to apply consistent code layout


ialarmedalien:
Suggestion: use a code formatter such as black and a linter like flake8 to apply consistent formatting to the source code of the linkml repos. Looking at the source code, I notice there's no standard for using `'` vs `"` (both are present in the same files), line indentation for follow-on lines varies, data structures are laid out differently, blah blah blah... 

These are fairly petty concerns, but using a code formatter and linter will keep the codebase looking uniform and consistent and make it easier to read for newcomers, plus it is easy enough to add black and flake8 (or whatever options you prefer) as a pre-commit/pre-push hook to keep the codebase tidy.

cmungall:
great idea - I think we had a PR to do this before but it got stale quickly

@hrshdhgd / @sujaypatil96 I suggest we don't try and do this all in one shot, let's iterate, and do this in a way that won't break all the open PRs. I suggest first making the tox and making it optional to conform, then we will make some PRs to reformat, and we will give open PRs a certain amount of time to rebase and apply before making it strict

We also have to be aware that linkml-runtime contains autogen'd code that does not conform. Maybe we want to add linting as an optional step in python/pydanticgen

hrshdhgd:
I have implemented listing by `black` and `isort` in PR #902 . It is a 1 commit PR which only formats the code. I will implement `flake8` and `mypy` in future PRs since they'll be time consuming to implement.

sujaypatil96:
`black` code formatting has been applied retroactively on the linkml library till a point in time, and no code committed after that conforms to the formatting standard. GitHub Actions needs to be capable of checking for conformance to black formatting standard. Close this issue out, and make a new issue for adding this capability to GitHub Actions?

CC: @hrshdhgd

hrshdhgd:
[This line ](https://github.com/linkml/linkml/blob/06134b13acc9cf8d64dc006d46475804765770c7/.github/workflows/main.yaml#L70) in GH actions should take care of it.

dalito:
If I see it right the changes made by the linters in GH actions are not committed to the repo so will be lost.

In other projects the linting/code-formatting is quite often solved with pre-commit. That works well in my experience.

dalito:
linkml uses ruff/black and also has a pre-commit configuration now. Maybe there could be even more ruff rules enabled but the general setup is in place. So I close this.

PR #902, opened 2022-08-10
Linting using black and isort

hrshdhgd:
Adresses #898 

 - [x] Lint using `black`
 - [x] Lint using `isort`


dalito:
Some comments:
* If someone else looks add the changes: tox.ini was change in the huge commit (last file listed)
* Instead of adding tox to run the linter, you could also use pre-commit and create a new CI job. As inspiration you could look at [pint](https://github.com/hgrecco/pint/blob/master/.pre-commit-config.yaml).
* You have to add some configuration to make black compatible with isort, see [isort manual](https://pycqa.github.io/isort/docs/configuration/black_compatibility.html).
* Neither ìsort` nor `black` have been added to `pyproject.toml` so far.

cmungall (approved):

cmungall:
Putting on agenda for linkml call

@dalito - good suggestions. I would like to ultimately align with the workflow we have in https://github.com/INCATools/ontology-access-kit/

PR #910, opened 2022-08-13
Adding support for enums in docgen.

cmungall:
Example: 

https://cmungall.github.io/linkml-phenopackets/HPOAbnormalityTerms/

image


sujaypatil96:
Looks good @cmungall. Looks like this PR is adding functionality to render tables on enum detail pages only if the enum has permissible values. Else indicate that it is a dynamic enum.

sujaypatil96 (approved):

PR #912, opened 2022-08-16
Add linkml-lint command

### LINKED ISSUE #909: Rename schema_fixer to linter and add additional checks and repairs
We have a class SchemaFixer:

https://github.com/linkml/linkml/blob/596bedfb0b5a3abd611fa627aaff04c3ae19510b/linkml/utils/schema_fixer.py#L20-L23

This should be renamed to something like Linter

We may want to move it out of utils, as utils is becoming a big grab-bag

Some additional functionality:

- report on missing descriptions
- naming conventions and style checks
     - CamelCase for classes
     - snake_case for slots
     - CamelCase for enums
     - ALL_CAPS or snake_case for PVs

We may want to have a mechanism to suppress this where deviation is intentional, e.g.

```
name: biolink
annotations:
   schema.style: TitlesAsNames
```

Not everything can be repaired by the linter - e.g missing descriptions. However, schema-automator can suggest missing descriptions (by looking up entities with the same IRI). So there is an argument for having the linter live in SA, but I think the downsides outweigh this

### LINKED ISSUE #916: linter should warn against xsd:int
I can't think of a situation where someone would use xsd:int over xsd:integer. It should be formally allowed but it's 99% likely it's a mistake

See: https://github.com/microbiomedata/nmdc-schema/pull/296

pkalita-lbl:
Fixes #909 
Fixes #916


cmungall (approved):

pkalita-lbl:
I think this is a good first cut at a schema linter. A little overview of what's going on here:

There is a new package called `linkml.linter`. Within that package:

* `cli.py` defines the `click` command line function. It is responsible for locating a configuration file (using default file names if not passed as a option), constructing `Linter` and `Formatter` objects, and running the linter on the specified schema(s).
* `linter.py` defines the `Linter` class which is responsible for constructing various subclasses of `LinterRule` and running the rules' `check` method against a provided schema.
* `rules.py` defines the abstract base class `LinterRule` and concrete subclasses which implement the checking and fixing logic for each rule. Each subclass must provide an `id` property, which is how the rule is identified in a configuration file and a `check` method which accepts a `SchemaView` object and a boolean indicating whether to attempt to fix issues (not all rules are capable of using the `fix` parameter). 
* `formatters` this package contains the implementations for the different report formats
* `config` this package contains a LinkML schema describing a linter configuration file, the Python code generated from that schema, and `default.yaml` which defines the default values for each slot. By default all rules are disabled (`level: disabled`). When a user provides a configuration file it is loaded and the result deeply merged with `default.yaml` to provide the final configuration that will be passed to the `Linter` and `LinterRule` objects. This package also contains a `.linkmllint.yaml` configuration file which can be used to lint the configuration LinkML model itself.

The general outline of how to use this would be something like this: say a project has a directory of LinkML schema files, like so:
```
my-project/
├─ schemas/
│  ├─ core.yaml
│  ├─ extras.yaml
│  ├─ types.yaml
├─ README.md
```
The project can decide on which rules to opt-in to and how to configure them by creating a `.linkmllint.yaml` file at the root of the project. The contents might look like this:
```yaml
# .linkmllint.yaml
no_xsd_int_type:
  level: error
standard_naming:
  level: warning
tree_root_class:
  level: error
  root_class_name: Things
  validate_existing_class_name: true
```
Now (assuming `linkml` is a dependency of the project) the linter can be run by:
```shell
linkml-lint schemas
```
This relies on `.linkmllint.yaml` being recognized as a default configuration file name. If multiple configurations need to be managed the `-c/--config` command line option can be used.

Follow-on work:
* Review currently implemented rule names and options. Decide if any changes are needed.
* See if there are any obvious rules that should be added now.
* Having all rules disabled by default is the most conservative option, do we want to be more prescriptive and enable some by default? (Users could still turn them off in their custom configuration file)
* Document usage in main LinkML docs?

PR #913, opened 2022-08-16
Enums on FHIR

joeflack4:
Link to model mappings: https://docs.google.com/spreadsheets/d/1Y35SpVv09phns3v6QZ5mlzf8NfXGzSU1vcodo4TR8yQ/edit#gid=1898715155

Updates
```
Feature: Enums-on-FHIR
- Description: Conversion of LinkML Enum to FHIR ValueSet.
- Add: Generator file
- Add: Test file
- Add: Test input file taken from linkml-model: dynamic-enums-example.yaml
- Add: linkml/generators/fhirgen/: Initialized w/ mapping
- Add: mapping__enum_valueset.csv

WIP
```

### DIFF
--- a/linkml/generators/fhirgen.py
+++ b/linkml/generators/fhirgen.py
@@ -1,0 +1,396 @@
+"""fhirgen: FHIR-related generators
+
+Currently, only supports Enum -> ValueSet
+
+Temp development notes
+  - Mapping path: /Users/joeflack4/projects/linkml/linkml/generators/fhirgen/mapping__enum_valueset.csv
+
+Resources
+ 1. Enum def: https://linkml.io/linkml-model/docs/EnumDefinition/
+ 2. ValueSet def: https://www.hl7.org/fhir/valueset.html
+ 3. Issue:https://github.com/linkml/linkml/issues/905
+ 4. PR: https://github.com/linkml/linkml/pull/913
+ 5. Mapping sheet source:
+   https://docs.google.com/spreadsheets/d/1Y35SpVv09phns3v6QZ5mlzf8NfXGzSU1vcodo4TR8yQ/edit#gid=1898715155
+ 6. Link to old CCDH Enum/CodeSet model (if useful):
+   https://raw.githubusercontent.com/cancerDHC/ccdhmodel/main/model/schema/crdch_model.yaml
+
+TODO's
+ 1. R5: Need to see if it is different than R4 for ValueSet.
+ 2. Delete all stuff from file that I don't need
+ 3. Finish mapping__enum_valueset.csv
+   - Add FHIR's Nested objects: e.g. `identifier`
+   - 1st: Do all IMPT fhir fields
+   - 2nd: Do all the IMPT linkml fields (need to go to dynamic-enums-example.yaml, from line 78, and add all the fields
+     that appear in the file as 'IMPT'.
+ 4. Evaluate 'abritrary data model mapping' standards/formats/tools to potentially use over my ad-hoc CSV approach:
+   (i) https://github.com/cmungall/linkml-transformer (ii) https://docs.google.com/presentation/d/
+   1ctgT1IfwPjnFQO2Q0sYlM8qk0wiB2_32JyeKyN4Uf8k/edit?usp=drive_web&ouid=109884413503012482953 (iii) YARRRML (rdf only?)
+todo later's
+ 1. XML output
+"""
+import csv
+import json
+import os
+# from copy import deepcopy
+# from enum import Enum
+from pathlib import Path
+from typing import Any, Callable, Dict, List, TextIO, Union
+# from typing import Callable, Dict, Iterator, TextIO, Tuple, Union
+
+import click
+
+# from linkml_runtime.dumpers import yaml_dumper
+from linkml_runtime.linkml_model.meta import EnumDefinition, SchemaDefinition
+# from linkml_runtime.linkml_model.meta import ClassDefinition, ClassDefinitionName, Definition, DefinitionName, \
+#     Element, EnumDefinition, SchemaDefinition, SlotDefinition, SubsetDefinition, TypeDefinition
+# from linkml_runtime.utils.formatutils import camelcase, underscore
+from linkml_runtime.utils.schemaview import SchemaView
+# from linkml.utils.generator import Generator, shared_arguments
+# todo: Do I need _ensure_ranked? If so, move out of docgen?
+# from linkml.generators.docgen import _ensure_ranked
+from linkml.utils.generator import Generator, shared_arguments
+
+# todo: make some of these class properties?
+FHIRGEN_DIR = os.path.join(os.path.dirname(__file__), 'fhirgen')
+MODEL_MAPPING_PATH = os.path.join(FHIRGEN_DIR, 'mapping__enum_valueset.csv')
+SUPPORTED_FORMATS = ['json']  # todo: later: xml
+FHIR_VALUESET_FIELD_MUTATIONS: Dict[str, Callable] = {
+    'status': lambda x: 'unknown'  # todo: temp
+}
+
+
+# TODO: when rebasing, this will be a dataclass. outdir will become a standard dataclass obj for example. don't make
+#  ...any of these changes before the dataclass PR is merged. (https://github.com/linkml/linkml/pull/924)
+class FhirValueSetGenerator(Generator):
+    """Generate FHIR ValueSet JSON from a LinkML YAML Enum schema"""
+    generatorname = os.path.basename(__file__)
+    generatorversion = "0.0.1"
+    valid_formats = SUPPORTED_FORMATS
+
+    def __init__(
+        self,
+        schema: Union[str, TextIO, SchemaDefinition],
+        outdir: str = None,
+        format: str = None,
+    ) -> None:
+        """
+        Creates a generator object that can write output
+
+        :param schema: path to schema file or schema object
+        :param outdir: path to directory save file or file(s)
+        :param genmeta:
+        """
+        self.sourcefile = schema
+        self.schemaview = SchemaView(schema)
+        self.schema = self.schemaview.schema
+        self.outdir = outdir
+        self.format = format
+
+    def serialize(self, save: bool = True, outdir: str = None, save_format='json') -> List[Dict]:
+        """
+        Serialize Enums in schema to JSON
+
+        :param save: if True, saves to disk.
+        :param outdir: path to directory save file or file(s)
+        :param save_format: Format to save.
+        :return: List of enums as dictionaries.
+        """
+        if save_format not in SUPPORTED_FORMATS:
+            raise ValueError(f'Format {save_format} not supported. Use one of: {",".join(SUPPORTED_FORMATS)}')
+        _outdir = outdir if outdir else self.outdir
+        if save and not _outdir:
+            raise RuntimeError('Tried to save, but did not supply `outdir`.')
+        sv = self.schemaview
+        linkml_dict_enums: Dict[str, EnumDefinition] = sv.all_enums()
+        fhir_dict_enums = []
+        for en, e in linkml_dict_enums.items():
+            d: Dict = self.serialize_enum(en, e)
+            fhir_dict_enums.append(fhir_dict_enums)
+            if save and save_format == 'json':
+                out_str: str = json.dumps(d)
+                self._write(out_str=out_str, outdir=_outdir, name=en, format=save_format)
+        return fhir_dict_enums
+
+
+    @staticmethod
+    def model_mappings(path: str = MODEL_MAPPING_PATH) -> Dict[str, str]:
+        """Get model mappings from CSV"""
+        with open(path) as csvfile:
+            reader = csv.reader(csvfile, delimiter=',')
+            rows = [x for x in reader][1:]  # don't need header
+        d = {
+            row[0]: row[1]  # easiest way; assumes linkml is always col 1, and fhir is col 2
+            for row in rows
+        }
+        d = {k: v for k, v in d.items() if k}  # removes anything left over from empty CSV rows
+
+        return d
+
+    # TODO: Most of the work should go here
+    # todo: Since I want to return a dict, should I call this dictify instead of serialize?
+    def serialize_enum(self, enum_name: str, enum: EnumDefinition) -> Dict:
+        """
+        Serialize Enums in schema to JSON
+
+        todo: use enum_name if needed
+        :param enum_name: The enum name
+        :param enum: The enum
+        :return:
+        """
+        d = {}
+        mappings: Dict[str, str] = self.model_mappings()
+
+        for linkml_field, target_field in mappings.items():
+            linkml_val: Any = getattr(enum, linkml_field)
+            if not linkml_val:
+                continue
+            # TODO: Add handling of nested vals
+            d[target_field] = linkml_val
+            if linkml_field in FHIR_VALUESET_FIELD_MUTATIONS:
+                d[target_field] = FHIR_VALUESET_FIELD_MUTATIONS[linkml_field](linkml_val)
+
+        return d
+
+    def _write(self, out_str: str, outdir: str, name: str, format: str = None) -> None:
+        """
+        Writes to disk
+
+        :param out_str: string to be written
+        :param outdir: path to directory save file or file(s)
+        :param name: base name - should correspond to element name
+        :param format: The data format
+        # todo: Move `save` params in here, expose to user, and add json.dump() here?
+        """
+        if not format:
+            format = self.format
+        Path(outdir).mkdir(parents=True, exist_ok=True)
+        file_name = f"{name}.{format}"
+        outpath = os.path.join(outdir, file_name)
+        with open(outpath, "w", encoding="UTF-8") as stream:
+            stream.write(out_str)
+
+    # def schema_title(self) -> str:
+    #     """
+    #     Returns the title of the schema.
+    #
+    #     Uses title field if present, otherwise name
+    #
+    #     :return:
+    #     """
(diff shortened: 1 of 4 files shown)


linkml/generators/fhirgen.py
```diff
@@ -0,0 +1,756 @@
+"""fhirgen: FHIR-related generators
```
joeflack4, line 1: Name of file: I called it `fhirgen.py`, but I only intend to do Enum->ValueSet. I can rename it to something more specific if desired.

linkml/generators/fhirgen.py
```diff
@@ -0,0 +1,380 @@
...
```
joeflack4, line 1: ### Files modified when doing `poetry install`?
@hrshdhgd Hey Harshad. If you don't mind, I'd like to lean on you for support a little bit. I have a question. I did `poetry install`, and I noticed that lots of tests ran, and they changed many output test files. Is this supposed to happen? I think all the tests passed. Should I commit these files?

joeflack4: @hrshdhgd never mind, i think this was the result of me running a docgen test

linkml/generators/fhirgen/mapping__enum_valueset.csv
```diff
@@ -0,0 +1 @@
+enum	valueset
```
joeflack4, line 1: I hope I don't create any ire w/ this pattern. This has been a highly successful model mapping pattern to me. But it seems a bit redundant/recursive. Doesn't LinkML itself do model mappings in yml? Yet, I vastly prefer doing this in a tabular file (TSV / CSV). Could I do such a mapping in SchemaView? Ought those best be done in GoogleSheets and either pulled down from the cloud, and/or saved/cached as CSV/TSV?

linkml/generators/fhirgen.py
```diff
@@ -0,0 +1,388 @@
...
+
+Currently, only supports Enum -> ValueSet
+
+Resources
+ 1. Enum def: https://linkml.io/linkml-model/docs/EnumDefinition/
+ 2. ValueSet def: https://www.hl7.org/fhir/valueset.html
+ 3. Issue:https://github.com/linkml/linkml/issues/905
+ 4. PR: https://github.com/linkml/linkml/pull/913
+
+TODO's
```
joeflack4, line 11: FYI: My main list of TODOs

linkml/generators/fhirgen.py
```diff
@@ -0,0 +1,388 @@
...
+        # todo: `en` and `n` below are redundant?
+        for en, e in enums.items():
+            n = self.name(e)
+            d: Dict = self.serialize_enum(n, e)
+            out_str: str = json.dumps(d)
+            if outpath:
+                self._write(out_str, outpath, n)
+
+    @staticmethod
+    def model_mappings(path: str = MODEL_MAPPING_PATH) -> Dict[str, str]:
```
joeflack4, line 96: Would like to use `pandas`, but it's not in the project requirements, so I used the standard library instead.

joeflack4:
@hrshdhgd Tagged you for review, but this is still heavily in the draft stages.

linkml/generators/fhirgen.py
```diff
@@ -0,0 +1,393 @@
...
+
+Resources
+ 1. Enum def: https://linkml.io/linkml-model/docs/EnumDefinition/
+ 2. ValueSet def: https://www.hl7.org/fhir/valueset.html
+ 3. Issue:https://github.com/linkml/linkml/issues/905
+ 4. PR: https://github.com/linkml/linkml/pull/913
+
+TODO's
+ 1. R5: Need to see if it is different than R4 for ValueSet.
+ 2. Delete all stuff from file that I don't need
```
joeflack4, line 13: There's still a lot of stuff in here from `docgen` that I don't need.

linkml/generators/fhirgen.py
```diff
@@ -0,0 +1,393 @@
...
+    Element, EnumDefinition, SchemaDefinition, SlotDefinition, SubsetDefinition, TypeDefinition
+from linkml_runtime.utils.formatutils import camelcase, underscore
+from linkml_runtime.utils.schemaview import SchemaView
+# from linkml.utils.generator import Generator, shared_arguments
+# todo: Do I need _ensure_ranked? If so, move out of docgen?
+from linkml.generators.docgen import _ensure_ranked
+from linkml.utils.generator import Generator
+
+
+# todo: make some of these class properties?
```
joeflack4, line 42: Haven't decided yet what all will be class/obj properties, class/static/normal methods or just top level variables.

linkml/generators/fhirgen.py
```diff
@@ -0,0 +1,393 @@
...
+FHIRGEN_DIR = os.path.join(os.path.dirname(__file__), 'fhirgen')
+MODEL_MAPPING_PATH = os.path.join(FHIRGEN_DIR, 'mapping__enum_valueset.csv')
+FHIR_VALUESET_FIELD_MUTATIONS: Dict[str, Callable] = {
```
joeflack4, line 45: FYI: This is how I handle mutations from one model field to the other, if needed. If it's complex enough, I'll use a function instead of a lambda.

linkml/generators/fhirgen.py
```diff
@@ -0,0 +1,393 @@
...
+        Serialize Enums in schema to JSON
+
+        :param directory: relative or absolute path to directory in which documents are to be written
+        :param outpath: path to save file or file(s)
+        :return:
+        """
+        sv = self.schemaview
+        enums: Dict[str, EnumDefinition] = sv.all_enums()
+        # TODO: This is super temporary; has to do with unit test and not all enums being well formed:
+        if 'EmploymentEventType' in enums:
```
joeflack4, line 88: I'm using `kitchen_sink.yml` for my test, but not all the Enums in it are well-formed yet.

linkml/generators/fhirgen.py
```diff
@@ -0,0 +1,393 @@
...
+            row[0]: row[1]  # easiest way; assumes linkml is always col 1, and fhir is col 2
+            for row in rows
+        }
+        d = {k: v for k, v in d.items() if k}  # removes anything left over from empty CSV rows
+
+        return d
+
+    # TODO: Most of the work should go here
+    # todo: Since I want to return a dict, should I call this dictify instead of serialize?
+    def serialize_enum(self, enum_name: str, enum: EnumDefinition) -> Dict:
```
joeflack4, line 114: FYI: This is where most of the work will go, apart from the mappings within the CSV.

linkml/generators/fhirgen.py
```diff
@@ -0,0 +1,393 @@
...
+            linkml_val = getattr(enum, linkml_field)
+            if linkml_field in FHIR_VALUESET_FIELD_MUTATIONS:
+                fhir_val = FHIR_VALUESET_FIELD_MUTATIONS[linkml_field](linkml_val)
+            else:
+                fhir_val = linkml_val
+            d[fhir_field] = fhir_val
+
+        return d
+
+    def _write(self, out_str: str, outpath: str, name: str) -> None:
```
joeflack4, line 138: Will probably change this to just do `json.dump()`.

linkml/generators/fhirgen.py
```diff
@@ -0,0 +1,393 @@
...
+
+        :return:
+        """
+        s = self.schemaview.schema
+        if s.title:
+            return s.title
+        else:
+            return s.name
+
+    def _file_suffix(self):
```
joeflack4, line 168: Not sure if I should keep this or not, since I'm only doing `json` now, and both `json` and `xml` have the same spellings for their file extensions.

linkml/generators/fhirgen.py
```diff
@@ -0,0 +1,393 @@
...
+                reverse=True,
+            )
+            for child in children:
+                # depth first - place at end of stack (to be processed next)
+                stack.append((depth + 1, child))
+
+
+# todo: Create an output file or just print to stdout as it is now?
+# TODO: test is generating an error: generator.decorator: default=g.valid_formats[0],
+# @shared_arguments(FhirValueSetGenerator)
```
joeflack4, line 371: Not sure why, but when I ran my test, I got an error error in the `Generator` super class, in the `decorator` method, on `default=g.valid_formats[0],` (`valid_formats` was empty)).

linkml/generators/fhirgen.py
```diff
@@ -0,0 +1,393 @@
...
+def cli(yamlfile, outpath, **args):
+    """Generate FHIR ValueSet JSON from a LinkML YAML Enum schema"""
+    # gen = FhirValueSetGenerator(
+    FhirValueSetGenerator(
+        yamlfile,
+        outpath=outpath,
+        **args,
+    )
+    # todo: print() was from docgen. Do I want to print to stdout? What if multiple files?
+    # print(gen.serialize())
```
joeflack4, line 389: @hrshdhgd I saw `print(gen.serialize())` in `docgen`. Is there a reason for it? I was thinking it could be useful to print the JSON file to `stdout`, but I think this is not a good idea for reasons: (i) multiple files being generated at once, (ii) only useful when using LinkML in the CLI to do something like `linkml-fhirgen myLinkml.yml > myFhir.json`, but not useful when using the Python API.

tests/test_generators/test_fhirgen.py
```diff
@@ -0,0 +1,42 @@
+"""Tests for: fhirgen: FHIR-related generators"""
+import unittest
+
+from linkml.generators.fhirgen import FhirValueSetGenerator
+from tests.test_generators.environment import env
+
+
+# TODO: Where to put output JSON?
```
joeflack4, line 8: @hrshdhgd I haven't looked at where I should put my outputs yet, but I think I can figure that out easily.

linkml/generators/fhirgen.py
cmungall: you may want to consider copying examples over from https://github.com/linkml/linkml-model/blob/main/examples/dynamic-enums-example.yaml

linkml/generators/fhirgen.py
cmungall: is there an equivalent of the python linkml stack for fhir? e.g. where you can instantiate objects and let the framework take care of serialization

linkml/generators/fhirgen.py
cmungall: OK, this is one of the confusing things about the old-school visitor-pattern based generators

check this:

https://github.com/linkml/linkml/blob/596bedfb0b5a3abd611fa627aaff04c3ae19510b/linkml/utils/generator.py#L144-L145

Yes, that is kind of horrible

for the new style schemaview-based generators there should be no need for this malarkey

linkml/generators/fhirgen.py
joeflack4: Very useful! Thanks.

linkml/generators/fhirgen.py
joeflack4: Not that I'm aware of yet; not like a general purpose library.

This Smart-on-FHIR (mainly something for web apps) client is the closest I've seen that has such models. Here's its [ValueSet](https://github.com/smart-on-fhir/client-py/blob/master/fhirclient/models/valueset.py). I think I might have seen Harold work on something familiar. Not sure how well this is maintained. Anyway looking at that particular example, it looks like it accepts a `jsondict`, not named args. And I don't see how it's setting values from that dict; looks like somewhere in the superclass calls.

Maybe something I can ask on the FHIR Zulip channel, if there's any more general purpose FHIR model library.

linkml/generators/fhirgen.py
joeflack4: Ahh, I see. After stripping away `docgen` of what I don't need, it doesn't look like schemaview is being used by the serialization or write functions. I probably will strip references to it away entirely. But if there's something I about it that I ought to use, let me know.

linkml/generators/fhirgen/mapping__enum_valueset.csv
hrshdhgd: You'll have to enlighten me with how this works. Unless @cmungall has an opinion about it should be before we decide on this.

linkml/generators/fhirgen.py
hrshdhgd: Yes, pandas is a heavyweight package hence while developing `linkml` the team made a decision not to rely on it and use lightweight alternatives.

linkml/generators/fhirgen.py
hrshdhgd: lambda should be fine.

linkml/generators/fhirgen.py
```diff
@@ -0,0 +1,395 @@
...
+ 4. PR: https://github.com/linkml/linkml/pull/913
+
+TODO's
+ 1. R5: Need to see if it is different than R4 for ValueSet.
+ 2. Delete all stuff from file that I don't need
+ 3. Finish mapping__enum_valueset.csv
+   - Nested objects: e.g. FHIR's `identifier`
+   - Mutation dict of func/lambda's
+   - Do all of the 1..x cardinalities in FHIR first
+ 4. Evaluate 'abritrary data model mapping' standards/formats/tools to potentially use over my ad-hoc CSV approach:
```
joeflack4, line 18: @cmungall I looked through the code and examples of `linkml-transformer` (and I've looked at the presentation a few times as well), and it still feels like something is missing with respect to arbitrary data model mappings. I intend to continue going down my route, while also periodically evaluate other options further.

tests/test_generators/input/dynamic-enums-example.yaml
```diff
@@ -0,0 +1,171 @@
+id: https://w3id.org/linkml/examples/enums
```
joeflack4, line 1: @cmungall Thanks for reminding me where this file is. I agree this is better to use. I copied it over. It's just a test input file, so that's probably alright, but if you'd like a DRYer approach where I import this from `linkml-model` somehow, let me know.

linkml/generators/fhirgen.py
```diff
@@ -0,0 +1,395 @@
...
+        :param genmeta:
+        """
+        self.sourcefile = schema
+        self.schemaview = SchemaView(schema)
+        self.schema = self.schemaview.schema
+        self.format = 'json'  # todo: support w/ CLI
+        self.outpath = outpath
+        self.genmeta = genmeta
+
+    def serialize(self, outpath: str = None) -> None:
```
joeflack4, line 82: @cmungall @hrshdhgd For now I'll keep the (1) initialize obj, (2) serialize & write pattern used by other generators. But my preference would be something like this:
1. Serialize to `dict` on object initialization. Can still be done in `serialize()`.
2. Have a `__new__` method that returns that `dict`
3. Have a non-private `write()` method to save to disk.

linkml/generators/fhirgen.py
joeflack4: @hrshdhgd Any idea on this? 
I copied the code from `docgen`. And its CLI has `@shared_arguments(DocGenerator)`, so I added `@shared_arguments(FhirValueSetGenerator)` to my code. But when I do this...

I get an error raised here: [`default=g.valid_formats[0]`](https://github.com/linkml/linkml/blob/main/linkml/utils/generator.py#L812)

because [`g.vaid_formats` is an empty list](https://github.com/linkml/linkml/blob/main/linkml/utils/generator.py#L36).

I wonder why this happens when I try to use `@shared_arguments()` for my new generator, but it works fine for docgen.

hrshdhgd: docgen has `valid_formats = ["markdown", "rst", "html", "latex"]` on line 100. Is there an equivalent for `fhirgen` ?

hrshdhgd: Almost all `*gen.py` has a value for `valid_formats` . I think declaring that in `fhirgen.py` may resolve your issue.

joeflack4: Fixed! thanks so much

linkml/generators/fhirgen.py
```diff
@@ -0,0 +1,396 @@
...
+        :param save_format: Format to save.
+        :return: List of enums as dictionaries.
+        """
+        if save_format not in SUPPORTED_FORMATS:
+            raise ValueError(f'Format {save_format} not supported. Use one of: {",".join(SUPPORTED_FORMATS)}')
+        _outdir = outdir if outdir else self.outdir
+        if save and not _outdir:
+            raise RuntimeError('Tried to save, but did not supply `outdir`.')
+        sv = self.schemaview
+        linkml_dict_enums: Dict[str, EnumDefinition] = sv.all_enums()
```
cmungall, line 105: no need to name a variable here, this is only used once

linkml/generators/fhirgen.py
```diff
@@ -0,0 +1,396 @@
...
+            d: Dict = self.serialize_enum(en, e)
+            fhir_dict_enums.append(fhir_dict_enums)
+            if save and save_format == 'json':
+                out_str: str = json.dumps(d)
+                self._write(out_str=out_str, outdir=_outdir, name=en, format=save_format)
+        return fhir_dict_enums
+
+
+    @staticmethod
+    def model_mappings(path: str = MODEL_MAPPING_PATH) -> Dict[str, str]:
```
cmungall, line 117: can you provide more details of what this is, and include an example in the unit test?

Also, it may be the case that this is not necessary - in theory the schema should provide all necessary information and mappings, no external out of band data should be required

joeflack4: Hey Chris. It seemed like this feature might have been a priority at the time, but since September it's been pretty far down the list of things needed for TIMS and TermHub. But eventually it will be useful for both.

This PR was heavily in flux. I was going to try out schema automator and some other stuff. It is in need of unit testing and a lot of other work.

If you want, we can close this one right now. I can always re open at any time.

Issue #923, opened 2022-09-01
Refactor generators


cmungall:
There are two issues with the current generators:

- they don't use dataclasses, which means it is hard to mentally reason about class variables, object attributes, and inheritance
- there is a mix of older visitor-style/SchemaLoader generators and newer non-visitor SchemaView classes

Neither of these affect the behavior or performance of the linkml stack. However, it does present a cognitive challenge to both new and existing developers

Some additional related minor issues:

- generators can parameterized both at the init level, and when passing args to `serialize`. While this flexibility is arguably nice for users of the object I think it actually makes it harder to reason about. This should be simplified
- the newer SV based generators intentionally don't use name mangling, but the difference can be confusing
- the older generators wrap calls with redirects of stdout, and use print statements, which redirect to a string. this is confusing
- the method of sharing click options via a decorator is a little confusing

1. [ ] Refactor generators to use dataclasses
     - some attributes like valid_formats must stay as ClassVars
2. [ ] Introduce intermediate subclasses for VistorGenerator and SimpleGenerator
     - VisitorGenerator would be an ABC with abstract methods for start/end_schema/class etc
     - VisitorGenerator would ultimately be deprecated, see below
     - SimpleGenerator would have some core wrapping to instantiate the SV object
3. [ ] Migrate all generators towards new framework
     - [ ] fully deprecate markdowngen once docgen has feature parity
     - [x] pydanticgen
     - [ ] jsonschemagen
     - [ ] owlgen
     - [x] sqltablegen (todo: fully remove the older sqlddlgen)
     - [x] javagen
     - [x] typescriptgen
     - [ ] graphqlgen
     - [ ] jsonldgen
     - [x] linkmlgen
     - [ ] shaclgen
     - [ ] shexgen
     - [ ] sssomgen

A lot of the old style visitor-based generators should be easy to move across as they don't intercept intermediate elements, and all logic is contained within end_schema, it should be a simple fix for these

sierra-moxon:
04/21/2023- we still have some generators to migrate to the schemaview method

cmungall:
This is still high priority. The schemaloader classes are hard to reason about, particularly when it comes to passing state such as `base_dir` between generators and schemaloader and `load_raw_schema`. This gets even more complicated as most of the methods are overly postelian in accepting a string, dict, object, or url.

Let's look at what happens when we create a `PythonGenerator` objects. This can be done as

```python
PythonGenerator(schema)
```

where schema is a string, url, dict, ...

[...]

PR #924, opened 2022-09-01
WIP: refactoring generators to use dataclasses. Part 1 of #923

cmungall:
- See #923

The general methodology here is:

- remove __init__
- ensure all args for __init__ are reflected in dataclass attribute (dataclasses does all the work of initializing)
- add a __post__init__ if required

Additionally, this also makes the distinction between ClassVars and ObjectVars are clearer. All attributes should now be explicitly documented as per standard dataclasses -- nothing buried in init methods.

This also introduces a new classvar:

https://github.com/linkml/linkml/pull/924/files#diff-0c1805c1851e79f4e48679169b9de8a93dacc5a44f1de3abc654cc3c46bfbae9R76-R77

`uses_schemaloader: ClassVar[bool]` (defaults to True)

This is used by older generators that rely on a pass through the SchemaLoader to inject builtins


linkml/utils/generator.py
```diff
@@ -30,73 +54,152 @@
...
+    """metamodel compliant schema.  Can be URI, file name, actual schema, another generator, an
+        open file or a pre-parsed schema"""
+
+    generatorname: ClassVar[str] = None
+    """ Name of the generator. Override with os.path.basename(__file__)"""
+
+    generatorversion: ClassVar[str] = None  # Generator version identifier
+    """Version of the generator. Consider deprecating and instead use overall linkml version"""
+
+    uses_schemaloader: ClassVar[bool] = True
```
cmungall, line 76: note this important change

PR #946, opened 2022-09-08
Fix 404 to the csvs.md page

joshmoore:
Link under https://linkml.io/linkml/data/conversion.html?highlight=csvs#loading-from-and-dumping-to-csvs goes to https://linkml.io/linkml/data/csvs.md (not HTML)


cmungall (approved):
Thanks, very much appreciated!

PR #949, opened 2022-09-08
show slot documentation pages with more detail 

sujaypatil96:
- [x] add more information under _Properties_ section of slot documentation pages, like multivalued, required, recommended, etc.
- [x] if slot page is mixin, show helpful table with information about what slots it's mixed into, etc.
  - [ ] should mixin usage logic be propagated to class jinja too?
- [ ] In the NMDC Schema docs, we show the various classes that use a given slot. For example: https://microbiomedata.github.io/nmdc-schema/alkalinity/
  - [ ] Is that useful functionality in the main docgen templates? I can add it if is, if not, this PR is ready for review


linkml/generators/docgen/slot.md.jinja2
```diff
@@ -20,9 +20,35 @@ URI: [{{ gen.uri(element) }}]({{ gen.uri(element) }})
...
+
 ## Properties
 
- * Range: {{gen.link(element.range)}}
+* Range: {{gen.link(element.range)}}
+* Multivalued: {{ element.multivalued }}
+{% if element.aliases %}
+* Aliases:
+    {% for alias in element.aliases -%}
+        * {{ alias }}
```
cmungall, line 40: is the indentation correct?

sujaypatil96: Seems like it: https://microbiomedata.github.io/nmdc-schema/dissolution_reagent/?

sujaypatil96: But I'll correct it anyway

linkml/generators/docgen/slot.md.jinja2
```diff
@@ -20,9 +20,35 @@ URI: [{{ gen.uri(element) }}]({{ gen.uri(element) }})
 <!-- no inheritance hierarchy -->
 {% endif %}
 
+{% if schemaview.is_mixin(element.name) %}
+## Mixin Usage
```
cmungall, line 24: I suggest col headers:

- child slot
- description
- range
- domain

maybe switch order it's more conventional to do domain then range

domain links should be hyperlinked

(all this can be addressed in future PR)

cmungall (approved):

Issue #931, opened 2022-09-02
mermaid diagram wishlist


sierra-moxon:
The following mermaid additions are essential:
- [x] include non is_a relationships 
- [ ] Use standard UML to round out slots inside a class (including type, cardinality)
- [x] Arrows should have cardinality in addition to arrow direction
- [x] diagram should show parents and children 

Nice to haves would be:
- [x] links in the classes themselves that allow the user to click on a class image and be redirected to the class page they click on, thus centering the mermaid diagram on the class they click on. 
- [x] contextual suppression of the slots in a class image if the slot list is very long. (slot table below is redundant and good enough if the slot list is very long) (see NMDC Biosample class)
- [x] edge labels on class diagrams

cmungall:
Implementation:

- should we continue to do this in jinja, or have a method in docgen? maybe there should even be a mermaidgen that gets called by docgen? see how the old markdowngen + yumlgen worked together

cmungall:
should we have a separate standalone module for making mermaid class diagrams?

sujaypatil96:
I vote for making a separate mermaingen.py file too because users might want to quickly make a mermaid diagram for some of their classes, etc.

sierra-moxon:
04/21/2023-  in progress.

cmungall:
@sujaypatil96 is this done?

sujaypatil96:
>  Use standard UML to round out slots inside a class (including type, cardinality)
 Arrows should have cardinality in addition to arrow direction

* Do we want this? We are heavily constrained on space on our documentation pages. Adding these two pieces of information will only "bloat" the class diagrams further
* Also does _type_ here mean _range_ of a slot?

sujaypatil96:
Putting this on my plate as a high priority item to work on.

sujaypatil96:
I have implemented " links in the classes themselves that allow the user to click on a class image and be redirected to the class page they click on, thus centering the mermaid diagram on the class they click on" in the NMDC schema. I will make the code generic and push it up.

sujaypatil96:
> edge labels on class diagrams
We are showing slot names on certain edges ("dependency" relationship type). The unfilled triangle arrow ("generalization"/is-a relationship type) is already understood that it is a sub-class-of relationship so no need to specify anything on those edges.

sujaypatil96:
> contextual suppression of the slots in a class image if the slot list is very long. (slot table below is redundant and good enough if the slot list is very long) (see NMDC Biosample class)

[...]

PR #957, opened 2022-09-12
modify class jinja template for middle layer elements

sujaypatil96:
Add jinja template logic to handle mermaid diagram rendering of classes with **both** parents and children.


cmungall (approved):

sierra-moxon (approved):

sierra-moxon:
fixes item in https://github.com/linkml/linkml/issues/931

Issue #959, opened 2022-09-14
schemasheets examples for ICBO 2022


turbomam:
https://github.com/linkml/linkml/tree/issue-959-schemasheets-icbo22

- [ ] Use https://github.com/linkml/linkml/blob/main/examples/tutorial/tutorial05/personinfo.yaml or similar/related file as input into linkml2sheets
  - [ ] will require templates
- workshop participants will be provided with a sample data file with some new fields that keeps it from validating against any of the personinfo.yaml versions in the examples/tutorial path
- will require participants to modify schema
- I will think of a modification pattern that is more suitable to shcemasheets that direct YAML editing
- We will need to discuss issues of round tripping, SOT (YAML, shared Google Sheet, checked-in TSVs...)

turbomam:
linkml/linkml isn't liking my attempt to make schemasheets a dependency

I might move all of this to linkml/schemasheets

sierra-moxon:
closing this for now; post-ICBO.  Please of course reopen if I got this wrong. :)

PR #1005, opened 2022-10-12
stragglers

turbomam:
#959


cmungall (changes requested):
I don't think you meant to make a PR on this repo @turbomam?

turbomam:
yeah, this isn't useful here and wouldn't be useful anywhere else

PR #1007, opened 2022-10-12
configure pydanticgen with use_enum_values set to true

kevinschaper:
Starting with a draft PR on the simple change to see how tests are affected


kevinschaper:
Going to retry from a fresh branch

PR #1010, opened 2022-10-12
show label name in inheritance tree when `use_slot_uris=True`

sujaypatil96:
When use_slot_uris is enabled, show label names in the inheritance tree instead of the id portion from the CURIEs.

CC: @turbomam, just tagging you on here, no need for review unless you want to.


cmungall (changes requested):
there are 105 files changed here. A PR should only include intended changes

sujaypatil96:
Most of the changed files there were because of some output test files I was adding, but then I decided to route the output to a temporary directory and then remove it.

This is one thing we need to keep in mind when we are cleaning up the tests folder too. To get rid of all or most of the output files in the test directories, by simply routing the output into temporary directories.

cmungall (approved):

Issue #1034, opened 2022-10-25
Add a consistent version option to all linkml command line tools


hsolbrig:
All Linkml command line tools should have a version function.

I'm looking at the the [LinkML Tutorial](https://linkml.io/linkml/intro/tutorial01.html), where one tests the install by running `linkml-convert --help`.   This creates a screenfull of stuff.  IMO, any linkml command line tools should be able to emit a short version line that includes the particular pypi version it came from as well as, perhaps, a link to its github data.  Something similar to what goes in one of the github software version badges.

**Proposal**
All functions include "-V / --version", and we create a bit of common code that formats the information.  Might even consider doing something clever with the terminal fonts and coloration :-)

**Priority**
Not essential, but would add to the consistency.

dalito:
Related issue fo linkml: #378

dalito:
Some tools have their own internal version info, e.g. https://github.com/linkml/linkml/blob/f899d36eb78bf00fc435413f4229dcccde50c625/linkml/generators/jsonschemagen.py#L52 - In #1052 the only linkml-version is returned.

PR #1048, opened 2022-10-26
Add dynamic versioning and --version command line option

### LINKED ISSUE #378: Add `--version` flag

dalito:
This addresses partly #1034. Click by default supports only the long format `--version` but not `-V`. ~~In my opinion the short form is not important to have and I suggest to stick with click´s default.~~ I added both '-V' and '--version' in a later commit.

Still to do: Adapt all other command line tools. I just added this to linkml-convert as an example to get feedback before continuing.

Closes #378

### DIFF
--- a/linkml/_version.py
+++ b/linkml/_version.py
@@ -1,0 +1,7 @@
+from importlib.metadata import version, PackageNotFoundError
+
+try:
+    __version__ = version("linkml")
+except PackageNotFoundError:
+    # package is not installed
+    __version__ = "0.0.0"
--- a/linkml/utils/converter.py
+++ b/linkml/utils/converter.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import sys
 from pathlib import Path
 from typing import List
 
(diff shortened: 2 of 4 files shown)


cmungall:
Thanks so much @dalito 🙏 ! @hsolbrig is trying this out and will write a test for it - I think he has some opinions about `-V` vs `--version` too!

hsolbrig (approved):
Looks great!  Thanks.

Appreciate the insights on how poetry can get the actual version from github.  We should really consider doing that for model versions as well ...

cmungall:
Good idea @hsolbrig! I made an issue #1051

PR #1049, opened 2022-10-26
docgen method to get class from which slot was inherited

sujaypatil96:
This method seeks to add more information to the slots table that appears on a class's documentation page, by telling us about how a particular slot was inherited in the context of a class.

### DIFF
--- a/linkml/generators/docgen/class.md.jinja2
+++ b/linkml/generators/docgen/class.md.jinja2
@@ -33,21 +33,15 @@ URI: {{ gen.uri_link(element) }}
 | Name | Cardinality and Range | Description | Inheritance |
 | ---  | --- | --- | --- |
 {%- if gen.get_direct_slots(element)|length > 0 %}
-{%- for d_slot in gen.get_direct_slots(element) %} | {{ gen.link(schemaview.get_slot(d_slot)) }} | {{ gen.cardinality(schemaview.get_slot(d_slot)) }} <br/> {{ schemaview.get_slot(d_slot).range }} | {{ schemaview.get_slot(d_slot).description }} | direct |
-{%- endfor -%}
-{%- endif -%}
-{%- if gen.get_indirect_slots(element)|length > 0 %}
-{%- for i_slot in gen.get_indirect_slots(element) %} | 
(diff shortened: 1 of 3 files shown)


sujaypatil96:
@cmungall: I've made the changes as requested. I'm not teasing out the mixin inherited slots separately from the other induced slots.

Also, it is not possible to link the induced parent class pages using `gen.link()` because we do not have a jinja equivalent of the Python `map()` method.

linkml/generators/docgen/class.md.jinja2
```diff
@@ -32,17 +32,16 @@ URI: {{ gen.uri_link(element) }}
...
+{%- if gen.get_direct_slots(element)|length > 0 %}
+{% for d_slot in gen.get_direct_slots(element) -%}
+| {{ gen.link(schemaview.get_slot(d_slot)) }} | {{ gen.cardinality(schemaview.get_slot(d_slot)) }} <br/> {{ schemaview.get_slot(d_slot).range }} | {{ schemaview.get_slot(d_slot).description }} | direct |
+{% endfor -%}
 {% endif -%}
-{%- if s.name in gen.get_indirect_slots(element) -%}
-| {{gen.link(s)}} | {{ gen.cardinality(s) }} <br/> {{ gen.link(s.range) }} | {{ s.description }}  | inherited |
+{% if gen.get_indirect_slots(element)|length > 0 -%}
+{% for i_slot in gen.get_indirect_slots(element) -%}
+| {{ gen.link(schemaview.get_slot(i_slot)) }} | {{ gen.cardinality(schemaview.get_slot(i_slot)) }} <br/> {{ schemaview.get_slot(i_slot).range }} | {{ schemaview.get_slot(i_slot).description }} | induced <br/> Class: _{{ gen.get_slot_inherited_from(element.name, i_slot)|join(', ') }}_ |
```
cmungall, line 42: You can get rid of `induced <br/> Class:`

And why not just make a docgen method that is the equivalent of link for lists of names?

sujaypatil96: Done. I've added a method called `links()` in the documentation generator. I'll merge the PR in once the tests go through.

cmungall (approved):

PR #1052, opened 2022-10-27
Add -V / --version option to all remaining cmd scripts

### LINKED ISSUE #1034: Add a consistent version option to all linkml command line tools

### LINKED ISSUE #1078: Need to incorporate a github id into the unit tests for test_linkml_files
Github limits the number of accesses via the API to 60 per hour if you don't have credentials.

We need to figure out how to fix this for testing.

dalito:
You were too fast with merging #1048 or I should have marked it as WIP...

Here is the rest: The same change applied to the other scripts plus a change to stay compatible with Python 3.7.

I also added one command line tool that was missing to `pyproject.toml`: gen-summary (`linkml.generators.summarygen:cli`).

Closes #1034


cmungall:
just this one test failing

```python
    def test_validate_yaml(self):
        """Test YAML file validation"""
        self.do_test(
            [env.input_path("yaml_validate_clean.yaml"), "-v"],
            "clean.txt",
            add_yaml=False,
        )
        with self.assertRaises(ValueError) as e:
            self.do_test(
                [env.input_path("yaml_validate_invalid.yaml"), "-V"],
                "invalid.txt",
                add_yaml=False,
            )
        self.assertIn("slot: k - unrecognized range (none)", str(e.exception))
```

I'm puzzled as why your changes would change the expected behavior here...

dalito:
@cmungall - The test was using uppercase V instead of lowercase v. But it is still surprising that only Linux pipelines failed but Windows passed. Locally I could reproduce the error on Win10 with Python 3.10. The only difference to the pipeline is that I always use pytest to run tests.

cmungall (approved):
Much appreciated!

PR #1053, opened 2022-10-27
Update pypi-publish.yaml for poetry-dynamic-versioning

dalito:
FYI: With [poetry-dynamic-versioning](https://github.com/mtkennerly/poetry-dynamic-versioning) this line (see diff) should not be required anymore. But please double-check before merging. I don't want to break your release pipeline.


hrshdhgd (approved):
Only one way to find out.

Issue #606, opened 2022-02-11
Add wrapper for jsonschema validator to ensure dates are validated


cmungall:
In validator.py we simply call the jsonschema validator

note this is not sufficient for validating dates as described here:
https://stackoverflow.com/questions/20264146/json-schema-date-time-does-not-check-correctly

The top answer provides an approach that will work for python. We should incorporate this so date checking works out-the-box for everyone. however, we should still provide an option for a vanilla mode.

cmungall:
It looks like we do this now:

https://github.com/linkml/linkml/blob/700cf37a8dfe0c70d2968dfa0ff0d5d1611543bd/linkml/validators/jsonschemavalidator.py#L51

Issue #368, opened 2021-09-24
Ensure all validators produce output in the same format and data model


cmungall:
Currently the differennt validators each produce their own output - we should make them produce standard objects (see linkml_model.validation) and standard output

cmungall:
We may want to follow the OAK model here

The interface returns an iterator over Validation objects:

https://github.com/INCATools/ontology-access-kit/blob/63df41569f2f62f1fe67c4ff81be770ca3c620d6/src/oaklib/interfaces/validator_interface.py#L48-L59

- this uses an iterator. This is useful if we later want to have say a BigTable validator that validates row by row in a streaming fashion
- it uses a Validation data model that is similar to the LinkML one (which itself came from shacl)

Data model: 

https://incatools.github.io/ontology-access-kit/datamodels/association/index.html

LinkML model:

https://github.com/linkml/linkml-model/blob/main/linkml_model/model/schema/validation.yaml

j-simmons-phd:
+1 to addressing this issue.

j-simmons-phd:
@sierra-moxon thanks for the instructions.  Official upvote added.

cmungall:
Covered by #1555 (sorry to split the vote!)

PR #1054, opened 2022-10-27
Refactor jsonschema generator to use schemaview

### LINKED ISSUE #1033: Refactor jsonschema generator to use schemaview
Currently jsonschemagen uses the older restricted visitor pattern, which makes it harder to implement things like jsonschema composition, which are inherently recursive.

There are a few other things that it would be good to change, the order I would do this

- [ ] Extend the [compliance cases](https://github.com/linkml/linkml/blob/main/tests/test_generators/input/kitchen_sink_failtest_inst_01.yaml)
- [ ] make the existing tests more modular
- [ ] switch from visitor to schemaview
- [ ] avoid dict manipulation and use a model of json schema; also avoid string manipulation of the form `f"#/$defs/{...}`

The last one may be non-critical. However, I think some lightweight abstracting out of common patterns may be useful.

pkalita-lbl:
These changes address the four points outlined in #1033. They do not address the other issue mentioned in that one (#368); I think that would be better addressed separately. 

One extra change is here is upgrading to the 4.x series of the `jsonschema` package and including the `format` extras. This allows validating data instances against JSON schemas including things like `{ "format": "date-time" }` (which in turn results from LinkML schemas that include, for example, `range: xsddatetime`)

Fixes #1033


pkalita-lbl:
I think this also resolves #606

linkml/generators/jsonschemagen.py
```diff
@@ -32,7 +33,81 @@
     "xsdtime": ("string", "time"),
 }
 
-WITH_OPTIONAL_IDENTIFIER_SUFFIX = "__identifier_optional"
+class JsonSchema(UserDict):
```
cmungall, line 36: I like how this is separated out

linkml/generators/jsonschemagen.py
```diff
@@ -49,95 +124,71 @@ class JsonSchemaGenerator(Generator):
 
     # ClassVars
     generatorname = os.path.basename(__file__)
-    generatorversion = "0.0.2"
+    generatorversion = "0.0.3"
```
cmungall, line 127: we haven't been doing this very consistently, need a strategy here. maybe overall version of linkml is good enough

linkml/generators/jsonschemagen.py
```diff
@@ -49,95 +124,71 @@ class JsonSchemaGenerator(Generator):
...
     valid_formats = ["json"]
     visit_all_class_slots = True
     visit_all_slots = True
-    uses_schemaloader = True
-    #requires_metamodel = False
+    uses_schemaloader = False
```
cmungall, line 131: 🍾

linkml/generators/jsonschemagen.py
pkalita-lbl: Okay yeah I wasn't sure. I saw it there so I figured I'd increment it. The `gen-json-schema` CLI is already hooked up to use the main LinkML version number when you pass the `--version` flag. I don't see any place the `generatorversion` variable is used within the `linkml` project. I'd be happy to just remove it unless you think there's a possibility of a downstream project using that variable.

linkml/generators/jsonschemagen.py
```diff
@@ -154,136 +205,110 @@ def extract_permissible_text(pv):
...
         if (
             self.top_class is not None and camelcase(self.top_class) == camelcase(cls.name)
         ) or (self.top_class is None and cls.tree_root):
-            self.schemaobj.properties[underscore(aliased_slot_name)] = prop
+            self.top_level_schema.add_property(aliased_slot_name, prop, slot_is_required)
+
+    def serialize(self, **kwargs) -> str:
+        self.start_schema(**kwargs)
+        for enum_definition in self.schemaview.all_enums().values():
+            self.visit_enum(enum_definition)
```
cmungall, line 307: it's maybe a bit odd keeping the method names as visit_X if it's not truly using the VP anymore... but YMMV

tests/test_generators/input/kitchen_sink_compliance_inst_01.yaml
```diff
@@ -40,28 +37,25 @@
...
 - description: additional properties is closed
-  skip: false
   dataset:
     persons:
       - id: P
         name: john smith
         made_up_property: "abc"
 - description: optional open properties
-  skip: true
+  skip: not_closed value does not propagate to nested objects. By design?
```
cmungall, line 58: hmm, perhaps Bill designed it that way, I can see the use case but a little unintuitive

tests/test_generators/input/kitchen_sink_compliance_inst_01.yaml
```diff
@@ -1,37 +1,34 @@
+# By default these cases are expected to *not* validate. Use valid: true
```
cmungall, line 1: `skip: true`?

cmungall: ah wait I see the structure now

cmungall (approved):

linkml/generators/jsonschemagen.py
pkalita-lbl: I think you're right. I've started working on refactoring and adding methods for handling rules and I'll probably rename some existing methods as part of that.

PR #1083, opened 2022-10-28
Fix pypi publish pipeline, Follow-up to #1053

hrshdhgd:
Follow-up to #1053

### DIFF
--- a/.github/workflows/pypi-publish.yaml
+++ b/.github/workflows/pypi-publish.yaml
@@ -18,10 +18,12 @@ jobs:
           python-version: 3.9
 
       - name: Install Poetry
-        run: pip install poetry poetry-dynamic-versioning
+        run: |
+          pip install poetry
+          poetry self add "poetry-dynamic-versioning[plugin]"
 
-      - name: Install dependencies
-        run: poetry install --no-interaction
+      # - name: Install dependencies
+      #   run: poetry install --no-interaction
 
       - name: Build source and wheel archives
         run: |
--- a/.github/workflows/pypi-publish.yaml
+++ b/.github/workflows/pypi-publish.yaml
@@ -18,12 +18,10 @@ jobs:
           python-version: 3.9
 
       - name: Install Poetry
-        run: pip install poetry
+        run: pip install poetry poetry-dynamic-versioning
 
       - name: Install dependencies
-        run: |
-          poetry self add "poetry-dynamic-versioning[plugin]"
-          poetry install --no-interaction
+        run: poetry install --no-interaction
 
       - name: Build source and wheel archives
         run: |
--- a/.github/workflows/pypi-publish.yaml
+++ b/.github/workflows/pypi-publish.yaml
@@ -12,23 +12,18 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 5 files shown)


dalito:
@hrshdhgd - Maybe you want to pipx-install poetry like in https://github.com/linkml/linkml-runtime/pull/214? Then poetry is isolated from the package. pipx is available by default.

hrshdhgd:
@dalito , I am trying it with `pip install -U poetry` in #1083

hrshdhgd:
Even when I do `poetry install` locally, I get:
```
Installing dependencies from lock file

No dependencies to install or update

Installing the current project: linkml (0.0.0)
```

Is this different for you? It blindly grabs the version from the toml file.

dalito:
@hrshdhgd - Can you add a line in CI code to list the poetry plugins for debugging:
```
poetry self show plugins
```

hrshdhgd:
I get `Too many arguments.` in red ..lol!

dalito:
With poetry-1.2.2?

hrshdhgd:
Ok, problems I noticed:

 - I was using an older version of `poetry` locally
 - Due to the above `poetry update` did not update the lock file with the plugins.

I think we're inching close towards the solution

hrshdhgd:
Nope: I still get:

```
• poetry-plugin-export (1.1.2) Poetry plugin to export the dependencies to various formats
      1 application plugin

      Dependencies
        - poetry (>=1.2.0,<2.0.0)
        - poetry-core (>=1.1.0,<2.0.0)
```

which does not show what you are getting

dalito:
If this fails I next guess is that builing before installing could help:
```
        run: |
          poetry build
          poetry install -E docs
          poetry self show plugins```

**Comment by hrshdhgd** [comment]
I tried this locally, still get what I got before. No change.

**Comment by dalito** [comment]
Hmm. Does the release pipeline re-use something from the main pipeline? Because there the somewhat problematic `snok/install-poetry` is still used.

**Comment by hrshdhgd** [comment]
ok, I didn't do Step 1 explained in the https://github.com/mtkennerly/poetry-dynamic-versioning repo

 - `poetry self add "poetry-dynamic-versioning[plugin]"`

File: .github/workflows/pypi-publish.yaml [hunk 1]
```diff
@@ -12,23 +12,18 @@ jobs:
...
-      #   with:
-      #     virtualenvs-create: true
-      #     virtualenvs-in-project: true
+      - name: Install Poetry
+        run: pip install poetry
 
       - name: Install dependencies
-        run: poetry install --no-interaction
+        run: |
+          poetry self add "poetry-dynamic-versioning[plugin]"
```
dalito, line 25: You could move the installation the pipx / poetry section. The command is: 
```
pipx inject poetry poetry-dynamic-versioning
```

hrshdhgd:
This did the trick! Thank you so much for the help @dalito!

.github/workflows/pypi-publish.yaml
hrshdhgd: Done! Thanks again!

cmungall (approved):

PR #1086, opened 2022-10-31
mixin usage block on class and slot pages

sujaypatil96:
Show `Mixin Usage` section with tables on class and slot pages.


cmungall (approved):
in general there should be tests accompanying these

PR #1092, opened 2022-11-02
Add slots section to enum Jinja template and link to slots

### LINKED ISSUE #998: gen-docs pages about enumerations should show the slots that use the enum
this is another case of showing inbound relationships

amc-corey-cox:
With PR https://github.com/linkml/linkml-runtime/pull/222 this Closes #998


amc-corey-cox:
@sujaypatil96 and @cmungall, what I could really use some guidance on is how to create a test-case for the changes to the enum jinja template. If you could point me to some code to copy or a snippet to start from, I'd really appreciate that. If you don't think this needs it's own test-case (the functionality is all implemented in linkml-runtime) then this is ready to go.

cmungall:
I'd change the edits to ks to avoid mixing atts and slots

for tests, see the patterns in test_docgen

cmungall:
oops, looks like conflicts need resolved

sujaypatil96:
@amc-corey-cox: looks like you need to resolve some conflicts in the kitchen_sink.yaml file?

sujaypatil96:
As for test cases, ideally, all of our docgen test cases can be found in the [test_docgen.py](https://github.com/linkml/linkml/blob/main/tests/test_generators/test_docgen.py) file. So what you can do is, to add a new method, say `test_get_slots_by_enum()` and use the `assert_md_file_contains()` method like how the other test cases use to check if the enum markdown pages indeed contain the output that you would expect.

sujaypatil96:
@amc-corey-cox: apologies for adding additional changes in your branch. This is excellent work, thank you so much. With the additional changes I think the PR looks ready, with functionality, tests, etc. Once we see that all the tests pass, we should be good to merge the PR.

Issue #101, opened 2021-01-14
Stack dump doesn't belong in value error messages


hsolbrig:
Errors in the model shouldn't be accompanied by a stack dump - only the value error line should be printed

Issue #65, opened 2020-08-31
Empty list entries aren't detected


hsolbrig:
```yaml
   in_subset:
      - abc
      -
```
Isn't detected until late in the process and then only by a mysterious failure.  Need to detect `None` in lists early in the process

sierra-moxon:
Would be helpful to see the error message, and if possible submit a simple test that shows the issue?

hsolbrig:
This turns out to be a real edge case.  If the empty entry isn't followed by a carriage return, the python yaml module misses it and it breaks in another part of the code.  I suppose that this could occur just often enough, however, that we probably need to fix it.

I'm weakly inclined to put a hack in the raw_loader that makes sure there is a carriage return at the end of the file...

I'll issue a PR for the test case and proposed fix.

PR #1110, opened 2022-11-16
Add stacktrace option w/ default no trace.

hsolbrig:
Common parameter for all CLI tools. "--stacktrace/--no-stacktrace"

tests/test_issues/input/issue_65_new.yaml is a sample w/ an error in it.

Fix for issue #101 .  Also addresses part of issue #65.  Notes:
1) There isn't a unit test for this fix -- the test file, `issue_65_new.yaml`, creates a really ugly error message if there is stack
trace, and a one-liner with this fix.  The problem, however, is the UnitTest harness overrides those fixes, so you can't see them.

### DIFF
--- a/linkml/utils/generator.py
+++ b/linkml/utils/generator.py
@@ -18,6 +18,7 @@ import abc
 import logging
 import os
 import re
+import sys
 from contextlib import redirect_stdout
 from dataclasses import dataclass, field
 from io import StringIO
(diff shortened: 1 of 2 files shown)


cmungall (approved):

linkml/utils/generator.py
```diff
@@ -153,13 +154,17 @@ class Generator(metaclass=abc.ABCMeta):
...
+
     def __post_init__(self) -> None:
         if not self.logger:
             self.logger = logging.getLogger()
         #    logging.basicConfig()
         #    self.logger = logging.getLogger(self.__class__.__name__)
         #    self.logger.setLevel(log_level)
-
+        if not self.stacktrace:
+            sys.tracebacklimit = 0
```
cmungall, line 167: I don't think the generator object should be messing with globals

Instead this should be controlled entirely at CLI level

See https://github.com/linkml/linkml/issues/1227

PR #1120, opened 2022-11-21
Add workflow for Docker build and publish 

### LINKED ISSUE #1102: Auto-release to dockerhub with each release

pkalita-lbl:
These changes:

* Update the Dockerfile so that the `linkml` package is installed from source in the image. Previously it was installed from PyPI, but going forward we don't want to have to wait for the package to be published to PyPI before building a Docker image and publishing it to Docker Hub. 
* Add a GitHub Actions workflow for building a Docker image and publishing it to Docker Hub. It will run on PRs and pushes to `main` and pushes to `v*` tags. The push to Docker Hub will be skipped for PRs, but it will still verify that the Docker image build works. Pushes to `main` or `v*` tags will publish the image to Docker Hub. 

Fixes #1102

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -1,0 +1,51 @@
+name: Build Docker Image
+
+on:
+  push:
+    branches:
+      - "main"
+    tags:
+      - "v*"
+  pull_request:
+    branches:
+      - "main"
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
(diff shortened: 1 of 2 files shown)


pkalita-lbl:
This is ready to review now. You can see the multi-arch image build happening in the Build Docker Image action. For PRs it's not pushing to Docker Hub. The push is a little harder to test. We'll either have to watch it when we make the next release or make a dummy `vX.Y.Z` tag, verify the push to Docker Hub works correctly, then remove the tag and image.

cmungall (approved):

Issue #880, opened 2022-07-22
Running the test suite should not change any version-controlled files


dalito:
Running the test suite changes many files which messes up `git status` (2687 modified files). While shell scripts are available for resetting the repo back to original state or for ignoring the changes it is somewhat inconvenient especially on Windows where the unix shell tools are typically not available. Moreover, linkml is deviating from most other Python projects in this regard. Normally running the test suite does not mess up your git repository.

**How important is this feature?** Select from the options below:
• Low - it's an enhancement but not crucial for work

cmungall:
I would love to get this sorted out. Some related issues here: https://github.com/orgs/linkml/projects/11

sierra-moxon:
duplicate of #587

PR #1121, opened 2022-11-21
improve contribution guidelines for writing unit tests

sujaypatil96:
More detailed documentation to point developers to for writing unit tests in the linkml testing framework.


docs/contributing/contributing.md
```diff
@@ -45,19 +55,12 @@ DEFAULT_LOG_LEVEL_TEXT: ERROR
 ```
 
 * Never hardcode any file paths. Always use import variables such as `INPUT_DIR`, `OUTPUT_DIR` and use `os.path.join()` to make file paths. This ensures that the tests will run independent of the OS they are running on.
+* Running the test suite changes many files in `..\output\..` folders which messes up `git status`.
```
dalito, line 58: Could you add a note about why it is done like that? Could the output not be instead written to a temporary dir and then compared to the expected result in "output"? Alternatively, the output could be written to a git-ignored folder. (related #880)

cmungall: Issue 880 is a big priority, @hsolbrig is working on this. I think it may detract from this document to include too much historic context especially when we get rid of this.

For now, we can just add a comment this is a temporary situation

cmungall (approved):

docs/contributing/contributing.md
sujaypatil96: Yup, it's probably not the best idea to add a lot of historic context to the document, only to deprecate it soon after.

I've added a comment to address your concern @dalito.

PR #1152, opened 2022-12-06
slots table breaking in few schemas

sujaypatil96:
The jinja templates in this PR have been tested on the kgcl and MIxS schemas, both of which had breaking slots table on class pages with the jinja templates from v1.3.15


cmungall (approved):
I still think the surrounding `if` is entirely superfluous, but we can leave that to a separate PR

sujaypatil96:
@cmungall: docgen.get_direct_slots() and docgen.get_indirect_slots() both return lists, but in the scenario where they return empty lists docgen will still try to apply the logic underneath the slots table and loop over an empty list. Which is why the `if` can prove to be useful. It simply renders an empty slots table and doesn't crash docgen.

Issue #1151, opened 2022-12-06
markdown documentation page names not properly cased


sujaypatil96:
When generating the documentation pages for the MIxS schema: https://github.com/GenomicsStandardsConsortium/mixs/tree/main/model/schema, we noticed that for some of the classes, for example `soil` class, the markdown and HTML links are being resolved properly.

So what's currently happening is this:
For `soil` class
* markdown page is generated as `soil.md`
* docgen is try to find a `Soil.md` page though and it's not able to find such a page

This probably has to do with code in this block: https://github.com/linkml/linkml/blob/main/linkml/generators/docgen.py#L299-L318 and `underscore()` being applied for class names.

sujaypatil96:
I guess one thing that's implicitly enforced by LinkML is the fact that classes are required to be camelcase and that slots are required to be snakecase.

On the off chance that the schema does not assert elements following the aforementioned criteria, it will produce broken links in the web docs.

Do we want to add handlers for the above scenario? Say, for example, when the class names are all lowercase, which is the case in MIxS.

sierra-moxon:
If we don't, Biolink model will have broken links; my vote is to support.  It will be hard for me to change Biolink to match the linting requirements (but I'd love to eventually! :))

sujaypatil96:
So it turns out the root cause of this is the fact that MacOS does not allow us to have two files with the same name but different case in the same directory, i.e., the MacOS file system is case insensitive by default: https://unix.stackexchange.com/questions/237685/how-can-i-create-two-files-with-same-name-with-different-case-in-mac-osx

So what's happening in the MIxS schema is that there is an imported schema called `soil`, basically `soil.yaml` and there also happens to be a class called `soil`. So docgen is trying to create a markdown file for the imported schema as `soil.yaml` which it does so first, and then another file for the class `Soil` as `Soil.md` which the filesystem does not accept.

cmungall:
this is a duplicate issue, we have a solution for this

sujaypatil96:
Closing in favour of duplicate issue #632

Issue #632, opened 2022-03-03
generated markdown has clashes if elements differ only in case on Mac OS


cmungall:
Some schemas like phenopackets and mixs have slot names that are case-insensitive-equal to class names. E.g. metaData (slot) and MetaData (class).

When generating markdown files (either gen-doc or gen-markdown) this will result in one overwriting the other -- but only on macs whose filesystem is case insensitive. This causes further confusion in mkdocs ending up with broken links.

One workaround is to build docs in docker... but the problem may persist or get weird if a volume is mounted

I think we need an option in gen-docs such that types can be prefixed with a folder or prefix, e.g. classes go in classes/Foo, slots in slots/foo. (of course, this assumes case-insensitive uniqueness within a metamodel type but you would be crazy to try and violate this). gen-docs should also warn if you don't do this and see clashes.

quick test to confirm behavior, on a mac:

```
✗ echo 1 > FOO && echo 2 > foo && cat FOO && cat foo
2
2
```

contrast with:

```
✗ docker run -it ubuntu:16.04 /bin/bash
root@b5211d0d3437:/# echo 1 > FOO && echo 2 > foo && cat FOO && cat foo
1
2
```

cmungall:
Does this problem go away if we switch to building the docs via github actions? I think the problem only arises if the mkdocs framework is executed on a mac

galund:
(Just got bitten by this, so plus one to being able to have separate folders. In my case I'd want classes in a separate folder, but I'd want slots at the root level, because I have a need for backward-compatible URIs for slots but I can get away with moving the class files.)

cmungall:
The current solution to this is to always use github actions to build the docs - don't build manually on a Mac

sujaypatil96:
Similar issue #1151

sujaypatil96:
What if users want to test the documentation locally on their Mac machines? This wouldn't be scalable for that I think? I don't think users should have to wait for GitHub Actions to publish their docs for them to see how it looks.

Also, I think @turbomam built and deployed the MIxS docs on a Linux machine, but links are still broken: https://genomicsstandardsconsortium.github.io/mixs/Agriculture.md

cmungall:
We can have an option for meta class specific folders but this of course
has major implications for URIs. I don’t see a way around this (other than
of course to avoid the clash in the first place, which I think is easy for
mixs since we can use whatever we like for class names eg
AgriculturePackage)

sierra-moxon:
[...]

PR #1155, opened 2022-12-07
workaround for clashing markdown file names

sujaypatil96:
Fix for issue #1151

### DIFF
--- a/linkml/generators/docgen.py
+++ b/linkml/generators/docgen.py
@@ -170,7 +170,7 @@ class DocGenerator(Generator):
             out_str = template.render(
                 gen=self, schema=imported_schema, schemaview=sv, **template_vars
             )
-            self._write(out_str, directory, imported_schema.name)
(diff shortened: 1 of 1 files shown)


cmungall (changes requested):
This will change all URLs and w3ids for all existing projects. This has to be a non-default option

sujaypatil96:
Not quite sure I understand how it would break existing URLs for projects and w3ids? As long as we regenerate the documentation with `gen-doc` once again, it should be fine?

But yes, the one thing it would change is the structure of the folder where the target documentation pages are output.

sujaypatil96:
Close this PR. See comments on #632 and #1155

sujaypatil96:
Message from @cmungall:

> it’s not a bad idea, but it would have to be opt-in

sierra-moxon:
@sujaypatil96 - I am going to close this PR per the feedback here.  I'll leave the branch there however, in case we can build on the already good start! :)

Issue #1160, opened 2022-12-12
AttributeError: 'SiteId' object has no attribute '_sa_instance_state'


turbomam:

turbomam:
I think the`nu_obj = self.to_sqla(element)` in  `linkml.utils.sqlutils.SQLStore.dump` is returning a near-empty, malformed object

turbomam:
I am continuing to see this issue, most recently in https://github.com/beril-model/beril-model

PR #1161, opened 2022-12-12
testing linkml-sqldb dump -> "AttributeError: 'SiteId' object has no attribute '_sa_instance_state'"

turbomam:
#1160

### DIFF
--- a/tests/test_issues/input/issue_1160_data.yaml
+++ b/tests/test_issues/input/issue_1160_data.yaml
@@ -1,0 +1,31 @@
+biosample_list:
+- id: GOLD:Gb0305833
+  name: Root microbial communities from poplar common garden site in Clatskanie, Oregon,
+    USA - BESC-13-CL1_35_33 endosphere
+  description: Root microbial communities from poplar common garden site in Clatskanie,
+    Oregon, USA
+  collected_from: GOLD:BESC-13-CL1_35_33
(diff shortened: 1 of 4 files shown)


nlharris:
What's the status of this? Seems like at this point it should be merged or closed.

turbomam:
`linkml-sqldb dump` still doesn't work for nmdc-schema, but maybe this test doesn't really illuminate that.

amc-corey-cox:
@turbomam this PR looks like it hasn't had any work in a while. Based on our recent discussion in the LinkML meeting I'm going to close this PR unless there is some activity in the next week or so.

turbomam:
> @turbomam this PR looks like it hasn't had any work in a while. Based on our recent discussion in the LinkML meeting I'm going to close this PR unless there is some activity in the next week or so.

I haven't tried any of the linkml sql tools in a while. I suspect that dumping and loading nmdc-schema data still won't work but could revisit that in the future. Just not in the next week or two.

turbomam:
> What's the status of this? Seems like at this point it should be merged or closed.

Yes, that's true. But there's really nothing to merge here.

amc-corey-cox:
I'm going to close this. It is linked to the issue so when someone is ready to address that they can either re-open this PR, use the test-case in their own or start over. Thanks for the work @turbomam !

PR #1163, opened 2022-12-13
Upgrade to latest linkml-runtime release candidate

pkalita-lbl:


pkalita-lbl:
The thing that looked possibly concerning to me is represented by the change in this test output file: https://github.com/linkml/linkml/pull/1163/files#diff-9f2f3a5e527d1ec4eeeb3b5cae189c310e6c2e1ea35904e43b07689a86734594. It looks like metamodel slots like `generation_date`, `metamodel_version` got removed from the OWL subset and `classes`, `enums`, etc got added -- and they have some sort of ugly serialized values.

The original change that updated the subset name and changed which slots are part of the subset is here: https://github.com/linkml/linkml-model/pull/113/commits/14e3da64dd55e76786ef0f79b107f1d21ffc44b7#diff-5fc807b5f16aafb921180aea704aa97a4c5a869eeff86344975963ba6b0bb49d

pkalita-lbl:
Also, just to note we **should not** merge these changes as they stand. I mostly wanted to use this PR as a way to offload testing from my own machine. We don't want the change to `pyproject.toml` that allows pre-release versions of `linkml-runtime` to go into `main`.

pkalita-lbl:
@cmungall if you have time can you advise on whether the changes in the `linkml-model` commit referenced above or [this logic in owlgen](https://github.com/linkml/linkml/blob/06134b13acc9cf8d64dc006d46475804765770c7/linkml/generators/owlgen.py#L510) needs to be rethought? Or some combination of model and owlgen changes?

PR #1169, opened 2022-12-15
WIP: test pythongen forward reference bug.

### LINKED ISSUE #1168: detection of forward references may sometimes be incorrect in pythongen
This replaces:

- #1166. 

It looks like there are some situations where foward ref detection falsely assumes no forward reference when it should use one

This is because `_sort_classes` is applied prior to writing classes, but the following code operates over the asserted sort order

https://github.com/linkml/linkml/blob/5ecd9c402b7e8f5ad1775904a9247df26a1918fa/linkml/generators/pythongen.py#L921-L938

I am working on replicating this issue - it seems to occur rarely in the wild

The workaround is to order elements with your schema accordingly, but this should not be necessary

cmungall:
When this PR is merged, it fixes #1168

### DIFF
--- a/tests/test_issues/input/linkml_issue_1168.yaml
+++ b/tests/test_issues/input/linkml_issue_1168.yaml
@@ -1,0 +1,19 @@
+id: http://example.org/issue-1168
+name: issue_1168
+description: Tests forward reference edge case
+see_also:
+  - https://github.com/linkml/linkml/issues/1168
+
+classes:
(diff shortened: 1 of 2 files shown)


sierra-moxon:
closing this PR as it is a yaml example file and a test with no asserts -- I think this is because this was already fixed in another PR and this is redundant.  please of course reopen if I got this wrong.

PR #1171, opened 2022-12-16
Issue 675

hsolbrig:
When I run unit tests locally, I get two errors - 

```
test_generators/test_sqltablegen.py:176: 
test_get_sql_range:
E           TypeError: not all arguments converted during string formatting

test_issues/test_issue_338.py:13: 
test_biolink
E           TypeError: not all arguments converted during string formatting
```

These come from the following lines in SchemaLoader:
```
                if mangled_slot_name in self.schema.slots:
                    # mangled names are overwritten if a schema with attributes is passed in
                    # TODO: handle this in a more graceful way
                    #  see https://github.com/linkml/linkml/issues/872
                    logging.warning(
                        f'Class: "{cls.name}" attribute "{attribute.name}" - '
                        f"mangled name: {mangled_slot_name} already exists",
                        attribute.name,
                    )
```
(The logger.warning line...)  I'm sort of doubtful that this is due to this fix...


cmungall (approved):

linkml/utils/ifabsent_functions.py
```diff
@@ -80,25 +68,16 @@ def default_ns_for(loader: SchemaLoader, cls: ClassDefinition) -> str:
...
-    ("slot_uri", True, lambda _, loader, ___, ____: "None"),
-    ("class_curie", True, lambda _, loader, ___, ____: "None"),
-    ("slot_curie", True, lambda _, loader, ___, ____: "None"),
-    # TODO: If you assign a range in the constructor, mergeutils has no way of knowing whether the range
-    #       was overridden or just defaulted.  We need to let the old code continue to work until we get
-    #       this bit resolved
-    # ("default_range", False, lambda _, loader, __, ____: f"ElementName({strval(loader.schema.default_range)})"),
-    ("default_range", False, lambda _, __, ___, ____: "None"),
+    (r"int\(([-+]?[0-9]+)\)", False, lambda m, __, ___, ____: int(m[1])),
+    ("class_uri", False, lambda _, __, class_definition, ____: 'class_class_uri'),
```
cmungall, line 72: I don't think this is correct

Issue #1099, opened 2022-11-08
Question concerning is_a inheritance, designates_type metamodel slot usage


j-simmons-phd:
If I have a schema like the one shown in https://linkml.io/linkml/intro/tutorial07.html#inheritance where `Person` and `Organization` both inherit from `NamedThing`, can I create a multivalued slot `things` with Range of `NamedThing` and add `Person` and `Organization` objects to the slot `things`?  Something like this:

```
# schema addition
Container:
    tree_root: true
    attributes:
      things:
        multivalued: true
        inlined_as_list: true
        range: NamedThing
```
```
# data file
things:
  - id: ORCID:1234
    full_name: Superman
    has_organizational_relationships:
      - related_to: ROR:1
        relationship_type: MEMBER_OF
  - id: ROR:1
    full_name: Justice League
```

I tried a version of this with one of my own schemas and got a validation error about one of the additional slots being unknown when I tried to create one of the subclasses as an item in the list (e.g. has_organizational_relationships).  Error message below.  Note, not all of this code is committed yet.  

If this sounds like something that *should* work I can refile this as a bug with fully committed code backing it up. 

```
[~/Repos/m30ml]$ linkml-validate -s dist/m30ml.yaml src/data/examples/009-MixedElements.yaml 
Traceback (most recent call last):
  File "/usr/local/bin/linkml-validate", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/linkml/validators/jsonschemavalidator.py", line 114, in cli
    obj = loader.load(source=input, target_class=py_target_class, **inargs)
  File "/usr/local/lib/python3.8/dist-packages/linkml_runtime/loaders/loader_root.py", line 85, in load
    results = self.load_any(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/linkml_runtime/loaders/yaml_loader.py", line 25, in load_any
[...]

Issue #945, opened 2022-09-08
Add support for designates_type in all generators


cmungall:
[designates_type](https://w3id.org/linkml/designates_type) is a metamodel boolean slot that can indicate that a slot is intended to hold the value of the instantiated class.

This should minimally constrain things such that the value is an enum of allowed classes (descendants of the holder class)

It can also dynamically determine how a field can be instantiated. Consider data

```yaml
objects:
  - type: my:Person
    ssn: NNN
  - type: my:Pizza
    topping: ...
```

with a schema:

```yaml
  Object:
    attributes:
      type:
         designates_type: true
         range: uriorcurie
   Person:
     is_a: Object
     attributes:
       ssn:
   Pizza:
     is_a: Object
     attributes:
       topping:
```

where objects' range is the parent of Person and Pizza (or linkml:Any). The type field determines how each object is instantiated to a class, and thus what to check (people shouldn't have toppings and pizzas don't have ssns).

Just as with any other slot, the range of the slot that is designated as holding the type is determined by `range`, with the additional conventions:

- if range is `uri` then the full URI would be used
- if range is `uriorcurie` as above then either MAY be used but the CURIE should be the canonical form (we should have a separate type added for strict CURIEs)
- if the range is `string` then the value should be the class name

Currently linkml-owl has a limited form of this

kervel:
i created another branch with added support for dataclasses and json-schema. it's here https://github.com/Kapernikov/linkml/tree/polymorphism_dataclasses . i'd love to get some advice over the code before i submit a PR, its quite messy now.

kervel:
hello @cmungall if we are using the full URI, we still have the option to use the native one or the one that is defined in the "uri". I took the native one for now.

nlharris:
This is a popular issue!

cmungall:
Done in 1.6

PR #1180, opened 2022-12-20
[WIP] Pydanticgen: support polymorphism using type designators

kervel:
This makes the following modifications:
* The type designators are const=True now in pydantic so that pydantic can use them to determine which class needs to be instantiated
* We specify a union of all possible types (using class_descendants) for the range of an slot that refers to a class having a type designator

References:
* https://github.com/pydantic/pydantic/issues/503
* https://github.com/linkml/linkml/issues/1099

Open questions:
* I use two APIs that are slow (i think): the induced slots and the descendants of the class. is this okay ?
* I now only enable polymorphism for things that have a type designator. Actually, in some cases it should be possible to enable polymorphism for things without a type designator (pydantic will simply try all classes and takes the first one that validates). However, i think this is a safe approach for now
* I have no idea (and i would need help) in order to add tests and documentation.
* I think the python dataclasses generator needs a similar modification

### DIFF
--- a/linkml/generators/common/__init__.py
+++ b/linkml/generators/common/__init__.py
@@ -1,0 +1,17 @@
+from linkml_runtime.linkml_model.meta import SlotDefinition, ClassDefinition
+from linkml_runtime.utils.schemaview import SchemaView
+from linkml_runtime.utils.formatutils import camelcase
+
+def get_type_designator_value(sv: SchemaView, type_designator_slot: SlotDefinition, class_def: ClassDefinition) -> str:
+    """
+        retuns the correct value for a type designator field for a given class, depending on its range
+        this implements the logic described in https://github.com/linkml/linkml/issues/945:
+    """
+    target_value = None
+    if type_designator_slot.range == 'uri':
+        target_value = sv.get_uri(class_def,expand=True,native=True)
+    elif type_designator_slot.range == 'string':
+        target_value = camelcase(class_def.name)
+    else:
+        target_value = sv.get_uri(class_def,expand=False)
+    return target_value
--- a/linkml/generators/pydanticgen.py
+++ b/linkml/generators/pydanticgen.py
@@ -26,7 +26,7 @@ default_template = """
 from __future__ import annotations
 from datetime import datetime, date
 from enum import Enum
-from typing import List, Dict, Optional, Any
+from typing import List, Dict, Optional, Any, Union
 from pydantic import BaseModel as BaseModel, Field
 from linkml_runtime.linkml_model import Decimal
 
--- a/linkml/generators/pydanticgen.py
+++ b/linkml/generators/pydanticgen.py
@@ -209,6 +209,7 @@ class PydanticGenerator(OOCodeGenerator):
                             + slot_values[camelcase(class_def.name)][slot.name]
                             + "]"
                         )
+                    slot_values[camelcase(class_def.name)][slot.name] = slot_values[camelcase(class_def.name)][slot.name] + ", const=True"
                 # Multivalued slots that are either not inlined (just an identifier) or are
                 # inlined as lists should get default_factory list, if they're inlined but
                 # not as a list, that means a dictionary
@@ -272,7 +273,10 @@ class PydanticGenerator(OOCodeGenerator):
             sv.get_identifier_slot(range_cls.name) is None
             and not sv.is_mixin(range_cls.name)
         ):
-            return f"{camelcase(slot.range)}"
+            if len([x for x in sv.class_induced_slots(slot.range) if x.designates_type]) > 0:
+                return f"Union[" + ",".join([camelcase(c) for c in sv.class_descendants(slot.range)]) + "]"
+            else:
+                return f"{camelcase(slot.range)}"
 
         # For the more difficult cases, set string as the default and attempt to improve it
         range_cls_identifier_slot_range = "str"
--- a/linkml/generators/pydanticgen.py
+++ b/linkml/generators/pydanticgen.py
@@ -269,7 +269,7 @@ class PydanticGenerator(OOCodeGenerator):
 
         # Inline the class itself only if the class is defined as inline, or if the class has no
         # identifier slot and also isn't a mixin.
-        if slot.inlined or (
+        if slot.inlined or slot.inlined_as_list or (
             sv.get_identifier_slot(range_cls.name) is None
             and not sv.is_mixin(range_cls.name)
         ):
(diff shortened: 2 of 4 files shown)


sierra-moxon:
Hi @kervel  - thanks so much for this PR! :)  I see a few open questions on the PR; may I convert this to a "draft" PR while we work through them?

kervel:
yes off course. i'm used to gitlab where when prefixing with WIP its draft automatically, sorry.

sierra-moxon:
Oh that makes sense - I haven't worked much with gitlab :) converting now.

kervel:
thanks. i guess the next step would be getting some feedback/review on this ? is there somebody who could help me out on that ?

sierra-moxon:
Hi @kervel - definitely :) - we are just about into the holidays in the US now, so our replies may be a bit slower than usual.  

Open questions:

```I use two APIs that are slow (i think): the induced slots and the descendants of the class. is this okay ?```

- yep, this is our intent with schemaview, that we use its methods to help traverse the model.  we can work on the speed of these methods in a separate ticket, if they are too slow?

```I now only enable polymorphism for things that have a type designator. Actually, in some cases it should be possible to enable polymorphism for things without a type designator (pydantic will simply try all classes and takes the first one that validates). However, i think this is a safe approach for now```

👍 

```I have no idea (and I would need help) in order to add tests and documentation.```

- actually, we are in the process of refactoring the testing architecture to make it more approachable.  :). Did you happen to find this [document](https://linkml.io/linkml/contributing/contributing.html)?  It has some good details on setting up a new test.  We can also do a meeting to work through this together.  

In fact, we'd love to have you at our LinkML community meeting in the new year, and I'd be happy to add you to our slack and see if we can iterate more quickly there.  Expect an invite shortly! :) 

```I think the python dataclasses generator needs a similar modification.```

- yes - I think #945 will be the grouping ticket for adding this kind of modification to all our generators.

kervel:
Hello!

* About the slow API's: i don't think they are too slow, but every time i generate code for a slot with a class range, i check if that class has an attribute or slot with a type designator. This will make code generation slower even for models that don't have type designators. For me it seems to be unnoticeable, but maybe for big models it might be a problem. I was just checking if this is acceptable or not.

* I am going to read the document you mentioned, and come up with a test

kevinschaper:
Thanks so much for this PR! Thank you especially for catching that the designates_type field wasn't getting populated anymore. 

I wasn't sure about the semantics of `inlined` and `inlined_as_list` when I wrote the logic that you fixed up, and I have to admit that I set both to true in my schemas where I wanted to inline as a list. I think this interpretation makes sense, if inlined isn't set or is false, and inlined_as_list is true, that should be enough to get the inlined_as_list behavior.  

The only thing that I'm not sure about is populating with the `class_uri` rather than `prefix:camelcase(class_def.name)` 

I know that for my own usage, I want a class_curie rather than a class_uri, but I don't have a good argument for why that should stay as the default behavior. What do you think @sierra-moxon @cmungall ?

kervel:
i opted for the full URI because i tought its more logical, since the generated pydantic code doesn't know anything about prefixes, a curie seems to be something more that is valid in a local scope where the prefix definitions are made explicit...

sierra-moxon:
Likely tangential, but I just wanted to give the example here of identifiers.org changing their base URIs from HTTP to HTTPS,  meaning that those using HTTP identifier.org URIs as keys have a conundrum.  Because of all the churn in this (at least in the biological domain), I would be more confident in storing curis as keys in a datastore than URIs.

cmungall:
@kevinschaper 
> I wasn't sure about the semantics of inlined and inlined_as_list when I wrote the logic that you fixed up, and I have to admit that I set both to true in my schemas where I wanted to inline as a list. I think this interpretation makes sense, if inlined isn't set or is false, and inlined_as_list is true, that should be enough to get the inlined_as_list behavior.

See: 

- #1199

Do you think we should also document the intended behavior of code here too: https://linkml.io/linkml/schemas/inlining.html

kevinschaper:
@kervel Here is an example of a standalone issue-specific test for pydanticgen: https://github.com/linkml/linkml/blob/main/tests/test_issues/test_linkml_issue_1094.py

To avoid generating new test artifacts, my approach has been to run a specific test like this locally, but let the broader test suite run in a PR.

cmungall:
@kevinschaper 
> The only thing that I'm not sure about is populating with the class_uri rather than prefix:camelcase(class_def.name)

I updated #945 to state what the defaults should be. This should be driven by the `range` of the slot that is designated to hold the type

sierra-moxon:
Hi @kervel  - can I ask a favor?  Can you change the destination of this PR from `main` to `kerval_pydantic`?  I created this new branch so that @kevinschaper @cmungall and I can round out the tests on this great addition to the project.  I can give you LinkML branch access as well, just to make development easier.

kervel:
I think my explanation of why i think URI's are better suited as type designators than CURIEs was not very clear, so i'm going to try again.

For me the type designator field is actually a way of expressing the is_a relationship in serialisation formats that don't have it out of the box, like json and yaml. I'd say we should strive to encode the type information in the most universal way possible: a json file generated by pydantic code should be interpretable by other code that is not pydantic. And i think this gives some arguments for using the full URIs.

* A CURIE is actually a short hand notation for an URI, but the CURIE in itself lacks the context needed to expand it to a full URI. Typically, the serialisation formats that don't support is_a out of the box also don't support defining prefixes out of the box. So, we have no way to encode what our prefix means, and the type designator fields contain opaque information that is only readable by code that has a built-in list of prefixes.
* Suppose we have two schemas (say A and B) that import a common third schema (C) with a different prefix. If we use URIs then objects of schema C expressed in one schema can be interpreted in the other, otherwise they can't.

Curies could be used as a "fix" for schema's that change their URIs. But i think the use of this fix is limited (only yaml / json could be "fixed" like this, won't help for rdf or json-ld), and it won't help in case other parts of the URI change.

So in my opinion URIs are more suitable than CURIES in this case. However, if the consensus is to use CURIES anyway, in my specific case it would not pose problems, and i can change the PR back to CURIES.

Awaiting your feedback.

kervel:
> Hi @kervel excited to see this WIP PR! I'm wondering if or how this might affect the linkml validator. In other words, would additional work need to be done to the `linkml-validate` command to support validating schemas that use this approach for modeling polymorphism?
> 
> Also tagging @cmungall

Hello,
this PR in itself won't fix the validator yet. for this, changes are needed in the json schema gen and the python dataclasses gen. I have prepared these changes in [another branch](https://github.com/Kapernikov/linkml/tree/polymorphism_dataclasses), will make a separate PR for that, once

kevinschaper (approved):

Issue #911, opened 2022-08-15
Provide unified validation and repair/coerce behavior


cmungall:
BACKGROUND: 

Currently there are a number of different validation strategies. These are currently all implemented:

1. Use `linkml-validate`, which is a part of the main linkml distribution. This uses jsonschema validator behind the scenes
2. Use https://github.com/linkml/linkml-validator, written by @deepakunni3 
3. Convert a schema to SPARQL queries and run these over an RDF representation of the data
4. Convert a schema to an external representation for framework X, and use X tools to validate
     - Example: convert a schema to SHACL and use pyshacl
     - Example: convert a schema to ShEx and use pyshex

Additionally, when using the loader framework to instantiate dataclasses, validation at the python level will be performed. Complicating the picture, the current dataclass generation framework creates classes that perform **silent coercion** to correct data. While this follow's [Postel's law](https://en.wikipedia.org/wiki/Robustness_principle) it can be confusing, as it means that any validator what uses python objects as an intermediate will not detect classes of error that are silently repaired/coerced/fixed.

@deepakunni3 wrote design principles for validation here
https://linkml.io/linkml-validator/design/

These are great, I will regurgitate verbatim here:

- **Schema Agnostic:** The validator must be schema agnostic and must not make any assumptions of the data outside of the scope of LinkML metamodel. i.e. The validator runs for any linkml schema
- **Plugin Architecture:** Each type of validation is its own plugin. So any subsequent vendor-specific or technology-specific validation scenarios must be its own plugin. For example, `JsonSchemaValidationPlugin` performs JSONSchema validation on one or more objects. There are two types of plugins that are supported:
    - *in-house plugins:* These are plugins that are defined in the linkml-validator repository
    - *external plugins:* These are plugins that live outside of the linkml-validator repository. But one main key characteristic of these external plugins is that they implement the linkml-validator's `BasePlugin` abstract class. This is to ensure that the external plugin class is compatible and plays well with the validator and other plugins
- **Easy to Configure**: Each plugin is instantiated with default arguments but these arguments can be overridden by providing the arguments at runtime
- **Flexible Generators:** Plugins use default generators provided by LinkML. But plugins can also provide ways to accept custom generators. This ensures that if a project is using an extended version of a LinkML generator then it is possible with existing plugins
- **Parseable Validation Messages:** The Validator returns easy to parse validation messages with a defined structure.
    - Each time a validation is run on an object, the Validator returns a `ValidationReport` for that object
[...]

Issue #567, opened 2022-01-19
schemaloader accepts formally incorrect schemas - Add optional jsonschema validation to warn modelers


cmungall:
Continued from https://github.com/microbiomedata/nmdc-schema/pull/239#discussion_r785167685

The schemaloader allows:

 - descriptions to be specified as singleton lists in yaml (even though it is formally singlevalued)
 - comments to be specified as non-list strings, with autoconversion to lists

This is the result of the current post_init design which intentionally performs repairs

Arguably this is handy - the be liberal in what you accept and conservative in what you produce maxim

however, it's potentially confusing and modelers should at least be made aware this is happening

I recommend an optional step in schemaloader that passes through the jsonschemavalidor, default behavior is to warn, but with a strict mode that will fail

sujaypatil96:
@cmungall: @sierra-moxon and I were reviewing our understanding of this issue.

Is the expectation that the JSON Schema generated from the LinkML YAML is validated by a JSON Schema Validator like https://github.com/java-json-tools/json-schema-validator?

Can this issue be closed as a dupe of https://github.com/linkml/linkml/issues/1148?

cmungall:
It’s a duplicate, yes

sierra-moxon:
closing as dupe of https://github.com/linkml/linkml/issues/1148

PR #1189, opened 2022-12-21
Issue 1188

hsolbrig:
Added unit test w/ multiple points of failure.

Fixes will be spread across pythongen.py AND linkml_runtime loaders.


hsolbrig:
I just realized that I have several changes that have to get into linkml-runtime before this can be committed.  We might consider making linkml-runtime a submodule here so that we can get changes of this sort in and tested in one go-round.

cmungall:
Let me try and summarize this PR:

This PR introduces a change to generated dataclasses; instead of `as_dict`, a new `_normalized_assignment` method in YAMLRoot is called (https://github.com/linkml/linkml-runtime/pull/236/), replacing previous uses of `as_dict`.

The motivation for this change is the desire to have auto-repair happen on instantiation, such that if you have a class

```
  C1:
    description: class with an inline list of strings
    attributes:
        thelist:
            multivalued: true
```

You can do a shortcut instantiation:

```python
C1(['a', 'b', ...])
```

such that it is auto-repaired/coerced into the equivalent of a call to

```python
C1(thelist=['a', 'b', ...])
```

I think it's worth discussing such a feature, but this would be a new enhancement that should be introduced with a new top level issue in this repo which the group can discuss.

Personally, I am generally in favor of Postel's principle, but I think here it is clearer if there is no repair magic/coercion happening. Such coercion can lead to hard to predict behavior and hard code to debug; see

 - #911
 - #567

In this case I don't think it is asking a lot for the user of dataclasses to instantiate using the correct form:

```python
C1(thelist=['a', 'b', ...])
```

hsolbrig:
There are at least two motivations behind these proposed changes:

1) (IMO) The syntax for loading a root node and the target of a slot should be identical.  Say one has the following model:

```yaml
classes:
   ListOStrings:
        attributes:
            entry:
                multivalued: true
                range: string

   StringsAndAnINT:
         attributes:
              the_strings:
                  range: ListOStrings
              the_int:
                  range: integer
```
IF one is able to load the following YAML using `YAMLLoader().load_any(text, StringsAndAnINT)`

```yaml
the_strings: [s1, s2, s3]
```

One should also be able to load the follwing (using `YAMLLoader().load_any(text, ListOStrings)`
```yaml
[s1, s2, s3]
```

The notion being that the property `the_strings` identifies the inner type. 


2) This raises the  question, however, as to why we should try to infer anything.  We could _require_:
```yaml
the_strings:
    entry: [s1, s2, s3]
```

in the first example and
```yaml
entry: [s1, s2, s3]
```
And the examples above would be flagged as errors.

Similarly, we could require that lists of dictionaries be exactly that -- lists, meaning that LinkML classes would have to be defined as:
```yaml
classes:
    - name: ListOStrings
      attributes:
      - name: entry
         multivalued: true
         range: string

      ...
```
And that two element classes have to be named, meaning that the LinkML prefixes definition would appear as:
```yaml
prefixes:
   - prefix_prefix: linkml
      prefix_reference: https://w3id.org/linkml/
   - prefix_prefix: skos
      prefix_reference: http://www.w3.org/2004/02/skos/core#
       ...
```

This is a very different philosophy to the one that I've been using to date. My goal had been to make the _YAML_ (or _JSON_ or _CSV_ or _..._) easy to write, where the tooling does its best to try to understand the author's intent.  An alternate philosophy is, instead of trying to guess the author's intent, establish a set of rigorous, documented rules that allow for one way to accomplish a task.

Perhaps the best way to compare the two approaches would be to compare what goes _in_ to the yaml loader in many of the models we are using with what comes _out_.

I suspect that some of the differences may be that we are writing to different audiences.  My target audience had been the JSON, YAML and/or SpreadSheet authors -- folks that already had an implicit model and had already created data around it -- who realized that they needed more documentation and rigor.  A different audience entirely would be application developers who want to start with a fixed, rigorous model before a line of data is generated.

Both approaches have their uses (IMO) and both have their drawbacks.  

One should note that this enhancement didn't break anything -- with one exception, everything that worked before continued to work as advertised.   It simply allowed some of the new models that I'd created to behave as expected.
The one exception is in [test_loader_dumpers.py](https://github.com/linkml/linkml-runtime/pull/236/commits/2e57d41ff9097acbe1188c305c67795eb5fb56da#diff-30cf2b0e44842a9b981a1e035e3e4960077594f1974de8739618a0efe2b97cb0)

Where the the test code stated that both a single instance and a _list_ of instances could satisfy a model of a given class.  Interestingly, this is the one place where I propose _tightening_ the rules.  Given the model:
```yaml
classes:
    Person:
         attributes:
             name:
                range: string
                identifier: true
             age:
                range: int

     Position:
         attributes:
               id:
                  range: string
                  identifier: true
               performer:
                  range: Person
```
The pre-PR loader, invoked with: `YAMLLoader().load_any(text, Position)`
```yaml
id: 17
performer:
    - name: Sam
      age: 22
    - name: Jane
      age: 43
```
as an error, yet would allow `YAMLLoader().load_any(text, Person)`

```yaml
  - name: Sam
    age: 22
  - name: Jane
    age: 43
```
The PR would require the modeler to differentiate between a single person and a _list_ of persons.

cmungall:
I realize the mistake in my previous comment. Of course, this shortcut instantiation should be allowed:

`C1(['a', 'b', ...])`

 this is just standard positional argument passing. This should already be treated the same as `C1(thelist=['a', 'b', ...])`

What I don't understand is why we need this PR to make this work - however, I will take another look, I just wanted to report my initial incorrect assumption here.

sierra-moxon:
I definitely appreciate the "shortcut" syntax I'm able to write for classes, slots, etc. in the model.   However, when I go to use a new (to me) metamodel component it is a bit tricky to figure out how to do so.  

For example, when trying to add "annotations" to classes in the Biolink Model, I went through several syntax variations -- starting with the most verbose to match what I interpreted the spec to be (this causes an error, `ValueError: tag must be supplied`):

```
  acts upstream of:
    is_a: related to at instance level
    domain: gene or gene product
    range: biological process
    annotations:
      - extension_tag: biolink:canonical_predicate
        extension_value: true
    exact_mappings:
      - RO:0002263 
 ```

then I tried this (I think I went looking around for tests and examples of fields that were similarly modeled to annotations/extensions and then for schemas/example schemas that were using those components.  it might've been [this one that I found](https://github.com/linkml/linkml/blob/a4a21e96e0883b787a1f9a5c6815fca48a8afe8c/tests/test_generators/output/personinfo.relational.yaml#L195), but I can't remember).  This generates an error: `
ValueError: Unrecognized entry: canonical: [{'tag': 'biolink:canonical_predicate', 'value': True}]`

```
  acts upstream of:
    is_a: related to at instance level
    domain: gene or gene product
    range: biological process
    annotations:
       canonical:
         - tag: biolink:canonical_predicate
           value: true
    exact_mappings:
      - RO:0002263 
 ```

then with more help, I got to here, which works, but is a little awkward.

```
  acts upstream of:
    is_a: related to at instance level
    domain: gene or gene product
    range: biological process
    annotations:
      - tag: biolink:canonical_predicate
        value: true
    exact_mappings:
      - RO:0002263 
 ```

and finally with Chris's help, to here:

```
  acts upstream of:
    is_a: related to at instance level
    domain: gene or gene product
    range: biological process
    annotations:
      canonical_predicate: true
    exact_mappings:
      - RO:0002263
```

probably more than one issue going on here and maybe this example is only tangentially related to this PR, but maybe documentation of the transformation rules would be a step towards having it both ways?

**Comment by cmungall** [comment]
Thanks both for your comments!

Let's start by summarizing *existing* shortcut patterns and means of inlining collections.

- the ability to inline as dictionaries is incredibly useful. This was one of the first innovations @hsolbrig made on the original biolinkml framework, I can't imagine living without this!
     - I think this is fairly well documented: https://linkml.io/linkml/schemas/inlining.html#inlining-as-a-dictionary
     - an additional FAQ entry and notebook would be welcome though!
     - overall I don't think this is so contentious, it adds only minimal complexity to the framework and the gain is large
     - I think we do need an option in the existing dumpers that will remove the redundant key though
- the trick to inline as single-value dictionary elements is also incredibly useful
    - This is a bit more "clever" and *not* so well documented, and the odyssey you went through is not unique @sierra-moxon 
    - However, this is easily fixed! Let's make better documentation for it!
    - #1205
    - note that without this feature we'd have no way to handle dicts of primitives in pre-existing yaml/json, which would be a major limitation!

I don't think we need to revisit any of these. They will obviously stay as part of the framework, since changing this would be a major backwards incompatible change.

The subject of this PR is a *new* shortcut. As a general SOP, it would be great if when proposing a new big change like this we started from an issue rather than a PR, with all of the rationale laid out! We can then discuss this as a group in the developers call.

There are always trade-offs here. I think we all want the YAML to be easy to author, and the existing dictionary inlining behavior are great examples. We also want to be able to schematize all possible YAML/JSON shapes, and the current behavior is necessary for that.

However, I think the tradeoffs get more nuanced the more we apply Postel's principle, and try and do automatic repairs that guess the users intent. If there are too many valid ways to say the same thing, then it can be confusing for the user. It can also complicate the overall framework (for example: the json schema generator has to accommodate all the alternative ways of saying things, if we deem these to be valid). In my experience with the LinkML user base so far, people actually appreciate consistent rules. Rather than trying to guess intent, we should just document the different ways that something can be done, make things easy, but have simple consistent rules.

The subject of this PR is a new shortcut rule that  would make

`load_any([1,2,3], ListOfStrings)`

valid, given a schema:

```yaml
classes:
   ListOStrings:
        attributes:
            entry:
                multivalued: true
                range: string
```

as things currently stand, the user would have to pass in a dict `{entry:[1,2,3]}`

My own position on this is that this is not something I have seen a user requesting, and I think introducing this can complicate things without any real gain.

However, I may be in the minority, or missing something - we can still put a proposal together and open this up more broadly. The proposal should lay out the benefits and state precisely what the repair rules are. If there is broad desire from the community we can go ahead and add this in linkml 1.6, resource permitting. I think if we did it, we would want to increase the minor version as this proposed change could potentially change behavior.

Regarding existing `load_any` behavior - the current behavior is intentional, but could be better documented. The use case here is that a user may have a yaml/json file that is a serialization of *either* a valid YAMLObject *or* a list of YAMLObjects (which do not necessarily have a container). The previously existing `load`/`loads` was not able to handle top level lists before.

hsolbrig:
"My own position on this is that this is not something I have seen a user requesting, and I think introducing this can complicate things without any real gain."

This fix exists because of a related problem that I (as a "user") had encountered that prevented me from using a model the way I had intended.  In the process of fixing the other issue, I wanted to make sure that I caught the edge cases.  It would actually be kind of difficult to fix the issue I'd encountered *without* fixing this one.

To be clear - I didn't submit this set of fixes on whim.  I'd seen @kevinschaper wrestling with what appeared to be the same bug on conference calls and, when I encountered it in my own use cases I went ahead and tried to fix the the general case.   So there is "real gain", in that I can proceed to use LinkML to represent my model (vs. changing my model to meet LinkML's view of the world). IMO, another "real gain" is that LinkML is more likely to work "out of the box", vs. an end user having to file bug reports, change requests, etc.

PR #1199, opened 2023-01-03
Test case for pydantic inlining

cmungall:


kevinschaper (approved):
I'm not sure if I should be approving this or requesting extra review, but I approve of the test part and now it's passing.

PR #1202, opened 2023-01-04
Fix non-stopping CI on windows

### LINKED ISSUE #1200: windows testing on PR does not register errors
**Describe the bug**
Running tests on windows platforms via github actions register "successful" even when tests fail on windows.

An example: https://github.com/linkml/linkml/actions/runs/3832945778/jobs/6523847485 

Note the exact same failures in the ubuntu version of this PR:
https://github.com/linkml/linkml/actions/runs/3832945778/jobs/6523847298

but the coverage report is still generated in windows tests instead of erroring out on the test failures, which leads to a "green check" in github actions interface. 
Screen Shot 2023-01-03 at 2 42 18 PM

**Expected behavior**

windows OS github action fails when tests fail on windows platform via github actions.
it might be that we just need to capture a different error code on the windows platform than is being captured on ubuntu?

dalito:
As suggested in https://github.com/actions/runner-images/issues/6668

Update: This fixes the problem but reveals that the tests on Windows produce (overlooked) errors which should be fixed before merging this. --> Done in #1206

Closes #1200

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -77,7 +77,7 @@ jobs:
           poetry run coverage run -m unittest discover
           poetry run coverage xml
           poetry run coverage report -m
-
+        shell: bash
       #----------------------------------------------
       #           upload coverage results
       #----------------------------------------------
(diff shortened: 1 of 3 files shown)


cmungall:
looks like the sqlite tests aren't passing - not sure if this is due to bad practice on our side, e.g. not releasing a lock...?

dalito:
Yes. Windows is less tolerant on missing clean-up of sqlAlchemy sessions/pools.

sierra-moxon (approved):

PR #1204, opened 2023-01-04
Add a mermaid ERDiagram generator.

### LINKED ISSUE #943: Add generator for ER diagrams (using mermaid or erdantic)
@ialarmedalien uses https://erdantic.drivendata.org/stable/ to generate ER direct from generated python

I suspect this would have some issues with our generated dataclasses, but I think it would do well with pydanticgen

It's also possible to do directly with mermaid
https://mermaid-js.github.io/mermaid/#/entityRelationshipDiagram

- The python route has the advantage for python-centric projects of fitting in well with that workflow
- mermaid would have similar look and feel

however neither seem to support the full ER set of symbols

image

In particular diamons for relationships - it would be good to show these for  reified relationships like biolink associations or PersonInfo:FamiliarRelationship

From #611

cmungall:
Fixes #943


linkml/generators/erdiagramgen.py
```diff
@@ -0,0 +1,307 @@
...
+
+        :return: mermaid string
+        """
+        sv = self.schemaview
+        structural_roots = [cn for cn, c in sv.all_classes().items() if c.tree_root]
+        if self.structural and structural_roots:
+            return self.serialize_classes(structural_roots, follow_references=True)
+        else:
+            diagram = ERDiagram()
+            for cn in sv.all_classes(imports=False):
```
sierra-moxon, line 155: should this be `for cn in sv.all_classes(imports=True):` ?

sierra-moxon (approved):

linkml/generators/erdiagramgen.py
cmungall: yes, this should probably be the default

PR #1211, opened 2023-01-06
slots modifying classes table

amc-corey-cox:
Creating a draft PR for testing

### DIFF
--- a/linkml/generators/docgen/slot.md.jinja2
+++ b/linkml/generators/docgen/slot.md.jinja2
@@ -39,8 +39,8 @@ URI: {{ gen.uri_link(element) }}
 
 | Name | Cardinality and Range | Description | Inheritance |
 | ---  | --- | --- | --- |
-{% for c in schemaview.get_classes_by_slot(element, include_induced=True) -%}
-{{ gen.link(c) }} | {{ schemaview.get_class(c).description }}
+{% for c in schemaview.get_classes_by_slot(element) -%}
+{{ gen.link(c) }} | {{ schemaview.get_class(c).description|enshorten }}
 {% endfor %}
 
 {% if schemaview.is_mixin(element.name) %}
--- a/linkml/generators/docgen/slot.md.jinja2
+++ b/linkml/generators/docgen/slot.md.jinja2
@@ -35,6 +35,14 @@ URI: {{ gen.uri_link(element) }}
 
 {% endif %}
 
+## Slot Modifying Classes
+
+| Name | Cardinality and Range | Description | Inheritance |
+| ---  | --- | --- | --- |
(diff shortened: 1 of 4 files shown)


linkml/generators/docgen/slot.md.jinja2
```diff
@@ -32,6 +32,14 @@ URI: {{ gen.uri_link(element) }}
...
 {% endif %}
 
+## Slot Modifying Classes
+
+| Name | Description |
+| ---  | --- |
+{% for c in schemaview.get_classes_by_slot(element) -%}
+{{ gen.link(c) }} | {{ schemaview.get_class(c).description|enshorten }}
+{% endfor %}
+
```
sujaypatil96, line 42: Great, you're showing the important columns: name of the class where the slot is being modified and its description, which is also useful.

sujaypatil96: @amc-corey-cox: I was looking at this PR once again, and I just noticed that you're making a call to `get_ckasses_by_slot()`. I think you were meaning to use `get_classes_modifying_slot()` instead?

This might require a change in your tests too.

sujaypatil96 (changes requested):
You might need to update the poetry.lock file on this branch so that it uses the latest version of linkml-runtime that implements this method.

amc-corey-cox:
I updated the poetry.lock file (to what main was using) and fixed the get_classes_modifying_slot function now. I'm still getting a test failure on my system but we'll see if it fails on GitHub. If it still fails, I may need a little help figuring out how to fix the test failure... it is in a file I didn't touch.

amc-corey-cox:
This depends on changes I just made in linkml-runtime so I expect builds will fail until we have that checked in.

linkml/generators/docgen/slot.md.jinja2
amc-corey-cox: Yes, I had several things that I needed to fix. I've taken care of this and fixed a few other things. It appears to be working now. All that's left is to figure out how to get all the test suite to pass on GitHub.

amc-corey-cox:
I think this is ready to go. Let me know if you see anything else.

sujaypatil96 (approved):
Looks good to me @amc-corey-cox, thank you for working on this. I can see this changing further in the future, so I'm happy to collaborate on improving this. With this PR the _applicable classes_ table on slot documentation pages will start showing additional information about whether or not a class modifies a slot. Now the next natural question that a user might ask is how is a particular slot being modified in the context of that class? For that we will need to prioritize some properties like requiredness, cardinality, multivaluedness, range, etc. and show them as part of the _applicable classes_ table on slot documentation pages.

PR #1212, opened 2023-01-09
Polymorphism using type designator: add support in python and jsonschema generators, so that linkml-validate works with polymorphic data

kervel:
This is a follow-up on PR #1180 which needs to be merged first.

some open questions:

* i created linkml.generators.common to put one common function. is this good or should it go elsewhere ?
* i shared data and schema between my two tests. is this good ? or should i merge all tests together in one file ?


kervel:
rebased

kervel:
will close this one and create a new one based on a linkml branch.

Issue #1213, opened 2023-01-09
Add unique_key validation to JsonSchemaGenerator


turbomam:
I will write a test if nobody has an example

cmungall:
Unfortunately this might not be possible: https://stackoverflow.com/questions/24763759/make-sure-item-property-in-array-is-unique-in-json-schema

PR #1223, opened 2023-01-12
#1213 SchemaBuilder disregards requested range

turbomam:
Tests for #1213

see tests/test_issues/test_issue_3.py

please disregard deleted "files changed" in `tests/test_issues/output/curie_case.context.jsonld`


turbomam:
If I really have a need for using SchemaBuilder, then I should just learn how to use SchemaBuilder.

Issue #1214, opened 2023-01-10
gen-doc has bad runtime characteristics


jankatins:
**Describe the bug**

I've a big schema: CGMES, generated from multiple UML diagrams (see e.g. http://sogno.energy/cimpy/cimpy.cgmes_v2_4_15.html and https://ontology.tno.nl/IEC_CIM/, who also generated most of it from a simialr source). The yaml file has 4.700 lines, gen-doc produces 600+ files. Most attributes are added via slots and the class hierarchy uses lot's of inheritance where slots are defined on the parent. 

Running `gen-doc --metadata -d eq/docs eq/all.yaml` takes >10min.

I spied on it with `py-spy` and this is the observed trace:
  
![image](https://user-images.githubusercontent.com/890156/210811182-63d08769-f7b4-4d0c-97f1-8571567dd6c7.png)

I interpret that that the two deepcopies are at fault here in `SchemaView.induced_slot()`:

```
induced_slot = deepcopy(slot)
...
return deepcopy(induced_slot)
...
```

I'm unsure why this needs two deepcopies, at least the last on the return line seems unneeded (at least to my eys: every assignment is of a scalar or deepcopied itself).

The `__bool__` line seems to be 

```
        if class_name:
            cls = self.get_class(class_name, imports, strict=True)
        else:
            cls = None
...
        if cls: # <- THIS, maybe if cls is not None
...
            if not cls: # <- THIS, maybe if cls is None

```

I updated to the 1.3.16 and did the above changes (remove second deepcopy, change to is None checks instead of conversion to bool) and it reduced one runtime from ~17min to ~11min. Unfortunately there were differences ("just" changed order of attributes, as far as I can see), but I have similar changes with the unchanged verison as well (seems to be either unstable sorting in slots or a difference to an old version which we used to check in these generated doc files).

<details><summary>Stack details!</summary>
[...]

PR #1232, opened 2023-01-13
Let both sides of a linking tables include pk ids

jankatins:
Up to now, if the target table had an identifier column, such linking tables
had one column with pk ids (targetA_pkid) and one without (just targetB).
This now let both sides include the PK id.

Also add a test case for the "target table has identifier column" case

### DIFF
--- a/linkml/transformers/relmodel_transformer.py
+++ b/linkml/transformers/relmodel_transformer.py
@@ -305,6 +305,9 @@ class RelationalModelTransformer:
                             comments=[f"Linking class generated from {cn}.{slot_name}"],
                         )
                         slot.name = sn_singular
+                        # On the linking table, it's inlined.
+                        # This triggers that the slot.name gets appended with the pk column name on the target side
+                        slot.inlined = True
                         add_attribute(linker_class.attributes, backref_slot)
                         add_attribute(linker_class.attributes, slot)
                         slot.slot_uri = "rdf:object"
@@ -328,7 +331,7 @@ class RelationalModelTransformer:
                                 MultivaluedScalar(
                                     source_class=cn,
                                     source_slot=src_slot.name,
-                                    target_slot=slot.name,
+                                    target_slot=sn_singular,
                                     join_class=linker_class.name,
                                 )
                             )
--- a/tests/test_generators/test_sqltransform.py
+++ b/tests/test_generators/test_sqltransform.py
@@ -1,4 +1,5 @@
 import unittest
+import itertools
 
 from linkml_runtime.dumpers import yaml_dumper
 from linkml_runtime.linkml_model import SlotDefinition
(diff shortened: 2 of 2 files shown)


jankatins:
This is part of changes we had to do to generate SQLAlchemy files for a 120+ Tables CGMES data model (see https://github.com/linkml/linkml/issues/1214)

jankatins:
Ok, something is strange here: In my dataset, all classes have identifiers and so the target class has an "identifier slot" and therefore the code below has `is_inlined` as `False`, as `source_sv.get_identifier_slot(tc.name)` returns a slot and therefore the second part of the OR becomes "not True". And therefore the pk column name is not added via `a.alias = f"{a.name}_{tc_pk_slot.name}"` and so the `_id` is missing for one column.

([link to code in main](https://github.com/linkml/linkml/blob/main/linkml/transformers/relmodel_transformer.py#L368-L384))
```

                    is_inlined = a.inlined or not source_sv.get_identifier_slot(tc.name)
                    if (
                        fk_policy == ForeignKeyPolicy.INJECT_FK_FOR_NESTED
                        and is_inlined
                        and not a.multivalued
                    ) or (fk_policy == ForeignKeyPolicy.INJECT_FK_FOR_ALL_REFS):
                        # if it is already an injected backref, no need to re-inject
                        if "backref" not in a.annotations:
                            del c.attributes[a.name]
                            if "forwardref" not in a.annotations:
                                add_annotation(a, "original_slot", a.name)
                            a.alias = f"{a.name}_{tc_pk_slot.name}"
                            a.name = a.alias
                            c.attributes[a.name] = a
```

In the [unit tests](https://github.com/linkml/linkml/blob/main/tests/test_generators/test_sqltransform.py#L169), the dataset is setup to not contain an identifier attribute. So the [code invents one](https://github.com/linkml/linkml/blob/main/linkml/transformers/relmodel_transformer.py#L228-L230) but for some reason, this invented identifier is not returned as "identifier slot" and so the code jumps into the if block and adds the `_id` to the column name.

If I change the setup to include an `id` column with identifier=True, the test pass again (or would fail with the current code):
```
        b = SchemaBuilder()
        slots = ["name", "description", "has_ds"]
        b.add_class("c", slots).add_class("d", ["id", "name"]).set_slot(
            "has_ds", singular_name="has_d", range="d", multivalued=True, inlined=False
        ).set_slot("id", identifier=True)
```

No clue whats going on here, but I can see three sources and I've no clue right now what is wrong here:

1. Should `source_sv.get_identifier_slot` return the invented primary key (`RelationalModelTransformer.get_direct_identifier_attribute(target_sv, tc.name)` does reuturns something, which not `source_sv.get_identifier_slot`?)
2. Should `RelationalModelTransformer.add_primary_key()` do something differently so that `source_sv.get_identifier_slot()` returns the slot?
3. Should the logic be changed to ignore if there is an "indentifier slot"? E.g. use `RelationalModelTransformer.get_direct_identifier_attribute(target_sv, tc.name)` instead.

Any advise would be helpful, as right now the above test setup is IMO equally valid as the one with "invented" IDs.

jankatins:
I added the mentioned test case and removed my changes to see if this triggers the failure as I expect.

jankatins:
Or just use the most easy fix: the slot on the linking table is inline, so the code which creates the link table should change the inline attribute accordingly. And now hopefully the tests work :-)

cmungall:
Thanks @jankatins!

Let me check I understand the motivation for the PR

Let's say we have a schema with a multivalued link between `c` and `d`, where (1) `c` does not have a native primary key and (2) `d` does

i.e:

```yaml
name: test-schema
id: http://example.org/test-schema
default_prefix: http://example.org/test-schema/
classes:
  c:
    attributes:
      name: {}
      description:
      has_ds:
        singular_name: has_d
        multivalued: true
        range: d
  d:
    attributes:
      id:
        identifier: true
      name: {}
```

currently this generates SQL where a surrogate primary key is injected into `c`, and a new linking table/class is created:

```sql
CREATE TABLE c (
	id INTEGER,
	name TEXT,
	description TEXT,
	PRIMARY KEY (id)
);
CREATE TABLE d (
	id TEXT,
	name TEXT,
	PRIMARY KEY (id)
);
CREATE TABLE c_has_d (
	c_id TEXT,
	has_d TEXT,
	PRIMARY KEY (c_id, has_d),
	FOREIGN KEY(c_id) REFERENCES c (id),
	FOREIGN KEY(has_d) REFERENCES d (id)
);
```

I think the *structure* that is generated here is as intended.

Your test case seems to impose a *terminological* expectation, that the FK in the linking table should be named `d_id` to be consistent with `c_id`.

I think this is a perfectly reasonable terminological expectation. Currently, the transformation attempts to preserve the original schema names as far as possible, even if when moving a multivalued slot to generated linking table. But I can see this this leads to a kind of asymmetrical terminology choice in the linking table.

Note you do have some degree of terminological control here, in that you could make the `singular_name` be `d_id`, but I think this is a bit unsatisfying (there may be other uses of `singular_name` outside the scope of SQL).

We could explore making the default be to always use `d_id` in the case above. However, note that we would need to deal with the self-linking scenario.

My preference would be to keep the default behavior as-is, but introduce a mechanism for giving you more fine-grained control of terminological choices for generated classes and slots. These could be any of:

1. a configuration option at the level of the generator
2. per-slot choices, via annotations or aliases on the slots themselves

If I am on the right track, then my preferred course of action here would be to first amend this PR to remove the logical change to the transformed and to amend the test to:

```python
      self.assertCountEqual(["c_id", "has_d"], list(c_has_d.attributes.keys()))
      self.assertCountEqual(["c_id", "has_d"], get_primary_key_attributes(c_has_d))
      self.assertDictEqual(
            {"c_id": "c.id", "has_d": "d.id"}, get_foreign_key_map(c_has_d)
        )
```

this would serve as additional docs/tests on the current behavior. We can then iterate on the solution for terminological fine grain control

It's possible I have misunderstood something and this isn't a terminological/naming thing, and in fact there is a desired change to the structure of the transformed model/DDL.

lmk if you want to jump on a zoom to discuss. Our developer calls are probably a bit late for you (1pm pacific). We have a community call this coming thursday at 8am pacific which you're welcome to join.

cmungall (changes requested):
needs further discussion

jankatins:
Not quite:

In my opinion, it should not matter if the original two classes already have an "id" identifier attribute or not (as the code invents one). So in the setup also both could have a id identifier column, the issue only happens when the second has one, the first column gets the id appended in another place.

> Your test case seems to impose a terminological expectation, that the FK in the linking table should be named d_id to be consistent with c_id.

Nope, I just want to have the second column also have the "_id" suffix, like every other generated fk column. In your example above, you have 

```
CREATE TABLE c_has_d (
	c_id TEXT,
	has_d TEXT, # <= HERE
	....
);
```

I want it

```
CREATE TABLE c_has_d (
	c_id TEXT,
	has_d_id TEXT, # <= HERE
	....
);
````

And this is also how it comes out, if there is no initial identifier "id" column on d (so it gets invented). So this pr removes this inconsistency.

jankatins:
I changed the current test to test all 4 cases of one or both of the involved tables having an id identifier column or not. All 4 cases should result in the same SQL, which is now true after this change. I also removed the change and the expected two cases were failing.

cmungall:
Thanks for your patience!

Yes, you are quite correct. There original slot names are only preserved in the single-valued range case. In the multivalued case where a linking table is created, it is indeed de-novo creation of slots.

I like the approach in the test of using itertools to test all the different combinations. We will soon have an improved specification of some of the generators, where the different combinations of possibilities are laid out.

This PR is good to go. I tested it with some other combinations of possibilities and will add these later after this is merged.

THANK YOU for you contribution 🙏 🙏 🙏

cmungall (approved):

Issue #1239, opened 2023-01-17
test: check slots/values retained by merge in SchemaView


turbomam:
see also #1238

matentzn:
Please reopen if still relevant!

PR #1240, opened 2023-01-17
report of induced schema def attribs for #1238

turbomam:
- #1238

### DIFF
--- a/tests/test_issues/test_issue_1238.py
+++ b/tests/test_issues/test_issue_1238.py
@@ -1,0 +1,78 @@
+import unittest
+
+from linkml_runtime import SchemaView
+
+META_URL = "https://raw.githubusercontent.com/linkml/linkml-model/main/linkml_model/model/schema/meta.yaml"
+SCRUTINIZED_CLASS = "schema_definition"
+
+EXPECTED_INDUCED_ATTRIBS = ['aliases',
+                            'alt_descriptions',
+                            'annotations',
+                            'broad mappings',
+                            'categories',
+                            'classes',
+                            'close mappings',
+                            'comments',
+                            'conforms_to',
+                            'default_curi_maps',
+                            'default_prefix',
+                            'default_range',
+                            'definition_uri',
+                            'deprecated',
+                            'deprecated element has exact replacement',
+                            'deprecated element has possible replacement',
+                            'description',
+                            'emit_prefixes',
+                            'enums',
+                            'exact mappings',
+                            'examples',
+                            'extensions',
+                            'from_schema',
+                            'generation_date',
+                            'id',
+                            'id_prefixes',
+                            'imported_from',
+                            'imports',
+                            'in_language',
+                            'in_subset',
+                            'keywords',
+                            'license',
+                            'local_names',
+                            'mappings',
+                            'metamodel_version',
+                            'name',
+                            'narrow mappings',
+                            'notes',
+                            'prefixes',
+                            'rank',
+                            'related mappings',
+                            'see_also',
+                            'settings',
+                            'slot_definitions',
+                            'slot_names_unique',
+                            'source',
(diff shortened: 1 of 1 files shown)


tests/test_issues/test_issue_1238.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class SchemaDefInducedAttribsCase(unittest.TestCase):
+    def test_induced_attrib_list(self):
+        meta_view = SchemaView(META_URL)
+
+        induced_schema_def = meta_view.induced_class(SCRUTINIZED_CLASS)
+        induced_schema_def_attribs = induced_schema_def.attributes
+        induced_schema_def_attrib_names = [v.name for k, v in induced_schema_def_attribs.items()]
+        induced_schema_def_attrib_names.sort()
```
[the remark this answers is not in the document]
turbomam: Thanks. How about composition/list members equal?

tests/test_issues/test_issue_1238.py
```diff
@@ -0,0 +1,78 @@
...
+                            'todos',
+                            'types',
+                            'version']
+
+
+class SchemaDefInducedAttribsCase(unittest.TestCase):
+    def test_induced_attrib_list(self):
+        meta_view = SchemaView(META_URL)
+
+        induced_schema_def = meta_view.induced_class(SCRUTINIZED_CLASS)
```
cmungall, line 69: consider naming the variables to be consistent with the type. This sounds like it should hold a SchemaDefinition object, not a ClassDefinition object.

tests/test_issues/test_issue_1238.py
```diff
@@ -0,0 +1,78 @@
...
+        induced_schema_def_attribs = induced_schema_def.attributes
+        induced_schema_def_attrib_names = [v.name for k, v in induced_schema_def_attribs.items()]
```
cmungall, line 71: This is simpler:
`list(induced_schema_def_attribs.keys())`

tests/test_issues/test_issue_1238.py
```diff
@@ -0,0 +1,78 @@
+import unittest
+
+from linkml_runtime import SchemaView
+
+META_URL = "https://raw.githubusercontent.com/linkml/linkml-model/main/linkml_model/model/schema/meta.yaml"
```
cmungall, line 5: tests should be self contained and have no network dependencies

Additionally this test will break as soon as the metamodel changes and a new metaslot is added to ClassDefinition

turbomam: OK, I'm adding to this now but will ultimately remove the test.

If I did want to retain this test, is the metamodel available from linkml/linkml?

Also, what does it mean to add a metaslot to `ClassDefinition`, as opposed to a slot? Does "metaslot" emphasize that it's a slot from the metamodel, as opposed to a user schema? Is metaslot defined anywhere?

turbomam: Thanks for the review

turbomam: This was a starting point for 
- #1239

cmungall: > OK, I'm adding to this now but will ultimately remove the test.

I see. I encourage PRs for test cases but if the goal is to do entirely local testing best not to make a PR

> Also, what does it mean to add a metaslot to ClassDefinition, as opposed to a slot? Does "metaslot" emphasize that it's a slot from the metamodel, as opposed to a user schema? Is metaslot defined anywhere?

a metaslot is a slot in the metamodel; see 
https://linkml.io/linkml-model/docs/specification/03schemas/#schema-basics

turbomam:
Whoa, this PR has great comments. But I don't have time to work though it any more now so will close.

PR #1241, opened 2023-01-17
Additional test for SchemaView induced_slot #1238

turbomam:
- #1238


cmungall:
See my comments on #1240 - did you mean to make two PRs? They seem identical?

turbomam:
> See my comments on #1240 - did you mean to make two PRs? They seem identical?

Thanks for comments. 

I'll make a push to the other PR in a few minutes. You'll see that they aren't intended to be identical, but rather build upon each other.

They're also still both in draft.

turbomam:
I think my schema view induction needs are met now. Closing.

PR #1252, opened 2023-01-20
Support ifabsent in pydanticgen

kevinschaper:
Starting as a draft with just the failing test, and many many thanks to @djarecka for laying out the test case so nicely.

### DIFF
--- a/docs/schemas/slots.md
+++ b/docs/schemas/slots.md
@@ -115,6 +115,16 @@ A class must not have more than one identifier (asserted or derived). `identifie
 If you need to mark additional fields as unique, or a collection of slots that when considered as a tuple are unique, use
 `unique_keys` (see the [constraits](constraints.md) section of the docs).
 
+## Type designator
+
+The `designates_type` slot can be used to indicate the type or category of instances of a class.
+
+```yaml
+slots:
+  category:
+    designates_type: true
+```
+
 ## Slot cardinality
 
 ### multivalued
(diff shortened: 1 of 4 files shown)


kevinschaper:
Looking at using the existing ifabsent code, I'm a little confused about the mapping

```
    (r"[Tt]rue", False, lambda _, __, ___, ____: "True"),
    (r"[Ff]alse", False, lambda _, __, ___, ____: "False"),
    (r"int\(([-+]?[0-9]+)\)", False, lambda m, __, ___, ____: int(m[1])),
    ("class_uri", False, lambda _, __, class_definition, ____: 'class_class_uri'),
    ("class_curie", False, lambda _, __, class_definition, ____: 'class_class_curie'),
    ("slot_uri", True, lambda _, loader, ___, slot_definition: f'slots.{slot_definition.name}.uri'),
    ("slot_curie", True, lambda _, loader, ___, slot_definition: f'slots.{slot_definition.name}.curie'),
    ("default_range", False, lambda _, loader, __, ____: f"{strval(loader.schema.default_range)}"),
    ("bnode", False, lambda _, __, ___, ____: "bnode()"),
    (r"string\((.*)\)", False, lambda m, __, ___, ____: strval(m[1])),
    (r"uri\((.*)\)", False, lambda m, loader, _, __: uri_for(m[1], loader)),
    ("default_ns", False, lambda _, loader, __, ____: f"{strval(loader.schema.default_prefix)}"),
```

It seems to want integer values to be wrapped with `int()` and string values to be wrapped with `string()`  - but I assume in the schema yaml it would just be expressed as a raw value. I guess I need to dig deeper.

kevinschaper:
The one bit of extra attention that I want this PR to get is on the documentation changes, I included a section on designates_type that isn't strictly related to the PR, but seemed like handy thing to add while I was there.

cmungall (approved):

Issue #1249, opened 2023-01-20
Issue validating RDF with LinkML-generated SHACL due to the shapes being closed


vemonet:
**Describe the bug**

Validating BioLink RDF with SHACL generated by LinkML (for the biolink-model) fails due to the shapes being `sh:closed`.

Because the SHACL specs specify that an entity should complies with the shapes of all its parent classes.

For example when trying to validate a `biolink:Drug`, the SHACL engine will check if the drug entity complies with the `biolink:NamedThing` shape, which is closed. The `biolink:Drug` shape can have new properties that are not in the `biolink:NamedThing`, so validating a Drug will always fail due to its parents classes being closed

I discussed this issue with the pySHACL developers in https://github.com/RDFLib/pySHACL/issues/173, and since it is specified in the SHACL specs our best option would be to fix our shapes to make them `sh:closed false ;`

A first fix could be to make all NodeShape `sh:closed false` here: https://github.com/linkml/linkml/blob/5ecd9c402b7e8f5ad1775904a9247df26a1918fa/linkml/generators/shaclgen.py#L62

And if necessary in the future we could fine tune the generator to make specific classes `sh:closed true` (e.g. classes that don't have subclasses)

**To reproduce**

1. Install dependencies:

```bash
pip install pyshacl rdflib
```

2. Run this script:

```python
import pyshacl
from rdflib import Graph

validate_rdf = {
    "@context": {
        "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
        "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
        "dct": "http://purl.org/dc/terms/",
        "biolink": "https://w3id.org/biolink/vocab/"
    },
    "@type": "biolink:Drug",
    "biolink:category": "biolink:Drug",
    "biolink:id": "drugbank:DB00001",
    "rdfs:label": "Lepirudin",
    "dct:description": "Lepirudin is a protein-based direct thrombin inhibitor."
}

validate_g = Graph()
validate_g.parse(
    data=validate_rdf,
    format="ttl"
)

shacl_g = Graph()
shacl_g.parse(
    "https://raw.githubusercontent.com/vemonet/biolink-model/add-shacl-gen/biolink-model.shacl.ttl",
    format="ttl"
)

biolink_g = Graph()
biolink_g.parse(
    "https://raw.githubusercontent.com/biolink/biolink-model/master/biolink-model.owl.ttl",
    format="ttl"
)

[...]

PR #1258, opened 2023-01-23
Change SHACL gen to make the shapes `sh:closed false` by default

vemonet:
Change SHACL gen to make the shapes `sh:closed false` instead of `sh:closed true`

The motives for this changes have been discussed with @cmungall in https://github.com/linkml/linkml/issues/1249


vemonet:
New pull request https://github.com/linkml/linkml/pull/1264

Issue #1259, opened 2023-01-23
NMDC Biosample rules not enforced, trying tests here


turbomam:

cmungall:
can you provide more context?

PR #1260, opened 2023-01-23
overly tolerant rule #1259

turbomam:
- #1259 

I added a rule that should reject people who have more daughters than children, but that case is still considered valid.


turbomam:
Actually, the test really just requires that the Person has at least two children if they have two or more daughters

turbomam:
Closing. Now that the nmdc-schema can use run-linkml-examples, I can work though my issues there.

PR #1264, opened 2023-01-25
Fix shaclgen by making mixins and classes without parents not closed

vemonet:
Fix shaclgen by making mixins and abstract classes `sh:closed false` in the shaclgen, more details in the discussion with @cmungall in https://github.com/linkml/linkml/issues/1249


sierra-moxon:
I like the solution here.

cmungall (approved):

PR #1273, opened 2023-01-30
Pydantic any_of range support

### LINKED ISSUE #1272: Unions of ranges not supported in genenator-pydantic [the presence of 'default_range: string' with Union ranges raises the 'ValueError: Slot cannot have both range and any_of defined']
The current implementation of gen-pydantic does not seem to support [any_of](https://w3id.org/linkml/any_of) accurately. The generator fails to produce Union and only `str` type is generated. Although [Unions of ranges](https://linkml.io/linkml/schemas/advanced.html#unions-as-ranges) has been identified as an advanced feature, its use seems to be trivial.

For example, consider the following use of `any_of`
```
classes:  
  A:
    class_uri: ...
  B:   
    class_uri: ...
  C:    
    class_uri: ...
    slots:
      - dummy_slot
    slot_usage:url
      dummy_slot: 
        multivalued: true
        any_of:
          - range: A
          - range: B
```
For `class C(...):`, instead of generating `dummy_slot: Optional[List[Union[A, B]]]`, the gen-pydantic generates `dummy_slot: Optional[List[str]]`.

• Important - it's a blocker and can't do work without it.

• Short-term - 2-4 weeks

kevinschaper:
Starting this out as a draft PR with just a test in it.

Initially, I'm actually hitting an error in schemaview when calling get_identifier_slot on the slot that is defined with the any_of range of classes.

(eventually) closes #1272


kevinschaper:
It's finally running the test without errors, but still two bugs to go:

When the any_of ranges are the same, they need to be collapsed (just need to make a unique list or make it a set). This is the straightforward one.
```
E           AssertionError: assert '    not_inlined_things: Optional[List[Union[str, str]]] = Field(default_factory=list)' == '    not_inlined_things: Optional[List[str] = Field(default_factory=list)'
E             -     not_inlined_things: Optional[List[str] = Field(default_factory=list)
E             +     not_inlined_things: Optional[List[Union[str, str]]] = Field(default_factory=list)
E             ?                                       +++++++++++    ++
```

and much more weird and frustrating is that something is somehow double backwards here:

```
E           AssertionError: assert '    inlined_things: Optional[List[Union[A, B]]] = Field(default_factory=dict)' == '    inlined_things: Optional[[Dict[str,Union[A, B]]] = Field(default_factory=list)'
E             -     inlined_things: Optional[[Dict[str,Union[A, B]]] = Field(default_factory=list)
E             ?                              ^^ ^  ----                                      ^ ^
E             +     inlined_things: Optional[List[Union[A, B]]] = Field(default_factory=dict)
E             ?         
```

kevinschaper:
Ugh, I just noticed that the kitchen_sink pydantic model now has: 

```
    aliases: Optional[List[Union[]]] = Field(default_factory=list)
```

So I think there's more work to do. I sure hope I get a test failure from that.

kevinschaper:
This is failing right now with 

```
E   pydantic.error_wrappers.ValidationError: 1 validation error for Dataset
E   code_systems -> id
E     value is not a valid dict (type=type_error.dict)
```

The slot definition for code system within a dataset is defined with:
```
      code systems:
        range: CodeSystem
        inlined: true
        multivalued: true
```

The generated python for a code system within a dataset is:
```
    code_systems: Optional[Dict[str, CodeSystem]] = Field(None)
```

The yaml that is getting loaded for the code system includes:
```
code_systems:
  - id: DiseaseCodes
    name: disease codes
  - id: ProcedureCodes
    name: disease codes
```

In the test, the yaml is being loaded with: 

```
            ds1 = Dataset(**dataset_dict)
```

It looks like this is stuck right now because Pydantic's loader share linkml's friendly approach to converting a list into a dictionary.

kevinschaper:
This is ready for review @cmungall @amc-corey-cox 

I did some work to de-nest the code for generating python ranges, for readability and so that it can be re-used to generate multiple pyranges if multiple ranges are specified with any_of.

tests/test_generators/test_pydanticgen.py
```diff
@@ -27,9 +28,12 @@ def test_pydantic(self):
         """Generate pydantic classes"""
         gen = PydanticGenerator(SCHEMA, package=PACKAGE)
         code = gen.serialize()
+        print(code)
```
cmungall, line 31: oops

(I know I am guilty of this.. we can start tidying these in future PRs)

cmungall (approved):

PR #1275, opened 2023-01-31
Pydantic gen: switch to external uri/curie by default, still accept native uri/curie.

kervel:
as discussed with @hsolbrig 

* Use external curie/uri by default if available. Native uri/curie is still accepted
* When the range is uri, we still accept the known curies, and when the range is uriorcurie, we also still accept uri's.

generated code: 

```python
class NamedThing(ConfiguredBaseModel):
    
    id: str = Field(None)
    full_name: Optional[str] = Field(None)
    thingtype: Literal["http://example.org/NamedThing","http://example.org/NamedThing","x:NamedThing","x:NamedThing"] = Field("x:NamedThing")
    


class Person(NamedThing):
    
    height: Optional[int] = Field(None)
    id: str = Field(None)
    full_name: Optional[str] = Field(None)
    thingtype: Literal["http://example.org/Person","http://testbreaker/not-the-uri-you-expect","x:Person","http://testbreaker/not-the-uri-you-expect"] = Field("x:Person")
    


class Organisation(NamedThing):
    
    number_of_employees: Optional[int] = Field(None)
    id: str = Field(None)
    full_name: Optional[str] = Field(None)
    thingtype: Literal["http://example.org/Organisation","http://example.org/Organisation","x:Organisation","x:Organisation"] = Field("x:Organisation")
```

### DIFF
--- a/linkml/generators/common/type_designators.py
+++ b/linkml/generators/common/type_designators.py
@@ -25,7 +25,13 @@ def get_accepted_type_designator_values(sv: SchemaView, type_designator_slot: Sl
             sv.get_uri(class_def, expand=False, native=True),
             sv.get_uri(class_def, expand=False, native=False),
         ]
-    if type_designator_slot.range == 'uri':
+    # unique, but with order preserved (https://stackoverflow.com/a/17016257)
+    accepted_uri_values = list(dict.fromkeys(accepted_uri_values))
+
+    slot_types = set(sv.type_ancestors(type_designator_slot.range))
+    uri_types = ['uri', 'uriorcurie']
+
+    if slot_types.intersection(uri_types):
         return accepted_uri_values
     elif type_designator_slot.range == 'string':
         return [camelcase(class_def.name)]
--- a/linkml/generators/common/type_designators.py
+++ b/linkml/generators/common/type_designators.py
@@ -1,24 +1,29 @@
 from linkml_runtime.linkml_model.meta import SlotDefinition, ClassDefinition
 from linkml_runtime.utils.schemaview import SchemaView
 from linkml_runtime.utils.formatutils import camelcase
-from typing import List
+from typing import List, Tuple
 
 def get_type_designator_value(sv: SchemaView, type_designator_slot: SlotDefinition, class_def: ClassDefinition) -> str:
     """
         retuns the correct value for a type designator field for a given class, depending on its range
         this implements the logic described in https://github.com/linkml/linkml/issues/945:
     """
-    target_value = None
-    if type_designator_slot.range == 'uri':
-        target_value = sv.get_uri(class_def,expand=True,native=False)
-    elif type_designator_slot.range == 'string':
-        target_value = camelcase(class_def.name)
+    slot_types = set(sv.type_ancestors(type_designator_slot.range))
+    if 'uri' in slot_types:
+        return sv.get_uri(class_def,expand=True,native=False)
+    elif 'uriorcurie' in slot_types:
+        return sv.get_uri(class_def,expand=False, native=False)
+    elif 'string' in slot_types:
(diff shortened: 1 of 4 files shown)


kervel:
started modifying this as per @hsolbrig feedback. generated pydantic code now looks like this. note that here we don't have a `CurieNamespaces` class since the pydantic code doesn't depend on linkml-runtime.


```python
class NamedThing(ConfiguredBaseModel):
    
    id: str = Field(None)
    full_name: Optional[str] = Field(None)
    thingtype: Literal["http://example.org/NamedThing","http://example.org/NamedThing","x:NamedThing","x:NamedThing"] = Field("x:NamedThing")
    


class Person(NamedThing):
    
    height: Optional[int] = Field(None)
    id: str = Field(None)
    full_name: Optional[str] = Field(None)
    thingtype: Literal["http://example.org/Person","http://testbreaker/not-the-uri-you-expect","x:Person","http://testbreaker/not-the-uri-you-expect"] = Field("x:Person")
    


class Organisation(NamedThing):
    
    number_of_employees: Optional[int] = Field(None)
    id: str = Field(None)
    full_name: Optional[str] = Field(None)
    thingtype: Literal["http://example.org/Organisation","http://example.org/Organisation","x:Organisation","x:Organisation"] = Field("x:Organisation")
```

kervel:
hello @kevinschaper could you give this one a look ? it think its ready for review now.

kevinschaper:
Hi @kervel, I checked on my use case, and it looks like it's losing the prefix on the CURIE version of the type designator field, so the default value won't actually be allowed.

```
    category: List[Literal["Gene"]] = Field(["biolink:Gene"], description="""Name of the high level ontology class in which this entity is categorized. Corresponds to the label for the biolink entity type class.

```

kervel:
hello @kevinschaper in your test case, what is the range of your type designator ("category") slot ?
 i tought the range was always filled in to be either uri / uriorcurie / string (and empty ranges seem to get expanded to string). the behaviour you are getting looks like what would happen if the range is neither or those

cmungall:
We should probably add something to SchemaView to make this kind of thing
easier. But the type should be found from the reflexive ancestors of the
type.

kevinschaper:
@kervel I took a shot accounting for type ancestors, and removed duplicate acceptable types to tidy up the generated output a bit.

kevinschaper (approved):

kervel:
i'm still busy trying to turn the fail you got into a test and also to also use type_ancestors for the default values (not only for the accepted values) ... so not completely ready yet

kevinschaper:
Oh, you're still up and working on it! I thought I'd put that small change in, and then wait to see what you thought of it before actually merging.

kervel:
Hello @kevinschaper , i tried to turn this part of biolink into a test for this. I plan to use the same tests for the data classes later, so i think its useful to add it.

However, the type designator field is multivalued in your example. I'm not quite sure what that should mean and if we even treat this correctly. I don't think we can really do that with pydantic ? now the type designator field accepts a list (see my new test) but it won't make sense if the list contains more than one value.

kervel:
hmm seems @cmungall answered the question on how to treat multivalued: take the most specific one (see https://github.com/linkml/linkml/pull/1291/files). this is definately not supported right now with the current code, but i think it still makes sense to merge this, and to add proper support for multivalued in a separate PR.

cmungall:
> but i think it still makes sense to merge this, and to add proper support for multivalued in a separate PR.

agreed

kervel:
rebased to fix merge conflicts

kevinschaper:
I guess I slightly blurred the reviewer line by making that small change, but I'm going to just assume that it's ok for me to bring it in and go for it.

PR #1277, opened 2023-01-31
Update tox.ini file

hsolbrig:
Apparently this hadn't been run since the poetry changeover.

The following errors were detected while running it:

1) Some issue with notebooks/examples.ipynb - we probably need to address this separately 
2) test_sqlddlgen - missing plugin: sqlalchemy.dialects:postgresql.pygresql 
3) ERDiagramGeneratorTestCase.test_follow_references 4) ERDiagramGeneratorTestCase.test_max_hops
4) ERDIagramGeneratorTestCase.test_serialize
5) ERDIagramGeneratorTestCase.test_serialize_all
6) gen-prolog --help  - missing linkml.generators.lpgen 
7) gen-terminusdb --help - missing terminusdb_client

Also - what is the policy on updating the poetry.lock file?   It will change whenever anyone does a `poetry update` command


sujaypatil96:
@hsolbrig yes, the `poetry.lock` file will change everytime someone runs the `poetry update` command. Here is official guidance from the developers of poetry on whether or not we should actually version control the `poetry.lock` file: https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control

The short answer is no, we need not worry about keeping our `poetry.lock` file properly version controlled since linkml is a library. The `pyproject.toml` file is the reference for correct **build** instructions, and the lockfile is the reference for a single successful **deployment**.

Although it's not the recommended guidance, the current direction for `poetry.lock` in linkml is, you should run `poetry update` and whatever lockfile you generate in your branch you should git add, commit and push it so we can get it into `main`. And the reason for that is because already having a `poetry.lock` file while contributing to the codebase helps cut down on build time for developers. We can and I think we should, stop versioning the `poetry.lock` file in the future, but that is up for debate still. But at the moment, please push whatever `poetry.lock` file you generate after running `poetry update`.

dalito:
Could you make updating poetry.lock part of the release process? Even if linkml is a library it is helpful to have poetry.lock to reproduce the tested setup exactly in case of problems.

Issue #1282, opened 2023-02-06
docgen.py index.md not respect --no-mergeimports


noelmcloughlin:
**Describe the bug**

DocGenerator does not use value of self.mergeimports 

I  expect `poetry run gen-doc --no-mergeimports -d docs src/myschema/schema/myschema.yaml` to only list classes, slots, types from  "myschema" on index page (generated by make serve) but I see imported classes listed too. 

**To reproduce**

```poetry run gen-doc --no-mergeimports -d docs src/myschema/schema/myschema.yaml ```

**Expected behavior**

**Additional context**
```
$ pip freeze | grep linkml
linkml==1.3.17
linkml-dataops==0.1.0
linkml-runtime==1.4.3```

sujaypatil96:
Hm, the latest version of docgen in linkml does respect `--mergeimports`. The fix came in a v1.4.x release of linkml, perhaps you could upgrade to the latest version and test documentation generation with that? Let me know if it still doesn't work and I'll look into it.

noelmcloughlin:
I am using cookiecutter and noticed packages are old so I upgraded linkml to latest version (see also https://github.com/linkml/linkml-project-cookiecutter/pull/44)

I  expect `poetry run gen-doc --no-mergeimports -d docs src/myschema/schema/myschema.yaml` to only list classes, slots, types from  "myschema" on index page (generated by make serve) but I see imported classes listed too. 

So after experimenting I found #1283 worked best in achieving correct index page.

turbomam:
see also
- https://github.com/linkml/linkml-project-cookiecutter/pull/44

noelmcloughlin:
See https://github.com/linkml/linkml/pull/1283#issuecomment-1435409515

sujaypatil96:
See PR #1309

noelmcloughlin:
Currently not fixed by https://github.com/linkml/linkml/pull/1309

Issue #969, opened 2022-09-19
gen-linkml CLI tool does not honor (default) --mergeimports parameter


turbomam:
`gen-linkml` advertises a `--mergeimports` parameter and even states that merging is the default state, as compared to `--no-mergeimports`

>   --mergeimports / --no-mergeimports
                                  Merge imports into source file
                                  (default=mergeimports)

However, the output of running the MIxS schema through `gen-linkml` does not contain the imports

```shell
poetry run gen-linkml \
		--format yaml \
		--mergeimports \
		--no-materialize-attributes model/schema/mixs.yaml
```

---

```yaml
name: MIxS
description: Minimal Information about any Sequence Standard
id: http://w3id.org/mixs
imports:
  - linkml:types
  - checklists
  - core
  - agriculture
  - food_animal_and_animal_feed
  - food_farm_environment
  - food_food_production_facility
  - food_human_foods
  - symbiont_associated
  - host_associated
  - microbial_mat_biofilm
  - miscellaneous_natural_or_artificial_environment
  - plant_associated
  - sediment
  - soil
  - wastewater_sludge
  - water
  - human_associated
  - human_gut
  - human_oral
  - human_skin
  - human_vaginal
  - air
  - built_environment
  - hydrocarbon_resources_cores
  - hydrocarbon_resources_fluids_swabs
prefixes:
  linkml:
    prefix_prefix: linkml
    prefix_reference: https://w3id.org/linkml/
  mixs.vocab:
    prefix_prefix: mixs.vocab
    prefix_reference: https://w3id.org/mixs/vocab/
  MIXS:
    prefix_prefix: MIXS
    prefix_reference: https://w3id.org/mixs/terms/
  MIGS:
    prefix_prefix: MIGS
    prefix_reference: https://w3id.org/mixs/migs/
default_prefix: mixs.vocab
```

turbomam:
@sujaypatil96 if this issue can be addressed this week, it would help in resolving https://github.com/microbiomedata/nmdc-schema/issues/509

capsulecorplab:
`--mergeimports` flag seems to be working for `gen-yaml`. It's what I'm currently using as part of a build script https://github.com/Mach30/m30ml/blob/2164d51bbcc7b3c2562f641c84f396a14ac2177a/build.sh#L11

sujaypatil96:
[...]

PR #1283, opened 2023-02-06
fix(docgen):  pass mergeimports to schemaview.all_xxx methods

noelmcloughlin:
This PR passes  `imports=self.mergeimports` as parameter to relevant methods so correct documentation is generated when external imports are used:  https://github.com/linkml/linkml/issues/1282


noelmcloughlin:
Is failing CI caused by my PR

sujaypatil96:
Right, yes, the Actions are erroring out on your branch specifically.

sujaypatil96:
In response to the functionality you have submitted as part of this PR, could you make sure https://github.com/linkml/linkml-runtime/pull/233 doesn't solve the problem already?

I believe the issue you are trying to address is a duplicate of: https://github.com/linkml/linkml/issues/969. An issue that we had noticed before too.

The way we got around this was by passing the value of `--mergeimports` globally to the SchemaView object itself. Does that make sense?

```python

sv = SchemaView("~/path/to/linkml.yaml", mergeimports=True)
```

linkml/generators/docgen.py
```diff
@@ -656,7 +656,7 @@ def all_class_objects(self) -> Iterator[ClassDefinition]:
         Ensures rank is non-null
         :return: iterator
         """
-        elts = self.schemaview.all_classes().values()
+        elts = self.schemaview.all_classes(imports=self.mergeimports).values()
```
sujaypatil96, line 659: Is this necessary? Similar comment for the other elements, i.e., slots, enums, types, etc.

The reason I ask is because if you look at PR https://github.com/linkml/linkml-runtime/pull/233/files which added the ability to honor `--mergeimports`, it makes a call to `merge_imports()` which calls the closure that looks through all the imported schemas and fetches the classes, slots, enums, etc. appropriately and associates with the SchemaView object.

sujaypatil96: So what I mean to say is that:

```python

sv = SchemaView("~/path/to/linkml.yaml", merge_imports=True)

sv.all_classes()  # same result as below
sv.all_classess(imports=True)  # same result as above
```

noelmcloughlin: I will add new comment as I might have figured out the issue.

noelmcloughlin:
I noticed the JINJA2 template invokes class methods not instance method:  https://github.com/linkml/linkml/blob/5ecd9c402b7e8f5ad1775904a9247df26a1918fa/linkml/generators/docgen/index.md.jinja2#L12

So possibly this PR is correct solution?  What do you think?

sujaypatil96:
@noelmcloughlin: I don't think that's the reason? You can see the definition for `all_class_objects()` for here: https://github.com/isamplesorg/metadata/blob/main/tools/docgen.py#L624-L634

sujaypatil96:
Apologies, the test output log is hard to look at. But we can see the Action is not going through on `test_docgen.py`. Perhaps you could run the tests on your local machine this way and look at what unit tests exactly are causing Actions to error out:

```
$ poetry run python -m unittest tests/test_generators/test_docgen.py
```

noelmcloughlin:
You are right - its not Jinja problem.  I see that `all_class_objects()` invokes `elts = self.schemaview.all_classes().values()` without passing parameters so **imports=True** is used by `linkml_runtime/utils/schemaview.py` here:

```    def all_classes(self, ordered_by=OrderedBy.PRESERVE, imports=True) -> Dict[ClassDefinitionName, ClassDefinition]:```

https://github.com/linkml/linkml-runtime/blob/163d459b91f0aecec8a1dd36528602167546be29/linkml_runtime/utils/schemaview.py#L250

Is there some problem in **schemaview.py** instead?

sujaypatil96:
The fix to your problem will be passing the value for `merge_imports` to `SchemaView` during initialization here: https://github.com/linkml/linkml/blob/main/linkml/generators/docgen.py#L145

If you want to take this up, I'm happy to help you make this bugfix?

noelmcloughlin:
Can you do it, I can test locally and let you know

sujaypatil96:
This seems to be the way to go @noelmcloughlin. You can use the tests from PR #1309 to add to this PR, and we should be good to merge after that.

sujaypatil96:
I'm happy to incorporate whatever you have in this PR into mine #1309

sujaypatil96:
Let me know if it's okay for me to copy over the contents of this PR into #1309 and I will do so. I don't want to take away from your contributions, which is why I was encouraging you to complete it.

noelmcloughlin:
Feel free, goal is to fix docgen for community,  thanks

noelmcloughlin:
Hi @sujaypatil96  
I tested https://github.com/linkml/linkml/pull/1309 but generated Index.md is still incorrect.  I see you added `self.mergeimports` in these places shown below, but my PR updated method calls associated with Jinja2 template processing.

How should I proceed. Should I leave my PR as is?  
Do you want me to revert lines below as I presume you intended to incorporate my PR. 

image

image

image

image

image

image

noelmcloughlin:
Actually I think lines above need to be reverted because when I browse my documentation the parent class was not generated.

image

sujaypatil96:
Thank you for testing the PR @noelmcloughlin, and bringing this to our notice. Could you close out this PR, and we can make continue discussing this on a separate issue?

Feel free to make the issue about the `index.md` file not getting generated properly upon enabling or disabling mergeimports, or I can make one for you, and we can continue our discussion on there.

noelmcloughlin:
See https://github.com/linkml/linkml/issues/1311

PR #1288, opened 2023-02-08
Configure use_enum_values to True on pydanticgen BaseModel

kevinschaper:
Setting use_enum_values = True in the BaseModel configuration to look at possible test failures that result. This PR needs a positive confirmation test to show that it's working and keep it working as well before it goes in.


cmungall (approved):

cmungall:
I converted to draft since this appears to be a trial balloon?

kevinschaper:
I think we could bring this in, this gives us good serialization to strings at the cost slightly harder comparison against the enum

PR #1291, opened 2023-02-14
Adding type-designator docs (not linked yet)

cmungall:


docs/schemas/type-designators.md
```diff
@@ -0,0 +1,44 @@
...
+        range: string
+  Business:
+    is_a: Organization
+  EducationalOrganization:
+    is_a: Organization
+   ...
+```
+
+This means that if the `type` slot is present in data for an instance of a class
+Organization, then the value of the `type` slot MUST either "Organization" or any
```
dalito, line 26: `MUST either`--> `MUST be either`

PR #1293, opened 2023-02-17
yumlgen: use POST and add ability to name output file

jankatins:
We needed these two changes to 
a) generate bigger diagrams
b) have more control over how the filenames end up


tests/test_scripts/output/genyuml/help
```diff
@@ -6,6 +6,8 @@ Options:
   -c, --classes TEXT              Class(es) to emit
   -d, --directory TEXT            Output directory - if supplied, YUML
                                   rendering will be saved in file
+  --diagram-name TEXT             Name of the diagram in the output directory
+                                  (without suffix!)
   -f, --format [yuml|png|pdf|jpg|json|svg]
```
jankatins, line 11: When I ran the tests, there were actually more changed but they were unrelated and I left them out.

cmungall (approved):
Many thanks! We are focusing efforts on mermaid but this still doesn't support everything supported by yumlgen, so this is most welcome

jankatins:
Mermaid would definitely be nice, we also would want that: our goal in the end is to import the whole generated doc into a azure wiki (based on markdown files in a git, so simply comitting would work)

PR #1294, opened 2023-02-17
RFC: Improvements/changes to sqlalachemygen

jankatins:
These are changes we had to do to make it easier for us to automatically generate and use models as sqlalchemy models. The goal was that we could reuse the python files "as is" without manual changes afterwards. This is an attempt to upstream these changes to see if any of them are of interest. If not, we will simply keep our forked scripts :-)

I tried to split this into multiple commits, but e.g. the changes to the final sqlalchemy jinja template could probably be split further.

There are a few more changes, like adding a base class to declaritive_base so one can add mixins from outside, but I can do these easily via some postprocessing (search&replace):

```
try:
    from .base import {{classname(model_path)}}Base as _Base
except ImportError:
    class _Base:
        pass

# Each model needs to have its own Base to not get clashes in the alembic handling
# when some classes are defined multiple times!
Base = declarative_base(cls=_Base, ...)
```

Let me know if there is anything interesting in these so I can add tests...

### DIFF
--- a/linkml/generators/sqlalchemygen.py
+++ b/linkml/generators/sqlalchemygen.py
@@ -64,7 +64,8 @@ class SQLAlchemyGenerator(Generator):
     ) -> str:
         # src_sv = SchemaView(self.schema)
         # self.schema = src_sv.schema
-        sqltr = RelationalModelTransformer(self.schemaview)
+        # use two underscores to distinguish it from the snake case underscore
+        sqltr = RelationalModelTransformer(self.schemaview, join_table_separator="__")
         if foreign_key_policy:
             sqltr.foreign_key_policy = foreign_key_policy
         tgen = SQLTableGenerator(self.schemaview.schema)
--- a/linkml/generators/sqlalchemygen.py
+++ b/linkml/generators/sqlalchemygen.py
@@ -126,6 +126,9 @@ class SQLAlchemyGenerator(Generator):
             mappings=tr_result.mappings,
             backrefs=backrefs,
             classname=camelcase,
+            schemaname=underscore,
+            tablename=underscore,
+            columnname=underscore,
             no_model_import=no_model_import,
             is_join_table=lambda c: any(
                 tag for tag in c.annotations.keys() if tag == "linkml:derived_from"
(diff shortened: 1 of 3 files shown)


linkml/generators/sqlalchemygen.py
```diff
@@ -119,6 +126,9 @@ def generate_sqla(
             mappings=tr_result.mappings,
             backrefs=backrefs,
             classname=camelcase,
+            schemaname=underscore,
```
jankatins, line 129: We are actualy using a even snakecase function here, but I guess the underscore function is enough for now

cmungall:
The underscore function actually snakecasifies, and should probably be
called snakecase

jankatins:
> The underscore function actually snakecasifies, and should probably be
called snakecase

I think I remmeber we had to change that becasue there were a lot of cases where it didn't properly snakecase it. We ended up with this:

```
import inflection
import linkml_runtime.utils.formatutils

def snakecase(txt: str) -> str:
    snakecased = inflection.underscore(txt)
    return cap_at_63_chars(snakecased)

def camelcase(txt: str) -> str:
    # preserve double underscores!
    txt = txt.replace("__", "§§")
    txt = linkml_runtime.utils.formatutils.camelcase(txt)
    return cap_at_63_chars(txt.replace("§§", "__"))

def cap_at_63_chars(txt: str) -> str:
    """Cap identifier at 63 chars to make it compatible with PG"""
    if len(txt) <= 63:
        return txt
    # This is a string of length 8 of all digits
    # The hash should not change across different runs
    hash = str(int(hashlib.sha256(txt.encode("utf-8")).hexdigest(), 16) % 10**8)
    return txt[: (63 - 8 - 1)] + "_" + hash
```

cmungall:
I think we're definitely interested in these changes. I prefer doing a lot of this kind of thing formally as part of the generation rather than post-processing

jankatins:
There is also now a big update how sqlalchemy classes should be build (See e.g. https://docs.sqlalchemy.org/en/20/changelog/whatsnew_20.html#migrating-an-existing-mapping). for that it would actually be nice if we could set an argument to chose the template so one could point it to something home grown instead of needing to postprocess/patch the one included in linkml.

jankatins:
@cmungall Whats your suggestion how to proceede? Push all of the commits as small PRs? Group them somehow?

jankatins:
We have stopped using linkml so I will not work on this PR anymore :-(

PR #1295, opened 2023-02-18
Handle local prefixes special when merging linkml files

jankatins:
This can e.g. happen if we have this kind of filestructure:

```
lib.yaml
folder/subfolder/more_stuff.yaml -> references as ../../lib.yaml
folder/all.yaml -> references both ../lib.yaml and ./subfolder/more_stuff.yaml
```

Now we want to use `gen-yaml` on `folder/all.yml` to create a combined linkml file

### DIFF
--- a/linkml/utils/mergeutils.py
+++ b/linkml/utils/mergeutils.py
@@ -1,4 +1,5 @@
 import dataclasses
+import logging
 from copy import deepcopy
 from typing import Dict, List, Optional, Union, cast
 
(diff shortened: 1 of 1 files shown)


cmungall:
it would be great to have a unit test for this - is this something you'd be able to take on or shall we have some of the core team handle it?

jankatins:
huih. My only idea would be to write out some very basic linkml structure with three files and then run yamlgen on it. I tried to find some tests for `merge_namespaces()` but couldn't find any.

Would that work for you?

cmungall:
that would be ideal, thanks!

linkml/utils/mergeutils.py
```diff
@@ -73,6 +74,37 @@ def merge_namespaces(
...
+                    and target.prefixes[prefix.prefix_prefix].prefix_reference
+                    != prefix.prefix_reference
+                ):
+                    logging.info(
+                        "Ignoring different relative prefix for %s from %s, "
+                        + "as we cannot adjust it relative to the final file. "
+                        + "Assuming the first found location is correct: %s!",
+                        prefix.prefix_prefix,
+                        mergee.name,
+                        namespaces[prefix.prefix_prefix],
```
sujaypatil96, line 104: A good place to use Python [f-strings](https://peps.python.org/pep-0498/)?

linkml/utils/mergeutils.py
```diff
@@ -1,4 +1,5 @@
 import dataclasses
+import logging
```
sujaypatil96, line 2: Probably a good idea to follow syntax like this: https://github.com/linkml/linkml/blob/main/linkml/generators/linkmlgen.py#L14?

sujaypatil96 (commented):
Left a couple of very trivial style guide review comments. Great work on the PR @jankatins

linkml/utils/mergeutils.py
cmungall: yes, see also: https://stackoverflow.com/questions/45965007/multiline-f-string-in-python

but we can do this after merging

jankatins: the linters won't like it: no fstring in the `log.*()` functions, I think for performance reasons (I actually converted it to use log and had a print(f"...") here in my q&d code :)

Issue #1299, opened 2023-02-20
shaclgen not generating nodeKind/datatype


noelmcloughlin:
**Describe the bug**

The artifacts generated by  `shaclgen.py` are missing `nodeKind` and `datatype` for linkml types.  During debugging this conditional is always false and inner block is skipped:
https://github.com/linkml/linkml/blob/ec9362352892ff2f949d1b20fe1897923efe33f1/linkml/generators/shaclgen.py#L112

**To reproduce**
Using linkml-project-cookiecutter you can run gen-shacl and example schema and examine types ("sh:property") nodes.

**Expected behavior**

I'd expect to see `sh:nodeKind` and `sh:datatype` generated for linkml types.

noelmcloughlin:
Draft PR: https://github.com/linkml/linkml/pull/1300

PR #1300, opened 2023-02-20
feat(shacl): improve type generation

noelmcloughlin:
This PR improves shaclgen.py by adding nodeKind and datatype for Linkml types (fixes https://github.com/linkml/linkml/issues/1299)

TODO: I am not sure if `sh:nodeKind sh:IRI` is correct and maybe should be `sh:nodeKind sh:Literal`.

Example output from this PR:
```
sh:property [ sh:datatype "xsd:anyURI" ;             <=========== added
            sh:description "Specifies a reference to a resource outside of the UCO." ;
            sh:nodeKind sh:IRI ;              <=========== added
            sh:order 3 ;
            sh:path core:externalReference ],
        [ sh:datatype "xsd:string" ;              <=========== added
            sh:description "The name of a particular concept characterization." ;
            sh:maxCount 1 ;
            sh:nodeKind sh:Literal ;               <=========== added
            sh:order 6 ;
            sh:path core:name ]
```

### DIFF
--- a/linkml/generators/shaclgen.py
+++ b/linkml/generators/shaclgen.py
@@ -109,10 +109,11 @@ class ShaclGenerator(Generator):
                         prop_pv(SH.nodeKind, SH.IRI)
                     else:
                         prop_pv(SH.nodeKind, SH.BlankNode)
-                elif r in sv.all_types().values():
+                elif r in sv.all_types():
+                    prop_pv(SH.nodeKind, SH.Literal)
                     rt = sv.get_type(r)
                     if rt.uri:
-                        prop_pv(SH.datatype, rt.uri)
+                        prop_pv(SH.datatype, Literal(rt.uri))
                     else:
                         logging.error(f"No URI for type {rt.name}")
                 elif r in sv.all_enums():


cmungall:
Thanks for taking this on!

I believe sh:datatype should only be used for literals https://www.w3.org/TR/shacl/#validator-DatatypeConstraintComponent

ideally the unit tests would use pyshacl to test that validation works as expected but let's hold off on adding that as a dependency for now. For now we can just check the resulting shacl triples map as extended, and we can do an out of band test to ensure the shacl validates as expected

noelmcloughlin:
Thanks @cmungall  Good point - we should pair datatype with Literals only.

I am checking my linkml generated shacl against a control rdfs+shacl file someone else created which only declares sh:IRI and sh:Literal nodeKinds. However my PR is naive.

From docs: "_The values of sh:nodeKind in a shape are one of the following six instances of the class sh:NodeKind: sh:BlankNode, sh:IRI, sh:Literal sh:BlankNodeOrIRI, sh:BlankNodeOrLiteral and sh:IRIOrLiteral._ "

Is there a way calculate correct nodeKind from value of  `str(Literal(rt.uri))` ?

cmungall:
If the range is a type, it’s a literal

If the range is a class with an identifier, iri

Class with no identifier, blank

Enums:

If all PVs have a meaning, iri
If none have meaning, literal
Otherwise iriorliteral

linkml/generators/shaclgen.py
```diff
@@ -109,10 +109,11 @@ def prop_pv_literal(p, v):
                         prop_pv(SH.nodeKind, SH.IRI)
                     else:
                         prop_pv(SH.nodeKind, SH.BlankNode)
-                elif r in sv.all_types().values():
+                elif r in sv.all_types():
+                    prop_pv(SH.nodeKind, SH.Literal)
                     rt = sv.get_type(r)
                     if rt.uri:
-                        prop_pv(SH.datatype, rt.uri)
+                        prop_pv(SH.datatype, Literal(rt.uri))
```
cmungall, line 116: I don't think this is right - even though the value of sh:datatype is used for validating literals, it is not itself a literal

See:

https://www.w3.org/TR/shacl/#DatatypeConstraintComponent

example:

```turtle
sh:property [
		sh:path ex:age ;
		sh:datatype xsd:integer ;
	] .
```

cmungall (changes requested):

nlharris:
Looks like this still has changes requested.

amc-corey-cox:
It looks like this may have been closed with https://github.com/linkml/linkml/pull/2102 and this PR has been idle for a while. I'm going to close this for now. If you are still working on this you are welcome to re-open the PR and make updates so we can resolve the issue.

PR #1308, opened 2023-02-23
replace `pkg_resources` methods with `importlib` methods

sujaypatil96:
pkg_resources is deprecated and the setuptools docs recommends moving to importlib https://setuptools.pypa.io/en/latest/pkg_resources.html

The `pkg_resources` module has a dependency on `setuptools`. Taking a look at the `pyproject.toml` file, we're not installing `setuptools`, either as a dev dependency or as a regular dependency. We don't have `setuptools` in our environment, so that means no access to the `pkg_resources` module either. 

We need to use equivalent methods from the standard `importlib` module in place of `pkg_resources` methods that we are currently using.

### DIFF
--- a/linkml/generators/docgen.py
+++ b/linkml/generators/docgen.py
@@ -8,7 +8,7 @@ from typing import (Callable, Dict, Iterable, Iterator, List, Optional, Set,
                     TextIO, Tuple, TypeVar, Union)
 
 import click
-import pkg_resources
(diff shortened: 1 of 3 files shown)


sujaypatil96:
@kevinschaper: I made a couple of quick edits to the Java generator, could you review that really quick?

kevinschaper (approved):

PR #1309, opened 2023-02-23
fix `mergeimports` issue in docgen

sujaypatil96:
Merge PR #1308 first

This PR seeks to solve the problem that @noelmcloughlin was facing in PR #1283

### DIFF
--- a/linkml/generators/docgen.py
+++ b/linkml/generators/docgen.py
@@ -8,7 +8,7 @@ from typing import (Callable, Dict, Iterable, Iterator, List, Optional, Set,
                     TextIO, Tuple, TypeVar, Union)
 
 import click
-import pkg_resources
+import importlib.util
 from jinja2 import Environment, FileSystemLoader, Template
 from linkml_runtime.dumpers import yaml_dumper
 from linkml_runtime.linkml_model.meta import (Annotation, ClassDefinition,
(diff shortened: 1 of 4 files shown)


sujaypatil96:
@noelmcloughlin: if you agree with the changes in this PR, then we can merge this in and close out PR #1283

linkml/generators/docgen.py
```diff
@@ -159,6 +158,7 @@ def __post_init__(self):
         if self.example_directory:
             self.example_runner = ExampleRunner(input_directory=Path(self.example_directory))
         super().__post_init__()
+        self.schemaview = SchemaView(self.schema, merge_imports=self.mergeimports)
```
noelmcloughlin, line 161: I see `self.mergeimports=False` here but https://github.com/linkml/linkml/issues/1282 is not fixed.

noelmcloughlin: Wait, let me check something.

sujaypatil96: I think you might be right @noelmcloughlin, we might have to pass the value of `self.mergeimports` to `all_classes()`, `all_slots()`, etc. in docgen.py like you have in PR #1283

noelmcloughlin: I will check. I'm having an inconvenient bug in my parent schema ;-( just trying to fix that and I can test for sure.

sujaypatil96: Awesome, thank you. I've updated this PR.

cmungall (approved):

noelmcloughlin:
https://github.com/linkml/linkml/issues/1311

PR #1310, opened 2023-02-24
Remove camelcasing of class name for type designator values

kevinschaper:


cmungall (approved):
looks like this was discussed on the call, thanks!

kevinschaper:
heads up @kervel - I was showing your type designator changes on the developer call and @hsolbrig noted that it would be better to leave populate the type designator with the class name as it occurs in the schema rather than converting to camelcase. It doesn't sound like this will affect you at all, but I wanted to make sure you saw just in case.

kervel:
thanks, makes sense indeed!

PR #1316, opened 2023-03-01
add dependency relationship between classes linked by a slot

sujaypatil96:
Examples of what this PR introduces in the NMDC Schema, say:

1. Beneficial rendering: can easily visualize the linking

Screen Shot 2023-03-01 at 3 16 10 PM

2. Not so useful rendering: too many slots for a class to class relationship compressed into a small space. This is a mkdocs limitation

Screen Shot 2023-03-01 at 3 17 19 PM

### DIFF
--- a/linkml/generators/docgen/class_diagram.md.jinja2
+++ b/linkml/generators/docgen/class_diagram.md.jinja2
@@ -12,6 +12,7 @@
       {% endfor %}
       {% for s in schemaview.class_induced_slots(element.name)|sort(attribute='name') -%}
         {{ gen.name(element) }} : {{gen.name(s)}}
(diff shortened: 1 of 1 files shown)


cmungall:
looks great. Don't love the duplication of code. Perhaps in future this should be in pure python I don't see the advantage of jinja2 for something like this. Have a look at how I did it for mermaid ER diagrams.

cluttered images: let's iterate on this. Maybe we can have options to cut off a certain number of subclasses and show with `...`s etc

sujaypatil96:
Sounds good, I can put the conversion of jinja code into Python code similar to how it's being done in erdiagramgen.py on my plate.

cmungall (approved):

Issue #1147, opened 2022-12-01
duplicate `domain_of` values in specification of meta model


vincentvialard:
**Describe the bug**
`domain_of` in meta model contains duplicates

**To reproduce**
1. Go to [Direct LinkML Specification of SchemaDefinition](https://linkml.io/linkml-model/docs/SchemaDefinition/#direct)
2. Click on 'Details'
3. Scroll down to `slot_usage`
4. See error (duplicate 'element')

```
slot_usage:
  name:
    name: name
    description: a unique name for the schema that is both human-readable and consists
      of only characters from the NCName set
    domain_of:
    - element
    - element
    range: ncname
tree_root: true
```
This occurs in other places, for instance [Direct LinkML Specification of SchemaDefinition](https://linkml.io/linkml-model/docs/SlotDefinition/#direct)

**Expected behavior**
Duplicates are not expected. 

*It seems that domains from `slots` and `slot_usage` are appended instead of merged.*

turbomam:
This effects the nmdc-schema too

sujaypatil96:
Unit test for the fix is in PR #1319

PR #1319, opened 2023-03-02
test case for ensuring `domain_of` for slot does not have dups

sujaypatil96:
This is a unit test for the fix in https://github.com/linkml/linkml-runtime/pull/252

The goal of both the above PR and this unit test is to fix issue #1147


sujaypatil96:
Please don't merge this PR until https://github.com/linkml/linkml-runtime/pull/252 has been merged in. After that we need to ensure that we are pulling in the latest version of linkml-runtime on the Actions running in this repo to make sure we have the fix.

cmungall (approved):

PR #1327, opened 2023-03-07
fix class slot dependency relationships in mermaid diagram

sujaypatil96:
TODO:

- [x] exclude type classes from dependency relationships [<s>work in progress</s>] [completed]

CC: @cmungall


sujaypatil96:
Here's what an example Mermaid diagram will look like with these changes:

Screen Shot 2023-03-08 at 2 00 03 PM

cmungall (approved):
merge as soon as tests pass

PR #1329, opened 2023-03-09
Fix errors in Actions

sujaypatil96:
Errors in null change PR #1328 

- [x] `tests/test_issues/test_issue_273.py` [fixed]
- [x] `tests/test_generators/test_json_schema_gen.py`

### DIFF
--- a/linkml/generators/jsonschemagen.py
+++ b/linkml/generators/jsonschemagen.py
@@ -298,17 +298,18 @@ class JsonSchemaGenerator(Generator):
         })
         self.top_level_schema.add_def(enum.name, enum_schema)
 
-    def get_type_info_for_slot_subschema(self, slot: AnonymousSlotExpression, slot_is_inlined: bool) -> Tuple[str, str, Union[str, List[str]]]:
+    def get_type_info_for_slot_subschema(self, slot: AnonymousSlotExpression, parent_slot_is_inlined: bool) -> Tuple[str, str, Union[str, List[str]]]:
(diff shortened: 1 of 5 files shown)


cmungall (approved):

linkml/generators/jsonschemagen.py
```diff
@@ -298,17 +298,18 @@ def extract_permissible_text(pv):
         })
         self.top_level_schema.add_def(enum.name, enum_schema)
 
-    def get_type_info_for_slot_subschema(self, slot: AnonymousSlotExpression, slot_is_inlined: bool) -> Tuple[str, str, Union[str, List[str]]]:
+    def get_type_info_for_slot_subschema(self, slot: AnonymousSlotExpression, parent_slot_is_inlined: bool) -> Tuple[str, str, Union[str, List[str]]]:
```
cmungall, line 301: parent or ancestor?

pkalita-lbl: Poorly chosen variable name on my part because I'm not referring to a parent (or ancestor for that matter) in an inheritance sense. What I'm talking about is:

```yaml
slots:
  a_slot_with_range_union:
    description: this is what I'm referring to as the "parent" SlotDefinition
    any_of:
      # This is one implied "child" SlotExpression
      - range: C1
      # Another implied "child" SlotExpression
      - range: C2
```

pkalita-lbl: Perhaps `container_slot_is_inlined` would be more appropriate?

PR #1330, opened 2023-03-09
Trigger on push to main in addition to PR

dalito:
This should help to avoid breaking main. (related to #1328, #1329 - I think #1323 was the breaking change)

On `workflow_dispatch` is for manually starting the pipeline.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -4,7 +4,14 @@
 
 name: Build and test linkml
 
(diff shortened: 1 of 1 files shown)


cmungall:
I think the fails here can be fixed with a rebase...?

dalito:
Done! Tests failed because I started when main was in broken state.

cmungall (approved):

PR #1342, opened 2023-03-16
jsonschemagen: use anyOf rather than oneOf for range class descendants

kervel:
This prevents an error if json data satisfies the schema of more than one range class descendants. This now happens more than it should because the type designator values are not yet part of the json schema, but it can still happen in other cases if there is no type designator field.


cmungall (approved):
This makes perfect sense. Yes there are many scenarios where there would indeed be multiple matches.

I am happy to merge this. @pkalita-lbl should we have a test case that specifically checks for the multimatch scenario? I am happy to do this as a future PR however.

pkalita-lbl (approved):
> should we have a test case that specifically checks for the multimatch scenario?

Definitely. I'm happy to take care of that next time I'm in the JsonSchemaGenerator tests. Otherwise this looks good!

PR #1348, opened 2023-03-18
Update typescriptgen.py

glass-ships:
Attempt to properly describe slots as abstract, if labeled as such


glass-ships:
> Or is that not an accurate representation of what this change will generate?

No, that's a valid assessment actually... we kind of had the assumption that we'd only be using abstract slots within abstract classes, but that may not necessarily _always_ be the case

cmungall:
@glass-ships what's the status of this?

glass-ships:
Thanks for the bump @cmungall, I think given the comments I sort of abandoned this idea. 

Patrick makes a good point, there's some [stackoverflow discussion](https://stackoverflow.com/a/72278792/9878785) about it, but it didn't seem worth exploring at the time. 

This can probably be closed as unplanned if you're ok with that

PR #1351, opened 2023-03-21
fix(docgen): emphasizing multiline descriptions

### LINKED ISSUE #1350: Multiline descriptions not being correctly rendered on the documentation
**Describe the bug**

Documentation generated fails to properly emphasize multiline descriptions. See screenshot below.

**To reproduce**
1. Go to <https://linkml.io/linkml-model/docs/MinimalSubset/>
2. See multiline description not being emphasized and starting and ending with an underscore.

**Expected behavior**

Multiline descriptions are properly emphasized with italic font.

**Screenshots**

image

**Additional context**

The issue is in the Jinja2 templates and therefore afects not only the LinkML online documentation, but any other generated documentation.

Silvanoc:
The MarkDown markup for emphasizing being used for descriptions fails if the description is multiline. This patch splits first the description and then adds the emphasis to each single line.

Fixes #1350

### DIFF
--- a/linkml/generators/docgen/class.md.jinja2
+++ b/linkml/generators/docgen/class.md.jinja2
@@ -6,7 +6,10 @@
 
 
 {% if element.description %}
-_{{ element.description }}_
+{% set element_description_lines = element.description.split('\n') %}
+{% for element_description_line in element_description_lines %}
+_{{ element_description_line }}_
+{% endfor %}
 {% endif %}
 
 {% if element.abstract %}
--- a/linkml/generators/docgen/enum.md.jinja2
+++ b/linkml/generators/docgen/enum.md.jinja2
@@ -1,7 +1,10 @@
 # Enum: {{ gen.name(element) }}
 
 {% if element.description %}
-_{{ element.description }}_
+{% set element_description_lines = element.description.split('\n') %}
+{% for element_description_line in element_description_lines %}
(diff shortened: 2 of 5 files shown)


cmungall:
🙏 🙏 🙏

sujaypatil96:
Thank you so much for the contribution @Silvanoc. I was looking through the code and saw that this code block was necessary in addition to the code you provided:

```
{%- if element_description_line|d('', true) %}
_{{ element_description_line }}_
{% endif -%}
```

This is to avoid printing whitespaces or empty strings if they somehow managed to seep into the description.

Silvanoc:
> This is to avoid printing whitespaces or empty strings if they somehow managed to seep into the description.

You might not want to emphasize empty lines, but you don't want to remove them either. That's the MarkDown way to create paragraphs, what I in fact use a lot in big descriptions.

sujaypatil96:
@Silvanoc right, but if you look at the generated markdown, there's unnecessary `__` that show up in there when there's empty strings right?

Silvanoc:
> @Silvanoc right, but if you look at the generated markdown, there's unnecessary `__` that show up in there when there's empty strings right?

@sujaypatil96 good that you asked me to look at the generated markdown. I would have expected to get the unnecessary `__` that you mention, but I don't!

PR #1368, opened 2023-03-27
Fix markdown multiline description rendering

sujaypatil96:
Thank you @Silvanoc for working on PR #1351 

I just wanted to propose a quick addition to your work in this PR.

### DIFF
--- a/linkml/generators/docgen/class.md.jinja2
+++ b/linkml/generators/docgen/class.md.jinja2
@@ -8,7 +8,9 @@
 {% if element.description %}
 {% set element_description_lines = element.description.split('\n') %}
 {% for element_description_line in element_description_lines %}
+{%- if element_description_line|d('', true) %}
 _{{ element_description_line }}_
+{% endif -%}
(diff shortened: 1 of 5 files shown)


Silvanoc:
In reply to [a comment of PR 1351](https://github.com/linkml/linkml/pull/1351#issuecomment-1486295659):

> This is to avoid printing whitespaces or empty strings if they somehow managed to seep into the description.

@sujaypatil96 you might not want to emphasize empty lines, but you don't want to remove them either. That's the MarkDown way to create paragraphs, what I in fact use a lot in big descriptions.

Silvanoc:
After having checked the resulting MarkDown:

1. I don't see the issue that you appear to be facing (empty lines resulting in `__`).
2. In fact the resulting Markdown on my side is exactly the same without and with your patch.

sujaypatil96:
Oh wow, okay, thank you for confirming that @Silvanoc. I'll check if there's some weirdness with what I'm doing.

Issue #1317, opened 2023-03-02
as of linkml-runtime 1.4.7, schemaview can't build a  view of a root schema and it's imports from a URL


turbomam:
For example, trying to make a view of https://raw.githubusercontent.com/linkml/linkml-model/main/linkml_model/model/schema/meta.yaml

> urllib.error.HTTPError: HTTP Error 404: https://w3id.org/home/mark/.cache/pypoetry/virtualenvs/submission-schema-DC6HKp4p-py3.9/lib/python3.9/site-packages/linkml_runtime/linkml_model/model/schema/units.yaml

I works with linkml-runtime 1.4.5

On my Ubuntu 20 system, I am told that

> Because submission-schema depends on linkml-runtime (1.4.6) which doesn't match any versions, version solving failed.

cmungall:
likely culprit: https://github.com/linkml/linkml-runtime/pull/248

turbomam:
I continue to have this problem in linkml-runtime > 1.4.7

turbomam:
@cmungall this is what we just discussed in our 1-on-1 meeting

PR #263, opened 2021-07-07
Adding markdown pages for subsets and adding markdown tests Fixes #262 Fixes #250 See also #245

cmungall:

PR #1379, opened 2023-03-31
Test case for imports using a schema fetched from a URL

turbomam:
- #1317


sierra-moxon (approved):
@turbomam - both of these tests pass locally for me.  Do you want to keep them in the repo for future testing?  I see that they are skipped b/c the schema could change.  For what it's worth, the tests you wrote seem good to me; you're counting classes and checking for a named schema (these seem least likely to change :)).

sujaypatil96:
@sierra-moxon @turbomam: since we're checking if schemas are being imported correctly or not, wouldn't it better to check if known classes from an imported schema were showing up in the SchemaView object rather than checking generally if the number of classes is greater than 0?

So basically in the case of [mixs.yaml](https://github.com/GenomicsStandardsConsortium/mixs/blob/main/model/schema/mixs.yaml), we know that [soil.yaml](https://github.com/GenomicsStandardsConsortium/mixs/blob/main/model/schema/soil.yaml) is an imported schema. We could check that the classes from _soil.yaml_, i.e., [soil](https://github.com/GenomicsStandardsConsortium/mixs/blob/main/model/schema/soil.yaml#L12) class is showing up in the resultant SchemaView object.

cmungall:
This PR was replaced by this one: #263 so I will close this

Note the test is skipped as it is fragile

PR #1420, opened 2023-05-03
jsonschemavalidator caching keyed on schema and entry point

### LINKED ISSUE #1419: jsonschema validation caching should be keyed on both schema and entry point
The changes made in https://github.com/linkml/linkml/pull/1363 consider only the _schema_ when implementing the caching logic.

However, the JSON Schema that the instance data needs to be validated against is dependent on two parameters - the schema, as well as the entry point, which is the `target_class`.

sujaypatil96:
Fixes #1419


pkalita-lbl (commented):
The change looks good. Is it possible to add a test for this? It doesn't necessarily have to verify the actual caching behavior, but even something that just reuses a single validator instance to validate multiple different classes would be really useful.

pkalita-lbl (approved):

PR #1445, opened 2023-05-12
python 3.7 compatibility for pydanticgen

kevinschaper:
linkml_runtime still supports 3.7, but was failing on pydanticgen's import of Literal from typing. This PR changes the import for Literal to typing_extensions.


pkalita-lbl (changes requested):
I don't see `typing_extensions` listed as a dependency in `pyproject.toml`. Probably best to not rely on transitive dependencies like that.

Given that it's only needed in older python versions it might be good to use https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies and then try to import it from `typing` and fallback to `typing_extensions` if it fails

pkalita-lbl (approved):
\*chef's kiss\*

PR #1475, opened 2023-06-01
Fix LinkML types appearing in generated JSON Schema

### LINKED ISSUE #1465: Getting "uriorcurie" as JSON-Schema type, what is invalid
**Describe the bug**
With 7cd4ea4ea26343566e7f5edd6ab3fead77affedf the [calculation of a slot type](https://github.com/linkml/linkml/blob/2b29658ce67f6507a8d510383ed18c31dab97159/linkml/generators/jsonschemagen.py#L330) results in invalid type for classes with identifier.

If the slots appoints to a class that has an `uriorcurie` identifier, then `id_slot.range` becomes `uriorcurie` and as a consequence `uriorcurie` is returned as type. But `uriorcurie` is not a valid JSON-Schema type.

I've added [this comment](https://github.com/linkml/linkml/commit/7cd4ea4ea26343566e7f5edd6ab3fead77affedf#r115207021) before creating this issue.

pkalita-lbl:
Fixes #1465

### DIFF
--- a/linkml/generators/jsonschemagen.py
+++ b/linkml/generators/jsonschemagen.py
@@ -306,7 +306,7 @@ class JsonSchemaGenerator(Generator):
         })
         self.top_level_schema.add_def(enum.name, enum_schema)
 
-    def get_type_info_for_slot_subschema(self, slot: AnonymousSlotExpression, parent_slot_is_inlined: bool) -> Tuple[str, str, Union[str, List[str]]]:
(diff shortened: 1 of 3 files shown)


cmungall (approved):

Silvanoc:
I've tested this branch on my side and it seems to fix the issue without introducing any new regressions.

turbomam (approved):
Thanks for doing this and including me as a reviewer. I trust the two reviewer before me.

Issue #932, opened 2022-09-06
Making equivalentClass assertions in linkml


deepakunni3:
Not sure if this is already possible with the current metamodel but it would be very helpful to be able to mark a class in a schema be equivalent to another class, where the another class is in the same schema or an external schema or an external ontology.

For example, using linkml I would like to say:

```yaml

my gene:
  description: "My gene in my schema"
  equivalent_to: "SO:0000704"

```

and get the following via gen-owl:

```
ex:MyGene owl:equivalentClass SO:0000704 .
```

@turbomam suggested `class_uri` but that seems to translate to `ex:MyGene skos:exactMatch SO:0000704`.

deepakunni3:
Also, @matentzn raised an observation that shouldn't `class_uri` be used to replace the class in the generated owl instead of saying `ex:MyGene skos:exactMatch SO:0000704`? 

I seem to recall that is how it was in 2020/2021. Was the change intentional or a bug in the owl generator?

ialarmedalien:
See also #922 -- could be adapted to these purposes.

cmungall:
Yes as @ialarmedalien says, the generic mapping proposal could encompass a wide variety of mapping and logical relationships.

@deepakunni3 - this has always been the behavior but this is a sensible enhancement request, I made a separate issue: #933

deepakunni3:
Thanks @ialarmedalien and @cmungall for the pointers! 👍 

Thank you for clarifying about the behavior for `class_uri`.

deepakunni3:
So the only way to define a class as owl:equivalentClass to another class would be to use the (planned) structured mappings?

cmungall:
Two things that could be done v quickly

1 option on owlgen to make EC axioms for class uri to def uri
2 same but for custom annotation

deepakunni3:
Makes sense! I think option 1 is easy to implement. And I am looking forward to seeing the structured mapping implementation :)

cmungall:
Fixed by #1484 

Additional tests added in #1843

PR #1484, opened 2023-06-08
Add option to assert equivalentClass between a class and its class_uri

deepakunni3:
This PR adds the ability to add an equivalentClass between a class and its class_uri.

Addresses concern raised in https://github.com/linkml/linkml/issues/932

Currently the class_uri is added as a `skos:exactMatch` by default.

When running gen-owl with `--assert-equivalent-classes` flag, the default behavior is changed and there is an `owl:equivalentClass` between class and its class_uri.

Happy to tweak this PR further as needed and appropriate :)


sierra-moxon (approved):

deepakunni3:
Thank you for the quick review and merge @sierra-moxon @cmungall 🚀

PR #1490, opened 2023-06-09
Codespell: config, workflog + some typos fixed

yarikoptic:
Currently I skip `tests/` entirely for codespell testing ~~but in the last commit I did add fixing of the most common typos. But I had to exclude the `tests/data/hp.dill` binary file so who knows -- may be fixing not thoroughly would bring some tests failures.  The last commit could be omitted then.~~

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,22 @@
+---
+name: Codespell
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Codespell
+        uses: codespell-project/actions-codespell@v2
--- a/SPECIFICATION.md
+++ b/SPECIFICATION.md
@@ -163,7 +163,7 @@ prefixes:
   qud: http://qudt.org/1.1/schema/qudt#
 ```
 
-The CURIE `wgs:lat` will exand to http://www.w3.org/2003/01/geo/wgs84_pos#lat.
+The CURIE `wgs:lat` will expand to http://www.w3.org/2003/01/geo/wgs84_pos#lat.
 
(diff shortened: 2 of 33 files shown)


cmungall:
This is awesome!

@hrshdhgd should we add this to our cookiecutters?

Not 100% sure of the implications of a merge - will future PRs be blocked if there are errors? How are false positives removed, is there a dictionary file?

yarikoptic:
> This is awesome!

Thank you! `codespell` is indeed quite nice!  An update: I decided to take 2 steps/commits back -- force pushed changes before I touched `tests/` so we could get some stable grounds and then I would follow up  with a PR on top.


> @hrshdhgd should we add this to our cookiecutters?

FWIW I already submitted and @hrshdhgd merged https://github.com/linkml/linkml-project-cookiecutter/pull/71 which adds codespell to [cookiecutter](https://github.com/linkml/linkml-project-cookiecutter/blob/main/.github/workflows/codespell.yml) but not to [the template itself](https://github.com/linkml/linkml-project-cookiecutter/tree/main/%7B%7Bcookiecutter.project_name%7D%7D/.github/workflows)

> Not 100% sure of the implications of a merge - will future PRs be blocked if there are errors?

yes

> How are false positives removed, is there a dictionary file?

yes. Config in this repo is in [pyproject.toml](https://github.com/linkml/linkml/pull/1490/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R147) and you can add `skip` globs for files, `ignore-word-list` for individual words (lowercased), or `ignore-regex` for regexes to ignore altogether.  If there is no `pyproject.toml`, then it goes to new `.codespellrc` like https://github.com/linkml/linkml-project-cookiecutter/blob/main/.codespellrc

yarikoptic:
I rewrote history again to address that code change in https://github.com/linkml/linkml/pull/1490/commits/204e13ff49e10596025919697f7245a35461c81b which was picked up in https://github.com/linkml/linkml/pull/1493/files#r1227051840 by @sierra-moxon sharp eye.  CI would need another round of blessings to be ran.

sierra-moxon (approved):
thanks again!

PR #1493, opened 2023-06-11
codespell+: extra on top of #1490 with some coverage of the tests

yarikoptic:
#1490 but with extra commits to partially fix some typos (some files renamed too) in `tests/`. Primarily to see if tests would be happy now


linkml/generators/erdiagramgen.py
```diff
@@ -203,11 +203,11 @@ def serialize_diagram(self, diagram: ERDiagram) -> str:
         :param diagram: ER Diagram
         :return:
         """
-        ser = str(diagram)
+        set = str(diagram)
```
sierra-moxon, line 206: I think I would feel more comfortable if we limited codespell to the documentation as a first pass?  We could slowly add it back in, e.g. this change is not one we want codespell to do for us, but I agree the variable name could be updated to be more descriptive (I think this is `er` diagram, not `set`)

sierra-moxon (changes requested):
Thanks so much for all the work here!  I am absolutely dependent on codespell in Biolink Model :D -- would you be willing to limit the automatic changes codespell produces to only the .md files as a first pass?  Some of the changes here are good suggestions for more verbose variable names, but I think we want to do those intentionally?  (e.g. limit codespell to .md files, .rst files, exclude .py files as a first pass?)

sierra-moxon:
Also - thank you so much for opening the fork to maintainers - what a lovely feature!!  
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

yarikoptic:
I don't know how to "whitelist", and only how to "blacklist"/skip files ATM .   Since there is an incentive to go piece meal, I will just close this PR for now, and fix for `ser` in #1490 which ATM covers only 33 files  with following extensions

```shell
❯ git diff --name-only origin/main... | sed -e 's,.*\.,,g' | sort | uniq -c | sort -n
      1 jsonld
      1 toml
      1 yaml
      1 yml
      2 rst
      3 ipynb
     10 py
     14 md
```
so probably nothing really to exclude from those.  After we are done there, we can get back here ;)

linkml/generators/erdiagramgen.py
yarikoptic: thank you for catching this!  I have addressed it by renaming `ser` to `er` in 204e13ff49e10596025919697f7245a35461c81b within #1490

Issue #1494, opened 2023-06-12
Refreshed design for LinkML validation


deepakunni3:
# Refreshed design for LinkML validation

This ticket serves as a way of capturing the design decisions for enhancing the validation functionality in LinkML.

The main goal of the validator is to take one (or more) object and validate against a given LinkML schema, while ensuring that the validator has a common interface, is configurable, and easy to extend.

After taking into consideration the functionalities of `linkml-validate` (LinkML's in-built validator) and linkml-validator (a prototype independent of linkml-validate), the proposed structure of this refreshed LinkML validation can be divided into 3 parts:
- **CLI**
- **Validators**
- **Plugins**

The flow of control is from CLI > Validator > Plugin where,
- A CLI utility is invoked from the command line
- Depending on the input, the CLI calls the appropriate validator instance
- Validator calls one or more plugins
- Each plugin generates one or more validation results
- Validator is responsible for collecting all validation results into a validation report
- Validator emits the validation report after all checks are completed for that object

## CLI

A CLI utility that enables validation via the command line. This will be quite similar to the interface proivded by `linkml-validate`.

This CLI utility is responsible for calling the appropriate validator and plugins based on input arguments.

### Input

- Input File, File URL or an endpoint
- Input Format (JSON, JSON Lines, TSV, CSV, RDF-Turtle, N Triples, etc.)
- Input Stream (I/O stream that contains objects in one of the above formats)
- Schema (schema file or schema URL)
- Generator Args (Arguments to the generator that will be used)
- Plugins + Plugin Args (Plugins to use and Arguments to the plugins)

### Output

- Output File
- Output Format
- Output Stream (I/O stream where validation results are streamed)

## Validators

Validator is responsible for dealing with:
- generation of an appropriate schema artifact
- fetching objects for validation (file or endpoint)
- serializing validation reports (file or stream)

Structure:
- `BaseValidator`: The base validation class
- `JsonschemaValidator`: Validating JSON against a JSONSchema
- `SparqlValidator`: Validating data from a SPARQL endpoint
- `TableValidator`: Validating data from a delimiter-separated file (CSV, TSV)
- ...

The different types of validator are to ensure that I/O and generation of schema artifacts are handled accordingly (plus any use-case specific optimization that might be required).

Each validator can have its own default choice of plugins and plugin arguments. But this can be overridden at runtime.

Methods for Validator:
- `init`: initialize the validator; Accepts a schema YAML, a generator instance, and generator arguments
[...]

PR #1502, opened 2023-06-16
Iterate over all errors found in JsonSchemaDataValidator

### LINKED ISSUE #891: linkml-validate only shows first erorr
there should be a mode to report all errors

### LINKED ISSUE #1038: `linkml-vallidate` emits the text "`None`" when validation is successful.
[Part 1 of the LinkML tutorial](https://linkml.io/linkml/intro/tutorial01.html) asks you to run `linkml-validate -s personinfo.yaml data.yaml ` stating that "You should see no errors".  What you _do_ see is:
```
> linkml-validate -s personinfo.yaml data.yaml 
None
> 
```

The "None" shouldn't be there.

I would _also_ propose that we might consider some sort of summary that shows that the validator actually did its job.  This could either be a default that you could disable if you really wanted silence, or something that you called for.  We also need to decide whether, in this context, errors should go to `stdout` or `stderr`.  The summary definitely should go to `stdout`.

### LINKED ISSUE #1337: CSV Validation Errors
**Describe the bug**
`linkml-validate` accepts `csv` as `--input-format` but returns unexpected validation results for csv files. Same data expressed in `tsv` format gets validated as expected.

**To reproduce**
For this [sample linkml schema](https://github.com/Shalsh23/linkml-tutorial/blob/experimentation/linkml-tutorial/personinfo.yaml) , the following sample files should throw validation errors but they get successfully validated.
1. [Missing identifier](https://github.com/Shalsh23/linkml-tutorial/blob/experimentation/linkml-tutorial/csv_samples/invalid-person-data-1.csv)
2. [Missing required field](https://github.com/Shalsh23/linkml-tutorial/blob/experimentation/linkml-tutorial/csv_samples/invalid-person-data-2.csv)
3. [Invalid pattern](https://github.com/Shalsh23/linkml-tutorial/blob/experimentation/linkml-tutorial/csv_samples/invalid-person-data-3.csv)

Sample command: `linkml-validate -s personinfo.yaml -S persons csv_samples/valid-person-data-1.csv`

**Expected behavior**
CSV data files with missing id and required fields should throw validation errors.

**Additional Context**
It is possible to circumvent these errors by converting `csv` to `tsv` and then using `linkml-validate`. However, csv files are pervasive in our [community](http://essdive.lbl.gov/)  and it is important for us to be able to use linkml tools with csv files directly.

pkalita-lbl:
The main purpose of these changes is to allow `JsonSchemaDataValidator` to report more than the first error it finds. The current code has two main failure modalities:

1) Data is deserialized via `linkml-runtime`'s loader classes into dataclasses from produced by pythongen. If, for example, a string like `"string"` is found where an integer is expected, an exception is raised by the dataclass instantiation and the validation process halts. These changes use the `load_as_dict` from the latest `linkml-runtime` if the selected loader implements it so that dataclasses are bypassed altogether.
2) The JSON Schema validation is currently done using a [method](https://python-jsonschema.readthedocs.io/en/stable/api/jsonschema/validators/#jsonschema.validators.validate) that raises an exception on the first problem it encounters. With this change a different [method](https://python-jsonschema.readthedocs.io/en/stable/api/jsonschema/protocols/#jsonschema.protocols.Validator.iter_errors) that iterates through all errors is used. 

A few other miscellaneous additions here:

1) Currently there is some confusion around loading TSV vs CSV data. This has been resolved by using the new `TSVLoader` and `TSVDumper` classes from the latest `linkml-runtime`.
2) `JsonSchemaDataValidator` currently does some caching to only generate the JSON Schema specification when needed. However the caching was only happening in one of the validation methods and missed one of the parameters (`closed`) that should have been part of the cache key. These changes generally overhaul that caching mechanism to make it more robust. 

Caveats:

1) Currently the `JsonSchemaDataValidator.validate_object` and `JsonSchemaDataValidator.validate_dict` methods return nothing. With these changes they return a list of strings representing validation problems. It would probably be better for these to return lists of objects with more details about the problem, but I'm deferring that to a larger [validator refactor](https://github.com/linkml/linkml/issues/1494).

Fixes #1337
Fixes #1038
Fixes #891


cmungall:
oops, sorry. I merged my docs PR yesterday when we had the false positive github actions, fail thinking _I should never merge a failing PR but it's a docs change, what could possibly go wrong_

cmungall (approved):

PR #1517, opened 2023-06-30
Pluggable validation module

pkalita-lbl:
This is a first pass at the ideas outlined in #1494. The work is implemented in a new package, `linkml.validator`. This is not yet the recommended replacement for the classes in `linkml.validators`. But I think it has reached enough critical mass to merge it in and keep adding functionality in smaller PRs.


pkalita-lbl:
@deepakunni3 this branch is getting a little long-lived. Do you have any concerns about getting this merged into `main` at this point and continuing to work on the refactor in smaller PRs? I added a "danger! experimental!" warning to `linkml/validator/__init__.py` for whatever that's worth, and everything in the new package is nominally functional (if possibly incomplete). I already pitched this to @cmungall and he was on board with that plan.

cmungall (approved):

deepakunni3 (approved):
After reviewing the code I just realized that we pass validation context to the process method each time it is called. Is there a reason why it is passed to the process method instead of just once to the Validation Plugin class when its instantiated (i.e. to the init method)?

Do we expect the validation context to change between objects? 

Of course this doesn't have to hold up the merging of this PR :)

pkalita-lbl:
> Is there a reason why it is passed to the process method instead of just once to the Validation Plugin class when its instantiated

Yeah, it's because the class you're validating against (`target_class`) is part of the context. And by providing the `target_class` at "validate time" instead of at initialization you can do things like:

```python
my_validator.validate(a_person, "Person")
my_validator.validate(a_pet, "Animal")
```

The context is also intended to be sort of internal plumbing between a `Validator` instance and plugin instances so ideally an end user wouldn't even need to know it exists.

PR #1527, opened 2023-07-06
Documentation updates

cmungall:


sierra-moxon (approved):

docs/schemas/annotations.md
```diff
@@ -0,0 +1,157 @@
...
+attached to schema elements via annotations.
+
+Other schema elements can also instantiate the `Reviewable` class:
+
+```yaml
+classes:
+  Person:
+    instantiates: 
+      - mymetamodel:Reviewable
+    annotations:
```
sierra-moxon, line 105: how does this `annotations` component relate to the Reviewable class? 
if I think of it like a mixin, then I would expect to only see `annotations` in the `attributes` section?

cmungall: Added a section:

Note that here the `attributes` section of the `Person` definition is the attributes that an
instance of `Person` may have. The `annotations` section pertains to the *class* Person, not instances.

PR #1528, opened 2023-07-07
update markdown generator documentation

sujaypatil96:
Updating the markdown generator docs on the linkml documentation site clarifying the use of `gen-doc` over `gen-markdown`.


cmungall:
I think the windows fail is spurious... but why do we even do codecov in windows?

cmungall (approved):

PR #1531, opened 2023-07-12
Add pre-commit config with black, ruff

dalito:
@pkalita-lbl This is a follow-up of your #1520 which added extended checks to CI/tox. 

With [pre-commit](https://pre-commit.com/) code can be tested (and fixed) before the commit and before it hits CI-testing.

Installation and use is simple even on Windows. The following commands were run in the root directory of the git clone. They show the typical steps including installation, update and running all checks (AKA "hooks").
```bash
$ pipx install pre-commit
'pre-commit' already seems to be installed. Not modifying existing installation in...

$ pre-commit install
pre-commit installed at .git\hooks\pre-commit

$ pre-commit autoupdate
[https://github.com/pre-commit/pre-commit-hooks] already up to date!
[https://github.com/psf/black] already up to date!
[https://github.com/astral-sh/ruff-pre-commit] already up to date!

$ pre-commit run
```

Pre-commit runs by default only for staged files. To check all files:
```bash
$ pre-commit run --all-files
```

To run only one specific hook, for example ruff, use:
```bash
$ pre-commit run ruff
```

To commit with failing checks pre-commit can be skipped by `--no-verify`. For example:
```bash
$ git commit . -m 'quick fix' --no-verify
```
~~I intentionally left out `codespell` because I found it to noisy. But in principle it can also run as a pre-commit hook.~~ (see update below)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,3 +18,9 @@ repos:
   hooks:
   - id: ruff
     args: [--fix, --exit-non-zero-on-fix]
+- repo: https://github.com/codespell-project/codespell
+  rev: v2.2.4
+  hooks:
+  - id: codespell
+    additional_dependencies:
+      - tomli
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,20 @@
+default_language_version:
+  python: python3
+repos:
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v4.4.0
+  hooks:
+  - id: end-of-file-fixer
+  - id: check-yaml
+  - id: check-toml
+  - id: trailing-whitespace
+- repo: https://github.com/psf/black
+  rev: 23.7.0
+  hooks:
+  - id: black
+    args: ['--target-version', 'py38']
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  rev: v0.0.278
+  hooks:
+  - id: ruff
+    args: [--fix, --exit-non-zero-on-fix]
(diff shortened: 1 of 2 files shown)


cmungall (approved):

pkalita-lbl:
Cool! I've seen other projects using `pre-commit` and it seems interesting. I'll test it out in this branch shortly. If it looks good I can also add a note about it to our contributing guide since it seems like it requires a bit of setup.

rly:
I have been enjoying pre-commit on a separate project: https://github.com/hdmf-dev/hdmf/blob/dev/.pre-commit-config.yaml

There is a hook for codespell as well.

dalito:
Initially I had codespell in the pre-commit config. However, since there is not yet a linkml-specific configuration for `codespell` in `pyproject.toml` it reports many errors and makes pre-commit fail. I would suggest to add it later (if at all).

This adds it if you want to try:
```yaml
- repo: https://github.com/codespell-project/codespell
  rev: v2.2.4
  hooks:
  - id: codespell
    additional_dependencies:
      - tomli
```

dalito:
> If it looks good I can also add a note about it to our contributing guide since it seems like it requires a bit of setup.

@pkalita-lbl - It is completely optional which means it is not a hard development dependency. If people don't like it they can just ignore it.

pkalita-lbl:
I tried it out and had some concerns that `pre-commit` would potentially check/fix files in directories that weren't checked/fixed by the tox testenvs. So I created a new PR based on this one: https://github.com/linkml/linkml/pull/1537. @dalito can you please check if you agree with what I did there?

pkalita-lbl:
Closing in favor of #1537

PR #1537, opened 2023-07-14
Add pre-commit config

pkalita-lbl:
This is building on top of https://github.com/linkml/linkml/pull/1531 with additional configuration to exclude Python files in `/docs`, `/examples`, and `/notebooks` from `black` and `ruff`. Previously this was accomplished by telling the commands to _only_ run against the `linkml` and `tests` directories in `tox.ini`. Configuring it with excludes instead allows `pre-commit` to also understand which directories we don't want to check. For more info on Black's `force-exclude` option see https://github.com/psf/black/issues/438).

This also adds documentation to the developer's guide on how to use black, ruff, and pre-commit.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,26 @@
...
+  hooks:
+  - id: end-of-file-fixer
+  - id: check-yaml
+  - id: check-toml
+  - id: trailing-whitespace
+- repo: https://github.com/psf/black
+  rev: 23.7.0
+  hooks:
+  - id: black
+    args: ['--target-version', 'py38']
```
dalito, line 15: Since the target version is already set in pyproject.toml we don't need it here and should remove the line.

dalito (approved):
Looks good! Just remove the args line for black from `.pre-commit-config.yaml` to ensure that black runs with exactly the same config options via tox and via pre-commit.

.pre-commit-config.yaml
pkalita-lbl: Good catch! Thank you

cmungall (approved):

sierra-moxon (approved):

PR #1543, opened 2023-07-18
feat(validators): add option include descendants

### LINKED ISSUE #1544: Add `--include-range-class-descendants` option to "linkml-validate"

`linkml-validate` doesn't accept children of a class specified in a range as valid. Being used to the behavior of inheritance in OOP that also applies to the generated Python code, I find it misleading and frustrating.

IMO (point of view of an OOP programmer, might be different for an "ontologist") accepting children should be the default behavior. Since `gen-json-schema` already has the option `--include-range-class-descendants`, having such an option for `linkml-validate` would be also fine for me.

• Medium - As of now I have to ignore validation errors that appear due to this limitation of `linkml-validate`

• Short-term - 2-4 weeks

**Additional context**
I'm providing a PR to implement this feature.

Silvanoc:
"gen-json-schema" has the option "--include-range-class-descendants" so that children of a class specified in a range are added to the generated JSON-Schema as allowed classes.

This patch adds a similar option to "linkml-validate", with the same result in the JSON-Schema being used for validation as the equally named option for "gen-json-schema".

Closes #1544

### DIFF
--- a/linkml/validators/jsonschemavalidator.py
+++ b/linkml/validators/jsonschemavalidator.py
@@ -26,7 +26,7 @@ class HashableSchemaDefinition(SchemaDefinition):
 
 
 @lru_cache(maxsize=None)
-def _generate_jsonschema(schema, top_class, closed):
+def _generate_jsonschema(schema, top_class, closed, include_range_class_descendants):
     logging.debug("Generating JSON Schema")
     not_closed = not closed
     return JsonSchemaGenerator(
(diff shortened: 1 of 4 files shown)


Silvanoc:
I wonder why tests are failing on the resulting JSON-LD context, since changes in `linkml-validate` shouldn't change anything in `gen-jsonld-context`.

Silvanoc:
Set to draft to write some unittests for it.

pkalita-lbl (approved):
Looks great!

PR #1559, opened 2023-07-31
Generate permissible JSON schema for empty enum

lkuchenb:
This patch changes the JSON schema generator such that it creates an unrestricted schema for enums for which no permissible values are specified in LinkML (which may be the case if a reachability query is specified instead).

We are running into this issue in the context of [dynamic enums](https://linkml.io/linkml/schemas/enums.html#dynamic-enums). The documentation says

> At this time, tooling support for dynamic enums is maturing, but you can still go ahead and use them in your schemas. The default behavior will be too permissive – however, you still gain additional clarity in your schema documentation.

However, currently, the JSON schema generator does not create permissible schemas, it creates schemas for which no values can be provided at all (`"enum" : []`).

### DIFF
--- a/linkml/generators/jsonschemagen.py
+++ b/linkml/generators/jsonschemagen.py
@@ -319,10 +319,11 @@ class JsonSchemaGenerator(Generator):
         enum_schema = JsonSchema(
             {
                 "type": "string",
-                "enum": permissible_values_texts,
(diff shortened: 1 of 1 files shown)


pkalita-lbl (approved):
Looks good! Thank you!

PR #1561, opened 2023-08-01
make line length consistent between black and ruff

sierra-moxon:


cmungall:
I guess this needs to be accompanied by actual reformatting?

sierra-moxon:
fixed formatting so that we can make these consistent.

pkalita-lbl (approved):

PR #1563, opened 2023-08-01
Port tests/test_issues directory to pytest framework

pkalita-lbl:
Summary of changes:

* All tests in `tests/test_issues` have been refactored to use the pytest framework instead of unittest. Generally speaking this entails:
  * What were previously `test_*` methods of `unittest.TestCase` subclasses become top-level `test_*` functions in each module.
  * Pytest embraces and enhances `assert` statements so they are used in place of `assert*` methods from `unittest.TestCase` 
  * Tests that previously used `TestEnvironmentTestCase` to locate test input files and compare results with saved output files now use two custom pytest fixtures: `input_path` and `snapshot` which are defined in `tests/conftest.py`. The "known, good files" used for comparison are now stored in `__snapshots__` directories alongside tests instead of `output` directories to make their purpose more clear. 
* The Makefile target and GitHub workflows that run tests have been updated to use the pytest CLI
* Updated the contributor guidelines to emphasize using pytest, capture the current state of the transition to pytest, and document our custom pytest fixtures.


pkalita-lbl:
@cmungall after working through a few cross-platform issues I think this is now ready for review. I'm not totally sure why your compliance test suite changes are showing up here as co-authored commits after I merged `main` into this branch; so apologies in advance for a sightly confusing diff because of that.

One other thing to note: when #1551 was merged in, the GitHub testing workflow was still using the `unittest` test runner. So those new tests written for the `pytest` framework weren't being run automatically. I've switched the workflow (and `Makefile`) over to the `pytest` runner in this branch. With the compliance tests, It now takes about 50 to 60 minutes for the build and test action to complete. I don't know if we should figure out some way to reduce that time before cutting over to `pytest` or get these changes in and then figure it out? Thoughts?

cmungall (approved):

cmungall:
np about the mixed commit streams

I was worried about the compliance tests taking so long - it's a lot less on my laptop, but as expected running on actions will be slower.

I think we should merge then figure out a plan. It's not like the tests were speedy before. I'll make an issue for this

dalito:
Thanks for doing this!

PR #1572, opened 2023-08-08
Add a schema transformer to make logical constraint models.

cmungall:


linkml/transformers/logical_model_transformer.py
```diff
@@ -0,0 +1,710 @@
...
+from linkml_runtime.linkml_model import ClassDefinitionName, SchemaDefinition
+from linkml_runtime.linkml_model.meta import (
+    AnonymousSlotExpression,
+    AnonymousTypeExpression,
+    SlotDefinition,
+    SlotDefinitionName,
+    TypeDefinition,
+)
+
+from linkml.transformers.model_transformer import ModelTransformer
```
pkalita-lbl, line 30: Looks like the tests can't run because this import can't be resolved. I don't see a a `model_transformer.py` module here. Maybe that didn't get checked in?

pkalita-lbl (approved):

Issue #1545, opened 2023-07-19
update pydantic to v2.0


ialarmedalien:
[pydantic 2.0](https://github.com/pydantic/pydantic) is now out, with various fixes and improvements, plus [some breaking changes to the API](https://docs.pydantic.dev/2.0/migration/). v1.x ships as part of v2, so it's possible to update the pydantic package to the latest version and continue to use v1 under the hood, e.g.

```py
from pydantic.v1 import BaseModel
```

That also makes incremental conversion to the full 2.0 API much easier.

Since package managers like poetry will report incompatibility in dependencies if you have packages that rely on v1 and v2, the best update policy (assuming that linkml wants to migrate to v2 at some point) would be to update all the pydantic-using/requiring linkml repos to use v2 at once, ideally with a version bump.

cmungall:
Thanks, agreed!

linkml-runtime is now unpinned:

- https://github.com/linkml/linkml-runtime/pull/272

I think this could cause problems with repos that include generated pydantic models? This should be resolvable in the short term by those repos pinning to v1. (or by providing an easy way of modifying the imports as above)

But I think what we need to do ASAP is modify pydanticgen to allow pydantic2 natively.

It's unfortunate that the python import system forces us to walk lockstep in this way...

sierra-moxon:
We support both v1 and v2 now; as a next step, we can set our dev default to v2, but we're not quite there yet.  Closing as complete.

PR #1575, opened 2023-08-10
Add option to generate Pydantic v2

glass-ships:
Addresses #1545 

There aren't too many significant [changes](https://docs.pydantic.dev/latest/migration/) between v1 and v2

This PR proposes a new flag for the `gen-pydantic` command (currently `--pydantic_version [1|2]`, but open to suggestions) which leads to small changes in the default template (now a function returning a string)

Note this PR doesn't (yet) update Linkml's dependency on Pydantic to ^2.0, only allows model generation in v2

### DIFF
--- a/linkml/generators/pydanticgen.py
+++ b/linkml/generators/pydanticgen.py
@@ -30,7 +30,11 @@ from linkml.generators.oocodegen import OOCodeGenerator
 from linkml.utils.generator import shared_arguments
 from linkml.utils.ifabsent_functions import ifabsent_value_declaration
 
-default_template = """
+
+def default_template(pydantic_ver: str = "1") -> str:
+    """Constructs a default template for pydantic classes based on the version of pydantic"""
+    ### HEADER ###
+    template = """
 {#-
 
   Jinja2 Template for a pydantic classes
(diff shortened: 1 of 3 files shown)


cmungall:
Thanks!

Looks like this mixes in an unrelated PR?

Also make sure to link to the main issue!

glass-ships:
> Thanks!
> 
> Looks like this mixes in an unrelated PR?
> 
> Also make sure to link to the main issue!

Thanks Chris, updated the comment to mention the related issue! 

Also yeah, I pulled in some changes i saw in another branch that seemed to fix some failing json related test, but only because i like nice green checkmarks

linkml/generators/pydanticgen.py
```diff
@@ -547,6 +579,7 @@ def default_value_for_type(self, typ: str) -> str:
 
 @shared_arguments(PydanticGenerator)
 @click.option("--template_file", help="Optional jinja2 template to use for class generation")
+@click.option("--pydantic_version", help="Pydantic version to use (1 or 2)", default="1")
```
pkalita-lbl, line 582: Check out https://click.palletsprojects.com/en/8.1.x/options/#choice-options and then you should be able to remove the bespoke value checking in lines 597-600. I believe using that will also show the valid options in the `--help` output.

pkalita-lbl (commented):
Looks good to me. Just one minor suggestion to make the CLI part better.

pkalita-lbl (approved):

kevinschaper:
The one change I could see making is passing the pydantic version into jinja and keeping all of the template conditionals in jinja, rather than having some in jinja and some happening as string concatenation while building the template. Let's bring it in though, so that we can support v2 ASAP, and if I find that frustrating later I can refactor it.

kevinschaper (approved):

PR #1580, opened 2023-08-15
3 new FAQ questions from NWD, fixes #1568

### LINKED ISSUE #1568: FAQ pages for common questions
Three questions that came up from community members during a LinkML presentation that we could think of adding FAQ entries for:

Q1) We allow terms from various ontologies in OBO, bioportal, etc. to be specified and mapped to various LinkML elements as CURIEs in our schemas/data models. Is there a _materialization_ step that checks the validity of the URIs that are resolved from those CURIEs?

Q2) Let's say a data entry tool, like DataHarmonizer is capturing data in a column that maps to a _slot_ in a LinkML schema that has a range of _uriorcurie_. Does LinkML validation check if the CURIEs being captured/recorded through the interface resolve into valid URIs?

Q3) When we choose to import a relatively large data model like biolink into our LinkML schemas, is it possible for us to import only a subset of biolink? As in, only select classes, slots, enums, etc. One of the reasons as to why we might want to import only a subset of elements is to significantly reduce documentation generation time. Currently generating documentation pages for the biolink schema is in order of hours, and this can help reduce that. But I'm unsure if it would go against linked data design principles since all data is so interconnected.

sierra-moxon:


docs/faq/modeling.md
```diff
@@ -259,6 +259,26 @@ slots:
 
 See [dynamic enums](https://linkml.io/linkml/schemas/enums.html#dynamic-enums) for more details.
 
+### Are CURIEs used in schema definitions checked for expandability and resolution?  
+
+No, not at this time.  However, linkml_runtime does have methods to help you expand the CURIEs in your data
+using the prefixes in your model (see: linkml_runtime.utils.namespaces.py) into URIs and then checking them is
```
cmungall, line 265: when we say checking do we mean:

- checking for syntactic validity according to the W3C URI specification
- link checking

docs/faq/modeling.md
```diff
@@ -259,6 +259,26 @@ slots:
 
 See [dynamic enums](https://linkml.io/linkml/schemas/enums.html#dynamic-enums) for more details.
 
+### Are CURIEs used in schema definitions checked for expandability and resolution?  
+
+No, not at this time.  However, linkml_runtime does have methods to help you expand the CURIEs in your data
+using the prefixes in your model (see: linkml_runtime.utils.namespaces.py) into URIs and then checking them is 
+a simple bit of python like [this](https://github.com/biolink/biolink-model/blob/master/scripts/verify_infores.py).
```
cmungall, line 266: it's not totally clear how to adapt this - maybe link to curies library?

cmungall (approved):

docs/faq/modeling.md
sierra-moxon: 👍 . removed.

pkalita-lbl:
I think this branch may be behind `main` on some of the test snapshot files. Try merging or rebasing `main` into/onto this branch and see if that clears up the test failures.

PR #1586, opened 2023-08-22
Convert tests/test_generators to pytest

pkalita-lbl:
* Add one directory-level fixture for locating the kitchen_sink schema used in many tests
* Turn TestCase methods into top-level test functions
* Replace TestCase assertion methods with assert statements
* Remove the output directory and any code that wrote into it. Testing is done in memory as much as possible. If actual files are needed, temporary files are used (via the tmp_path fixture)


cmungall (approved):

sujaypatil96 (approved):
Great refactor @pkalita-lbl!

* Checked some generator tests that previously used the kitchen_sink.yaml schema, to make sure they use the kitchen_sink_path fixture now, and they do ✅
* I'm assuming the tests that were using TestCase were part of some of the older generators like test_markdowngen.py, test_pythongen.py, test_rdfgen.py, etc. so I looked through those to make sure they all have top-level test functions, and they do ✅
* _assert_ statements present in all tests ✅

PR #1587, opened 2023-08-22
Update snapshots with changes from latest metamodel

pkalita-lbl:
The [latest change](https://github.com/linkml/linkml-model/commit/448146e53d5730b5e6735dac5ce40df205dd5681) to the metamodel has caused the expected output of some snapshot tests to change.

This represents a sort of loose, external dependency of these tests. We'll have to keep an eye out for often this happens and if it becomes onerous we might want to figure out how to make these tests more isolated.

### DIFF
--- a/tests/test_issues/__snapshots__/issue_167b.yaml
+++ b/tests/test_issues/__snapshots__/issue_167b.yaml
@@ -23,6 +23,10 @@ types:
     name: string
     definition_uri: https://w3id.org/linkml/String
     description: A character string
+    notes:
+    - In RDF serializations, a slot with range of string is treated as a literal or
+      type xsd:string.   If you are authoring schemas in LinkML YAML, the type is
+      referenced with the lower case "string".
     from_schema: https://w3id.org/linkml/types
     imported_from: linkml:types
     exact_mappings:
@@ -33,6 +37,9 @@ types:
     name: integer
     definition_uri: https://w3id.org/linkml/Integer
     description: An integer
+    notes:
+    - If you are authoring schemas in LinkML YAML, the type is referenced with the
+      lower case "integer".
     from_schema: https://w3id.org/linkml/types
     imported_from: linkml:types
     exact_mappings:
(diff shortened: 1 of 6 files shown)


cmungall:
I've run into this before, apologies for merging the model too early. Usually I am conservative and follow the pipeline

- make a new linkml-model rc with compiled model copied over
- test linkml core using rc

even here there is potential for mishap in any code that dynamically imports types.yaml rather than the pypi bundled version

sierra-moxon:
wait: I am confused about how the model merge to main over in that GH repo (no release yet in `linkml-model`) has caused errors here (or is this preemptive)?  (I also have not copied over the model to linkml repo, yet either - though this appears to do some of that?)

pkalita-lbl:
If your schema includes:

```yaml
prefixes:

imports:
  - linkml:types
```

It's going to go fetch `https://w3id.org/linkml/types.yaml` which redirects to `https://linkml.io/linkml-model/linkml_model/model/schema/types.yaml`. The `linkml-model` repo [settings](https://github.com/linkml/linkml-model/settings/pages) has GitHub Pages set up to deploy that from the `main` branch (❗). So the end result is that changes to `main` immediately affect imports. 

I could imagine a world where `https://w3id.org/linkml/types.yaml` points to the latest stable _release_. It seems like it would be useful to have old versions available as well (e.g. something like `https://w3id.org/linkml/v1.5.1/types.yaml`) but that would need some technical investigation in terms of both GitHub pages deployment and w3id redirecting.

dalito:
@pkalita-lbl - I have done such multi-version backwards linking in gh-pages incl. w3id.org integration just recently for [voc4cat](https://github.com/nfdi4cat/voc4cat). There we link "dev", "latest" (release) and all tagged releases. You might be interested in the workflow files.

pkalita-lbl:
Nice, I'll take a look at that. @cmungall also pointed me to https://github.com/jimporter/mike which seems useful here.

sierra-moxon (approved):

PR #1591, opened 2023-08-24
convert plantUML generator tests to `pytest`

sujaypatil96:
The tests written for the plantUML generator were originally written in pytest. This PR seeks to convert those tests to pytest.

### DIFF
--- a/tests/test_generators/test_plantuml.py
+++ b/tests/test_generators/test_plantuml.py
@@ -1,17 +1,11 @@
-import logging
 import os
-import shutil
-import tempfile
-import unittest
-from copy import copy
-from typing import List
+
+from xml.dom import minidom
+
+import pytest
 
 from linkml.generators.plantumlgen import PlantumlGenerator
-from tests.test_generators.environment import env
 
-SCHEMA = env.input_path("kitchen_sink.yaml")
-PLANTUML_OUT = env.expected_path("kitchen_sink.plantuml.md")
-SVG_DIR = env.expected_path("kitchen_sink_svg")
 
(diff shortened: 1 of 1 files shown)


tests/test_generators/test_plantuml.py
```diff
@@ -45,114 +37,88 @@ class "Person" {
...
+        ("Person", MARKDOWN_FOOTER),
+        # check that expected plantUML class diagram blocks are present
+        # when diagrams are generated for different classes
+        ("Person", PERSON),
+        ("Dataset", DATASET2PERSON),
+        ("MedicalEvent", PERSON2MEDICALEVENT),
+        ("FamilialRelationship", FAMILIALRELATIONSHIP2PERSON),
+    ],
+)
+def test_serialize_selected(input_class, expected, kitchen_sink_path):
```
pkalita-lbl, line 55: I'm not sure this is the best use of `parametrize`. Especially because this will end up re-generating for the `Person` class a few times. I think the intention of the test might be more clear without parameterizing and doing something like:

```
serialize Person class
assert ...
assert ...
assert ...

serialize Dataset class
assert ...

serialize MeticalEvent class
assert ...
```

pkalita-lbl (approved):
One minor comment if you have time, but otherwise looks good!

tests/test_generators/test_plantuml.py
sujaypatil96: That's a great suggestion @pkalita-lbl 😁 will keep that in mind while using `parametrize`. For the sake of this test though, I actually think it's okay to remove the lines that create the class multiple times.

PR #1609, opened 2023-09-08
can't populate container with sub-classes of range if the sub-classes use slots that are not assigned to the parent

turbomam:
I was under the impression that a LinkML collection (Class `Database` in this case), which might correspond to one MongoDB collection, could use an aggregating slot (`vehicles`) to aggregate all of the subclasses of the named range (`Vehicle`)

This would be very useful for minimizing the number of collections in the NMDC MongoDB, but allowing dDifferent kinds of metabolomics `SampleOperation`s to be aggregated in one collection, or for both `Studies` and `SampleCollectionConsortium`s to be aggregated in a single `ResearchInitiative` collection.

In this test case, LinkML `yaml_loader.loads` doesn't seem to like instantiating `Boat` and an `Airplane` instances, because they each use a slot that is not associated with the parent, `Vehicle`

### DIFF
--- a/tests/test_issues/input/linkml_issue_1608_data.yaml
+++ b/tests/test_issues/input/linkml_issue_1608_data.yaml
@@ -1,0 +1,10 @@
+vehicles:
+  - exterior_color: red
+    max_passengers: 4
+    type: Boat
+    waterline: 30.0
+  - exterior_color: blue
+    max_passengers: 5
+    type: Airplane
+    wingspan: 100.0
+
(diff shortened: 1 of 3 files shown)


turbomam:
Maybe I need to set the range of the `vehicles` slot something like AnyOf(Vehichle, Boat, Airplane)? That just doesn't seem tot take advantage of the hierarchy.

cmungall (approved):

PR #1617, opened 2023-09-12
allow polymorphism for dataclasses using type designators

### LINKED ISSUE #1608: pythongen should use type_designator to determine specific subtype during instantiation
When instantiating a class that has a type_designator, use the type_designator to determine the subclass.

See pydanticgen for the reference implementation

[note: summary provided by @cmungall ]

cmungall:
This PR is a continuation of #1257 

- allow polymorphism for dataclasses using type designators (specification of range of type designators in #945)
- Add a test for python dataclasses polymorphism.
- Adding type_designators to pythongen. Fixes #1608


cmungall:
I am not sure why an unrelated test has started failing in windows...

pkalita-lbl:
The test failures are in fact related to the changes here. These changes introduce `SchemaView` into `PythonGenerator` and the failure happens as `SchemaView` is loading the imports of the schema, kicked off by [this line](https://github.com/linkml/linkml/pull/1617/files#diff-baca17b6d0f9403fdc99778a5fb1eb858b910da0fddd4e77dd9d9ce2d8aff214R721). The failing tests are pre-existing `PythonGenerator` tests.

It is a bit odd that tests for `SchemaView` in `linkml-runtime` or the tests of other generators that use `SchemaView` haven't tickled this error condition before. I haven't fully characterized what's going on enough to suggest a solution, but in the meantime I don't think we should merge this quite yet.

cmungall:
Perhaps the most conservative thing to do is to get the entailed type designators via schemaloader (since pythongen already uses this framework), then we can back out the schemaview dependency.

We should also try and isolate what is happening with windows and imports as a separate issue/PR

pkalita-lbl:
https://github.com/linkml/linkml/issues/1627

cmungall:
On re-reviewing, while it is unusual to mix the two modes, I think on balance it makes sense to use schemaview in this PR. I believe a new runtime release will fix the issue.

pkalita-lbl (approved):

PR #1630, opened 2023-09-15
feat(converters): add option include descendants

### LINKED ISSUE #1629: Add `--include-range-class-descendants` option to "linkml-convert"

`linkml-convert` doesn't accept children of a class specified in a range as valid. Being used to the behavior of inheritance in OOP that also applies to the generated Python code, I find it misleading and frustrating.

IMO (point of view of an OOP programmer, might be different for an "ontologist") accepting children should be the default behavior. Since `gen-json-schema` and `linkml-validate` already have the option `--include-range-class-descendants`, having such an option for `linkml-convert` would be also fine for me.

• Medium - As of now I have to ignore validation errors that appear due to this limitation of `linkml-convert`

• Short-term - 2-4 weeks

**Additional context**
I'm providing a PR to implement this feature.

Silvanoc:
"gen-json-schema" and "linkml-validate" have the option "--include-range-class-descendants" so that children of a class specified in a range are added to the generated JSON-Schema as allowed classes.

This patch adds a similar option to "linkml-convert", with the same result in the JSON-Schema being used for validation as the equally named option for "gen-json-schema" and "linkml-validate".

Closes #1629

### DIFF
--- a/linkml/utils/converter.py
+++ b/linkml/utils/converter.py
@@ -150,19 +150,25 @@ def cli(
                 raise Exception("--index-slot is required for CSV input")
         inargs["index_slot"] = index_slot
         inargs["schema"] = schema
-    obj = loader.load(source=input, target_class=py_target_class, **inargs)
-    if infer:
-        infer_config = inference_utils.Config(use_expressions=True, use_string_serialization=True)
-        infer_all_slot_values(obj, schemaview=sv, config=infer_config)
+    try:
+        data_as_dict = loader.load_as_dict(source=input, **inargs)
+    except NotImplementedError:
+        print("load_as_dict has failed! Loading as object")
+        obj = loader.load(source=input, target_class=py_target_class, **inargs)
+        data_as_dict = as_simple_dict(obj)
     if validate:
         if schema is None:
             raise Exception(
                 "--schema must be passed in order to validate. Suppress with --no-validate"
             )
-        # TODO: use validator framework
-        validation.validate_object(
-            obj, schema, include_range_class_descendants=include_range_class_descendants
-        )
+        validator = JsonSchemaDataValidator(schema, include_range_class_descendants=include_range_class_descendants)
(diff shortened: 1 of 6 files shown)


Silvanoc:
In draft status until tests are provided. Functional changes already available (manually tested) for preview.

cmungall:
Thanks!

I think in general there is less need of user control over this now we have:

- #1613

This keeps the default as false, but always allows class descendants if there is a type designator, which essentially allows the jsonschema validator to "choose" the intended one based on the value of the type designator.

However, it's still good to have consistency in options and how options are passed through.

Further discussion on dynamic typing:
https://github.com/orgs/linkml/discussions/1548

Silvanoc:
@cmungall I've seen PR #1613, but I didn't know how to take advantage of it. And I've seen that applying the same pattern available for other tools with the option `--include-range-class-descendants` is very easy.

Silvanoc:
> @cmungall I've seen PR https://github.com/linkml/linkml/pull/1613, but I didn't know how to take advantage of it. And I've seen that applying the same pattern available for other tools with the option --include-range-class-descendants is very easy.

I think, I've understood it now!

The option `--include-range-class-descendants` is an external flag that applies subtype polymorphism on all class hierarchies. But it also means that the subtype can be only inferred based on [duck typing](https://en.wikipedia.org/wiki/Duck_typing).

The _type designators_ state explicitly of which subtype an object is, so no more duck typing needed. I haven't tested it yet, but I understand that once a _type desingator_ has been used, descendant classes are automatically included for the subtypes.

I like it!

Silvanoc:
I'm having difficulties to get this functionality working. I suspect that issue #1638 is the origin of the difficulties I'm facing.

amc-corey-cox:
@Silvanoc this is another PR that hasn't seen activity in a while. If you would still like to work on this please respond here and I'll keep it open. If I don't see activity in the next week or so, I'm going to close this PR. You are welcome to re-open this PR if you decide to work on it again.

PR #1631, opened 2023-09-15
Fix some typos

yarikoptic:
- spotted fresh typo in https://linkml.io/linkml/contributing/contributing.html `pytset` and decided to fix
- apparently codespell found more , not sure why action (https://github.com/linkml/linkml/actions/runs/6188045401/job/16799203411#step:8:1) didn't pick them up.
   - I did go with fixing for `implementer` since that one also was used already:
```
❯ git grep -i implementer
tests/test_biolink_model/input/biolink-model.yaml:      implementers should use one of the more specific subtypes of this generic property.
tests/test_biolink_model/input/biolink-model.yaml:      implementer can identify.  Performing a rigorous analysis of upstream data providers is expected; every effort
```

### DIFF
--- a/docs/contributing/contributing.md
+++ b/docs/contributing/contributing.md
@@ -26,7 +26,7 @@ There are two cases you need to consider while writing your unit test. The first
 To run a single test file using `pytest`:
 
 ```python
-poetry run pytset tests/test_issues/test_linkml_issue_NNN.py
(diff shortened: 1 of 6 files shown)


cmungall (approved):
Much appreciated!

Not totally sure why codespell didn't spot these earlier but appreciated either way!

PR #1636, opened 2023-09-16
Update json-schema.rst

RichardBruskiewich:
Tiny typo: is this supposed to be a JSON file for validation?


cmungall:
No, see https://linkml.io/linkml/generators/json-schema.html, generators take linkml schemas which are by convention in yaml and compile to some other form

PR #1663, opened 2023-10-06
Add option to JsonschemaValidationPlugin to provide pre-generated JSON Schema

### LINKED ISSUE #1659: Add ability to cache generated artefacts with new linkml validator CLI
The old linkml-validate had the ability to use a previously generated python module (`--module`). This is moot in the new implementation, which validates directly off the dict from the json.

However, the new one still has to recompile the schema to jsonschema each invocation (assuming using the default jsonschema plugin). This is usually not a big deal, but if someone is validating dozens of small data files over a very large schema like mixs, it can take hours.

Example from @turbomam:

```
✗ time poetry run python -m linkml.validator.cli --target-class AllSlotsTestClassCollection --schema generated-schema/mixs_6_2_rc.yaml extracted-data/mixs_6_2_rc.extracted-examples.yaml
No issues found!
poetry run python -m linkml.validator.cli --target-class  --schema    206.08s user 0.67s system 99% cpu 3:27.08 total
```

We can imagine two strategies:

1. the user builds the jsonschema upstream, and passes this as a path
2. the validator will take care of caching

1 is more explicit but makes the user (or cookiecutter-derived makefile) do more work. 2 is more convenient, but less transparent, and needs to have a mechanism to make sure the cached schema is not stale. 2 can be made more transparent by having the user pass a cache directory.

The 2 need not be mutually exclusive.

The overall strategy should be extensible to other implementations, which could involve compiled shacl or shex, sql tables, ...

pkalita-lbl:
This is a bare-bones solution to fix #1659. From a user perspective, if they're using the CLI they would need a config file that looks something like:

```yaml
plugins:
  JsonschemaValidationPlugin:
    json_schema_path: some-cached-schema.json
```

### DIFF
--- a/linkml/validator/plugins/jsonschema_validation_plugin.py
+++ b/linkml/validator/plugins/jsonschema_validation_plugin.py
@@ -1,4 +1,5 @@
-from typing import Any, Iterator
+import os
+from typing import Any, Iterator, Optional
 
 import jsonschema
 from jsonschema.exceptions import best_match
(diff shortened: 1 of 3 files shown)


cmungall (approved):
perfect, the general pattern is easily extended to other validators that require generated products

And every unittest->pytest change makes me happy!

Issue #1682, opened 2023-10-30
URI link on the page for an enum returns a 404


AlasdairGray:
**Describe the bug**
The generated URI for an enumeration is not a resolvable URI.

**To reproduce**
1. Go to 'https://linkml.io/linkml-model/1.6.x/docs/PvFormulaOptions/'
2. Click on '[PvFormulaOptions](https://linkml.io/linkml-model/1.6.x/docs/PvFormulaOptions/PvFormulaOptions)'
3. See error 404 not found as you have been taken to `https://linkml.io/linkml-model/1.6.x/docs/PvFormulaOptions/PvFormulaOptions`

**Expected behavior**
For enums that have a declared `enum_uri` property, I would expect to be taken to the corresponding page. If no such value is given I would expect that the link is not shown.

**Screenshots**
<img width="1400" alt="Screenshot 2023-10-30 at 16 45 38" src="https://github.com/linkml/linkml/assets/1141265/9bdee8cc-bee1-40bb-8318-ee0fe053a908">

**Additional context**
Bug found and reported at https://github.com/co-cddo/ukgov-metadata-exchange-model/issues/61 but discovered that it is an issue with LinkML rather than our definition.

cmungall:
@sujaypatil96 looks like you fixed this in the MIxS templates, can you also make the same fix for the generic ones? https://github.com/GenomicsStandardsConsortium/mixs/pull/688

AlasdairGray:
I've incorporated this fix into our template which was also causing the same problem. However, if an enum is linked to an existing vocabulary using the `enum_uri` property then this is not taken into account. The corresponding case of a class would link out to the external vocabulary class definition.

turbomam:
@sujaypatil96 has also been working on this.

sujaypatil96:
The difference between say, the class template (jinja template that controls the layout and content of class documentation pages) and the enum template in the way that they display URIs is:
* class URI: https://github.com/linkml/linkml/blob/main/linkml/generators/docgen/class.md.jinja2#L29 (uses DocGenerator's uri_link() method)
* enum URI: https://github.com/linkml/linkml/blob/main/linkml/generators/docgen/enum.md.jinja2#L10C29-L10C29 (uses DocGenerator's link() method)

[...]

PR #1686, opened 2023-10-31
Fix broken enum and subset URI links

sujaypatil96:
This PR is a fix for issue #1682 

The docgen method responsible for rendering resolvable uri links is [uri_link()](https://github.com/linkml/linkml/blob/main/linkml/generators/docgen.py#L346-L358) which depends on [uri()](https://github.com/linkml/linkml/blob/main/linkml/generators/docgen.py#L334-L344) which has special handling for enums and subsets because the underlying SchemaView method [get_uri()](https://github.com/linkml/linkml-runtime/blob/main/linkml_runtime/utils/schemaview.py#L904-L941) does not know how to make URIs for enums/subsets.

Ultimately, `uri_link()` should be the one method which handles the resolving/rendering of element URIs but that will require fixes in linkml-model/linkml-runtime, so until then, this can be a temporary fix for issue #1682

### DIFF
--- a/linkml/generators/docgen/enum.md.jinja2
+++ b/linkml/generators/docgen/enum.md.jinja2
@@ -7,7 +7,7 @@ _{{ element_description_line }}_
 {% endfor %}
 {% endif %}
 
-URI: {{ gen.uri_link(element) }}
+URI: {{ gen.link(element) }}
 
 {% if element.permissible_values -%}
 ## Permissible Values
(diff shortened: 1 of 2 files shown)


sujaypatil96:
There is a related issue on NMDC schema: https://github.com/microbiomedata/nmdc-schema/issues/870

sujaypatil96:
Apologies, this came PR from a fork. Will ensure future PRs come from main repo branches.

cmungall (approved):

PR #1700, opened 2023-11-06
Adds support for using typeof to generate the typescript type if type.base is not set

kevinschaper:
This fix addresses https://github.com/monarch-initiative/monarch-app/issues/449, where custom types brought in from the oak similarity schema didn't get the correct typescript type.


pkalita-lbl (approved):
This is good. Maybe in the future it should recursively walk up the `typeof` chain until it hits a base type.

kevinschaper:
> This is good. Maybe in the future it should recursively walk up the `typeof` chain until it hits a base type.

That would be great, maybe in a SchemaView method that can hide that detail.

PR #1704, opened 2023-11-09
Add support for Python 3.12

Silvanoc:
Add test to validate that "greenlet" is on a version that makes the project runnable on Python v3.12.

Update the Poetry locking file to install such a version. Should someone try to downgrade "greenlet" to a version that cannot run on Python 3.12, the test will fail.


cmungall:
It looks like older versions of prefixmaps over-pinned the version of greenlet:

 - https://github.com/linkml/prefixmaps/blob/v0.1.5/pyproject.toml#L27

So perhaps bumping the minimum prefixmaps might get around the issues here?

In general prefixmaps should be light and unopinionated, see also:
https://github.com/linkml/prefixmaps/pull/54

Silvanoc:
> So perhaps bumping the minimum prefixmaps might get around the issues here?

That's what I originally thought and why I started working on it. Unfortunately it is much more complex than that. Please have a look at this PR: https://github.com/linkml/prefixmaps/pull/53.

Initially it looked easy, until I've realized that CI caching was resulting in the object under test (package dependency resolution) not being tested. Once [I've introduced python version specific caches](https://github.com/linkml/prefixmaps/pull/53/commits/7f02b39de870bbb4e66d89dce2738ebf30740e2a), further issues has appeared. Unfortunately right now I don't have time to work on it.

Silvanoc:
Thanks to @cmungall PR linkml/prefixmaps#54 I've been able to resolve the issues I had on PR linkml/prefixmaps#53. Merging my PR, releasing a new prefixmaps version with the fix and upgrading prefixmaps to that version here might help easily fix it here too.

cmungall:
looks like still getting that pesky greenlet error...

Silvanoc:
> looks like still getting that pesky greenlet error...

I haven't been following this topic for some time. I'll give make sure to rebase all the PRs I've done WRT Python 3.12 and then update the status accordingly on the comments.

Silvanoc:
Having support for Python 3.12 in this project doesn't bring anything until `linkml-runtime` also supports it. I've had some issues to get it fixed there, but it now looks as if a working [PR](https://github.com/linkml/linkml-runtime/pull/286) would be ready. I'll try to prepare here also the required changes (for example, removing support for Python <3.9) for the moment when `linkml-runtime` supports it.

Silvanoc:
I'm having difficulties with the installation of `numpy` on Python 3.12 due to an incompatibility of a higher numpy version with Python 3.8. I suposse that the use of [Poetry dependency extras](https://python-poetry.org/docs/dependency-specification/#dependency-extras) would help resolving the dilemma, but I lack the expertise to get it done. @pkalita-lbl you seem to be quite fit in Python, could you please have a look at it? Thanks!

dalito:
The solution (I think) is just a little further down under [Multiple constraints dependencies](https://python-poetry.org/docs/dependency-specification/#multiple-constraints-dependencies).

Silvanoc:
> The solution (I think) is just a little further down under [Multiple constraints dependencies](https://python-poetry.org/docs/dependency-specification/#multiple-constraints-dependencies).

I thought the same right after my post, that's what I'm trying right now :-) Thanks for the hint anyway.

pyproject.toml
```diff
@@ -142,7 +143,10 @@ myst-parser = "*"
 [tool.poetry.group.tests.dependencies]
 pytest = "^7.4.0"
 pytest-subtests = "^0.11.0"
-numpy = ">=1.24.3"
+numpy = [
+  { "version" = ">=1.24.3", "python" = "<3.12" },
```
dalito, line 147: Shouldn't this be `{ "version" = "~1.24.3", "python" = "<3.12" },`? (to express < 1.25.0)

Silvanoc: Poetry takes care of selecting a version of `numpy` that is compatible with the rest of the dependencies. Therefore [for Python 3.8, it updates only up to `1.24.4`](https://github.com/linkml/linkml/blob/add-support-for-python3.12/poetry.lock#L1596) and not further up.
The tests will tell us, if I'm right or not 😄

Silvanoc: Confirmed, resolving this conversation.

Still stumbling upon a different issue...

Silvanoc:
On Python 3.12 following [test is failing](https://github.com/linkml/linkml/actions/runs/7944987517/job/21691084162?pr=1704): [test_data_types](https://github.com/linkml/linkml/blob/7ea1a33226bb6480c7e0bea5b1c5ad11347dcceb/tests/test_compliance/test_core_compliance.py#L344) with input [("date", "20210101", False)](https://github.com/linkml/linkml/blob/7ea1a33226bb6480c7e0bea5b1c5ad11347dcceb/tests/test_compliance/test_core_compliance.py#L328C9-L328C36).

But I fail to understand why the test has started failing on Python 3.12 and not on previous Python versions. Python 3.12 is removing lots of long deprecated functions, what could be the source of the failure.

dalito:
Since Python 3.11, `date.frromisoformat(...)` also accepts `YYYYMMDD` format, see https://docs.python.org/3/library/datetime.html#datetime.date.fromisoformat - so testing with a date in such format is no longer giving the expected failure ([link to test](https://github.com/linkml/linkml/blob/7ea1a33226bb6480c7e0bea5b1c5ad11347dcceb/tests/test_compliance/test_core_compliance.py#L328)).

Silvanoc:
It then means that the [failing test](https://github.com/linkml/linkml/actions/runs/7944987517/job/21691084162?pr=1704): [test_data_types](https://github.com/linkml/linkml/blob/7ea1a33226bb6480c7e0bea5b1c5ad11347dcceb/tests/test_compliance/test_core_compliance.py#L344) should accept `20210101` as a valid date and therefore get input `("date", "20210101", True)`... Let's try it out.

Silvanoc:
@dalito I don't think that this test should be testing which formats are accepted by Python, but rather that only those dates that are valid [`xsd:date`](https://books.xmlschemata.org/relaxng/ch19-77041.html) are accepted. And my proposed solution would be doing exactly the opposite: making sure that `CCYYMMDD` is accepted, though it's not a valid `xsd:date`.

It means that LinkML cannot rely on python `datetime` library to reject `CCYYMMDD` dates and will require some own business logic for date validation.

Silvanoc:
Now I realize that the issue is rather [in `linkml-runtime`](https://github.com/linkml/linkml-runtime/blob/c9adcaeb405804efcf19bd4fcfc950a8ddb173e1/linkml_runtime/utils/metamodelcore.py#L265) and should be fixed there.

dalito:
Yes I also had a look YYYYMMDD is never valid in jsonschema. So the test case cannot be generally changed just because py 3.11 is accepting another valid iso-format notation which is however not a [valid internet time format](https://www.rfc-editor.org/rfc/rfc3339.html#section-5.6).

Silvanoc:
The fix needs to be done in `linkml-runtime` and additional tests would be needed to avoid a regression. Pretty much the test that is failing here, but transfered to `linkml-runtime`.

Silvanoc:
I've added a regression test for the issue to https://github.com/linkml/linkml-runtime/pull/286 (which first fails before adding a fix) and then the needed change to the business logic. Once I get it merged into `linkml-runtime`and a new release is available, we should be able to move forward on this PR. Until then, I'm setting the draft status again.

sneakers-the-rat:
j2lyk `typing_extensions` is only needed for `Literal` in `python<3.8`, 3.8 has `Literal`, so if we are going `>=3.8` then we should drop the dep and remove it from eg. pydanticgen, and well everywhere it's used!

cmungall:
> Once I get it merged into linkml-runtimeand a new release is available, we should be able to move forward on this PR. Until then, this PR will be on-hold.

https://github.com/linkml/linkml-runtime/releases/tag/v1.7.2 is out, thanks for your help on this!

Silvanoc:
The branch of this PR has started failing right after upgrading `linkml-runtime` to version 1.7.2. I was wondering if my contribution to that release could have triggered the issue.

But the [failing test](https://github.com/linkml/linkml/actions/runs/8008852632/job/21879208746?pr=1704#step:11:544) is about imports and I haven't changed anything on imports...

Then I've realized that that release also got [a contribution from](https://github.com/linkml/linkml-runtime/pull/302) @sneakers-the-rat that apparently do have something to do with imports... Could it be the source of the issue?

sneakers-the-rat:
oop just seeing this lemme do a patch if this is from me

Silvanoc:
The tests claim that everything is set for merging. I would nevertheless recommend a careful review.

@cmungall I added Python 3.11 to the tests because the changes are mostly concerned about the support on the different Python versions and I wanted to be 100% that I'm not breaking support for any of the versions. Please feel free to remove it before merging, if desired.

Thanks to @dalito and @sneakers-the-rat on catching and fixing a couple of issues!

cmungall:
The only thing I am concerned about is the dependency on pydanticgen.

Not necessarily against this (we have previously discussed eliminating support for pydantic1 as part of *generation* but this is different). But I'm more comfortable if this can be deferred. I assume it was added there due to some combo of pydantic1 and python 3.12 not working? Maybe we can matrix this?

sneakers-the-rat:
Aha I missed that. Related to https://github.com/linkml/linkml/issues/1925

I think a bit of warning would be nice - its a bit different than usual with a runtime dependency, people will presumably still have their existing pydantic 1 models they generated, and ofc they can go back to old versions of linkml if they want more, but still I think deserves a warning.

I think we can do python version-specific pydantic deps, and I can write the warnings and handle the 2x2 python x pydantic compatibility grid in pydanticgen which should probably happen as part of this PR. Docs will need updating too.

In the warning we probably want to give some timeline for deprecation, whether that be in months or version numbers, and I think more generally it would be good to have guidelines for deprecation across the package, but thats another convo

Big in favor of deprecating pydantic 1. It doubles the complexity of any change to pydanticgen, is way slower and doesnt have a lot of nice pydantic 2 things. Its been out since last June. I had proposed 6 months in that other issue but I think we could give as little as 1 month.

dalito (approved):
Looks good to me. Open pydantic 1 related changes could be addressed separately I think.

sneakers-the-rat:
so just tbc, we're dropping pydantic 1 support here?

dalito:
@sneakers-the-rat Ah now I see, in dependencies pydantic 1 is dropped but not in test matrix. Sorry for the confusion.

sneakers-the-rat:
Right, which is why im confused! If we drop it in dependencies then people wont be able to generate them in environments that have linkml installed without overriding dep checker, so we would be dropping support for pydantic one with this PR, right? Works fine for me, just wanted to be clear bc im working on some stuff related to it. Also willing to help adjust the docs

cmungall (changes requested):
Dropping pydantic1 support needs to be justified

sneakers-the-rat:
I just ran tests for pydantic 1.10.14 in python 3.12 and most of the errors were from simple string mismatches where the `abc` module has changed the error message "with abstract methods ..." to "without an implementation for abstract methods ..."

the one substantive error was from generics, which I don't think we use in our generated models.

So Im not sure why we need to change the pydantic dependency here? I assumed (like in the python 3.12 issue in linkml-runtime) it was because there was some known incompatibility, but there doesn't seem to be? 

Simplest thing to do here would be to just leave pydantic dependency as-is and start a deprecation warning period in another PR, which i am happy to do, but would just like a lil guidance on how that's handled normally ( https://github.com/linkml/linkml/issues/1925 )

Silvanoc:
> Dropping pydantic1 support needs to be justified

@cmungall I've restored support for pydantic1 (55e51f6), since I dropped it due to a wrong assumption from my side.

Silvanoc:
Apparently there's still something missing to get it working with Pydantic v1 on Python 3.8. But I don't have currently any time to debug the issue...

sneakers-the-rat:
there ya go, just needed to merge in main. should be good to go now

sneakers-the-rat:
so we have the links on hand:

the parents of the commit are:
- https://github.com/linkml/linkml/commit/f148a1882df1fd162f24b6dc8b192491116dfa16 - the previous tip of [`add-support-for-python3.12`](https://github.com/linkml/linkml/commits/add-support-for-python3.12/)
- https://github.com/linkml/linkml/commit/ec35c43c8af1fed63b00a71dfd520f26219e468b - the current tip of [`main`](https://github.com/linkml/linkml/commits/main/) 

diffs:
- the current diff: https://github.com/linkml/linkml/compare/ad71735..ec35c43
- the diff between the previous commit in this PR and the current tip of main: https://github.com/linkml/linkml/compare/ec35c43..f148a18
- the diff between the current commit in this PR and the previous one https://github.com/linkml/linkml/compare/f148a18..ad71735

what are you seeing that should be here and is missing? sorry if i messed up the merge, i literally just merged from main after pulling from both

Silvanoc:
@sneakers-the-rat I never merge to synchronize, but rebase. Therefore I was puzzled by the amount of changes. If it's simply a sync with current main and it would be mergeable, it would be fine for me.

But trying to fix the conflicts mentioned in the PR is getting pretty tedious. I wonder if it's the double merging that is causing it. Or if it's me that I'm not used to sync-merge and then merge back...

linkml/generators/pydanticgen/pydanticgen.py
```diff
@@ -187,10 +187,7 @@ class PydanticGenerator(OOCodeGenerator):
             Path,
             PurePath as RenamedPurePath
         )
-        if sys.version_info >= (3, 8):
-            from typing import Literal
-        else:
-            from typing_extensions import Literal
```
sneakers-the-rat, line 193: oh my bad for missing this before! This is a comment demonstrating the usage and results of this object, not the template, so it's correct! 

I pre-emptively removed typing_extensions from the [`DEFAULT_IMPORTS`](https://github.com/linkml/linkml/blob/581bc8091bc87630d72a0930de12b8bad4d153d3/linkml/generators/pydanticgen/pydanticgen.py#L63) assuming this would be merged shortly after :) i'll fix rn

sneakers-the-rat:
> @sneakers-the-rat I never merge to synchronize, but rebased. Therefore I was puzzled by the amount of changes. If it's simply a sync with current main and it would be mergeable, it would be fine for me.

aha gotcha! ok glad that's cleared up, i got scared for a minute i had broken the whole thing <3

> But trying to fix the conflicts mentioned in the PR is getting pretty tedious. I wonder if it's the double merging that is causing it. 

just the same places in the code getting modified in main and here! but ya totally agreed, let's get this thing merged. dragged on for long enough that the changes have been whittled away in other PRs except for actually lifting the version cap!

just merged main again, so should b good to go.

sneakers-the-rat:
ping on this, let's roll this out!

Silvanoc:
> ping on this, let's roll this out!

I don't know whom you are trying to appeal. I'm not a maintainer and therefore I cannot merge. Even if I could, I wouldn't since the PR is blocked by a change request by @cmungall and he should give his green light. I have requested him to review it again, but I suppose that it's just one among multiple pending notifications...

sneakers-the-rat:
i am pinging whoever can merge :) i was just working on a PR for `linkml-runtime` that tests against upstream `linkml` and wanted to include 3.12 in that, but can't since this hasn't been merged yet. so just friendly reminder in case it had fallen off whoever's radar is capable of merging this, no impatience implied.

dalito (approved):

dalito:
Chris has to merge. I can't.

cmungall (approved):

PR #1705, opened 2023-11-09
Show headers for class, slot documentation pages similar to MIxS

sujaypatil96:
Examples for the way in which we're making the headers for class and slot documentation pages for the MIxS docs is like below:
* https://genomicsstandardsconsortium.github.io/mixs/MigsBa/ (class doc page)
* https://genomicsstandardsconsortium.github.io/mixs/0000122/ (slot doc page)


cmungall (approved):
why not to this in python?

sujaypatil96:
Yup, that's true. We can do this in Python.

It made sense to do this in the jinja for mixs because we didn't have "access" to docgen.py there, but we can do it in docgen.py here. I'll make the necessary modifications and update this PR.

PR #1709, opened 2023-11-11
[gen-doc] Allow to rename the index document.

### LINKED ISSUE #1708: `gen-doc` should allow renaming the index page
The `gen-doc` tool automatically generates an index page for the documentation it generates. That page is always named `index.html`. This is annoying as most web servers are configured by default to return `index.html` to a client that asks for a directory index.

This means, for example, that when someone visits the [website for SSSOM](https://mapping-commons.github.io/sssom/) (which is partially generated by LinkML), the page the visitor lands on first is the LinkML index, which is not very informative – the visitor then needs to manually go to the [“home” page](https://mapping-commons.github.io/sssom/home/) to learn what SSSOM actually is.

I would like a way to change the default behaviour of `gen-doc` so that the auto-generated index page may be named differently. For example, a command-line option as follows:

```sh
gen-doc --index-name schema-index ...
```

which would lead `gen-doc` to write the index in a file named `schema-index.html` instead of `index.html`.

There are at least two possible workarounds:

* changing the configuration of the web server to use a different filename for the directory index (not always possible depending on where and how the website is hosted);
* forcibly rename the generated `index.html` to something else and replace it with the intended home page after the `gen-doc` run (would work everywhere but definitely not ideal).

• Short-term - 2-4 weeks

gouttegd:
This PR adds a `--index-name` option to the documentation generator (`gen-doc`) to allow users to specify a custom name for the index document, instead of the default `index`.

closes #1708

### DIFF
--- a/linkml/generators/docgen.py
+++ b/linkml/generators/docgen.py
@@ -130,6 +130,9 @@ class DocGenerator(Generator):
     directory: str = None
     """directory in which to write documents"""
 
+    index_name: str = "index"
+    """name of the index document"""
+
     template_directory: str = None
     """directory for custom templates"""
 
@@ -188,7 +191,7 @@ class DocGenerator(Generator):
         }
         template = self._get_template("index")
         out_str = template.render(gen=self, schema=sv.schema, schemaview=sv, **template_vars)
-        self._write(out_str, directory, "index")  # TODO: make configurable
+        self._write(out_str, directory, self.index_name)
         if self._is_single_file_format(self.format):
             logging.info(f"{self.format} is a single-page format, skipping non-index elements")
             return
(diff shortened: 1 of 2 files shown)


sujaypatil96:
This is a good solution to allow users to specify the name of the index markdown file that is generated alongside the LinkML schema element markdown documentation pages, i.e., allow the option of calling it something other than `index.md` which is currently the default (and only option).

However, it is also important to keep in mind, the [linkml-project-cookiecutter](https://github.com/linkml/linkml-project-cookiecutter) repo which is what the LinkML community recommends you start off with while creating a new LinkML schema. The cookiecutter creates a project directory structure for you which includes an `mkdocs.yml` file, and the index/landing page is "hardcoded" in the mkdocs.yml file on this line: https://github.com/linkml/linkml-project-cookiecutter/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/mkdocs.yml#L15

At the very least we should make a _note_ on our documentation page for the markdown generator: https://linkml.io/linkml/generators/markdown.html that users will need to change the name of their index file in mkdocs.yml if they've created their schema repos using thecookiecutter.

sujaypatil96 (approved):

cmungall:
Thanks @sujaypatil96! I made an issue on cookiecutter

PR #1732, opened 2023-11-15
Add option to `gen-excel` to support splitting up workbook on a per class basis

### LINKED ISSUE #1706: Add option to allow generated Excel workbook to be split up based on worksheets
Currently, the Excel spreadsheet generator in the LinkML framework outputs one large monolithic Excel file with multiple worksheets, which is basically a spreadsheet representation of your LinkML schema. In this issue, we are requesting for a feature to be added to the excel generator so it can handle/support the ability to split up this one large monolithic Excel file into multiple Excel files, one file per class in the schema.

Add an optional argument (boolean flag) called `--split-worksheets` or `--split-workbook` which will facilitate the above behavior when enabled/set to True.

• Short-term - 2-4 weeks

**Additional context**
This issue request came out of a use case in the MIxS schema where we want to ship Excel artifacts for consumers, and given the number of classes in the schema, we don't want to ship out on big Excel file with a lot of worksheets that users/consumers have to scroll through.

sujaypatil96:
Fixes #1706 

This PR adds an option called `--split-worksheets` to the Excel generator (`gen-excel`), which on enabling/turning on splits up the one monolithic Excel file (workbook) which is usually created by default (with multiple worksheets) into multiple Excel files, one for each class.

### DIFF
--- a/linkml/generators/excelgen.py
+++ b/linkml/generators/excelgen.py
@@ -1,6 +1,6 @@
 import logging
 import os
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import List
 
 import click
@@ -8,11 +8,9 @@ from linkml_runtime.utils.schemaview import SchemaView
 from openpyxl import Workbook
 from openpyxl.utils import get_column_letter
 from openpyxl.worksheet.datavalidation import DataValidation
-from openpyxl.worksheet.worksheet import Worksheet
 
 from linkml._version import __version__
 from linkml.utils.generator import Generator, shared_arguments
-from linkml.utils.helpers import convert_to_snake_case
 
 
 @dataclass
@@ -23,6 +21,8 @@ class ExcelGenerator(Generator):
     valid_formats = ["xlsx"]
     uses_schemaloader = False
     requires_metamodel = False
+
+    split_worksheets: bool = field(default_factory=lambda: False)
 
     def __post_init__(self) -> None:
         super().__post_init__()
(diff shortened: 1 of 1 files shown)


dalito:
I find the name `split-worksheet` a bit misleading, because it does not split a worksheet, but splits/spreads the output to sheets. 
Maybe `ws-by-class` is clearer?

sujaypatil96:
Thank you for the feedback @dalito! I was contemplating the option name too, and like you said, it is misleading/confusion because it does not split worksheets, but in fact splits up one monolithic workbook into multiple worksheets.

`ws-by-class` is good but again, `ws-` is not a standardized suffix so it might be hard to understand what the expansion for `ws` is exactly. Perhaps something like `--split-workbook-by-class` might work better.

dalito:
I agree, "ws" may be too cryptic. I like `--split-workbook-by-class` as well or maybe `--one-worksheet-per-class`.

sujaypatil96:
I've made the change to the option name @dalito! it's now `--split-workbook-by-class`.

sujaypatil96:
I've tagged a number of reviewers on this PR. Please feel free to review/add comments/provide feedback on as much of it as you can.

linkml/generators/excelgen.py
```diff
@@ -37,49 +37,41 @@ def create_workbook(self, workbook_name: str) -> Workbook:
...
+                slots = [s.name for s in sv.class_induced_slots(cls_name, self.mergeimports)]
+                self.add_columns_to_worksheet(workbook, cls_name, slots)
+                workbook.save(output_path)
 
-    def create_schema_worksheets(self, workbook: str) -> None:
-        """
-        Creates worksheets in a given Excel workbook based on the classes in the
-        schema.
+                # Add enum validation for columns with enum types
+                enum_list = [e_name for e_name, _ in sv.all_enums(imports=self.mergeimports).items()]
```
dalito, line 65: You can simplify to `[e_name for e_name in sv.all_enums(imports=self.mergeimports)]` here and once again 3 lines below.

linkml/generators/excelgen.py
```diff
@@ -37,49 +37,41 @@ def create_workbook(self, workbook_name: str) -> Workbook:
...
 
-        :param workbook: The workbook to which the worksheet should be added.
-        """
-        sv = self.schemaview
-        for cls_name, cls in sv.all_classes(imports=self.mergeimports).items():
-            if not cls.mixin and not cls.abstract:
-                self.create_worksheet(workbook, cls_name)
+        workbook.save(output_path)
+        if self.split_workbook_by_class:
+            self.logger.info(f"The Excel workbook has been written to {output_path}")
```
dalito, line 74: Should this be plural? "The split Excel workbooks have been..."

linkml/generators/excelgen.py
```diff
@@ -129,48 +117,49 @@ def column_enum_validation(
...
-        self.create_schema_worksheets(workbook)
+        if self.split_workbook_by_class:
+            self.output = os.path.abspath(self.schema.name + "_worksheets") if not self.output else self.output
 
-        sv = self.schemaview
-        for cls_name, cls in sv.all_classes(imports=self.mergeimports).items():
-            if not cls.mixin and not cls.abstract:
-                slots = [s.name for s in sv.class_induced_slots(cls_name, imports=self.mergeimports)]
-                self.add_columns_to_worksheet(workbook, cls_name, slots)
+            os.makedirs(self.output, exist_ok=True)
```
dalito, line 131: Do you use still os or has linkML switched to pathlib?

linkml/generators/excelgen.py
```diff
@@ -129,48 +117,49 @@ def column_enum_validation(
...
+    is_flag=True,
+    default=False,
+    help="""Split model into separate Excel workbooks/files, one for each class""",
+)
 @click.option(
     "-o",
     "--output",
     type=click.Path(),
-    help="""Name of Excel spreadsheet to be created""",
+    help="""Name of Excel spreadsheet to be created, or name of directory to create split worksheets in""",
```
dalito, line 157: Should be "split workbooks" here too.

linkml/generators/excelgen.py
```diff
@@ -37,49 +37,41 @@ def create_workbook(self, workbook_name: str) -> Workbook:
         :return: An openpyxl Workbook object representing the newly created workbook.
         """
         workbook = Workbook()
-        workbook.title = workbook_name
+        workbook.save(workbook_name)
         return workbook
 
-    def get_workbook_name(self, workbook: Workbook) -> str:
+    def create_workbook_and_worksheets(self, output_path: str, classes: List[str]) -> None:
```
dalito, line 43: You have type `str` for a path her. If you use pathlib (of which I am not sure) it would be better to use `path` type.

linkml/generators/excelgen.py
cmungall: agreed

linkml/generators/excelgen.py
cmungall: I think the code is still a mix but I prefer using pathlib for cases like this.

Also: good idea to do parents=True too

cmungall (changes requested):
can you make the changes suggested by @dalito once you are settled Sujay, thx!

sujaypatil96:
From discussion with Mark on 12/7: do not generate non-functional pull downs and warn users via logging that they're going to be excluded.

cmungall (approved):

PR #1779, opened 2023-12-12
Pydantic gen pattern lists

plbremer:
continues https://github.com/linkml/linkml/pull/1772

This PR corrects the previous attempt at regex support for both pydantic v1 and v2. It does this by adding validators to the jinja templating. The previous approach attempted to use only pattern key:values in the Fields, which causes pydantic to expect non-iterables.

"deep learnings":  the test "test_pattern_compliance.py" generated code for each tuple in
```
[
        ("complete_match", r"^\S+$", "no_ws", "ab"),
        ("complete_match", r"^\S+$", "ws", "a b"),
        ("partial_match", r"ab", "partial_ab", "abcd"),
        ("partial_match", r"ab", "complete_ab", "ab"),
        ("partial_match", r"ab", "ws", "a b"),
],
```
upon testing, this caused a duplicate validator error

adding allow_reuse=True to the decorator allowed for this to happen. Ultimately, we were able to remove PYDANTIC from the list of tested frameworks, as indicated here

https://obo-communitygroup.slack.com/archives/C04EU7JL1NF/p1701911088218449?thread_ts=1701887197.532469&cid=C04EU7JL1NF

### DIFF
--- a/linkml/generators/pydanticgen.py
+++ b/linkml/generators/pydanticgen.py
@@ -145,7 +145,7 @@ class {{ c.name }}
     {% endfor %}
     {% for attr in c.attributes.values() if c.attributes -%}
     {%- if attr.pattern %}
-    @validator('{{attr.name}}')
+    @validator('{{attr.name}}', allow_reuse=True)
     def pattern_{{attr.name}}(cls, v):
         pattern=re.compile(r"{{attr.pattern}}")
         if isinstance(v,list):
--- a/linkml/generators/pydanticgen.py
+++ b/linkml/generators/pydanticgen.py
@@ -49,8 +49,8 @@ from typing import List, Dict, Optional, Any, Union"""
 from pydantic import BaseModel as BaseModel, Field, validator"""
     else:
         template += """
-from pydantic import BaseModel as BaseModel, ConfigDict, field_validator, Field"""
-
+from pydantic import BaseModel as BaseModel, ConfigDict,  Field
+from pydantic.functional_validators import field_validator"""
     template += """
 import re
 import sys
--- a/linkml/generators/pydanticgen.py
+++ b/linkml/generators/pydanticgen.py
@@ -44,10 +44,10 @@ from __future__ import annotations
 from datetime import datetime, date
 from enum import Enum
 from typing import List, Dict, Optional, Any, Union"""
-    if pydantic_ver == '1':
+    if pydantic_ver == "1":
         template += """
 from pydantic import BaseModel as BaseModel, Field, validator"""
-    elif pydantic_ver == '2':
+    elif pydantic_ver == "2":
         template += """
 from pydantic import BaseModel as BaseModel, ConfigDict,  Field, field_validator"""
     template += """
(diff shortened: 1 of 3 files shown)


plbremer:
@pkalita-lbl tag

linkml/generators/pydanticgen.py
```diff
@@ -144,6 +143,24 @@ class {{ c.name }}
     {% else -%}
     None
     {% endfor %}
+    {% for attr in c.attributes.values() if c.attributes -%}
+    {%- if attr.pattern %}
+    @validator('{{attr.name}}', allow_reuse=True)
+    def pattern_{{attr.name}}(cls, v):
+        pattern=re.compile(r"{{attr.pattern}}")
+        if isinstance(v,list):
+            # [raise ValueError(f"Invalid name format: {element}") for element in v if not pattern.match(element)]
```
pkalita-lbl, line 152: Is there any need for this commented-out code to be emitted by the generator?

linkml/generators/pydanticgen.py
```diff
@@ -182,6 +198,24 @@ class {{ c.name }}
     {% else -%}
     None
     {% endfor %}
+    {% for attr in c.attributes.values() if c.attributes -%}
+    {%- if attr.pattern %}
+    @field_validator('{{attr.name}}')
+    def pattern_{{attr.name}}(cls, v):
+        pattern=re.compile(r"{{attr.pattern}}")
+        if isinstance(v,list):
+            # [raise ValueError(f"Invalid name format: {element}") for element in v if not pattern.match(element)]
```
pkalita-lbl, line 207: Ditto above comment

linkml/generators/pydanticgen.py
```diff
@@ -144,6 +143,24 @@ class {{ c.name }}
...
+            for element in v:
+                if not pattern.match(element):
+                    raise ValueError(f"Invalid {{attr.name}} format: {element}")
+        elif isinstance(v,str):
+            if not pattern.match(v):
+                raise ValueError(f"Invalid {{attr.name}} format: {v}")
+        return v
+    {% endif -%}
+    {% else -%}
+    None
```
pkalita-lbl, line 162: This looks a little funky to me. I don't think this `else` is enclosed in any `if`/`endif`. I guess Jinja is forgiving enough to just drop this, but it might be better to remove it for good hygiene.

linkml/generators/pydanticgen.py
```diff
@@ -182,6 +198,24 @@ class {{ c.name }}
...
+    {% endfor %}
```
pkalita-lbl, line 218: Ditto above comment

pkalita-lbl (commented):
This looks great! It's awesome that you got this adhering to the compliance test. Just a couple of minor hygiene comments and then I think this will be good to go!

linkml/generators/pydanticgen.py
plbremer: error

linkml/generators/pydanticgen.py
plbremer: agreed

pkalita-lbl (approved):

PR #1783, opened 2023-12-12
Convert tests in tests/test_scripts to pytest 

### LINKED ISSUE #1723: Convert tests in `tests/test_scripts` to pytest
* Ensure all tests are written in `pytest` style (top-level test functions instead of `unittest.TestCase` subclasses, plain `assert` statements)
* Ensure tests only write to temporary files, if needed (by default running tests should not change checked-in files)
* Remove any `print` statements, use logging judiciously.

pkalita-lbl:
Apologies in advance for the horrific diff here!

### Summary

* The `tests/test_scripts/test_*.py` files are now pytest-based tests that use a "standard" combination of: click's `CliRunner` class, pytest parametrization, our own existing pytest fixtures like `input_path` and `snapshot`. Previously these tests were based on a custom `unittest.Testcase` subclass called `ClickTestCase` which is no longer used anywhere and has been removed.
* Moved one test function into `tests/test_generators/test_pythongen.py` because it wasn't testing the `gen-python` CLI at all.
* The file `tests/test_scripts/testnm_gen_graphviz.py` was never being run because of its name, and it tests a deprecated generator. I decided to just remove it since it wasn't really doing anything for us.
* A number of files shuffled between the old (and removed) `test/test_scripts/output` directory and the new `test/test_scripts/__snapshots__` directory.

### Thoughts for the future

All of these tests have at least one test that runs the metamodel through the CLI under test. A number of them run it through multiple times with different options (e.g. once for each output format the CLI supports). I can understand the rationale, but it makes these tests _slooowwww_. If we're looking for places to speed up our test suite this would be a great place to start. Perhaps limiting each test to one metamodel test function and others can operate on the kitchen sink model?

Fixes #1723


tests/test_scripts/test_gen_json_schema.py
```diff
@@ -1,67 +1,62 @@
...
+        ("", "meta.json"),
+        ("-f json", "meta.json"),
+        ("-i", "meta_inline.json"),
+    ],
+)
+def test_metamodel_valid_calls(arguments, snapshot_file, snapshot):
+    runner = CliRunner()
+    result = runner.invoke(cli, f"{arguments} {LOCAL_METAMODEL_YAML_FILE}")
+    assert result.exit_code == 0
+    assert result.output == snapshot(f"genjsonschema/{snapshot_file}")
```
pkalita-lbl, line 28: @sierra-moxon ☝🏻

sierra-moxon (approved):

sierra-moxon:
I like the changes to the code; just need windoze tests to pass.

pkalita-lbl:
Yeah something about resolving the path to the local metamodel YAML file. I got the Windows VM out and I'm investigating.

Issue #1689, opened 2023-11-03
verbose INFO logging from LinkML CLI tools like gen-python


turbomam:
https://docs.python.org/3/howto/logging.html

turbomam:
> INFO:root:Importing MIXS_yaml:agriculture as https://raw.githubusercontent.com/microbiomedata/mixs/main/model/schema/agriculture from source https://raw.githubusercontent.com/microbiomedata/mixs/1da849346a80b717810a02d7c8ed74a22bcd84de/model/schema/mixs.yaml; base_dir=https://raw.githubusercontent.com/microbiomedata/mixs/1da849346a80b717810a02d7c8ed74a22bcd84de/model/schema

turbomam:
> INFO:root:TRUE: OCCURS SAME: Database == WorkflowExecutionActivity owning: Database

PR #1785, opened 2023-12-13
comment out verbose log info statements that percolate out to model g…

sierra-moxon:
…enerations

### DIFF
--- a/linkml/generators/pythongen.py
+++ b/linkml/generators/pythongen.py
@@ -963,10 +963,11 @@ dataclasses._init_fn = dataclasses_init_fn_with_kwargs
             return True
         for cname in self.schema.classes:
             if cname == owning_class:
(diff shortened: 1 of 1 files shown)


cmungall:
This PR partially mitigates #1689 but we need to figure out what is going on and why these are leaking in the first place!

cmungall (approved):

dalito:
Which command produced the problem?

From a quick look: I miss at the top of the file on module level a line `logger = logging.getLogger(__name__)`. Then all logging messages should be created using that logger. So it should be `logger.info(...)` not `logging.info(...)`. The latter falls back to using a default logger.

pkalita-lbl:
Maybe close in favor of https://github.com/linkml/linkml/pull/1790?

pkalita-lbl:
Closing now that https://github.com/linkml/linkml/pull/1790 is merged.

Issue #1666, opened 2023-10-12
Pydantic generator Jinja template incorrectly uses `getattr` when item is absent


sneakers-the-rat:
**Describe the bug**

Here's a fun one. 

I have a class with an attribute named "keys." For some reason the pydantic model kept showing up like this:

```python
class MyModel(BaseModel):
    keys: SomeType = Field(<built-in method keys of dict object at 0x13696a100>)
```

This is because the check here merely checks for a non-falsy value:

https://github.com/linkml/linkml/blob/5c1cbeae9abb841b4d35e23d73e4945ea2685e29/linkml/generators/pydanticgen.py#L127

According to Jinja2's docs:

https://jinja.palletsprojects.com/en/3.0.x/templates/#notes-on-subscriptions

> For the sake of convenience, foo.bar in Jinja does the following things on the Python layer:
>
> - check for an attribute called bar on foo (getattr(foo, 'bar'))
> - if there is not, check for an item 'bar' in foo (foo.__getitem__('bar'))
> - if there is not, return an undefined object.
>
> foo['bar'] works mostly the same with a small difference in sequence:
>
> - check for an item 'bar' in foo. (foo.__getitem__('bar'))
> - if there is not, check for an attribute called bar on foo. (getattr(foo, 'bar'))
> - if there is not, return an undefined object.
> 
> This is important if an object has an item and attribute with the same name. Additionally, the attr() filter only looks up attributes.

So since there wasn't a predefined slot value, it did `getattr()` instead, and lo and behold the python `dict` class had an attribute `keys`

Doing this instead resolved the problem:

```
{%- if predefined_slot_values[c.name][attr.name] is string -%}
```

cmungall:
See also:

- #104

sneakers-the-rat:
> See also:
> 
>     * [Using reserved words such as 'null' generates undecipherable stack traces #104](https://github.com/linkml/linkml/issues/104)

i feel like we should split off the generators into separate packages lol, seems hard to triage issues 1500 back. it is still sort of surprising to me to have the generators in this package and not `linkml_runtime` (though i totally understand how the wandering history of software would lead to that point). Thinking of the refactoring that i want to do with the pydantic generator and imagine that ppl concerned with different domains probably would like to split those conversations off to their own spaces.

PR #1797, opened 2023-12-16
Fix 1666 - Dont use callables for generated pydantic field defaults

sneakers-the-rat:
Fix: https://github.com/linkml/linkml/issues/1666

Bug is described in that issue, problem was that if a field had the same name as a method/attr of a python dict, it would get included as a field like:

```python
class MyClass(BaseModel):
    keys: Optional[str] = Field(<built-in method keys of dict object at 0x10f1f13c0>)
```

which is no good. Simple fix in the template and a regression test :)

### DIFF
--- a/linkml/generators/pydanticgen.py
+++ b/linkml/generators/pydanticgen.py
@@ -125,7 +125,7 @@ class {{ c.name }}
     {%- endif %}
     {% for attr in c.attributes.values() if c.attributes -%}
     {{attr.name}}: {{ attr.annotations['python_range'].value }} = Field(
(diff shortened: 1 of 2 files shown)


cmungall:
https://github.com/linkml/linkml/actions/runs/7229034231/job/19699352896?pr=1797#step:11:466

I believe `model_fields` is introduced in pydantic2

sneakers-the-rat:
> I believe `model_fields` is introduced in pydantic2

oop, definitely right. fixed

cmungall (approved):

Issue #1782, opened 2023-12-12
In gen-typescript create utility functions for interfaces


amc-corey-cox:
**Is your feature request related to a problem? Please describe.**
This is a feature request. Currently, when gen-typescript is run it creates interfaces with required and optional fields for each of the classes within the schema. This appears to be the recommended approach for data objects. However, imported data is then as a native object. This results in having to declare the object is of the interface type to allow proper type checking in the IDE. Since types are compiled away in typescript during runtime, this has no effect at runtime unless data is not conformant to the interface, which can result in crashes. This is why we use the IDE to perform type-checking to reduce opportunities for type errors to cause crashes.

**Describe the solution you'd like**
In my current work-flow, I write two companion functions for hinting to the IDE that imported data is conformant to an interface or to assure that it is. Here is an example:
export interface QCReport {
  dangling_edges?: QCPart[]
  edges: QCPart[]
  missing_nodes?: QCPart[]
  nodes: QCPart[]
}

export function toQCReport(i: object = {}): QCReport {
  const o = <QCReport>i
  return {
    dangling_edges: o.dangling_edges ?? [],
    edges: o.edges ?? [],
    missing_nodes: o.missing_nodes ?? [],
    nodes: o.nodes ?? [],
  }
  // Note this should return some type of error for missing edges/nodes but I don't have a good way to do that yet.
}

export function isQCReport(i: object): i is QCReport {
  return "edges" in i && "nodes" in i
}

**How important is this feature?** Select from the options below:
• Low - I already have these portions implemented for the parts that I need and I don't know of anyone else currently using the typescript generation. However, it would be nice to have this as an option to build type-checking and save some development time in future changes to the schema.

**When will use cases depending on this become relevant?** Select from the options below:
• Short-term - 2-4 weeks
The use case above is already relevant to my work, especially as I'm planning on extending my schema soon.

**Additional context**
It is quite possible that I'm not going about this the right way at all. Perhaps classes are a better choice for type-checking? I avoided implementing data-classes as full classes because they don't need any class-based functions.

amc-corey-cox:
I would plan to add this as an option to gen-typescript. Something like --include-type-functions or --include-helpers.

amc-corey-cox:
Refer to future discussion on PR #1799, which is currently a proof of concept of this idea.

amc-corey-cox:
PR 1799 merged.

PR #1799, opened 2023-12-19
Add util functions for typescript in generator

amc-corey-cox:
This implements the utility functions in typescript that I mentioned in #1782. This is based on how I use these in a web app I've based on my schema. I'd appreciate any thoughts and feedback on whether I should manage this differently. It's important to remember that typescript compiles out all types during runtime, so types are only checked at compile-time. These helper functions can be used to ensure we conform to the interfaces during runtime.

### DIFF
--- a/linkml/generators/typescriptgen.py
+++ b/linkml/generators/typescriptgen.py
@@ -79,6 +79,7 @@ export interface {{gen.name(c)}} {%- if parents %} extends {{parents|join(', ')}
     {%- endfor %}
 }
 
+{% if gen.gen_type_utils %}
 export function is{{gen.name(c)}}(o: object): o is {{gen.name(c)}} {
     {%- set rcs = gen.required_slots(c) %}
     {%- set comp = "&&" if rcs else "||" %}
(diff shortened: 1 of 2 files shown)


amc-corey-cox:
Fixes 1782

amc-corey-cox:
I realized I should probably add a few tests on this... I'll add those after I get some feedback on the approach here and whether this is appropriate generation.

amc-corey-cox:
Okay, I also realized I'm not creating appropriate attributes for when the supplied object doesn't have an attribute. This would require checking the multivalued and type on the slot and creating the appropriate empty object or primitive. I'll add this if we decide to move forward.

Similarly, I think this should be optional and I'll add a flag to add these if desired. I'm currently planning to use --gen-type-utils unless someone tells me a more appropriate flag.

amc-corey-cox:
Okay, the things I mentioned earlier were easier to add than I thought... I'll want this output for my development anyway, even if we don't main-line this. I do think these functions are useful, especially for Typescript. Because Typescript treats types in an unusual way, having utilities to ensure type safety and type-checking is potentially more important than in other scenarios.

amc-corey-cox:
I can pull the slot inheritance into another PR if we want to consider it separately; it was small so I just threw it in.

cmungall (approved):

PR #1809, opened 2024-01-02
Allow non-object types in JSON Schema generated from unconstrained classes

### LINKED ISSUE #1691: Annotations breaks on v1.6 - "is not of type 'object"
**Describe the bug**
I'm trying to implement the examples listed on https://linkml.io/linkml/schemas/annotations.html.

When using linkml 1.6, annotations using `linkml-lint --validate` results in a `valid-schema error`. This happens whether or not I'm using `instantiates. 

**To reproduce**
With the following schema.yaml:
```
id: https://example.org/mymetamodel/
name: mymetamodel
classes:
  Person:
    description: testing
    annotations:
      review:
        value: A very useful class that is well defined
```
This works with v1.5 and not v1.6:
```
$ docker run -ti --rm -v $PWD:/linkml   linkml/linkml:1.5  linkml-lint --validate /linkml/schema.yaml
✓ No problems found

$ docker run -ti --rm -v $PWD:/linkml   linkml/linkml:1.6  linkml-lint --validate /linkml/schema.yaml
/linkml/schema.yaml
  error    In classes > Person > annotations > review > value: 'A very useful class that is well defined' is not of type 'object'  (valid-schema)

✖ Found 1 problem in 1 schema
```

With instantiates, I still get the same error on 1.6. schema.yaml:
```
id: https://example.org/mymetamodel/
name: mymetamodel
classes:
  Reviewable:
    description: testing
    attributes:
      review:
        description: an expert review of a schema element
        range: string
        required: true
  Person:
    description: testing
    instantiates:
      - Reviewable
    annotations:
      review:
        value: A very useful class that is well defined
```
```
$ docker run -ti --rm -v $PWD:/linkml   linkml/linkml:1.6  linkml-lint --validate /linkml/schema.yaml
/linkml/schema.yaml
  error    In classes > Person > annotations > review > value: 'A very useful class that is well defined' is not of type 'object'  (valid-schema)
```

The same error occurs with without the `value` key as well
```
      review: A very useful class that is well defined
```

However, no error occurs (`linkml-lint --validate` shows "no problems found") with the following format for `review`:
```
      review:
        value:
          value: A very useful class that is well defined
```

**Expected behavior**
The same annotations in v1.5 work the same in v1.6.

**About your computer (if applicable, please complete the following information):**
```
$ docker run -ti --rm -v $PWD:/linkml   linkml/linkml:1.6  linkml-lint --version
linkml-lint, version 1.6.1
$ docker run -ti --rm -v $PWD:/linkml   linkml/linkml:1.5  linkml-lint --version
linkml-lint, version 1.5.8
```

pkalita-lbl:
Fixes #1691 

The root cause of the issue lies in a metamodel change in the 1.6 release. [This commit](https://github.com/linkml/linkml-model/commit/799b549be473f9628f2ddc297f4d71dd6a181034) changed the range of an annotation value from string to a class with `class_uri: linkml:Any`. 

When interpreting a class with `class_uri: linkml:Any`, the JSON Schema generator previously asserted that the value must be an **object**, albeit one with any attributes. For example, with this schema (abridged):

```yaml
# schema.yaml
classes:
  Anything:
    class_uri: linkml:Any
  MyClass:
    tree_root: true
    attributes:
      anything:
        range: Anything
```

This instance was considered valid:

```yaml
# data-valid.yaml
anything:
  a: 1
  b: 
    c: hello
```

However this instance was considered invalid:

```yaml
# data-invalid.yaml
anything: no good
```

With these changes the JSON Schema generator will produce JSON Schema that admits **any** scalar type (object, string, number, boolean, or null) as the value for a range marked with `class_uri: linkml:Any`, and so both of the above instances would be considered valid. Likewise this means that the pre-1.6 behavior of allowing simple strings as an annotation `value` is allowed again. 

These changes also do a little cleanup where a utility class was previously inheriting from `UserDict` instead of `dict`, even though we weren't actually using any of the features of `UserDict`. On the contrary there were some edge cases where passing a `UserDict` subclass to the `jsonschema` library functions was problematic. 

* * *

### Update

The compliance test suite identified an issue with the initial fix. Namely when a LinkML schema contains:

```yaml
s1:
  range: Any  # A class with class_uri: linkml:Any
  any_of:
    - range: D  # Just some other class
    - range: integer
```

The test expects for validation to fail on an instance like:

```yaml
s1: some string
```

Since the `any_of` range union is more specific than the `range` itself that makes sense. That LinkML schema gets translated pretty directly into JSON Schema that looks like:

```json
"s1": {
  "$ref": "#/$defs/Any",
  "anyOf": [
    { "$ref": "#/$defs/D" }, 
    { "type": "integer" }
  ]
}
```

The crucial piece of the puzzle here is that the JSON Schema generator has been specifying that the output it generates is compatible with JSON Schema Draft 7 (i.e. `"$schema": "http://json-schema.org/draft-07/schema#"`). However the Draft 7 spec says that when `$ref` is seen in a subschema then its sibling keywords (`anyOf` in the above example) are **not** processed. 

So before these changes the `"$ref": "#/$defs/Any"` was checking for an object with any keys. If it found a string instead (as in the above example), it failed and then stopped without checking the `anyOf` subschema. Which means the test was passing, but not exactly for the right reason.

With the initial fix the `"$ref": "#/$defs/Any"` checks for any scalar type. A string passes that check, but it still stops without checking the `anyOf`. Fortunately the JSON Schema authors seem to also realize the limitation of this, and in the JSON Schema Draft 2019-09 spec they changed it so that the sibling keywords of `$ref` are checked as well (https://json-schema.org/draft/2019-09/release-notes#core-vocabulary).

So the final change here is to update the JSON Schema generator to advertise that this later version of the spec should be used in validation (i.e. `"$schema": "https://json-schema.org/draft/2019-09/schema"`). Previously in our validation code we used to hardcode the use of a Draft 7 validator. These changes update that to choose a validator based on what the JSON Schema itself asserts ([see also](https://python-jsonschema.readthedocs.io/en/stable/api/jsonschema/validators/#jsonschema.validators.validator_for)). With that final change when the JSON Schema of the above example is encountered the `"$ref": "#/$defs/Any"` part still passes validation for a string, but the `anyOf` subschema _is also_ checked and it fails as expected.

A number of test snapshot files also had to be updated as a consequence of changing the JSON Schema generator output.

### DIFF
--- a/linkml/generators/jsonschemagen.py
+++ b/linkml/generators/jsonschemagen.py
@@ -189,7 +189,7 @@ class JsonSchemaGenerator(Generator):
 
         self.top_level_schema = JsonSchema(
             {
-                "$schema": "http://json-schema.org/draft-07/schema#",
+                "$schema": "https://json-schema.org/draft/2019-09/schema",
                 "$id": self.schema.id,
                 "metamodel_version": metamodel_version,
                 "version": self.schema.version if self.schema.version else None,
--- a/linkml/generators/jsonschemagen.py
+++ b/linkml/generators/jsonschemagen.py
@@ -1,7 +1,6 @@
 import json
 import logging
 import os
-from collections import UserDict
 from copy import deepcopy
 from dataclasses import dataclass, field
 from typing import Any, Dict, List, Optional, Tuple, Union
@@ -43,7 +42,7 @@ json_schema_types: Dict[str, Tuple[str, Optional[str]]] = {
 }
 
 
-class JsonSchema(UserDict):
+class JsonSchema(dict):
     OPTIONAL_IDENTIFIER_SUFFIX = "__identifier_optional"
 
     def __init__(self, *args, **kwargs):
@@ -112,7 +111,7 @@ class JsonSchema(UserDict):
         return self.get("type") == "object"
 
     def to_json(self, **kwargs) -> str:
-        return json.dumps(self.data, default=lambda d: d.data, **kwargs)
+        return json.dumps(self, **kwargs)
 
     @classmethod
     def ref_for(cls, class_name: Union[str, List[str]], identifier_optional: bool = False):
@@ -204,13 +203,15 @@ class JsonSchemaGenerator(Generator):
         if cls.mixin or cls.abstract:
             return
 
+        subschema_type = "object"
         additional_properties = False
         if self.is_class_unconstrained(cls):
+            subschema_type = ["null", "boolean", "object", "number", "string"]
             additional_properties = True
 
         class_subschema = JsonSchema(
             {
-                "type": "object",
+                "type": subschema_type,
                 "additionalProperties": additional_properties,
                 "description": be(cls.description),
             }
@@ -513,7 +514,7 @@ class JsonSchemaGenerator(Generator):
             type_value = get_type_designator_value(self.schemaview, slot, cls)
             prop["enum"] = [type_value]
 
-    def generate(self) -> dict:
+    def generate(self) -> JsonSchema:
         self.start_schema()
         for enum_definition in self.schemaview.all_enums().values():
             self.handle_enum(enum_definition)
--- a/linkml/linter/linter.py
+++ b/linkml/linter/linter.py
@@ -1,5 +1,4 @@
 import inspect
-import json
 from copy import deepcopy
 from dataclasses import dataclass
 from functools import lru_cache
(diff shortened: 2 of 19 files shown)


cmungall (approved):
Thanks for all the background information, I agree with the analysis and the design decisions.

PR #1856, opened 2024-02-01
Account for `any_of` ranges in documentation pages

### LINKED ISSUE #1851: Markdown documentation pages bug when range expressed as unions using `any_of`
The range of a slot can be expressed as a union of enums using the `any_of` construct, as described here: https://linkml.io/linkml/schemas/advanced.html#unions-as-ranges

Currently, the documentation generator does not have the provision/functionality to correctly render the ranges of slots when they are specified using `any_of`. It simply displays it with the `default_range` which is _string_.

Proposals:

If the range of a slot is a union of enums, then express it as one of the following:
* <s> Enum1 ∪ Enum2 </s>
* <s> [Enum1, Enum2] </s>
* Enum1 or Enum2 ✅

Feel free to suggest other syntax for expressing this if you have something in mind.

sujaypatil96:
Fixes #1851

### DIFF
--- a/linkml/generators/docgen/class.md.jinja2
+++ b/linkml/generators/docgen/class.md.jinja2
@@ -8,6 +8,21 @@
     {%- endif -%}
 {%- endif -%}
 
+{% macro compute_range(slot) -%}
+    {%- if slot.any_of -%}
+        {%- for subslot in slot.any_of -%}
+            {%- if subslot.range -%}
+                {{ gen.link(subslot.range) }}
+                {%- if not loop.last -%}
+                    &nbsp;∪&nbsp;
+                {%- endif -%}
+            {%- endif -%}
+        {%- endfor -%}
+    {%- else -%}
+        {{ gen.link(slot.range) }}
+    {%- endif -%}
+{% endmacro %}
+
 # Class: {{ title }}
 
 {%- if header -%}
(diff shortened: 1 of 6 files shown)


sujaypatil96:
CC: @turbomam

linkml/generators/docgen/class.md.jinja2
```diff
@@ -8,6 +8,21 @@
     {%- endif -%}
 {%- endif -%}
 
+{% macro compute_range(slot) -%}
```
cmungall, line 11: this looks like logic that is already in schemaview?

linkml/generators/docgen/slot.md.jinja2
```diff
@@ -8,6 +8,21 @@
     {%- endif -%}
 {%- endif -%}
 
+{% macro compute_range(slot) -%}
```
cmungall, line 11: a bit duplicative

cmungall (approved):
I think this is basically fine but this only works for a narrow range of possible expressions.

The correct way to do this is to traverse the full tree rendering recursively as markdown.

See https://github.com/linkml/linkml/blob/main/tests/test_compliance/test_boolean_slot_compliance.py

linkml/generators/docgen/class.md.jinja2
sujaypatil96: Yes, you're right, the functionality is in schemaview here: https://github.com/linkml/linkml-runtime/blob/main/linkml_runtime/utils/schemaview.py#L1471-L1486

I'm using that now

sujaypatil96:
For some strange reason there seems to be an error in one of the ShEx tests 🤔

sujaypatil96:
As discussed with @cmungall I am disabling the test that is erroring out in `test_shexgen.py` to unblock this PR.

Issue has been documented in #1914

Issue #1857, opened 2024-02-02
gen-python generates classes in wrong order in some cases


kervel:
**Describe the bug**

When there are multiple schema's refering to each other, it is possible to generate syntactically incorrect python. For instance:

**To reproduce**

machine.yaml:

```yaml
id: https://w3id.org/a
name: m
title: m

prefixes:
  mdm: https://a/mdm/
  linkml: https://w3id.org/linkml/
  schema: http://schema.org/
default_prefix: mdm
default_range: string

imports:
  - linkml:types
  - ./equipment


classes:
    Machine:
      is_a: Equipment


    Container:
        tree_root: true
        attributes:
            machines:
                multivalued: true
                inlined_as_list: true
                range: Machine
```

equipment.yaml

```yaml
id: b
name: b
title: b

prefixes:
  mm: https://b/
  linkml: https://w3id.org/linkml/
  schema: http://schema.org/
default_prefix: mm
default_range: string

imports:
  - linkml:types

classes:
  Equipment:
    description:
      An individually identifiable piece of equipment
```

now `gen-python machine.yaml` generates python that is not correct. The reason is that the type hints in the `Container` class refer to a `Machine` which is only defined later. The fix is simple (quoting the type hints).

**Expected behavior**

The generated python compiles

**About your computer (if applicable, please complete the following information):**
 - OS: Linux ubuntu 22.04
 - Version Linkml 1.6.10

cc @vdwpsmt

PR #1859, opened 2024-02-05
attempt at fixing forward references when using imported schemas in p…

kervel:
The forward_reference function in pythongen doesn't sort classes the same way the class definition generator does, leading to wrong results and wrong quoting in the generated python.

This PR fixes this by using exactly the same way of sorting.

This should fix issue #1857

### DIFF
--- a/linkml/generators/pythongen.py
+++ b/linkml/generators/pythongen.py
@@ -966,7 +966,8 @@ dataclasses._init_fn = dataclasses_init_fn_with_kwargs
             return False
         if slot_range in self.schema.enums:
             return True
-        for cname in self.schema.classes:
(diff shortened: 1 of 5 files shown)


kervel:
it seems that there are unit tests that compare generations with a snapshot. I think the snapshots would need updating after this patch, i guess there is a way to do that ?

cmungall:
See https://linkml.io/linkml/contributing/contributing.html#custom-pytest-fixtures

kervel:
> See https://linkml.io/linkml/contributing/contributing.html#custom-pytest-fixtures

I did that and the tests are fine now. So i guess this PR is ready for review.. Who could i ask for a review ? 
Thanks!

cmungall (approved):

Issue #1862, opened 2024-02-05
Improve pydantic support for inlining as CompactDict SimpleDict


cmungall:
pydanticgen support `inlined` as dicts (added here some time ago #1094)

however, if the input dictionary/json/yaml uses SimpleDict or CompactDict it's necessary to use the normalized to pre-process. It would be more convenient if this was supported out the box.

pydantic2 now supports the equivalent of `__post_init__`, so we have the option of handling this the same way as dataclasses.

However, I am not sure if the dataclasses way of doing this is optimal. It creates awkward type signatures with lots of `Union` constructs (to handle the fact the inlined object mat have the key instantiated from the "outside").

Although it creates additional boilerplate, it may be better to separate the initialization signature (Postelian) from the access signature (coerced)

sneakers-the-rat:
I think we should really avoid adding `__post_init__` -like behavior to the pydanticgen, as it's not really how pydantic is supposed to work and makes the models fragile (no validation on assignment, no reflection in the generated schema, etc) - we should use validators for this. a `pre` validator can do coercion.

I'm not sure when this would happen in generated models, is this just for compatibility with lots of input types? I know that the annotation class there is used in the pydantic generation process, but is it also intended to be used by the models themselves? I guess the question is 'how would we make this generic' - do all dictionaries of this form use `tag` and `value`? if so we could just make that its own type with validator

```python
from typing import Dict, Union, Annotated, List
from typing_extensions import TypedDict
from pydantic import BaseModel, BeforeValidator

class MyDict(TypedDict):
    tag: str
    value: str

def coerce_simple_dict(v: Union[Dict[str,str], Dict[str, MyDict], List[MyDict]]):
    if isinstance(v, list):
        return {item['tag']: item for item in v}
    elif isinstance(v, dict):
        return {k: v if isinstance(v, dict) else MyDict(tag=k, value=v) for k,v in v.items()}
    return v

GenericDict = Annotated[
    Dict[str, MyDict],
    BeforeValidator(coerce_simple_dict)
]

class MyModel(BaseModel):
    a_dict: GenericDict

model = MyModel(a_dict={'foo': 'bar'})
model2 = MyModel(a_dict={'foo': MyDict(tag='foo', value='bar')})
model3 = MyModel(a_dict=[MyDict(tag='foo', value='bar')])
```

which we could do since we can inject classes and whatnot directly now

[...]

PR #1869, opened 2024-02-07
Adding support for serializing inlined tuples as simple dicts in pydantic.

cmungall:
See #1862


cmungall:
argh, last commit introduced test output changes, ...

PR #1887, opened 2024-02-08
[pydanticgen] Add list of lists array implementation

### LINKED ISSUE #1888: Add support for linkml:array in pydanticgen
Introduced in 1.7.0

* [ ] default should be to use `[List[List[...]]]` (pydantic will validate these)
* [ ] add ability to use other frameworks for both the array in-memory management and typing
     - [ ] numpy
     - [ ] nptyping
     - [ ] xarr
     - [ ] ...

rly:
Fixes #1888

@cmungall @sneakers-the-rat 

Code review TODOs:
- [x] Add test cases for `minimum_dimensions == maximum_dimensions` as equivalent to `exact_dimensions` - https://github.com/linkml/linkml/pull/1887/commits/89bd79c3336a41dda7b8df76b51d75a3c50043d5
- [x] Fix https://github.com/linkml/linkml/pull/1887#discussion_r1518658858 - https://github.com/linkml/linkml/pull/1887/commits/89bd79c3336a41dda7b8df76b51d75a3c50043d5
- [x] Write docs - https://github.com/linkml/linkml/pull/1887/commits/bcfeec0f7b7278132631f2671e1dc93a984fd85a
- [x] Make black optional dependency (but required for docs) - https://github.com/linkml/linkml/pull/1887/commits/5b14fd99120719ec1736dc1b02b92c44141dc3d8
- [x] convert `__add__` to `merge` in `BuildResult` - https://github.com/linkml/linkml/pull/1887/commits/856f1746dcc960e7a461c9990dbe446bdaade4be


linkml/generators/pydanticgen.py
```diff
@@ -525,6 +532,52 @@ def generate_collection_key(
             return list(collection_keys)[0]
         return None
 
+    def get_array_representations_range(self, slot: SlotDefinition) -> str:
+        """
+        Generate the python range for array representations
+        """
+        array_info = slot.array
+        if array_info is None:  # pragma: no cover
+            return ""  # this should not happen
```
sneakers-the-rat, line 541: should we raise `ValueError` here just to avoid silent bugs?

sneakers-the-rat: oh wait wasn't this supposed to be "any shaped array"?

sneakers-the-rat:
I gotta run, but i refactored into an ABC so we can keep track of what we hace done and need to do!!! write more later

linkml/generators/pydanticgen.py
cmungall: Yep!

cmungall:
Thanks!

Looks like this introduces:

```
FAILED tests/test_compliance/test_array_compliance.py::test_array[pydantic-1D array with float against 1D array float schema-1-object1-True] - TypeError: Can't instantiate abstract class ListOfListsArray with abstract methods any_shape, labeled_shape, mixed_shape
```

sneakers-the-rat:
The specification of unbounded shape arrays never stop being a barrel of laughs <3

Let's start with desired API. How about something like this

```python
class MyModel(BaseModel):
    any_dtype: AnyShapeArray
    int_dtype: AnyShapeArray[int]
```

Warmup: how do you even annotate an arbitrarily deep list of lists in python? Recursive types of course!

```python
RecursiveList = List[int | List['RecursiveList']]
```

And if we want to get generic

```python
T = TypeVar('T')
RecursiveListType = List[T | List['RecursiveListType']]
```

Problem: this throws a recursion error as pydantic tries to get an infinite depth of schemas.  The problem comes from both
* the generation of the `core_schema` used for validation and serialization and
* the generation of the JSON schema

Initially i tried to just say "ok the pydantic core schema is going to just be incorrect, but we can provide a validator function that checks the inner type of arbitrarily nested lists. but that was inelegant in a bunch of ways, because I don't want to have to run a big nasty recursive function to visit every subarray in python (slow), it would be better to be able to use the underlying Rust `pydantic_core` schema still. 

Thankfully, at the very bottom of the pydantic core schema docs, we find `definitions_schema` and `definitions_reference_schema` : https://docs.pydantic.dev/latest/api/pydantic_core_schema/#pydantic_core.core_schema.definitions_schema

This lets us declare the JSON schema and the core schema in one fell swoop.

Something like this:

```python
array_ref='recursive-array'
item_schema = core_schema.int_schema()
schema = core_schema.definitions_schema(
    core_schema.list_schema(core_schema.definition_reference_schema(array_ref)),
    [core_schema.union_schema(
        [
            core_schema.list_schema(
                core_schema.definition_reference_schema(array_ref)
            ),
            item_schema
        ], 
        ref=array_ref
    )]
)
```

New problem: now we can only declare one type of unbounded list per model! so we need to create different schema refs for each type by modifying the `array_ref` name to include the type.

```python
class AnyShapeArray(Generic[T]):
    @classmethod
    def __get_pydantic_core_schema__(
            cls, source_type: Any, handler: GetCoreSchemaHandler
    ) -> CoreSchema:
        # get the python type specified as the generic.
        # double-nested parameterized types here
        # source_type: List[Union[T,List[...]]]
        item_type = Any if get_args(get_args(source_type)[0])[0] is T else get_args(get_args(source_type)[0])[0]

        item_schema = handler.generate_schema(item_type)
        array_ref = f"any-shape-array-{item_type.__name__}"

        schema = core_schema.definitions_schema(
            core_schema.list_schema(core_schema.definition_reference_schema(array_ref)),
            [core_schema.union_schema([
                core_schema.list_schema(core_schema.definition_reference_schema(array_ref)),
                item_schema
            ], ref=array_ref)])
        
        return schema
```

OK that works, and the model validates, but the type checker hates that because it's expecting a `AnyShapeArray` type, not a list of lists. So now we need to combine these approaches using `Annotated`, where we tell the type checker one thing and tell pydantic another thing:

```python

T = TypeVar('T')
RecursiveListType = List[T | List['RecursiveListType']]

class AnyShapeArrayType(Generic[T]):
    @classmethod
    def __get_pydantic_core_schema__(
            cls, source_type: Any, handler: GetCoreSchemaHandler
    ) -> CoreSchema:
        # double-nested parameterized types here
        # source_type: List[Union[T,List[...]]]
        item_type = Any if get_args(get_args(source_type)[0])[0] is T else get_args(get_args(source_type)[0])[0]

        item_schema = handler.generate_schema(item_type)
        array_ref = f"any-shape-array-{item_type.__name__}"

        schema = core_schema.definitions_schema(
            core_schema.list_schema(core_schema.definition_reference_schema(array_ref)),
            [core_schema.union_schema([
                core_schema.list_schema(core_schema.definition_reference_schema(array_ref)),
                item_schema
            ], ref=array_ref)])
        
        return schema

AnyShapeArray = Annotated[RecursiveListType, AnyShapeArrayType]
```

And that ends up working:

```python

class RecModel(BaseModel):
    anytype: AnyShapeArray
    int_array: AnyShapeArray[int]
    str_array: AnyShapeArray[str]

a = RecModel(
    anytype=[[[[2], ["3"]]]],
    int_array=[[[[[[[2], [2]]]]]]],
    str_array=[[[["a"], ["b"], ["c"]]]],
)
```

Model correctly validates, and creates the following JSON Schema:

```json
{'$defs': {'any-shape-array-Any': {'anyOf': [{'items': {'$ref': '#/$defs/any-shape-array-Any'},
     'type': 'array'},
    {}]},
  'any-shape-array-int': {'anyOf': [{'items': {'$ref': '#/$defs/any-shape-array-int'},
     'type': 'array'},
    {'type': 'integer'}]},
  'any-shape-array-str': {'anyOf': [{'items': {'$ref': '#/$defs/any-shape-array-str'},
     'type': 'array'},
    {'type': 'string'}]}},
 'properties': {'anytype': {'items': {'$ref': '#/$defs/any-shape-array-Any'},
   'title': 'Anytype',
   'type': 'array'},
  'int_array': {'items': {'$ref': '#/$defs/any-shape-array-int'},
   'title': 'Int Array',
   'type': 'array'},
  'str_array': {'items': {'$ref': '#/$defs/any-shape-array-str'},
   'title': 'Str Array',
   'type': 'array'}},
 'required': ['anytype', 'int_array', 'str_array'],
 'title': 'RecModel',
 'type': 'object'}
```

So... anyway that's the working version, there is probably a more elegant way of doing that with proper use of generics, but we can optimize later. It's not pretty but it works! I think no matter what we do here, if we want to be able to properly validate with the pydantic models we'll need to bundle some additional type along with the generated schema, unfortunately. I'm going to make an additional patch to the pydantic generator that allows the imports and included classes to be customized depending on what the generated schema needed, so eg. if you're generating in List of Lists mode, inject this type and the necessary imports, but otherwise keep the generated code slim.

sneakers-the-rat:
> Thanks!
> 
> Looks like this introduces:
> 
> ```
> FAILED tests/test_compliance/test_array_compliance.py::test_array[pydantic-1D array with float against 1D array float schema-1-object1-True] - TypeError: Can't instantiate abstract class ListOfListsArray with abstract methods any_shape, labeled_shape, mixed_shape
> ```

Oh yes sorry didnt see this. not done. That is a correct error since those methods havent been written, just committed to show the structure.  Not a regression per se since I took the case that previous version handled and adapted it here. Ill have a working version in a bit, or at least one that doesnt raise because of the ABC.

sneakers-the-rat:
Since there will be some pretty significant Pydantic 1 v 2 differences in the implementation here, would it be alright to split this out into a few files? iirc. `pydanticgen` used to be a package with the templates split out into a separate file. It would be nice to be able to split some of the array generation classes out so that this file doesn't get massive, and there will also be substantial module-level `if:` switches for doing pydantic 1 vs 2 stuff that seems pretty error prone

cmungall:
Thanks @sneakers-the-rat!

Is there any way we could split this into two PRs? I agree that some kind of refactoring of pydanticgen will be necessary. But it will be easier to approach this incrementally.

sneakers-the-rat:
> Thanks @sneakers-the-rat!
> 
> Is there any way we could split this into two PRs? I agree that some kind of refactoring of pydanticgen will be necessary. But it will be easier to approach this incrementally.

totally ya. I could finish the list of lists implementation in this PR in place, and then save refactoring for a later PR if that works? or do you mean split this current one up? What i had in mind here for this PR was just doing the list of lists implementation + tests for all the different forms of arrays. 

To avoid a "half implementation," i'll first make a PR for injecting imports and classes into the pydanticgen and then rebase this PR off of that, does that work?

cmungall:
> totally ya. I could finish the list of lists implementation in this PR in place, and then save refactoring for a later PR if that works? or do you mean split this current one up? What i had in mind here for this PR was just doing the list of lists implementation + tests for all the different forms of arrays.

I had in mind originally saving the refactoring for a later PR

> To avoid a "half implementation," i'll first make a PR for injecting imports and classes into the pydanticgen and then rebase this PR off of that, does that work?

I think this also works!

sneakers-the-rat:
genuine question here - so the list of lists implementation is for a zero-additional-deps version, but the more I think about it the less sense it makes, are we thinking people will be working with array data without having even numpy installed?

Here's what i'm thinking:

Case for list-of-lists *only* implementation is to be able to generate a module whose only external dependency is pydantic, keeping in line with the existing generator. This would be for cases where someone is dealing with array data that comes from JSON, I think almost exclusively, because in any other context where you're using python then an array will be synonymous with at least numpy (if not more sophisticated array libs like zarr, dask, etc.). I can't think of another case where someone would be dealing with array data without an array library as a dependency. 

## Downsides to list of lists having an independent implementation: 
- a decent amount of additional complexity: expand implementations to 4 - (LoL, pydantic 1; LoL, pydantic 2; nptyping, pydantic 1; nptying, pydantic 2). 
- accompanying maintenance burden.
- potential for divergent functionality
- difficult to do complete array validation: to validate the shape of a LoL, one needs to do a recursive check - eg for an array with shape `(2,3,4,5)`, check top list is `len() == 2)`, check each of those lists is `len() == 3`, check each of those lists is `len() == 4` ... and so on. this is especially complex for the mixed constrainted/unconstrained arrays where we basically rely on python's recursion error checking being the backstop of infinite computation time.
- slow: this example for 100 frames of a 720p video:

```python
from pydantic import BaseModel
from pydantic.types import conlist
import timeit
import numpy as np
import sys

class MyModel(BaseModel):
    array: conlist(
        conlist(
            conlist(
                conlist(
                    int,
                    min_length = 100, 
                    max_length = 100
                ),
                min_length=3, 
                max_length=3
            ),
            min_length=720,
            max_length=720
        ),
        min_length=1280,
        max_length=1280
    )

lol = np.random.randint(0,255,(1280, 720, 3, 100), dtype=np.uint8).tolist()

if __name__ == "__main__":
    print('timing')
    print(sys.getsizeof(lol))
    print(timeit.timeit('validated = MyModel(array=lol)', globals=globals(), number=10))
```

takes 74s (7.4s per iteration) - pydantic has to visit each int and do type checking, and even in the rust `pydantic-core` backend this is super slow. i imagine it's even slower in pydantic 1.
- akward syntax or lots of boilerplate: i sort of hate the nested `conlist` implementation above, so we would need to make a wrapper class for that, and the one I have now is already relatively complex. 

## Requirements:

we still want to be able to 
- read from LoL
- serialize to LoL
- be able to control deps on generation. 

the `nptyping`-based ArrayProxy class i was working on should be able to do all of that - start from any format, serialize to any format (dependent on what packages you have installed), and it should also be able to generate the correct JSON-schema representation as a LoL, but it could then use the much faster/easier to program `shape` and `dtype` attributes offered by any of these libraries. We could even wrap that dependency and say "you can use this type without numpy or whatever installed, but numpy is necessary to be able to validate the shape and dtype of your arrays." I think that might ultimately be just as kind bc we can write that class s.t. it provides hooks to let you declare an interface for your favorite array library, so they can keep the deps they already use in their package. 

anyway wanted to check in before i continue working on that, since the last major thing to do is the mixed constrained/unconstrained shape arrays

sneakers-the-rat:
Ping on this q, im gonna work on the nptyping version today. Thinking about what we want re: extensibility, we will want to provide hooks to a bunch of different types of array backends, so im going to decouple the LoL schema generation part (which should be more or less constant across implementations) from the validation part, which is where other backends hook in.

sneakers-the-rat:
I am working on this now and i don't think we'll be able to do List of List implementation with pydantic 1 with validation - bumping request for pydantic 1 deprecation timeline - https://github.com/linkml/linkml/issues/1925

I'm going to try for about an hour more but the problem is with the 'any shape' arrays, all the stuff i did for recursive types for pydantic 2 doesn't apply in pydantic 1, and pydantic 1 doesn't have comparable control over the internal schema. the best i think i'll be able to do is a recursive validator function but that will only be able to validate the inner type, and it'll get really ugly for mixed specified and unspecified arrays.

cmungall:
I think the proposal here is to drop support for v1 *in pydanticgen*. In
which case, I am OK with this, but let's check more widely first, as
migration is not always easy due to transitive dependencies.

We can keep linkml-runtime neutral for the immediate future, AFAICT
(because linkml-model, for better or worse, uses dataclasses).

sneakers-the-rat:
Sorry a few different things here has i've been working:

https://github.com/linkml/linkml/issues/1925 is a proposal to set a deprecation timeline for V1 generally to allow for downstream migration if anyone depends on v1 models, since it is generally deprecated. 

I have I think found a way to do the array spec in v1 (and am almost done, hopefully by tonight) but it's _not pretty_ since v1 doesn't have some of the tools for custom types that v2 has. 

[separately](https://github.com/linkml/linkml/pull/1887#issuecomment-1942800865) i was wondering about the need for an implementation that doesn't depend on numpy at all (ie. is purely based on python lists of python lists) because it's substantially more complex than being able to use numpy for validation. my implementation that does use numpy is capable of doing all the validation, ingestion, and schema generation for lists of lists, just not having numpy is about ~500 lines of ugly code doing a bunch of manual list parsing that would have to be injected into the module anyway. i've been working on [other](https://github.com/linkml/linkml/pull/1927) [strategies](https://github.com/linkml/linkml/pull/1916) in parallel to make it somewhat maintainable, but ya it would be a very simplifying requirement to be able to just say "if you are using arrays you have to have numpy installed"

sneakers-the-rat:
OK Here's the first complete (i think?) draft of all combinations of array specifications with pydantic 1. like I said it is NOT PRETTY and also I have yet to inject the necessary accessory classes, but anyway here's this:

replicable from [9fbd99c](https://github.com/linkml/linkml/pull/1887/commits/9fbd99c088faa4808642bc3d87eeba23f17c1dad) with:

```python
from linkml.generators import PydanticGenerator
gen = PydanticGenerator('tests/test_generators/input/arrays/all_arrays.yaml')
print(gen.serialize())
```

which is based off the following schema:

Any shape arrays:

```yaml
AnyType:
  attributes:
    array:
      range: AnyType
      array: {}

Typed:
  attributes:
    array:
      range: integer
      array: {}
```

Anonymously shaped arrays:

```yaml
ExactDimensions:
  attributes:
    array:
      range: integer
      array:
        exact_number_dimensions: 3

MinDimensions:
  attributes:
    array:
      range: integer
      array:
        minimum_number_dimensions: 2

MaxDimensions:
  attributes:
    array:
      range: integer
      array:
        maximum_number_dimensions: 5

RangeDimensions:
  attributes:
    array:
      range: integer
      array:
        minimum_number_dimensions: 2
        maximum_number_dimensions: 5
```

Labeled axis arrays (note that i have not added the aliases, that will be something that will be possible to do after https://github.com/linkml/linkml/pull/1927

```yaml
LabeledArray:
  attributes:
    array:
      range: integer
      array:
        dimensions:
          - alias: max_card
            maximum_cardinality: 5
          - alias: min_card
            minimum_cardinality: 2
          - alias: range_card
            minimum_cardinality: 2
            maximum_cardinality: 5
```

Mixed labeled and anonymous arrays:

```yaml
MixedAnyShapeArray:
  attributes:
    array:
      range: integer
      array:
        maximum_number_dimensions: false
        dimensions:
          - alias: max_card
            maximum_cardinality: 5
          - alias: min_card
            minimum_cardinality: 2

MixedMaxShapeArray:
  attributes:
    array:
      range: integer
      array:
        maximum_number_dimensions: 5
        dimensions:
          - alias: max_card
            maximum_cardinality: 5
          - alias: min_card
            minimum_cardinality: 2

MixedRangeShapeArray:
  attributes:
    array:
      range: integer
      array:
        minimum_number_dimensions: 4
        maximum_number_dimensions: 5
        dimensions:
          - alias: max_card
            maximum_cardinality: 5
          - alias: min_card
            minimum_cardinality: 2

MixedExactShapeArray:
  attributes:
    array:
      range: integer
      array:
        exact_number_dimensions: 5
        dimensions:
          - alias: max_card
            maximum_cardinality: 5
          - alias: min_card
            minimum_cardinality: 2
```

Generates the following code (with manual formatting because i don't want to preoptimize for formatting if the underlying representation is going to change, plus i think we should just add black as a dependency and use it whenever we serialize everything instead of doing basically all of the stylistic formatting that's done in the generator and templates, but i digress)

```python
from typing import List, Dict, Optional, Any, Union, Literal
from pydantic import BaseModel as BaseModel, Field, validator, conlist

metamodel_version = "None"
version = "None"


class WeakRefShimBaseModel(BaseModel):
    __slots__ = "__weakref__"


class ConfiguredBaseModel(
    WeakRefShimBaseModel,
    validate_assignment=True,
    validate_all=True,
    underscore_attrs_are_private=True,
    extra="forbid",
    arbitrary_types_allowed=True,
    use_enum_values=True,
):
    pass

# -------------------------------------------
# Any Shape
# -------------------------------------------

class AnyType(ConfiguredBaseModel):
    array: Optional[AnyShapeArray[AnyType]] = Field(None)


class Typed(ConfiguredBaseModel):
    array: Optional[AnyShapeArray[int]] = Field(None)

# -------------------------------------------
# Anonymous Shape
# -------------------------------------------

class ExactDimensions(ConfiguredBaseModel):
    array: Optional[List[List[List[int]]]] = Field(None)


class MinDimensions(ConfiguredBaseModel):
    array: Optional[List[List[AnyShapeArray[int]]]] = Field(None)


class MaxDimensions(ConfiguredBaseModel):
    array: Optional[
        Union[
            List[int],
            List[List[int]],
            List[List[List[int]]],
            List[List[List[List[int]]]],
            List[List[List[List[List[int]]]]],
        ]
    ] = Field(None)


class RangeDimensions(ConfiguredBaseModel):
    array: Optional[
        Union[
            List[List[int]],
            List[List[List[int]]],
            List[List[List[List[int]]]],
            List[List[List[List[List[int]]]]],
        ]
    ] = Field(None)

# -------------------------------------------
# Labeled Dimensions
# -------------------------------------------

class LabeledArray(ConfiguredBaseModel):
    array: Optional[
        conlist(
            max_items=5,
            item_type=conlist(
                min_items=2, 
                item_type=conlist(
                    min_items=2, 
                    max_items=5, 
                    item_type=int
                )
            ),
        )
    ] = Field(None)

# -------------------------------------------
# Mixed Labeled & Anonymous
# -------------------------------------------

class MixedAnyShapeArray(ConfiguredBaseModel):
    array: Optional[
        conlist(
            min_items=2, 
            item_type=conlist(
                max_items=5, 
                item_type=AnyShapeArray[int]
            )
        )
    ] = Field(None)


class MixedMaxShapeArray(ConfiguredBaseModel):
    array: Optional[
        conlist(
            min_items=2,
            item_type=conlist(
                max_items=5,
                item_type=Union[
                    List[int], 
                    List[List[int]], 
                    List[List[List[int]]]],
            ),
        )
    ] = Field(None)


class MixedRangeShapeArray(ConfiguredBaseModel):
    array: Optional[
        conlist(
            min_items=2,
            item_type=conlist(
                max_items=5, 
                item_type=Union[
                    List[List[int]], 
                    List[List[List[int]]]
                ]
            ),
        )
    ] = Field(None)


class MixedExactShapeArray(ConfiguredBaseModel):
    array: Optional[
        conlist(
            min_items=2, 
            item_type=conlist(
                max_items=5, 
                item_type=List[List[List[int]]]
            )
        )
    ] = Field(None)


```

If you take a look at the implementation u will see many cleverness that it took to get there, and many cleverness that should be cleaned up. (i notice that did an unnecessary reversal of the dimension spec. 

anyway that's an implementation that I would call "unreadable" and "messy" and I would like to replace it with a much cleaner implementation but it would require being able to inject dependencies and i wanted to wait on doing that until we could chat. 

I'll follow up with tests and also finalize the other PRs that will allow this to reach a final form in the coming days <3

cmungall:
What are the best steps forward for this PR? We now have the refactored pydanticgen merged, does it make sense to rebase, or start a new PR...

Regarding the ugliness of the generated solutions: I think this is acceptable for now. If you want to have very flexible arrays and at the same time use strong direct typing with no dependencies, then you end up with these disjunctive definitions. These are will likely be in the minority for now.

If these start being a problem for people then there are two paths:

1. an option to make a less strongly typed signature
2. running pydanticgen with the option of compiling to something like numpy with signatures provided by nptyping

Shall we discuss on the Friday arrays call?

sneakers-the-rat:
Planning on continuing this today :)

I think we can keep this going, i'll merge in main and finish the last bits re: injecting the generated classes. The purpose of waiting on that was to be able to only inject these big classes when needed, so that'll help i think with the ugliness of it ;)

i dont think pydanticgen should ever make models that are less explicit than the schema - too many weird cases that would have to be documented, i think it's better to just say "pydantic models are as 1:1 with the schemas as we can make them even if they are ugly sometimes"

option 2 is the more fun one - it'll require a bit more than just the nptyping signatures to make them work with pydantic models (that's the purpose of my numpydantic package), but that "if you are using other array libraries then use this option" will i think we the one that actually gets used in practice. i think you'll like what i have in mind there, i have a pretty clear picture of how to make it hookable for lazy loading and passthrough functionality to any array library u like.

cmungall:
> "i dont think pydanticgen should ever make models that are less explicit
than the schema "

I think this is good in principle and I agree in this specific case of
arrays. But in general not generator is guaranteed to translate every
construct in LinkML and in many cases this is impossible due to
expressivity limits in the target framework. We need to document this a bit
better, this underpins a lot of ongoing discussions about monotonicity (a
generated representation is guaranteed to be no more restrictive than the
source schema, even if all components cannot be expressed)

sneakers-the-rat:
Totally, yes of course. and that's another thing I hope that having target-domain models will help with! def agreed on "as close as possible without extra restrictions"

sneakers-the-rat:
here's a fun one - for the list of lists implementation, if we are assuming numpy won't be present, then we also can't really handle if it _is_ present all that well, so eg. passing a numpy array to a LoL array will fail, which will probably be counterintuitive. I'm going to finish the test for it now and see what happens, but anyway ya i suspect this LoL implementation won't be all that useful and it will be dramatically simplifying to say "if you want to use arrays you need to have numpy"

cmungall:
I think the scientific array use case will dominate (in which case you
really want to use a lib like numpy). But there will be users who want to
do things like represent a list of tuples for which LoL is natural and
numpy is overkill…

sneakers-the-rat:
([rod serling voice](https://youtu.be/yeiz66uDBUA)): submitted for your approval

an implementation of the List of List arrays with a first round of tests for all the types of arrays that the specification allows:

- Any Shape
  - typed
  - untyped
- Anonymous shape
  - min dimensions
  - max dimensions
  - range dimensions
  - exact dimensions
- Labeled shape
  - min cardinality
  - max cardinality
  - range cardinality
  - exact cardinality
- Mixed shape
  - (product of the above)

## Test Schema

```yaml
id: all_arrays
name: all_arrays
title: all_arrays
imports:
  - linkml:types

classes:
  AnyType:
    attributes:
      array:
        range: AnyType
        array: {}

  Typed:
    attributes:
      array:
        range: integer
        array: {}

  ExactDimensions:
    attributes:
      array:
        range: integer
        array:
          exact_number_dimensions: 3

  MinDimensions:
    attributes:
      array:
        range: integer
        array:
          minimum_number_dimensions: 2

  MaxDimensions:
    attributes:
      array:
        range: integer
        array:
          maximum_number_dimensions: 5

  RangeDimensions:
    attributes:
      array:
        range: integer
        array:
          minimum_number_dimensions: 2
          maximum_number_dimensions: 5

  LabeledArray:
    attributes:
      array:
        range: integer
        array:
          dimensions:
            - alias: min_card
              minimum_cardinality: 2
            - alias: max_card
              maximum_cardinality: 5
            - alias: range_card
              minimum_cardinality: 2
              maximum_cardinality: 5
            - alias: exact_card
              exact_cardinality: 6

  MixedAnyShapeArray:
    attributes:
      array:
        range: integer
        array:
          maximum_number_dimensions: false
          dimensions:
            - alias: max_card
              maximum_cardinality: 5
            - alias: min_card
              minimum_cardinality: 2
            - alias: range_card
              minimum_cardinality: 2
              maximum_cardinality: 5
            - alias: exact_card
              exact_cardinality: 6

  MixedMaxShapeArray:
    attributes:
      array:
        range: integer
        array:
          maximum_number_dimensions: 6
          dimensions:
            - alias: max_card
              maximum_cardinality: 5
            - alias: min_card
              minimum_cardinality: 2
            - alias: range_card
              minimum_cardinality: 2
              maximum_cardinality: 5
            - alias: exact_card
              exact_cardinality: 6

  MixedRangeShapeArray:
    attributes:
      array:
        range: integer
        array:
          minimum_number_dimensions: 5
          maximum_number_dimensions: 7
          dimensions:
            - alias: max_card
              maximum_cardinality: 5
            - alias: min_card
              minimum_cardinality: 2
            - alias: range_card
              minimum_cardinality: 2
              maximum_cardinality: 5
            - alias: exact_card
              exact_cardinality: 6

  MixedExactShapeArray:
    attributes:
      array:
        range: integer
        array:
          exact_number_dimensions: 6
          dimensions:
            - alias: max_card
              maximum_cardinality: 5
            - alias: min_card
              minimum_cardinality: 2
            - alias: range_card
              minimum_cardinality: 2
              maximum_cardinality: 5
            - alias: exact_card
              exact_cardinality: 6
```

## Generated Models
(Pydantic 1)

The following has been formatted with black after generation to make it somewhat readable. it still really sucks to read because it has to go inside-out since I tried to keep injected classes at a minimum and use builtin python or pydantic types when possible since that is the design constraint of the LoL array type. The `numpydantic` classes will all look the same since we will have greater control over the classes. that'll be next PR.

```python
from datetime import datetime, date
from decimal import Decimal
from enum import Enum
import re
from pydantic.version import VERSION as PYDANTIC_VERSION
from pydantic import conlist

if int(PYDANTIC_VERSION[0]) >= 2:
    from pydantic import BaseModel, ConfigDict, Field, field_validator
else:
    from pydantic import BaseModel, Field, validator

from pydantic import conlist
from pydantic import conlist
from typing import (
    Any,
    List,
    Literal,
    Dict,
    Optional,
    Union,
    Generic,
    TypeVar,
    _GenericAlias,
)

metamodel_version = "None"
version = "None"


class WeakRefShimBaseModel(BaseModel):
    __slots__ = "__weakref__"


class ConfiguredBaseModel(
    WeakRefShimBaseModel,
    validate_assignment=True,
    validate_all=True,
    underscore_attrs_are_private=True,
    extra="forbid",
    arbitrary_types_allowed=True,
    use_enum_values=True,
):
    pass


_T = TypeVar("_T")


class AnyShapeArray(Generic[_T]):
    type_: Any

    def __class_getitem__(cls, item):
        alias = type(
            f"AnyShape_{str(item.__name__)}", (AnyShapeArray,), {"type_": item}
        )
        return alias

    @classmethod
    def __get_validators__(cls):
        yield cls.validate

    @classmethod
    def __modify_schema__(cls, field_schema):
        item_type = field_schema["allOf"][0]["type"]
        array_id = f"#any-shape-array-{item_type}"
        field_schema["anyOf"] = [
            {"type": item_type},
            {"type": "array", "items": {"$ref": array_id}},
        ]
        field_schema["$id"] = array_id
        del field_schema["allOf"]

    @classmethod
    def validate(cls, v: Union[List[_T], list]):
        if str(type(v)) == "<class 'numpy.ndarray'>":
            v = v.tolist()

        if not isinstance(v, list):
            raise TypeError(f"Must be a list of lists! got {v}")

        def _validate(_v: Union[List[_T], list]):
            for item in _v:
                if isinstance(item, list):
                    _validate(item)
                elif cls.type_.__name__ != "AnyType":
                    if not isinstance(item, cls.type_):
                        raise TypeError(
                            (
                                f"List items must be list of lists, or the type used in "
                                f"the subscript ({cls.type_}. Got item {item} and outer value {v}"
                            )
                        )
            return _v

        return _validate(v)


class MixedAnyShapeArray(ConfiguredBaseModel):
    array: Optional[
        conlist(
            max_items=5,
            item_type=conlist(
                min_items=2,
                item_type=conlist(
                    min_items=2,
                    max_items=5,
                    item_type=conlist(
                        min_items=6,
                        max_items=6,
                        item_type=Union[AnyShapeArray[int], int],
                    ),
                ),
            ),
        )
    ] = Field(None)


class MixedMaxShapeArray(ConfiguredBaseModel):
    array: Optional[
        conlist(
            max_items=5,
            item_type=conlist(
                min_items=2,
                item_type=conlist(
                    min_items=2,
                    max_items=5,
                    item_type=conlist(
                        min_items=6,
                        max_items=6,
                        item_type=Union[int, List[int], List[List[int]]],
                    ),
                ),
            ),
        )
    ] = Field(None)


class MixedRangeShapeArray(ConfiguredBaseModel):
    array: Optional[
        conlist(
            max_items=5,
            item_type=conlist(
                min_items=2,
                item_type=conlist(
                    min_items=2,
                    max_items=5,
                    item_type=conlist(
                        min_items=6,
                        max_items=6,
                        item_type=Union[
                            List[int], List[List[int]], List[List[List[int]]]
                        ],
                    ),
                ),
            ),
        )
    ] = Field(None)


class MixedExactShapeArray(ConfiguredBaseModel):
    array: Optional[
        conlist(
            max_items=5,
            item_type=conlist(
                min_items=2,
                item_type=conlist(
                    min_items=2,
                    max_items=5,
                    item_type=conlist(
                        min_items=6, max_items=6, item_type=List[List[int]]
                    ),
                ),
            ),
        )
    ] = Field(None)


class LabeledArray(ConfiguredBaseModel):
    array: Optional[
        conlist(
            min_items=2,
            item_type=conlist(
                max_items=5,
                item_type=conlist(
                    min_items=2,
                    max_items=5,
                    item_type=conlist(min_items=6, max_items=6, item_type=int),
                ),
            ),
        )
    ] = Field(None)


class ExactDimensions(ConfiguredBaseModel):
    array: Optional[List[List[List[int]]]] = Field(None)


class MinDimensions(ConfiguredBaseModel):
    array: Optional[List[AnyShapeArray[int]]] = Field(None)


class MaxDimensions(ConfiguredBaseModel):
    array: Optional[
        Union[
            List[int],
            List[List[int]],
            List[List[List[int]]],
            List[List[List[List[int]]]],
            List[List[List[List[List[int]]]]],
        ]
    ] = Field(None)


class RangeDimensions(ConfiguredBaseModel):
    array: Optional[
        Union[
            List[List[int]],
            List[List[List[int]]],
            List[List[List[List[int]]]],
            List[List[List[List[List[int]]]]],
        ]
    ] = Field(None)


class AnyType(ConfiguredBaseModel):
    array: Optional[AnyShapeArray[AnyType]] = Field(None)


class Typed(ConfiguredBaseModel):
    array: Optional[AnyShapeArray[int]] = Field(None)


# Update forward refs
# see https://pydantic-docs.helpmanual.io/usage/postponed_annotations/
MixedAnyShapeArray.update_forward_refs()
MixedMaxShapeArray.update_forward_refs()
MixedRangeShapeArray.update_forward_refs()
MixedExactShapeArray.update_forward_refs()
LabeledArray.update_forward_refs()
ExactDimensions.update_forward_refs()
MinDimensions.update_forward_refs()
MaxDimensions.update_forward_refs()
RangeDimensions.update_forward_refs()
AnyType.update_forward_refs()
Typed.update_forward_refs()

```

Some stuff to note
- I have started the process of cleaning up the pydanticgen with **Results models** - so each kind of thing that is intended to be built should be able to tell the rest of the generation process what it requires to be used (in this case, we want to both import as little as possible and inject as little as possible), but we also don't want to have a ton of squirrely methods for handling every case when different conditions combine or conflict with one another. This is the very first draft of adding results models so it's not pretty yet, but it's a hint of what will be
- I had to retire generating JSON schema for now because it turns out that pydantic 1 has a special `GenericModel` way of dealing with generics that is not correctly documented and likely never will be because Pydantic 1 is almost 9 months deprecated now. I am again pleading to set a pydantic 1 deprecation timeline because handling both has been by far the greatest labor sink in this implementation - we would have had a Pydantic 2-only version weeks ago. https://github.com/linkml/linkml/issues/1925

I doubt the tests pass with pydantic 2 since i've been focusing most of my effort on Pydantic 1 because it is subtlely but actually extraordinarily more difficult to use, and hell even pydantic 1 probably won't pass just yet, but the structure of it is there and that's something.

sneakers-the-rat:
last TODO here is docs, this is ready for review :)

sneakers-the-rat:
Starting on docs, but PR still ready for review. sorry if tests fail for a bit.

added black string formatting method that we can also work into the generator itself later, and since we now have a very neat way of rendering individual classes and whatnot I made some convenience functions for rendering them with the documentation (rather than freezing some fixed strings that inevitably go out of date) in some nice lil tab sets that show the linkml and pydantic versions back to back

linkml/generators/pydanticgen/array.py
```diff
@@ -0,0 +1,396 @@
...
+
+class ListOfListsArray(ArrayRangeGenerator):
+    """
+    Represent arrays as lists of lists!
+
+    TODO: Move all validation of values (eg. anywhere we raise a ValueError) to the ArrayExpression
+    dataclass and out of the generator class
+    """
+
+    REPR = ArrayRepresentation.LIST
```
rly, line 245: I'm curious - what is this for? Currently the classes map 1:1 to representations. Is the point of this variable to allow multiple classes to return the same array representation?

sneakers-the-rat: It was a way of having a firm notion of what representations exist for the purposes of testing, and also used by the generator to know which representations to emit, but it is overly restrictive in that it doesnt allow for additional representations to hook in as plugins. I think it should probably be replaced by introspecting `__subclasses__`

sneakers-the-rat: Also 1:1 class:representation is what we want. If someone wants to patch a representation we should make hooks, but didnt want to try and tackle that in this one PR, but good eye, agreed its not ideal

linkml/generators/pydanticgen/array.py
```diff
@@ -0,0 +1,396 @@
...
+                and ``dtype`` (default: ``False`` )
+
+        """
+        if self.dtype in ("Any", "AnyType"):
+            annotation = "AnyShapeArray"
+        else:
+            annotation = f"AnyShapeArray[{self.dtype}]"
+
+        if with_inner_union:
+            annotation = f"Union[{annotation}, {self.dtype}]"
```
rly, line 299: I'm a bit confused here. This is used below to handle the case where `array.maximum_number_dimensions` is False. Let's say the dtype is integer. The resulting annotation is `Union[AnyShapeArray[integer], integer]` as opposed to just `AnyShapeArray[integer]`. What is the "inner union"? It seems like this line makes it so that scalar integers are allowed, but I am surely missing something.

sneakers-the-rat: AnyShapeArray needs to be at least one level deep of a list when used on its own to validate. When specifying explicit dimensions, one needs to specify all those dimensions also as conlists or lists. This is to handle the inclusive part of the range - "three constrained dimensions or more" is 3 conlists, but if it was just an AnyShapeArray then it is minimum 4 dimensions. Adding the inner Union allows the "3" part of "3 or more"

sneakers-the-rat: Alternatively we could allow AnyShapeArray to validate with a scalar value (0 dimensional array) and coerce it to a length 1 list. It just makes the validation behavior of AnyShapeArray less explicit, but I think that would also be a totally reasonable way to do it if we prefer that

rly:
I started reviewing the PR (great work @sneakers-the-rat !) but need to go now, so I will finish the review by tomorrow evening. Just wanted to let you know so I'm not leaving you hanging.

sneakers-the-rat:
Take ur time, its a bazillion LoC on a Friday evening. And I havent even written the docs yet anyway lol :)

sneakers-the-rat:
lets put a live array representation generator in the docs page lmao https://github.com/orgs/linkml/discussions/1963

linkml/generators/pydanticgen/array.py
rly: Ah, I understand and also see the helper function `get_generator` on line 209 above now. I think this is fine for now.

docs/_includes/arrays/mixed_shape.yaml
```diff
@@ -0,0 +1,75 @@
...
+          dimensions:
+            - alias: max_card
+              maximum_cardinality: 5
+            - alias: min_card
+              minimum_cardinality: 2
+            - alias: range_card
+              minimum_cardinality: 2
+              maximum_cardinality: 5
+            - alias: exact_card
+              exact_cardinality: 6
```
rly, line 75: This is minor and not totally related to this PR, but I'm trying to imagine the use case for `minimum_number_dimensions < len(dimensions)`. 

I wonder when someone would want to specify a minimum number of dimensions but not even provide an alias hinting to the user what those dimensions are supposed to represent. 

Example 1: Let's specify an array with at least 3 dimensions and only the first one has a label. We could write:

```yaml
array:
  minimum_number_dimensions: 3
  maximum_number_dimensions: False
  dimensions:
    - alias: time
```

The user must recognize that `dimensions` does not provide information on neither the 2nd and 3rd required dimensions, nor the dimensions beyond that. 

Example 2: Example 1 is equivalent to this schema:

```yaml
array:
  minimum_number_dimensions: 3
  maximum_number_dimensions: False
  dimensions:
    - alias: time
    - {}
    - {}
```

The dimensions beyond the first 3 still have no information here. You could still provide information on those dimensions by adding more list items to `dimensions`. 

Example 2 is less compact but it might be more intuitive. It looks a little ugly with the `{}` but I think we might want to encourage schema writers to provide at least some information on the required dimensions. I don't see why they could not, and if they do not want to, then they have to explicitly say there is nothing with `{}`. 

But I've looked at this too long to be a good judge of what is and is not intuitive. So I guess my question is whether we want to allow the "shortcut" schema pattern shown in Example 1. What do you think @sneakers-the-rat @cmungall ? 

(In NWB, you can define the allowed shapes without providing any information about the shapes, but it is generally recommended that you label the dimensions. In some cases, it might feel unnecessary, e.g., a 1D array of timestamps has the first dimension labeled as "times" or "num_times".)

linkml/generators/pydanticgen/array.py
```diff
@@ -0,0 +1,396 @@
...
+            return self.any_shape()
+        elif array.maximum_number_dimensions:
+            min_dims = array.minimum_number_dimensions if array.minimum_number_dimensions is not None else 1
+            annotations = [
+                self._list_of_lists(i, self.dtype) for i in range(min_dims, array.maximum_number_dimensions + 1)
+            ]
+            # TODO: Format this nicely!
+            return SlotResult(annotation="Union[" + ", ".join(annotations) + "]")
+        else:
+            return SlotResult(
```
rly, line 317:
```suggestion
            # min specified with no max
            # e.g., if min = 3, annotation = List[List[AnyShapeArray[dtype]]]
            return SlotResult(
```

Just adding comments and examples to help me and future code readers

linkml/generators/pydanticgen/array.py
```diff
@@ -0,0 +1,396 @@
...
+
+    def anonymous_shape(self, array: ArrayExpression) -> SlotResult:
+        if array.exact_number_dimensions:
+            return SlotResult(annotation=self._list_of_lists(array.exact_number_dimensions, self.dtype))
+        elif not array.maximum_number_dimensions and (
+            array.minimum_number_dimensions is None or array.minimum_number_dimensions == 1
+        ):
+            return self.any_shape()
+        elif array.maximum_number_dimensions:
+            min_dims = array.minimum_number_dimensions if array.minimum_number_dimensions is not None else 1
```
rly, line 310:
```suggestion
            # e.g., if min = 2, max = 3, annotation = Union[List[List[dtype]], List[List[List[dtype]]]]
            min_dims = array.minimum_number_dimensions if array.minimum_number_dimensions is not None else 1
```

linkml/generators/pydanticgen/array.py
rly: Adding examples at the risk of the example becoming outdated and inconsistent, but I think this one will not...

linkml/generators/pydanticgen/array.py
```diff
@@ -0,0 +1,396 @@
...
+    def labeled_shape(self, array: ArrayExpression) -> SlotResult:
+        """
+        Constrained shapes using :func:`pydantic.conlist`
+
+        TODO:
+        - preservation of aliases
+        - (what other metadata is allowable on labeled dimensions?)
+        """
+        # generate dimensions from inside out and then format
+        range = self.dtype
```
rly, line 332:
```suggestion
        # e.g., if dimensions = [{min_card: 3}, {min_card: 2}], 
        # annotation = conlist(min_length=3, item_type=conlist(min_length=2, item_type=dtype))
        range = self.dtype
```

linkml/generators/pydanticgen/array.py
```diff
@@ -0,0 +1,396 @@
...
+                    "if exact_number_dimensions is provided, it must be greater than the parameterized dimensions"
+                )
+
+        elif array.maximum_number_dimensions is not None and not array.maximum_number_dimensions:
+            # unlimited n dimensions, so innermost is AnyShape with dtype
+            res = self.any_shape(with_inner_union=True)
+
+            if array.minimum_number_dimensions and array.minimum_number_dimensions > len(array.dimensions):
+                # some minimum anonymous dimensions but unlimited max dimensions
+                res += self._list_of_lists(array.minimum_number_dimensions - len(array.dimensions), res.annotation)
```
rly, line 361:
```suggestion
                # e.g., if min = 3, len(dim) = 1, then res.annotation = List[List[Union[AnyShapeArray[dtype], dtype]]]
                res += self._list_of_lists(array.minimum_number_dimensions - len(array.dimensions), res.annotation)
```

linkml/generators/pydanticgen/array.py
```diff
@@ -0,0 +1,396 @@
...
+        for dimension in reversed(array.dimensions):
+            range = self._labeled_dimension(dimension, range).annotation
+
+        return SlotResult(annotation=range, imports=_ConListImports)
+
+    def mixed_shape(self, array: ArrayExpression) -> SlotResult:
+        """
+        Mixture of labeled dimensions with a max or min (or both) shape for anonymous dimensions
+        """
+        if array.exact_number_dimensions is not None:
```
rly, line 342:
```suggestion
        # first process any unlabeled dimensions which must be the innermost level of the annotation,
        # then wrap that with labeled dimensions
        if array.exact_number_dimensions is not None:
```

linkml/generators/pydanticgen/array.py
```diff
@@ -0,0 +1,396 @@
...
+            initial_min = array.minimum_number_dimensions if array.minimum_number_dimensions is not None else 0
+            dmin = max(len(array.dimensions), initial_min) - len(array.dimensions)
+            dmax = array.maximum_number_dimensions - len(array.dimensions)
+
+            res = self.anonymous_shape(ArrayExpression(minimum_number_dimensions=dmin, maximum_number_dimensions=dmax))
+        else:
+            raise ValueError("Unsupported array specification! this is almost certainly a bug!")
+
+        # Wrap inner dimension with labeled dimension
+        for dim in reversed(array.dimensions):
```
rly, line 381:
```suggestion
        # e.g., if dimensions = [{min_card: 3}, {min_card: 2}] and res.annotation = List[Union[AnyShapeArray[dtype], dtype]] (min 3 dims, no max dims)
        # then the final annotation = conlist(
        #     min_length=3, 
        #     item_type=conlist(
        #         min_length=2, 
        #         item_type=List[Union[AnyShapeArray[dtype], dtype]]
        #     )
        # )
        for dim in reversed(array.dimensions):
```

linkml/generators/pydanticgen/array.py
```diff
@@ -0,0 +1,396 @@
...
+        else:
+            annotation = f"AnyShapeArray[{self.dtype}]"
+
+        if with_inner_union:
+            annotation = f"Union[{annotation}, {self.dtype}]"
+        return SlotResult(annotation=annotation, injected_classes=_AnyShapeArrayInjects, imports=_AnyShapeArrayImports)
+
+    def anonymous_shape(self, array: ArrayExpression) -> SlotResult:
+        if array.exact_number_dimensions:
+            return SlotResult(annotation=self._list_of_lists(array.exact_number_dimensions, self.dtype))
```
rly, line 304:
```suggestion
        if array.exact_number_dimensions or (
            array.minimum_number_dimensions and array.maximum_number_dimensions and 
            array.minimum_number_dimensions == array.maximum_number_dimensions
        ):
            return SlotResult(annotation=self._list_of_lists(array.exact_number_dimensions, self.dtype))
```

Since the case min==max should be the same as if exact is specified, it might be nice to combine these conditions and avoid going into the below logic. This would also avoid creating a Union of annotations of length 1.

linkml/generators/pydanticgen/array.py
```diff
@@ -0,0 +1,396 @@
...
+        Mixture of labeled dimensions with a max or min (or both) shape for anonymous dimensions
+        """
+        if array.exact_number_dimensions is not None:
+            if array.exact_number_dimensions > len(array.dimensions):
+                res = SlotResult(
+                    annotation=self._list_of_lists(array.exact_number_dimensions - len(array.dimensions), self.dtype)
+                )
+            elif array.exact_number_dimensions == len(array.dimensions):
+                # equivalent to labeled shape
+                return self.labeled_shape(array)
```
rly, line 349:
```suggestion
        if array.exact_number_dimensions or (
            array.minimum_number_dimensions and array.maximum_number_dimensions and 
            array.minimum_number_dimensions == array.maximum_number_dimensions
        ):
            exact_dims = array.exact_number_dimensions or array.minimum_number_dimensions
            if exact_dims > len(array.dimensions):
                res = SlotResult(
                    annotation=self._list_of_lists(exact_dims - len(array.dimensions), self.dtype)
                )
            elif exact_dims == len(array.dimensions):
                # equivalent to labeled shape
                return self.labeled_shape(array)
```

Again to combine the cases where exact dims are provided and where min and max are provided and are equal.

linkml/generators/pydanticgen/array.py
```diff
@@ -0,0 +1,396 @@
...
+                    "either use exact_number_dimensions > len(dimensions) for extra anonymous dimensions or set "
+                    "maximum_number_dimensions explicitly to False for unbounded dimensions"
+                )
+            )
+        elif array.maximum_number_dimensions:
+            initial_min = array.minimum_number_dimensions if array.minimum_number_dimensions is not None else 0
+            dmin = max(len(array.dimensions), initial_min) - len(array.dimensions)
+            dmax = array.maximum_number_dimensions - len(array.dimensions)
+
+            res = self.anonymous_shape(ArrayExpression(minimum_number_dimensions=dmin, maximum_number_dimensions=dmax))
```
rly, line 376: I believe it is possible here for `dmax = 0`, i.e., `array.maximum_number_dimensions` is specified and equals the length of the labeled dimensions. This results in an `ArrayExpression` where `maximum_number_dimensions==0` which is equal to False and I think it is interpreted as such (unlimited number of dimensions) in `self.anonymous_shape`. Maybe this case is caught above though.

rly:
```suggestion
            # dmin can be 0. if dmax = 1, then res.annotation = Union[dtype, List[dtype]]
            res = self.anonymous_shape(ArrayExpression(minimum_number_dimensions=dmin, maximum_number_dimensions=dmax))
```

Note that it is also possible here for `dmin = 0` but tracing through the code, I think that is OK and works out well.

linkml/generators/pydanticgen/array.py
rly, line 361: Does `+=` here work? `self._list_of_lists` returns a string, and `__add__` expects a `SlotResult`.

linkml/generators/pydanticgen/build.py
```diff
@@ -0,0 +1,44 @@
...
+    annotation: str
+    """The type annotation used in the generated model"""
+    field_extras: Optional[dict] = None
+    """Additional metadata for this slot to be held in the Field object"""
+
+    def __add__(self, other: "SlotResult"):
+        res = super(SlotResult, self).__add__(other)
+        # Replace with other's annotation
+        res.annotation = other.annotation
+        return res
```
rly, line 44: The replacement of `annotation` when using `+=` is a bit less than intuitive to me. It looks like it is used in only two places above: lines 361 and 382 of `array.py`. I get that the imports and injected classes are additive and that makes for cleaner code in `self._labeled_dimension` and elsewhere, but when seeing `+=`, I had the incorrect intuition that `annotation` was also additive, like it is a conlist, and `+=` would add a constraint to the conlist. Instead, it is used like `res += function(arg, dtype=res.annotation)` which nests/wraps the previous annotation and adds any relevant imports and injected classes needed by `function`. 

It makes sense now after tracing the code, and it is nicely compact, but I wonder if it could and should be refactored to be more intuitive, at the cost of more lines of code. Instead of using `+=`, we could explicitly create a new `SlotResult`. For example, instead of lines 381-382:

```python
for dim in reversed(array.dimensions):
    res += self._labeled_dimension(dim, dtype=res.annotation)
```
(note: this might add `_ConListImports` N times to the resulting `SlotResult`?)

Use:
```python
ann = res.annotation
for dim in reversed(array.dimensions):
    ann = self._labeled_dimension(dim, ann)
res = SlotResult(
    annotation=ann,
    injected_classes=res.injected_classes,
    imports=res.imports.copy() + _ConListImports,
)
```

Now that I write it out, I see that this is pretty verbose. I'm not sure if it is better, but it is more clear to me what's going on. I've just been scarred recently by unintuitive `+=` behavior elsewhere (see below)...

Note that on lines 317-322, we have something similar:
```python
return SlotResult(
    annotation=self._list_of_lists(array.minimum_number_dimensions - 1, self.any_shape().annotation),
    injected_classes=_AnyShapeArrayInjects,
    imports=_AnyShapeArrayImports,
)
```

Fun exercise:
```python
a = np.array([1, 2, 3])
b = a[1:3]
b += 2
print(b)  # [4 5]
print(a) # [1 4 5] !!!
# this happens because numpy slices are views of the array and `+=` acts in place
# compare that with:
a = np.array([1, 2, 3])
b = a[1:3]
b = b + 2
print(b)  # [4 5]
print(a) # [1 2 3]
# one might think that b += 2 is the same as b = b + 2, but no, b + 2 created a copy
```

linkml/generators/pydanticgen/pydanticgen.py
```diff
@@ -491,7 +501,28 @@ def _template_environment(self) -> Environment:
...
+
+        if len(array_reps) == 0:
+            raise ValueError("No array representation generated, but one was requested!")
+
+        return array_reps
+        # elif len(array_reps) == 1:
+        #     return array_reps[0]
+        # else:
+        #     # TODO: lets format this nicely :)
+        #     return f"Union[{','.join(array_reps)}]"
```
rly, line 523:
```suggestion
```

rly:
I reviewed everything but the tests. This is a lot of great work! A few minor comments / discussion above. Happy to jump on a call and discuss if that helps.

linkml/generators/pydanticgen/array.py
rly: That makes sense. I understand now. Thanks!

docs/_includes/arrays/mixed_shape.yaml
sneakers-the-rat: Assuming you mean `minimum_number_dimensions > length(dimensions)` at the top by context :)

In this case, what i was thinking about was providing a complete and predictable behavior based on the syntactic tools provided. We could make a rule like "minimum number dimensions must be equal to number of specified dimensions" but that would be an additional constraint on top of the basic rules of the syntax that would require additional documentation and special validation logic. 

The implementation burden is relatively low, and imo a simple syntax that does the things you expect it to is good to do, even if we can't think of times when it might be useful! I don't really like enforcing or nudging design patterns at the level of a syntax, that takes us out of the realm of 'neutral expressive tool that can accomodate many formats' and more into the realm of 'opinionated data format,' in this case both of those should be valid (and equivalent) - and probably should be added as test cases since they are good examples of edge cases.

linkml/generators/pydanticgen/array.py
sneakers-the-rat: i like comments :) and i think that a lot of these will be made clearer when i finish adding the docs, which will be a lot of "this syntax generates that type annotation" examples.

linkml/generators/pydanticgen/array.py
sneakers-the-rat: ah, good catch

sneakers-the-rat:
Merging some stuff from review and adding additional TODO items before merge in the top comment :)

linkml/generators/pydanticgen/array.py
sneakers-the-rat: will add to test cases and evaluate monday

linkml/generators/pydanticgen/build.py
sneakers-the-rat: Fair point!

So the expectation with these BuildResults is that they are going to be part of a potentially unknown processing chain in the rest of the generator. Each "build stage" should receive the existing BuildResult and return a modified one. Replacing the annotation is from the expectation that between stages, the build method wouldn't know about the internal structure of the generated annotation. This is sort of the cost of doing code generation - at some point it becomes a string unless you build a full-on ast. This first draft is not intended to be final, but just a starting place, so that can of course be varied (eg. it probably would be good to carry the `SlotDefinition` along with the result to, etc.). 

One thing's for sure - this needs to be documented better than i have it now, regardless of the behavior.

> when seeing +=, I had the incorrect intuition that annotation was also additive, like it is a conlist, and += would add a constraint to the conlist.

Also a fair assumption - this would be awesome, but super tricky to implement in a general way. I think the intuition for this will be clearer when/if used in the rest of the generator - the annotation is specifically the string that is going to be rendered in the final model, and each of the build stages can't know the programmatic format of the annotation. 

The main thing that the results are for is to contain all the side effects that happen throughout the build process - specifically knowing which imports and etc. need to happen to make each of the different slots possible to use without needing to inject everything every time. So I think we should avoid doing a ton of modification of the build result in the middle of the method because that sort of kills the generality of that as a container - eg. we have some other metamodel addition later that adds a field to the buildresult, now we need to go through and modify all the places where we made custom addition logic instead of just updating the `__add__` method.

> scarred recently by unintuitive += behavior elsewhere

youch! yes that is why this method always returns a new object instead of modifying in place.

cmungall (approved):

linkml/generators/pydanticgen/build.py
sneakers-the-rat: swapped `__add__` to `merge()`

linkml/generators/pydanticgen/array.py
sneakers-the-rat: was a bug, added test case and fixed :)

linkml/generators/pydanticgen/array.py
sneakers-the-rat: u can commit urself or make another suggestion comment and i'll merge it so it gets attributed right :)

docs/_includes/arrays/mixed_shape.yaml
sneakers-the-rat: does this deserve to be in 'linkml arrays, extended docs universe', a separate issue, or keep this open on the issue for history's sake?

sneakers-the-rat:
about to work on docs, and question: what do we want to call the two types of arrays? I had tentatively called them "NDArrays" and "Indexed arrays," but how were we referring to those?

rly:
Do you mean the arrays with `array_data_mapping` vs without? If so, I would just say "NDArray"/"N-dimensional array" and "Labeled array" (to go with Xarray nomenclature) Or do you mean something else?

sneakers-the-rat:
Ok ya thats what I meant.

Fine with me. If we call those "labeled arrays" then we probably want to call what I had termed "labeled arrays" something different, ie. The ndarrays with parameterized/named/etc. Dimensions

rly:
Naming is hard! I know you are starting the docs now so this suggestion may be a bit late, but what do you think about this:

> If we call those "labeled arrays" then we probably want to call what I had termed "labeled arrays" something different, ie. The ndarrays with parameterized/named/etc. Dimensions

I would prefer not to call ndarrays with parameterized dimensions "labeled arrays" because the spec can be more than just a label/name. Also I was confused with what "anonymous" meant initially (it's still a little confusing), but I see that "anonymous" makes sense when the alternative is "labeled". 

What do you think about calling the four types:

1. Any shape, no dimensions information = Any shape array
2. Min/max number of dimensions specified, no dimensions information = Unspecified dims array
3. No min/max number of dimensions specified, dimensions information (ndim = len(dimensions)) = Specified dims array
4. Min/max number of dimensions specified, dimensions information = Complex specified dims array

And then combine Types 3 and 4 and just call them both a "specified dims array". Type 3 ("labeled array") is a special case of Type 4 and might not need its own name? It looks decently straightforward to merge the code around Type 3 into Type 4. I like this because to me, the naming more explicitly describes the specification.

@cmungall any thoughts on naming?

> Edit: for the purposes of this PR im just putting labeled arrays as "coming soon"

Sounds good

rly:
I am happy to go through the code and adjust the naming however we want. Let me know if that would be useful or I'd just get in the way with merge conflicts.

sneakers-the-rat:
> Any shape, no dimensions information = Any shape array

agreed here :)

>  Min/max number of dimensions specified, no dimensions information = Unspecified dims array

this is a bit confusing to me, since we do specify the dimensions. It would be weird to say "an unspecified dimensions array is one in which only the number of dimensions are specified." 

The core of it is "numerically specified dimensions" but that's also awkward... 

How about this, since each of these types only has to do with the shape, what if we dropped the "arrays" qualifier and say "Numeric shape" ?

> No min/max number of dimensions specified, dimensions information (ndim = len(dimensions)) = Specified dims array

In re: above, how about "Parameterized shape" ?

> Min/max number of dimensions specified, dimensions information = Complex specified dims array

"Complex" shape works for me. I sort of like "mixed" because it's directly referring to the mixture of the previous types, but complex is close enough for me

sneakers-the-rat:
Alright that's docs. Ignore the naming for now, i'll change that all once we agree on the final names. I also wasn't sure what version this would be deployed in, so that's just a guess for now.

Here's a single-page render of it in case you don't have a sphinx env handy - 

[arrays.html.zip](https://github.com/linkml/linkml/files/14662479/arrays.html.zip)

lmk what else shoudl be in there/changed/etc.

sneakers-the-rat:
Boop boop may the tests pass and never unpass

rly:
I see your point about unspecified/specified. These categories might be too difficult to communicate in few words. 

> How about this, since each of these types only has to do with the shape, what if we dropped the "arrays" qualifier and say "Numeric shape" ?

I like having the arrays qualifier. Numeric shape array doesn't capture the min/max range of number of dimensions to me. How about: "Case 2: Array with bounded number of dimensions and unparameterized dimensions" or "Unparameterized, bounded number of dimensions array"? It's not punchy, but I like that it is explicit, and we can refer to it and the other cases by number. Shorthand in code could be "unparam_bounded_ndims". I can't imagine a single use case for this type, so I think it's OK to have a long name, or even remove it as a category. 

> > No min/max number of dimensions specified, dimensions information (ndim = len(dimensions)) = Specified dims array
> 
> In re: above, how about "Parameterized shape" ?

For this case, I think it's more the dimensions that are being parameterized. "Case 3: Array with parameterized dimensions"?

> > Min/max number of dimensions specified, dimensions information = Complex specified dims array
> 
> "Complex" shape works for me. I sort of like "mixed" because it's directly referring to the mixture of the previous types, but complex is close enough for me

"Case 4: Array with bounded number of dimensions array and parameterized dimensions"?

sneakers-the-rat:
> bounded number of unparameterized dimensions

what about "bounded dimensions array" and "parameterized dimensions array" ? 

> I can't imagine a single use case for this type

it's the only way to do optional dimensions (i think?). some examples might be in a parent class where one sets "2-4 dimensional array" where the children are an image, grayscale video, and multichannel video, and where the axes might be in a different order (x/y, y/x, etc.), or values in a vector space whose size and dimensions are unknown but need to be in a range for downstream usage, or making a schematization of an array library that can only handle arrays up to a certain dimensionality, etc.

> Case 3: Array with parameterized dimensions

yes yes you're right my b

> Case 4: Array with bounded number of parameterized dimensions

complex grew on me! i think it would be good to have terminology that can be matched in both the docs and the code, and asking for someone to have to read and write `bounded_dimensions_array`, `parameterized_dimensions_array`, `bounded_parameterized_dimensions_array` seems unkind ;). I think if we treat these 'titles' as slugs where the first sentence of the 'complex array' section is 'a `complex` NDarray has a bounded number of parameterized dimensions that gets us both clarity and a programmer friendly shorthand?

rly:
> what about "bounded dimensions array" and "parameterized dimensions array" ?

Yeah, that would work. We should spell it out where we have space in the docs but these would be fine shorthands.

> 
> > I can't imagine a single use case for this type
> 
> it's the only way to do optional dimensions (i think?). some examples might be in a parent class where one sets "2-4 dimensional array" where the children are an image, grayscale video, and multichannel video, and where the axes might be in a different order (x/y, y/x, etc.), or values in a vector space whose size and dimensions are unknown but need to be in a range for downstream usage, or making a schematization of an array library that can only handle arrays up to a certain dimensionality, etc.

I see. Good examples. Re: x/y, y/x. I was thinking that would be poor schema design and not useful downstream, but I guess you could have another field that specifies user-defined aliases for the dimensions. OK.

> complex grew on me! i think it would be good to have terminology that can be matched in both the docs and the code, and asking for someone to have to read and write `bounded_dimensions_array`, `parameterized_dimensions_array`, `bounded_parameterized_dimensions_array` seems unkind ;). I think if we treat these 'titles' as slugs where the first sentence of the 'complex array' section is 'a `complex` NDarray has a bounded number of parameterized dimensions' that gets us both clarity and a programmer friendly shorthand?

Sounds good to me - both "complex array" and "complex dimensions array"

Thanks for hashing out names with me!

sneakers-the-rat:
> I was thinking that would be poor schema design and not useful downstream

probably, but who are we to judge? <3

> Thanks for hashing out names with me!

always, how else would we figure out names that make sense outside our single-player brains!?!?!? 

will update PR tomorrow

sneakers-the-rat:
ok renamed, check my work! if docs looks good then i think we're good to go

docs/_includes/arrays/mixed_shape.yaml
rly: I think this is fine to resolve. Thanks for the response! Being a neutral expressive tool as opposed to opinionated data format makes sense here. LinkML could have a separate layer of best practices that suggests one of the above formats over the other for clarity of those reading the raw schema. I added an issue about that here https://github.com/linkml/linkml/issues/2025

.github/workflows/doc_pages.yaml
```diff
@@ -23,7 +23,10 @@ jobs:
           version: 1.3.2
 
       - name: install
-        run: poetry install
+        run: poetry install -E black
```
rly, line 26: Is `black` needed for the docs?

.github/workflows/docs-test.yaml
```diff
@@ -34,14 +34,17 @@ jobs:
         uses: actions/cache@v3
         with:
           path: .venv
-          key: venv-docs-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}
+          key: venv-docs-dummy2-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}
```
rly, line 37:
```suggestion
          key: venv-docs-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}
```
Do we still need this change?

tests/test_generators/input/arrays/complex_dimensions.yaml
```diff
@@ -0,0 +1,136 @@
...
+            - alias: range_card
+              minimum_cardinality: 2
+              maximum_cardinality: 5
+            - alias: exact_card
+              exact_cardinality: 6
+
+  ComplexImplicitExactShapeArray:
+    description: |
+      When minimum and maximum number of dimensions are equal, the result should be the same as specifying exact dimensions
+        attributes:
```
rly, line 120:
```suggestion
```

.github/workflows/doc_pages.yaml
sneakers-the-rat: it's highly desirable for the docs! 

i think it's in general a good move to not try and nail the formatting in the templates and generators, and just to get them into syntactically correct python and let black handle the rest.

so if `black` is present, we use it - https://github.com/linkml/linkml/blob/dd1ce611f83b00484929abdec157c28665d2fe7e/linkml/generators/pydanticgen/template.py#L85

but it also works without it. 

tested here: https://github.com/linkml/linkml/blob/pydanticgen_array_reps/tests/test_generators/test_pydanticgen.py#L1602-1691

(test was actually a pain in the ass to write, it's actually pretty hard to mock a module not being present once everything has already imported it, but i [made it work](https://github.com/linkml/linkml/blob/dd1ce611f83b00484929abdec157c28665d2fe7e/tests/conftest.py#L211-L252) lol)

.github/workflows/docs-test.yaml
sneakers-the-rat: nope :)

linkml/generators/pydanticgen/build.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+    The result of any build phase for any linkML object
+
+    BuildResults are merged in the serialization process, and are used
+    to keep track of not only the particular representation
+    of the thing in question, but any "side effects" that need to happen
+    elsewhere in the generation process (like adding imports, injecting classes, etc.)
+    """
+
+    # FIXME: PLACEHOLDER TYPES PENDING MERGE OF OTHER PULL REQUESTS
```
rly, line 20:
```suggestion
```
can we remove this now?

linkml/generators/pydanticgen/pydanticgen.py
```diff
@@ -491,7 +505,23 @@ def _template_environment(self) -> Environment:
...
+    def get_array_representations_range(self, slot: SlotDefinition, range: str) -> List[SlotResult]:
+        """
+        Generate the python range for array representations
+        """
+        array_reps = []
+        for repr in self.array_representations:
+            generator = ArrayRangeGenerator.get_generator(repr)
+            result = generator(slot.array, range, self.pydantic_version).make()
+            array_reps.append(result)
+            # TODO: Handle imports, injected classes
```
rly, line 517:
```suggestion
```
I believe this is handled downstream

linkml/generators/pydanticgen/template.py
```diff
@@ -454,6 +523,17 @@ def __add__(self, other: Import) -> "Imports":
...
+        # SPECIAL CASE - __future__ annotations must happen at the top of a file
+        # TODO: Make a sort method
+        futures = []
+        for item in imports[-2:]:
+            if item.module == "__future__":
+                futures.append(item)
+        if len(futures) > 0:
+            imports.remove(futures[0])
+            imports.insert(0, futures[0])
+
```
rly, line 536: Could you describe what this is doing more in the comments? It looks like it looks searches the last two imports for imports from `__future__` and if any are found, it moves the first import from `__future__` to the beginning of the imports list. But why search only the last two imports and why move only the first of those?

linkml/generators/pydanticgen/array.py
```diff
@@ -0,0 +1,455 @@
...
+
+        return SlotResult(annotation=range, imports=_ConListImports)
+
+    def complex_dimensions(self, array: ArrayExpression) -> SlotResult:
+        """
+        Mixture of parameterized dimensions with a max or min (or both) shape for anonymous dimensions.
+
+        A mixture of ``List`` , :class:`.conlist` , and :class:`.AnyShapeArray` .
+        """
+        if array.exact_number_dimensions or (
```
rly, line 382:
```suggestion
        # first process any unlabeled dimensions which must be the innermost level of the annotation,
        # then wrap that with labeled dimensions
        if array.exact_number_dimensions or (
```

docs/schemas/arrays.md
```diff
@@ -0,0 +1,533 @@
...
+
+(array-forms)=
+### Shape Forms
+
+The combinations of the different {class}`.ArrayExpression` properties imply four NDArray forms:
+
+- [Any Shape](AnyShapeArrays) - Arrays without limits to their shape
+- [Bounded Shape](BoundedShape) - Arrays with constraints on the number of dimensions without further parameterization
+- [Parameterized Shape](ParameterizedShape) - Arrays with parameterized dimensions
+- [Complex Shape](ComplexShape) - Arrays with both numeric constraints and parameterized dimensions
```
rly, line 262:
```suggestion
- [Complex Shape](ComplexShape) - Arrays with both constraints on the number of dimensions and parameterized dimensions
```
I think this phrasing is clearer

linkml/generators/pydanticgen/build.py
sneakers-the-rat: yessss thank you

linkml/generators/pydanticgen/pydanticgen.py
sneakers-the-rat: correct!

docs/schemas/arrays.md
```diff
@@ -0,0 +1,533 @@
...
+::::
+::::{tab-item} numpydantic
+:::{admonition} Coming Soon
+Direct support for array libraries like numpy, hdf5, dask, zarr, and xarray with pydantic validation <3
+::: 
+::::
+:::::
+
+The resulting pydantic models use a special {class}`AnyShapeArray` class injected by pydanticgen's 
+{mod}`~linkml.generators.pydanticgen.template` system when using the List of List representation
```
rly, line 304:
```suggestion
{mod}`~linkml.generators.pydanticgen.template` system when using the List of List (LoL) representation 
```

linkml/generators/pydanticgen/template.py
sneakers-the-rat: ah right. so there can only ever be 2 imports added at this stage in the method - usually there will only be one, but we can get two depending on if we are getting a conditional import and already have one.

`__future__` imports shouldn't be a conditional import, though, and should always merge down into a single import object, but it might be in the second to last or last position.

now that i'm looking at this though, i agree it's hideous, and i should just use `sorted` with a lambda like

`imports = sorted(imports, lambda i: i.module == '__future__', reverse=True)`

i'll do that after i merge your comments

docs/schemas/arrays.md
sneakers-the-rat: agreed

linkml/generators/pydanticgen/template.py
sneakers-the-rat: here ya go. "a single line takes 30 lines to test" should be some kind of folk idiom or something.

https://github.com/linkml/linkml/pull/1887/commits/df0c26714a540cfa4d83efd11b22feb21bf16755

rly: Thanks! Looks great and much more understandable

docs/schemas/arrays.md
```diff
@@ -0,0 +1,533 @@
...
+---
+sch = schemas / 'complex_shape.yaml'
+render_comparison(sch, 'ComplexAnyShapeArray')
+```
+
+## Generators
+
+### Support
+
+At releast, only pydanticgen supports arrays, but arrays will be implemented gradually for the rest of the generators.
```
rly, line 485:
```suggestion
At release, only pydanticgen supports arrays, but arrays will be implemented gradually for the rest of the generators.
```

docs/faq/tricky_choices.md
```diff
@@ -0,0 +1,12 @@
...
+
+Inspired by IPLD's [tricky choices](https://ipld.io/design/tricky-choices/) section,
+this page describes some of the challenging decisions that could have gone multiple ways,
+and the considerations we weighed to arrive at the form we did. 
+
+## Arrays
+
+```{admonition} Hey
+Document me plz
+```
```
rly, line 12: @sneakers-the-rat Do you want to document our thinking for arrays in this PR? I can help write this but not until the end of the week, so maybe a separate PR would be best.

linkml/generators/pydanticgen/template.py
sneakers-the-rat: I always think I am decreasing the proportion of lines I write that I later look back on and can say nothing but "what the fuck," but moments like this remind me its more of an asymptote to a long term average than trending to zero

rly:
Looks like there is a poetry lockfile issue. Aside from that and filling in `tricky_choices.md` (or replacing it with something like "TODO"), I think this is a good to go. 

Amazing work @sneakers-the-rat ! Very thorough. If I could change the author of this PR to you to give you full credit, I would. 

I'll leave the final review and merge to @cmungall

docs/faq/tricky_choices.md
sneakers-the-rat: Yeah lets do it later, I also wrote up my thoughts in our lab slack during the workshop bc I knew I would forget by the time it came to write the docs. Im itchin to get this merged so I can start on the next leg (no hurry tho!!!!!)

sneakers-the-rat:
I have been dealing with lockfile stuff all day, whats one more ;) and yes maybe TODO is a bit more professional.

cmungall (approved):

sneakers-the-rat:
hell ya. onto the next representation

Issue #1897, opened 2024-02-10
Issues with `any_of` in gen-jsonld-context


djarecka:
**Describe the bug**
A clear and concise description of what the bug is (including the error message that is printed, if any)

there are issues with context generator when `any_of` with `range` is used. In the context, in the slots description `"@type": "@id",` is missing.

**To reproduce**

Try a simple model:
```
name: reproschema
id: http://example.org/example
imports:
- linkml:types
prefixes:
  reproschema: http://schema.repronim.org/
  schema: http://schema.org/
  linkml: https://w3id.org/linkml/
default_range: string

classes:
  CreativeWork:
    class_uri: schema:CreativeWork

slots:
  isAbout:
    slot_uri: reproschema:isAbout
    any_of:
      - range: CreativeWork
```

when running `gen-jsonld-context simple_model.yaml` I get:

```
{
   "comments": {
      "description": "Auto generated by LinkML jsonld context generator",
      "generation_date": "2024-02-09T19:59:10",
      "source": "reproschema_testshort.yaml"
   },
   "@context": {
      "linkml": "https://w3id.org/linkml/",
      "reproschema": "http://schema.repronim.org/",
      "schema": "http://schema.org/",
      "@vocab": "http://example.org/example/",
      "isAbout": {
         "@id": "reproschema:isAbout"
      },
      "CreativeWork": {
         "@id": "schema:CreativeWork"
      }
   }
}
```

**Expected behavior**

I would expect for `isAbout`:
```
      "isAbout": {
         "@type": "@id",
         "@id": "reproschema:isAbout"
      },
```

And I get this if in the model is simply `range: CreativeWork`  instead of `any_of`

**Screenshots**

**About your computer (if applicable, please complete the following information):**

OSX

**Additional context**

PR #1898, opened 2024-02-10
[wip] fixing context generator for `any_of` with `range`

djarecka:
fixies #1897 

I've tried to fix the context generator, but I'm actually not sure if my approach is correct. Should `slot_def["@type"] = "@id"` be added if any of the ranges from `any_of` are in `self.schema.classes`?

### DIFF
--- a/linkml/generators/jsonldcontextgen.py
+++ b/linkml/generators/jsonldcontextgen.py
@@ -150,7 +150,7 @@ class ContextGenerator(Generator):
         else:
             slot_def = {}
             if not slot.usage_slot_name:
-                any_of_ranges = [any_of_el["range"] for any_of_el in slot.any_of]
+                any_of_ranges = [any_of_el.range for any_of_el in slot.any_of]
                 if slot.range in self.schema.classes or any(rng in self.schema.classes for rng in any_of_ranges):
                     slot_def["@type"] = "@id"
                 elif slot.range in self.schema.enums:
--- a/linkml/generators/jsonldcontextgen.py
+++ b/linkml/generators/jsonldcontextgen.py
@@ -150,7 +150,8 @@ class ContextGenerator(Generator):
         else:
             slot_def = {}
             if not slot.usage_slot_name:
-                if slot.range in self.schema.classes:
+                any_of_ranges = [any_of_el["range"] for any_of_el in slot.any_of]
+                if slot.range in self.schema.classes or any(rng in self.schema.classes for rng in any_of_ranges):
                     slot_def["@type"] = "@id"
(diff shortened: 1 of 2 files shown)


linkml/generators/jsonldcontextgen.py
```diff
@@ -150,7 +150,8 @@ def visit_slot(self, aliased_slot_name: str, slot: SlotDefinition) -> None:
         else:
             slot_def = {}
             if not slot.usage_slot_name:
-                if slot.range in self.schema.classes:
+                any_of_ranges = [any_of_el["range"] for any_of_el in slot.any_of]
```
cmungall, line 153: does this definitely work? All metamodel elements should be accessed as objects, e.g `any_of_el.range`.

cmungall:
As a next step for this PR, I recommend taking a look at:

https://github.com/linkml/linkml/blob/b6c62ba46abca35373af9b6aa2de60877a967118/tests/test_compliance/test_boolean_slot_compliance.py#L54-L116

and adding something here:

https://github.com/linkml/linkml/blob/b6c62ba46abca35373af9b6aa2de60877a967118/tests/test_compliance/test_boolean_slot_compliance.py#L100-L102

to test the jsonld context output

But let us know if you have issues or don't have time for that and we can take on the PR, thanks!

linkml/generators/jsonldcontextgen.py
djarecka: it worked, but I will change it to `any_of_el.range`

djarecka:
I've added a check for `JOSNOLD_CONTEXT`. I believe the result should be the same in any combination of `use_default_range` and  `use_any_type`, am I right?

FYI: I must say that it was a bit confusing to me when pytest was failing only for one combination of `data_name,value,is_valid`, it took me some time to realize that the checks from `_mappings` are only check for one combination and for the following test this is not being check at all.

cmungall (approved):

djarecka:
@cmungall - I've been using this branch for one of my project, and I have problems when I have `any_of` with `range` either `str` or some class, e.g.:
```
    any_of:
    - range: string
    - range: langString
```
In this specific case, `langString` is `langString: class_uri: rdf:langString` (but I believe it could be anything). 

If in the context I have `@type: @id` and the data is a simple string, I got an error: `Message: Node <http://localhost:8000/examples/activities/items/item1> does not conform to one or more shapes in [ sh:datatype xsd:string ] , [ sh:datatype rdf:langString ]`

Not sure if this is a common case to have range either string or class, and if there is a way that this PR should be changed?

cmungall (approved):

PR #1907, opened 2024-02-12
Foreign keys should be of same type as the primary key they reference

noahhl:
Ran into the same issue as linkml/linkml/#1407 -- if you have a schema like:

```YAML
ClassA:
   attributes:
      id:
          identifier: true
          range: integer
ClassB:
   attributes:
      other_model:
         range: ClassA
```

This was generating a foreign key on ClassB of type text in all SQL derivatives, even though the key that references is an integer. This pull request switches it so foreign keys are specified by the datatype of the identifier they link to.


cmungall (approved):
much appreciated!

PR #1921, opened 2024-02-13
Perf: Mark shex generator as slow, skip except for gh actions runner

sneakers-the-rat:
Continuing from https://github.com/linkml/linkml-runtime/pull/296 trying to improve DX

`test_scripts/test_gen_shex.py:gen_meta` singlehandedly takes 7 minutes (out of 50 total) because it's doing a gigantic rdflib graph equality test. 

Mark this as slow and only run on gh actions test runner.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -160,11 +160,7 @@ def pytest_addoption(parser):
         action="store_true",
         help="Generate new files into __snapshot__ directories instead of checking against existing files",
     )
-    parser.addoption(
-        "--slow",
-        action="store_true",
-        help="include tests marked slow"
-    )
+    parser.addoption("--slow", action="store_true", help="include tests marked slow")
 
 
 def pytest_collection_modifyitems(config, items):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -160,12 +160,12 @@ def pytest_addoption(parser):
         action="store_true",
         help="Generate new files into __snapshot__ directories instead of checking against existing files",
     )
-    parser.addoption("--slow", action="store_true", help="include tests marked slow")
+    parser.addoption("--with-slow", action="store_true", help="include tests marked slow")
 
 
 def pytest_collection_modifyitems(config, items):
-    if not config.getoption("--slow"):
-        skip_slow = pytest.mark.skip(reason="need --slow option to run")
+    if not config.getoption("--with-slow"):
+        skip_slow = pytest.mark.skip(reason="need --with-slow option to run")
         for item in items:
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -160,6 +160,15 @@ def pytest_addoption(parser):
         action="store_true",
         help="Generate new files into __snapshot__ directories instead of checking against existing files",
     )
+    parser.addoption("--slow", action="store_true", help="include tests marked slow")
```
cmungall, line 163: `--include-slow/--no-include-slow` seems a bit more in-keeping with clig.dev guidelines, but I am not so familiar with conventions around pytest

cmungall (approved):

pkalita-lbl (commented):
Instead of adding a custom option, did you consider just using pytest's `-m` flag? Keeping the proposed `slow` marker, we could add `addopts = -m 'not slow'` to the default config. Then in the `main.yaml` workflow we could override that with `-m ''`. Maybe that's a little less explicit, but it saves us from having a custom option.

tests/conftest.py
sneakers-the-rat: how about `--with-slow` ? I think `includes` is a mildly special word ( https://docs.pytest.org/en/6.2.x/reference.html?highlight=include#pytest-freeze-includes ), and `--with-slow` is maybe more conversational? "run pytest with slow tests," which matches poetry's `--with` 

no strong feelings, good with either, prefer both to just `--slow`, your call :)

pkalita-lbl: I like `--with-slow`. And I've come to peace with adding an custom option 😂 Whenever you have a chance to make the change I'm happy to merge.

sneakers-the-rat:
ok changed to `--with-slow` and also added a note in the docs about various marks i've seen and how `slow` is intended to be used

sneakers-the-rat:
I got curious about the rdf canonicalization in rdflib that's the source of the slowness, and noticed that it only takes a really really long time for the `metashex.ttl` case and am wondering why that is.... presumably these are equivalent graphs, since graph isomorphism is transitive. is it really the case that the .ttl representation of the graph is millions of times more expensive to canonicalize than the other representations?

In any case, it looks like rdflib's graph comparison implementation doesn't cache correctly and i'm gonna see if i can fix speed upstream

pkalita-lbl:
Okay, in the meantime I think this is still useful infrastructure to have. So I'll go ahead and merge.

pkalita-lbl (approved):

Issue #1920, opened 2024-02-13
Inferred `tree_root` is not used in JSON schema generation


candleindark:
**Describe the bug**
The inferred `tree_root` of a LinkML schema is not taken into account in the generation of the equivalent JSON schema of the LinkML schema.

**To reproduce**
1. Go to https://linkml.io/linkml/intro/tutorial03.html#example-schema
2. Create the file `personinfo.yaml` with the specified content in the example.
3. Run `gen-json-schema personinfo.yaml > personinfo.schema.json` as specified in the [exercise](https://linkml.io/linkml/intro/tutorial03.html#exercises) after the example.
4. Run `jsonschema -i bad-data.json personinfo.schema.json`, as specified in the exercise as well, to validate the `bad-data.json` (created in accordant with the exercise instruction) against `personinfo.schema.json`.
5. The validation will not produce the expected validation errors.

**Expected behavior**
The validation produces two validation errors that are equivalent to the validation errors produced in the validation process using the LinkML schema, i.e. `linkml-validate -s personinfo.yaml bad-data.yaml`

**About your computer (if applicable, please complete the following information):**
 - OS: MacOS 14.2.1
 - linkml  version 1.7.3
 - Pyton version 3.11.7

**Additional context**
A correct equivalent JSON schema can only be produced if the `Container` class is explicitly marked as root with the `tree_root` slot.

More specifically, `personinfo.yaml` needs to be the following to have a correct equivalent JSON schema generated by `gen-json-schema`

```yaml
id: https://w3id.org/linkml/examples/personinfo
name: personinfo
prefixes:
  linkml: https://w3id.org/linkml/
imports:
  - linkml:types
default_range: string

classes:
  Person:
    attributes:
      id:
        identifier: true     ## unique key for a person
      full_name:
        required: true       ## must be supplied
        description:
          name of the person
      aliases:
        multivalued: true    ## range is a list
        description:
          other names for the person
      phone:
        pattern: "^[\\d\\(\\)\\-]+$"   ## regular expression
      age:
        range: integer       ## an int between 0 and 200
        minimum_value: 0
        maximum_value: 200
  Container:
    tree_root: true
    attributes:
      persons:
        multivalued: true
        inlined_as_list: true
        range: Person
```

pkalita-lbl:
Thanks for reporting this issue @candleindark! 

[...]

PR #1931, opened 2024-02-16
Update gen-json-schema command in tutorial to use --top-class option

pkalita-lbl:
Fixes https://github.com/linkml/linkml/issues/1920


amc-corey-cox (approved):
I agree, let's tell them about the solutions and make this run properly rather than adding complexity to what should be a simple example.

PR #1934, opened 2024-02-17
macos support in CI

bendichter:


cmungall:
I'm all for more tests, but this does explode the number of tests that are run. I believe that macos tests are more expensive (but we are OK I think because this is a public repo?).

Do we think osx is sufficiently different enough from ubuntu this buys much (there is the lowercase vs uppercase filename thing we have been burned on before)?

Perhaps change the matrix such that we only test osx with one other combo rather than the full cross-product?

cmungall:
Hmm, and this would double the time taken:

image

cmungall:
How about this:

- we use pytest to mark tests that are suspected to have some different osx behavior (for example, anything related to dynamic file names, such as markdowngen)
- the osx tests *only* run these tests

cmungall (changes requested):
See previous comments, this will make the CI take too long as-is

bendichter:
Yes, I am used to projects where the matrix is simply os x python version. With the additional pydantic dimension it certainly does add a lot. It's free to you, so you don't need to worry about that, but it does get excessive at a certain point. One approach could be to only run osx on the latest version of Python.

That will not address the time issue. Since the tests run in parallel, you will have this issue even for a single osx test. You could address this by continuing to have the osx test but making it non-blocking for merging. That way you will still get the feedback of a failure, but it shouldn't slow down development.

bendichter:
I have less in favor of your suggestion for only running a subset of tests. In my experience you can end up with cross-OS differences that you don't expect.

pkalita-lbl:
I'll just chime in to say that I'm also hesitant on this. I'm having a hard time thinking of a case where test results differed between GitHub's Linux runner and my local MacOS environment. A little extra insurance wouldn't necessarily be a bad thing, but increasing the time it takes for PR checks to complete to nearly an hour gives me pause.

bendichter:
ok, well feel free to close if you think the cost outweighs the value.

cmungall:
Discussed on 2024-03-08 LinkML dev call, we'll close this for now, we can revisit if the macos builds get faster or if we start running into scenarios where macos and linux differ.

PR #1940, opened 2024-02-23
Deprecate SQLDDLGenerator and change references to SQLTableGenerator

sneakers-the-rat:
While doing https://github.com/linkml/linkml/pull/1939 I noticed a ton of deprecation warnings from the SQLDDLGenerator. I see that it was marked as deprecated in 2022, and it has been partially replaced with SQLTablegen (eg. in `pyproject.toml` scripts), but for the sake of tidying up I thought it would be worth finally removing it. Looking at the commit history, it hasn't been substantively changed since then (just formatting and other global changes like making staticmethods)

I think 2 years is plenty of advance warning, particularly when the functionality has been replaced by another class!

Just trying to help with a chore, if y'all don't want to drop it then np just close this.


sneakers-the-rat:
oop so it looks like the old DDLGenerator was being used in projectgen, was that intended? like i said i just saw something marked as deprecated, but if it wasn't supposed to be then close this!

cmungall:
you did the right thing, I think this may be a false positive will check later...

sneakers-the-rat:
yes so the current projectgen does use the old SQLDDLGenerator: https://github.com/linkml/linkml/blob/05adde14f32cc97dea69ca9b8f4cbe6a4a5e107d/linkml/generators/projectgen.py#L56

which uses the base `Generator.serialize` method. trivial problem here where the TableGenerator.serialize method passed `kwargs` to the relational transformer which doesn't accept arbitrary **kwargs. did a more specific unpacking and that fixed that test, we'll see if the rest of the tests still pass.

this will likely affect projects generated with the ProjectGenerator, since they were using the old DDL generator. not sure how much of a problem that is. taking a look now at the output of the old ddlgenerator and Tablegen and they are extremely different, so we may want to put some warning on this. 

I'm just trying to help out with tech debt, don't have strong feelings about the SQL schema generator. it does seem like a shame that most people may not have been seeing all the work that has been done on the tablegenerator if they were just using projectgen (as i was, from the cookiecutter project template makefile)

sierra-moxon (approved):

PR #1941, opened 2024-02-23
[QoL] Make generating output optional in compliance tests

sneakers-the-rat:
More test profiling :)

Writing and reading yaml turns out to be like really expensive when you do it thousands of times, who knew! 

Added another option (i know i am sorry) to be able to write the (very cool!) diagnostic output in the compliance tests.

This output is (as far as I can tell) not accessible in the CI runner, but is useful when you want to diagnose problems locally, so makes sense to me as a default option.

Did a little hacky monkeypatching to a module level variable to preserve the config option since you can't get access to the config fixtures from functions that aren't test functions, but it works. 

Total time before: 2200s (36.6m)
Total time after: 1620s (27m)
Difference: 580s (9.6m) (26%)

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -73,6 +73,9 @@ DEFAULT_MISMATCH_ACTION = eval(test_settings.get("DEFAULT_MISMATCH_ACTION", "Mis
 #  the final output
 USE_LOCAL_IMPORT_MAP = test_settings.getboolean("USE_LOCAL_IMPORT_MAP", False)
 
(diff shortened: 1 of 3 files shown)


cmungall:
Just FYI the roadmap is to use the generated output to make a website (standalone or part of main linkml docs) that shows a compliance matrix, how different frameworks support, are not able to support, or support by coercion, different language constructs https://github.com/orgs/linkml/discussions/1549

Issue #1925, opened 2024-02-14
Proposal: Deprecation timeline for pydantic v1?


sneakers-the-rat:
Maintaining two versions of the pydantic generator is decent amount of work for increasingly diminishing returns. Pydantic 2 is here to stay (and is really good!). It was released in [june 2023](https://docs.pydantic.dev/2.0/blog/pydantic-v2-final/), and it seems like a ~year-ish of migration time is plenty. So, since this project does rolling releases and it doesn't make sense to target a specific version for deprecation, shall we pick a date for dropping support for pydantic v1?

- [ ] Pick a date for deprecation - Loose suggestion would be 6 months after we add the deprecation warning
- [ ] Add a deprecation warning - When pydanticgen is used with pydantic v1, alert the user that they will need to update their downstream project to use pydantic v2!
- [ ] Survey affected projects - I am not sure what projects are out there that still depend on pydantic v1, it would be nice to have at least an incomplete list, for documentation if not to help triage migration work todo

**How important is this feature?** Select from the options below:
• Medium - can do work without it; but it's important (e.g. to save time or for convenience)

supporting v1 and v2 doubles a lot of work on pydanticgen!

**When will use cases depending on this become relevant?** Select from the options below:

Edit: this is now short term for me, without dropping pydantic 1 i'm going to work on a fork until we do

sneakers-the-rat:
Is there any compelling reason aside from being kind to give adaptation time for keeping pydantic 1? it's strictly deprecated at this point, 100% of the functionality is ported over to 2, anyone who is already using pydantic 1 models should be able to switch to pydantic 2 without any loss of function (from linkML at least, their downstream derivatives may be affected). cost of maintenance is pretty dang high and will only get higher

cmungall:
It's not always easy for everyone to migrate to pyd2 - all it takes is one recalcitrant direct or indirect dependency. Python really needs proper vendoring/shading, see [my comments on python.org](https://discuss.python.org/t/can-vendoring-dependencies-in-a-build-be-officially-supported/13622/11).

However, we don't need to support new features in pydanticgen for `1` as a target. If arrays are only allowed in `2` this is acceptable since we are not depriving anyone of functionality they once had. I realize this doesn't completely alleviate the complexity of supporting both.

I propose:

1. feature freeze on `pydantic_version=1` as a target in pydanticgen effective immediately
2. drop `pydantic_version=1` as a target in pydanticgen in linkml 1.8.0
[...]

Issue #1771, opened 2023-12-08
Include a deprecation guide in the release management guide


cmungall:
Release management how-to: #1765

We should add a section to this on deprecation. We can draw from:

- https://github.com/microbiomedata/nmdc-schema/blob/main/src/docs/schema_element_deprecation_guide.md
- https://biolink.github.io/biolink-model/maintaining-the-model/#deprecating-classes-and-slots

sujaypatil96:
Related PR: #2079

PR #1957, opened 2024-03-06
Add deprecations module and add deprecation warnings for pydantic v1

sneakers-the-rat:
Fix: https://github.com/linkml/linkml/issues/1925

I've actually been meaning to experiment with something like this for awhile, so figured i would take a stab at it. I have seen the decorator a la https://github.com/briancurtin/deprecation a lot before, but i don't necessarily like how the deprecation warnings are tied to the decorator itself, and you have to make sure that the module is imported and the relevant section is run to know they exist.

## Usage

This makes a single `deprecation` module where deprecations are defined, and then you have a simple `deprecation_warning` function that emits the warning and logs that it was emitted wherever it is relevant. So in this case since we aren't deprecating a function or class per se, but a version of a dependency and a value of a parameter, it is easy enough to drop in with those checks where appropriate. A warning is only emitted once bc it's tracked in the `EMITTED` set. 

## Testing

The classes in the `deprecation` module are tested, but i also test that things that are marked as having been removed should be removed in the version that we say they are! This should help us avoid problems like https://github.com/linkml/linkml/pull/1940 and force us to keep up to date. This works by ensuring that `test_removed_are_removed` is run at the very end of testing, where presumably any point where a deprecation warning would have been emitted would have been hit - if it's not, that's a test coverage problem!

### CI

To make this work, I had to make some changes to the CI, sorry if this is a lot

- install the dynamic versioning plugin
- fetch more commits, and the tags from the upstream repo
- install the package instead of using `--no-root`

Otherwise linkml thinks it is at version `0.0.0` and you can't test whether deprecations post that version have been done

## Docs

Made a docs page that looks like this:

Screenshot 2024-03-05 at 9 43 19 PM

Added some dummy/test deprecations just to make sure the plot works, but anyway ya that should make it good and obvious what is getting deprecated when.

I have been wanting to do this both for some of my projects and to demo to my lab, and it turns out that just using jinja is way easier than trying to make a custom sphinx directive. i still have never figured out how that's supposed to work.

## Caveats

- I don't necessarily love the use of module variables here, but i figure it's already overkill, and it sets us up with a tool we can use going forward.
- I added sphinx-jinja and matplotlib to the dev dependencies. sry about that. I could also split up the docs dependencies from the other dev dependencies if that would help soothe the additional weight
- the dependency cache on CI is only invalidated with the lockfile changes. since we don't always refresh the lockfile when the version is incremented, we might miss some deprecations until it is. imo we should always refresh the lockfile so that our testing env doesn't get out of sync with what someone would experience when installing from pip, but that's a separate conversation for a different PR!


sneakers-the-rat:
Somehow 3.8 is failing with the pydantic generator, how did that not happen on the PR? According to [the docs for 3.9](https://docs.python.org/3.9/library/functions.html?highlight=classmethod#classmethod):

> Changed in version 3.9: Class methods can now wrap other [descriptors](https://docs.python.org/3.9/glossary.html#term-descriptor) such as [property()](https://docs.python.org/3.9/library/functions.html?highlight=classmethod#property).

which is news to me. i'm going to mark this as ready for review bc aside from that ti's fine, and we're in the process of dropping 3.8 anyway in https://github.com/linkml/linkml/pull/1704

turbomam:
Thanks @sneakers-the-rat . Some of us are causally reviewing this int eh LinkML Developers meeting now.

There's a question about whether adding a visualization library like matplotlib is really necessary to communicate the version vs. functionality deprecation timeline. 

Can you same some more about the different colors and shapes in your sample deprecation visualization? If color and shape aren't hard requirements for communicating the important points, then maybe it could be communicated with a table or ASCII art?

sneakers-the-rat:
Haha oh boy I tried to do it with svg for a few hours to avoid adding deps before deciding I should just make a simple version that works and talk about deps later.

The dependency should just be necessary when building docs, and so in general I think of docs dependencies as being much freer than main or even testing deps. Matplotlib is heavy, yes, and a PNG is not ideal format for accessibility either. I didnt want to do too much here but imo we should split up the dev dependencies group into tests and docs at least, so then when running tests/CI no matplotlib or Sphinx etc  and vice versa.

I think the visualization would be a really nice touch, and imo worth the docs dependency, but im also super flexible on it. The goal was to have a single page where people could go to see at a glance what to prepare for, with additional information below. This is me "making the deprecation warnings I want to see in the world," so a bit of an experiment (vs. just using the one standard deprecations package)

The rest of the page is effectively table-like, but a sort of collapsed table that can be read on mobile since the deprecation warnings also have potentially paragraph length explanations and recommendations. The plot is useful to visualize relative timing of deps, lengths between deprecations and removals (eg. Pydanticgen support ends sooner than us cranking up the dependency spec), and also shows you that in relation to the current version. I think a table version of that would be comparatively awkward. An ASCII art version would be retro and very cool but potentially a good bit of work to make.

Colors are
green: still good to use
red: stop using - pay attention to this one
gray: already removed.

the lines vs diamonds are just to handle times when we will potentially not set a removal version (eg. To accommodate existing informal deprecations), but maybe that should just be a line that goes arbitrarily to the right.

If theres an easier way to communicate relative position and duration of deps vs. Current version then great! But I think this might also be an opportunity to split up deps so that we dont have to be limited for docs dependencies in the same way we want to be careful for runtime dependencies :)

amc-corey-cox:
I took a thorough look at this and unfortunately, I don't think I quite understand the parts of the code it is touching well enough to offer a review. However, I didn't see anything that jump out at me as concerning. Quite a lot of the files/lines touched are just introducing this throughout where it needs to be used. It looks like a good PR to me.

However, I do have a couple of questions about implementation. I'm not entirely sure what the plot adds to the deprecation documentation. It certainly looks nice and maybe it's more useful than I realize but it seems to me we might be able to get the same information across with a simple table. I don't have strong feelings either way on including this I just wanted to mention I'm not sure it is necessary. I also didn't find it obvious what all of the different features mean, i.e. the green diamond and the importance of green/red/gray lines.

sneakers-the-rat:
How about this. Look at the image above of the page. If that plot adds nothing of value - ie. It doesnt help you know when there are things that will be deprecated, then we can take it out. The section below it is all that information in a tabular form.

In my opinion the plot adds a great deal of "at a glance" clarity - I can tell what version im at now, green is a universal signifier for "ok," red is a universal signifier for "not ok," and then I can see matched headings that give me further information below. The page without the plot would be what it would look like with just "tabular form" information - lots to parse, but possible.

If we dont want it, fine, I dont really care! Main thing i need is someplace to go on this - I want to get it merged in some form so we can actually stay true to the deprecation tl and give advance warning, so just lmk which to do.

sneakers-the-rat:
The plot also can be changed, annotated, etc. So if thats what is missing that can also be done

amc-corey-cox:
I don't hate the plot. It does add a certain at-a-glance value. I just thought we should think about it a little before adding more dependencies for a fairly limited application. I'm ambivalent on keeping the plot so I'm not concerned if we do.

sierra-moxon:
@sneakers-the-rat - we talked in dev about this again - and while we _love_ the idea, we don't love the idea of supporting the utils method added here, just because we're limited in resources.  We wonder if you had thought about making this a standalone module?  Then we could get feedback from the wider python community, etc?  We think it should be some sort of standard.

I'd like to at least turn this into a draft PR, or better into a discussion so we don't lose your work, but we aren't ready to bring it in yet.

sneakers-the-rat:
> I don't hate the plot. It does add a certain at-a-glance value. I just thought we should think about it a little before adding more dependencies for a fairly limited application.

again, not adding a runtime dependency, this would be a dependency that is *only necessary when building the docs*, i suggested above that we split up `dev-dependencies` into `docs` and `tests` so one wouldn't even need to install it during development. pretty much the only time matplotlib would need to be installed are the two CI actions that build the docs.

---

> we don't love the idea of supporting the utils method added here

you mean the whole `deprecation.py` module? 

> We wonder if you had thought about making this a standalone module? Then we could get feedback from the wider python community, etc? We think it should be some sort of standard.

sure, could definitely do that, but then it would just be a dependency which seems like more of a maintenance burden IMO? 

I don't really care about the form of this PR - we can totally ditch the plot, we can make it a separate package, all fine, but I guess stepping back i'd be curious what y'all *do* want as far as deprecation warnings go? there is a bunch of refactoring work to be done, and i'd like to help with that, and to do that in a responsible way we should be leaving deprecation warnings as we go and doing gradual phase-outs. 

Elsewhere I've noticed a bunch of places where stuff is soft marked as deprecated but ends up hanging around for awhile (eg. https://github.com/linkml/linkml/pull/1940 and the `validators` module https://github.com/linkml/linkml/blob/6d739f706bf51d7ecb1c9340aa89d1055aa75356/linkml/validator/__init__.py#L3 ), so it seems like we are holding on to old code because we don't have a good deprecation system. So the purpose of this PR was to make a central place to keep track of and enforce deprecation so that we can iteratively streamline the package to decrease, rather than increase maintenance burden.

So eg. here: https://github.com/linkml/linkml/pull/1967 there's some consolidation that should be done that probably includes deprecating importing something from the current location either in `linkml` or `linkml-runtime` (see https://github.com/linkml/linkml-runtime/pull/305#issuecomment-1997991984 ). with this PR, we would be able to mark that here: https://github.com/linkml/linkml/blob/f5010ba9f061d0e50ab4762a4f0bbbe24f4ce4d9/linkml/validator/loaders.py#L52-L58

like:
```python
DEPRECATIONS = (
  # other ones...
  Deprecation(
    name="runtime-loaders",
    deprecated_in=SemVer.from_str("1.7.5"),
    removed_in=SemVer.from_str("1.8.0"),
    message="message explaining the change"
    # ...
  )
)
```

and then replace that warning with 

```python
deprecation_warning('runtime-loaders')
```

and then that deprecation becomes visible across the project. Currently it's sort of challenging to keep track of what and where things are being changed aside from keeping on top of all the PRs that are happening, but this PR lets me easily add a clear, global signal about an incoming change. Otherwise you would only see that deprecation warning in tests, which not everybody runs, and so someone else working on a different part of the package could be surprised when it's gone.

--- 

So again i'm down to do whatever, but I'm not really sure what the hesitance is with this module in particular - it's fully tested, relatively straightforward, isolated, and has no downstream utility (eg. we won't have people raising issues because they are trying to use it in their project). LMK what would be good to do bc [linkml needs *some* deprecation syustem](https://github.com/linkml/linkml/issues/1771) and i thought this was a relatively tidy way to do that.

sneakers-the-rat:
ping on this - list of list array PR is almost done, and i imagine that might be a minor version bump. we had set a tentative `1.8.0` deprecation date for pydantic generator but we haven't even deployed a warning yet, so that'll probably have to get pushed back, but ya ideally not much further bc i want to be able to simplify that generator at some point. still unsure what the hesitation is!

sneakers-the-rat:
pinging again on this - seems like we are moving closer and closer to pydantic v2 being default, and it would be really nice to give a long deprecation timeline to ppl who are still using pydantic 1 models. if not this PR, then lmk what we would want as far as deprecation warnings - this PR is basically a set of nice things that make it easier for us to manage the many deprecations that currently are leaving zombie code in the library or we might be hesitant to do because we don't have a good system for deprecations! but if we don't like this, then we can do something else, i'm just not really sure what the objections are and so i don't know what else to propose!

sierra-moxon:
from dev meeting:
- @sneakers-the-rat evaluated deprecation annotations/decoration (in use in linkml-runtime already) - https://pypi.org/project/deprecation/
  - the documentation generator needs to know about those annotations vs. just encountering this at runtime. 
- instead thinking about a Deprecation object, with methods that call it/display its output 
- this would be helpful for non-method-level deprecation. 

- we'd like python to give us a standard here, instead of rolling this ourselves, e.g. same as java. 
- might be something in typing that allows marking as deprecated. 
- scipy has one (in repo - not genericizable) 
- can this live as its own thing and generate a community around it?  (then we can import it as a dependency). 

what would the dev workflow be here?
-

sneakers-the-rat:
OK Docs pushed! here's a rendered copy for ur perusal

[deprecation.html.zip](https://github.com/linkml/linkml/files/14891949/deprecation.html.zip)

sneakers-the-rat:
It looks like using the `setup-python` action with the poetry cache can be used without doing `--no-root` (which previously could cause an old version of the package to be installed): https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages

and we need the version info for linkml to be able to test that things marked as removed are actually removed, so i changed to install whole package.

sneakers-the-rat:
wat in tarnation did I break. Need to do other stuff for now so marking as draft until I get back to it

sneakers-the-rat:
ok that was weird but were good now. redy for review

sierra-moxon (approved):
Thank you for the documentation @sneakers-the-rat - I appreciate it very much.

cmungall (approved):

PR #1601, opened 2023-08-31
Add any_of implementation to shaclgen.py 

anjastrunk:
closes #1600

PR #1964, opened 2024-03-12
Support ifabsent in the Shacl generator

vincentkelleher:
This Pull Request implements `ifabsent` attribute support in the Shacl generator as it's needed during the generation of our Gaia-X ontology.

### Remove Logic from SchemaLoader

First of all, I had toremove the generic `ifabsent` parsing and check in the `SchemaLoader` class as the implementation was mostly focused on Python generation whereas default values aren't managed the same way in different languages.

For example, Shacl does not have a specific semantic for default values as stated in [its specification](https://www.w3.org/TR/shacl/#defaultValue) :

> 2.3.2.4 sh:defaultValue
>
> Property shapes may have a single value for sh:defaultValue. The default value does not have fixed semantics in SHACL, > but may be used by user interface tools to pre-populate input widgets. The value type of the sh:defaultValue should align > with the specified sh:datatype or sh:class of the same shape.

From now, the `ifabsent` attribute can be generated specifically for each target language.

### Fix Python/Pydantic Generation Methods

When writing tests using the `kitchen-sink.yaml` file and testing with my own files, I noticed that some issues existed in generated Python code when `ifabsent` had a `date` range for example. This is why I proposed a few fixes to those generation methods.

Furthermore, some refactoring was done for the existing methods as I remove some exceptions in the above chapter.

### Evolve the Shacl Generator

Last but not least, the `ShaclGenerator` class has been updated to support the `ifabsent` attribute which led to some refactoring such as the introduction of the `ShaclDataType` enum for convenience and maintainability.

#### Features

The `ShaclGenerator` can now generate `sh:defaultValue` predicates that manage default LinkML types but also custom enum values.

An example from test files : 

```yaml
classes:
  IfAbsent:
    description: Example class composed of attributes using ifabsent
    attributes:
      ifabsent_string:
        range: string
        ifabsent: string(This works)
      ifabsent_boolean:
        range: boolean
        slot_uri: ks:ifabsent_boolean
        ifabsent: True
      ifabsent_int:
        range: integer
        ifabsent: int(123)
      ifabsent_decimal:
        range: decimal
        ifabsent: decimal(1.23)
      ifabsent_float:
        range: float
        ifabsent: float(1.23456)
      ifabsent_double:
        range: double
        ifabsent: double(1.234567)
      ifabsent_date:
        range: date
        ifabsent: date("2024-02-08")
      ifabsent_datetime:
        range: datetime
        ifabsent: datetime('2024-02-08T09:39:25Z')
      ifabsent_uri:
        range: uri
        ifabsent: uri('https://w3id.org/linkml/tests/kitchen_sink/ifabsent_boolean')
      ifabsent_not_literal:
        range: CordialnessEnum
        ifabsent: CordialnessEnum(heartfelt)
```

I've based myself on the [LinkML Default Value documentation](https://linkml.io/linkml/schemas/slots.html#default-values) so values such as the following are supported :
- `string(value)`
- `string('value')`
- `string("value")`
- `value`
- `'value'`
- `"value"`
- `MyEnum(value)`
- `MyEnum('value')`
- etc.

The generated Shacl `sh:defaultValue` has a `Literal` object with the corresponding XSD datatype when relevant.

#### Tests

I took time to write all the necessary tests with a specific `kitchen_sink_ifabsent.yaml` file and by isolating all the Shacl generation logic in a `IfAbsentProcessor` class for better responsibility division and testing.


vincentkelleher:
As Github doesn't offer the possibility to mention dependencies between PRs, this depends on https://github.com/linkml/linkml/pull/1601 being merged.

cmungall (approved):

vincentkelleher:
Thanks @cmungall for the very quick approval and merge 🙏

PR #1967, opened 2024-03-13
Remove `validator.loaders` - move to `linkml-runtime`

sneakers-the-rat:
Depends on: https://github.com/linkml/linkml-runtime/pull/305

Noticed that these seem to be duplicated and older, and most of the work on the loaders is over on linkml-runtime. Figure this was just an accident of history, since (from what i gather) linkml-runtime only came along later. Trying to help out with pruning old code as i read it.

Pretty straightforward 
- add all functionality from these loaders to the ones on linkml-runtime, 
- preserve import locations by changing `validator.loaders` from a package to a module,
- add deprecation warning. once https://github.com/linkml/linkml/pull/1957 merges then we can update to track it that way for final deprecation.

I didn't move the tests there because it seems like linkml-runtime only gets pytest as a passive dependency through `coverage` and wanted to ask first. These tests will obviously fail on CI until `linkml-runtime#305` merges, but tested locally they pass.


nlharris:
What does this "chore" label mean? This is the only PR (or issue) with that label.

sneakers-the-rat:
idk i was just making it up, basically function neutral changes like code cleanup. we can remove and delete it np

dalito:
@nlharris - "chore" is also used in other repos e.g. for action updates and other "boring" stuff that has to be done once in a while.

nlharris:
I don't object to adding a new label if it's useful! This was the only thing tagged "chore" (which is one of 50 labels -- I try to do cleanup of labels that are unused or rarely used).

sneakers-the-rat:
(go ahead and delete i have no strong feelings at all and it doesn't *exactly* fit - i was applying it here bc i figured it was an uncontroversial change, but it turns out it's not! so probably inaccurate)

amc-corey-cox:
@sneakers-the-rat My read on the above thread is that this is obsolete? Is that correct?

Would it be okay if I close this one? I'm trying to clear out some of our backlog of old PRs that haven't seen much activity.

sneakers-the-rat:
I think the prior discussion was about removing the `chore` label rather than the PR. This module is still present, i believe it is dead code, see discussion at https://github.com/linkml/linkml-runtime/pull/305 

but ya i don't have strong feelings about it - just trying to prune and consolidate some of the dead code to reduce maintenance burden, and I think unifying the loaders/dumpers is relatively important to the usefulness of the library, since currently that behavior is pretty scattered and it's core to being able to use data with linkml models and all that, but to be honest i still love yall & understand review labor is scarce, but i've lost the passion for tidying a bit and have moved on a bit since my prs started stacking up. so i am trying to neither insist on nor block anything but just comment if it's something i've been tagged in or is important to my uses of linkml.

sierra-moxon:
closing for now; keeping branch open

Issue #1958, opened 2024-03-07
[Proposal] Remove `tests/**/output` from vcs


sneakers-the-rat:
**Is your feature request related to a problem? Please describe.**

Whenever I run tests locally, I always end up with a lot of the files in various `output` folders changed, and then I need to be careful to ignore them when committing, or else run `git update-index --assume-unchanged` for all of them a la here: https://github.com/linkml/linkml/blob/0c3afa90701dd18ce874f6103cc7668684eb697d/hide_test_changes.sh#L3

that call doesn't persist across switching branches, which one needs to do all the time, so it's like playing whack-a-mole where inevitably i will end up committing one of them.

I am not quite sure what the purpose of having them versioned is - being able to look at the intended outputs all the time is nice, and in that case we'd expect those files to not be used in the tests since the test should be self-contained. Using those as expected outputs to compare tests to would also be good, but then since the tests also generate there then their status becomes ambiguous. 

It seems like `__snapshot__` is supposed to be the way to store expected outputs that are used in the tests, but many tests use `outputs` instead - eg. any test that uses `TestEnvironment` does, since `outdir` is `output` : https://github.com/linkml/linkml/blob/0c3afa90701dd18ce874f6103cc7668684eb697d/tests/utils/test_environment.py#L42

which is used by `expected_path`: 
https://github.com/linkml/linkml/blob/0c3afa90701dd18ce874f6103cc7668684eb697d/tests/utils/test_environment.py#L97

which is used in turn by `TestEnvironment.generate_single_file`, `GeneratorTestCase.single_file_generator`, which is used in `test_base` and `test_biolink_model` . 

So ultimately it isn't clear to me what the `output` directories are for, and seems like it has the potential to cause weird testing errors and make PRs more difficult to review. 

**Describe the solution you'd like**
I think we should remove `output` directories from being tracked by git and make a clearer distinction between the types of test directories. I like the way that test directory structure is mirrored across the test packages, so this hopefully shouldn't be too hard.
- `__snapshots__`  (versioned) - frozen output used to test
- `temp` (ignored) - output that is generated during testing that is cleared during cleanup, ideally by a fixture: if one wants to keep those outputs, we already have a `--with-output` flag for that.
- `examples` (versioned) - sample output that is used to be able to see expected results without them being used in testing, if desired.

We can then clarify in the developer docs what the directory structure is within the tests, as well as document how the various utility classes and fixtures are intended to be used.

[...]

PR #1968, opened 2024-03-13
Remove test outputs from git

sneakers-the-rat:
Fix: https://github.com/linkml/linkml/issues/1958

As of now this PR only removes the `output` files from git - they are already in `.gitignore` so nothing more needs to be done there.

Wanted to open this and ask what we want in the way of docs for this - should we describe the directory structure for each test package in this PR? 

I also didn't make keeping the `output` files dependent on `--with-output` because i wasn't sure if we wanted that. Since the files aren't tracked, this would mostly be a disk usage thing. The outputs are ~dozens-~hundreds of MB, and are sort of a pain to delete, but I'm not exactly sure how one would do something like what one would do with a pytest fixture with the unittest class.

eg. in pytest we would do something like this:

```python
def make_output_dir(request):
    dir = Path() / 'output'
    yield dir
    if not request.config.get('with-output'):
        shutil.rmtree(dir)

```

Anyway more tidying. This is mostly so I don't have to worry about accidentally committing more output files and also don't have to keep avoiding them every time i make a commit lol

Other changes:
- Make UnitTest environment use `__snapshots__` instead of `output`
- Make UnitTest environment respect `--generate-snapshots` when applicable


cmungall (approved):

pkalita-lbl:
I might be able to give a bit more context here. You've accurately identified a lot of the issues with the existing tests that are based around `TestEnvironmentTestCase` subclasses, so no need to rehash that. We want to get to a place where:

* All tests are pytest-based
* `TestEnvironmentTestCase` and `output` directories don't exist anymore
* Tests that need to compare results to "gold standard" files use the [snapshot fixture](https://github.com/linkml/linkml/blob/6d739f706bf51d7ecb1c9340aa89d1055aa75356/tests/conftest.py#L125) and should be done sparingly 
* Any tests that must write to disk do so using temporary files and directories using pytests standard fixtures for that

We started taking a directory-by-directory approach to this conversion and cleanup effort: https://github.com/linkml/linkml/issues?q=is%3Aissue+%22convert+tests+in%22 I'll admit that effort has sort of stalled out a bit as I haven't has as much time to devote to LinkML lately. 

Again, I 1000% feel your pain. I also hate having to jump through git hoops to tame the `output` directories. But my 2 cents is that I'd like to see effort being put towards the remaining test directory conversions. I get a little uncomfortable seeing the old and new testing infrastructures getting intermingled in this PR when the old infrastructure is slated for demolition.

sneakers-the-rat:
That's fair. I was trying to change as little as possible in this one PR while unifying behavior around `--generate-snapshots`. So what would be better to do? 

This PR is a step forwards/not in conflict with other efforts to replace the test cases classes, it just makes them behave like the pytest functions that use the `snapshot` fixture. It's only actually used in a few places:

`TestEnvironmentTestCase`
- `test_base/test_python.py`
- `test_enhancements/*`
- `test_notebooks/test_notebooks.py`

`GeneratorTestCase`
- `test_base/*`except `test_python.py`

so i could probably swap those out in about an hour if we'd rather do it in one shot. 

lmk what we want to do here.

sneakers-the-rat:
ya this doesn't seem hard. i'll have this done by EOD https://github.com/sneakers-the-rat/linkml/tree/tests-remove-testcase

sneakers-the-rat:
superceded by: https://github.com/linkml/linkml/pull/1976

Issue #1948, opened 2024-02-27
Consolidate metamodel tests


sneakers-the-rat:
**Is your feature request related to a problem? Please describe.**
Still trying to make the tests run faster :)

**Describe the solution you'd like**
Test metamodel correctness separately from generator and cli script correctness.

Currently the metamodel is generated and tested in a few places, and it seems like some are duplicates, and others don't need to be tested. In particular the `test_scripts` directory generates a bunch of versions of the metamodel and then compares them to snapshots. It seems like `test_scripts` is for testing the cli scripts, but they use the metamodel as the demo data rather than a lighter weight model, so they end up taking a long time!

It seems like we should test the metamodel correctness and generation in `test_base/test_metamodel.py` and use a much cheaper model in `test_scripts`. This would shave off another 5-10m depending on which of the tests we decide to keep

**How important is this feature?** Select from the options below:
• Low - it's an enhancement but not crucial for work

**Additional context**
I'm happy to do this if it would b useful. my goal is to make tests be fast enough that we can move quickly and confidently, we've already come down from an hour-ish to 25m, and we can keep going :)

cmungall:
Thanks, fully agree with this analysis and suggested course of action

sneakers-the-rat:
I can handle this - is there a script/tool already for creating and saving the snapshots? or how are those updated?

Issue #1956, opened 2024-03-05
Generate PlantUML locally or mark all tests flaky


sneakers-the-rat:
**Describe the bug**
PlantUML generation depends on a web API called "kroki": https://github.com/linkml/linkml/blob/0c3afa90701dd18ce874f6103cc7668684eb697d/linkml/generators/plantumlgen.py#L97-L105

however this is not a reliable way to generate plantUML, as it intermittently fails during test runs, which still take ~20-30m to do, eg. https://github.com/linkml/linkml/actions/runs/8153290936/job/22284373873?pr=1887

If there is any package that can generate plantUML locally, we should use it, otherwise we should mark any test that the plantUML class touches as [flaky](https://docs.pytest.org/en/stable/explanation/flaky.html)

dalito:
plantuml (the java package) would be required in CI to generate images. It can be done, but checking the generated "puml" code should be sufficient for linkML in my opinion. For sphinx I prefer to use the puml-file and to generate the svg image as part of the docs build.

So maybe image generation from puml via an external service should not be part of linkML´s plantuml support?

Silvanoc:
> So maybe image generation from puml via an external service should not be part of linkML´s plantuml support?

I would do something intermediate:

1. As proposed by @dalito, generate by default only PlantUML code and test the code generation.
2. Additionally make the already existing argument `--kroki-server` act simultaneously as a flag: if passed with a value, generate the SVG of the diagram using the provided Kroki server. Then test only the correctness of the generated Kroki API request without even interacting with a real server.
 
That way the generator doesn't have a real external dependency to be tested and all flakiness will disappear.

sneakers-the-rat:
oh ya 1) should definitely be done for all the generators, good unit tests in addition to the end to end generation tests. 
I'll PR 2) in a second - imo that's just polite in addition to the perf benefits and test robustness, and should generally apply to any tests that make network requests. we can use https://pytest-vcr.readthedocs.io/en/latest/

sneakers-the-rat:
for recordkeeping: the tests that still make network requests atm are - 

- test_base/test_metamodel::test_meta_markdown - 86 requests
- ''::test_meta_rdf - 15 requests
- test_generators/test_plantuml::test_generate_svg - 1 request
- test_generators/test_rdfgen:;test_annotation_extensions - 6 requests
- test_issues (all from 1-6 requests)
  - 112
  - 12
  - 163
  - 3
  - 381
  - 163
  - 384
  - 388
[...]

PR #1969, opened 2024-03-13
Replace metamodel with kitchen sink in script tests

sneakers-the-rat:
Fix: https://github.com/linkml/linkml/issues/1948

Another simple one that looks scarier than it is - replace the metamodel with the kitchen sink model and update the snapshots to match. 

Only editorial choices i made here were to use the `Person` and `Event` classes in place of the `schema_definition` and `element` classes.

To fix before merging:
- [ ] Marked `test_gen_rdf` tests as `xfail` for now because when rdflib tries to parse the JSON-LD string that gets generated, it makes a call to `https://linkml.io/linkml-model/linkml_model/jsonld/tests/kitchen_sink/core.context.jsonld/kitchen_sink/core.context.jsonld` which doesn't exist, and so it 404's and errors. Wasn't sure what to do there, so waiting for guidance.
- [ ] Decide which metamodel tests we want to have! Ideally we would test the correctness of the generators independently, but having some checks that the metamodel is consistent is good for double checking we're not breaking things unexpectedly. Are there any metamodel tests we want to keep from `test_scripts`, keeping in mind that comparing to snapshots is only as good as our snapshots, so it's really a 'test of change' - which again is good in the case of being careful about not changing metamodel accidentally, but might not be necessary for all generators.

Edit: swapping out the metamodel for kitchen sink made `test_meta` in `test_gen_shex` no longer 7 minutes! now completes in ~10 seconds, so removed the `slow` mark! woohoo

Edit2:
Test duration before: [25m](https://github.com/linkml/linkml/actions/runs/8251230764/job/22567793478)
Test duration this PR: [17m](https://github.com/linkml/linkml/actions/runs/8258775723/job/22591596670?pr=1969)
Difference: -8m (32%)

Edit3: tests are failing because of https://github.com/linkml/linkml/issues/1956 not because of this PR


sneakers-the-rat:
with this most recent batch of PRs, we have gone from the test taking an hour to run to <5m

cmungall:
My preferred course:

- merge this PR
- investigate jsonld context oddity independently
- for our "scattergun test" strategy, we proceed more intentionally along the following lines
    - select a fixed set of N blessed models to include (e.g metamodel, biolink)
    - run gen-project over these and use the snapshot strategy to check for changes (ensures broadness), but don't otherwise test anything combinatorial
    - document the SOP. E.g. if I am changing pydanticgen then I expect perhaps a large no of snapshots to change, but that's OK, I do a bit of TestByGuru, but largely trust regenerating snapshots, for any changes to generated pydantic (if other things change I stop and reconsider)
    - we freeze the version of each model, and make sure we expand imports or ensure imports are to a fixed version
    - we periodically update each model (regenerating snapshots)
    - we increase N where it doesn't adversely impact overall runtime of tests

sneakers-the-rat:
Good with that.

> run gen-project over these and use the snapshot strategy to check for changes (ensures broadness), but don't otherwise test anything combinatorial

Another thing we could do when comparing snapshots is to have a multi-modal testing strategy: do a string equality test first which should be very fast. If that fails, then test equality line by line to report diff. Then try doing modality-specific equality test: eg. Graph isomorphism for RDF, etc. If modality specific test fails, fail test, otherwise warn with location of string diff. Try to keep those warnings close to zero/modify the test to exclude mutable substrings like generation date.

Since the snapshot tests are tests of change, and they are all strings (right?) I think we dont need to do expensive equality comparisons. We could even modify the snapshots to be k:v stores with content hashes so we dont even need to do IO for the snapshot, just read and compare the hash.

> we freeze the version of each model, and make sure we expand imports or ensure imports are to a fixed version

I am not sure what u mean by this, could u clarify? In any case yes re: freeze version, and it would be good to have a separate test to check if the snapshot version is the most recent version, right?

> we periodically update each model (regenerating snapshots)

Ya ya re: above, have a cron job that checks for model updates. We'll catch changes to our generated version of model with tests, but we should also get notified when we need to update model because it changes upstream, esp in linkml-runtime and metamodel

> we increase N where it doesn't adversely impact overall runtime of tests

Yes yes, and we can mark all blessed model tests as slow (assuming they are big and complicated like biolink/metamodel) and only run on ci by default. The major low hanging fruit perf barrier that remains is the yuml generator making calls to an external API, but im about to PR a VCR'd version of those tests with frozen network responses so that should be good. 

From there, since the tests are so much lighter, we can think about other strategies for integration tests with eg. Linkml-runtime and if we want to work on splitting out generators (which I would love to do).

One last thought on this which should probs be its own issue: I feel like since metamodel and pythongen are so core to the whole thing, it might be good to, when metamodel -> python models change, we force using those regenerated models in the tests - ie. We monkeypatch those over the models in linkml-runtime. Anyway not for this PR, but wanted to float idea, will raise that (and any other separate issues) after this

sierra-moxon (approved):
discussed in dev call: we're ok with this, noticed that we move the metamodel out of all testing, which gave us a bit of pause.  But we should target the metamodel as a separate test and since this is a "copy" of the metamodel that isn't necessarily tied to a model release, we aren't worse off from this change.

sneakers-the-rat:
@sierra-moxon metamodel is still tested! all the metamodel tests are just in one test class now: tests/test_base/test_metamodel.py

each are marked as `slow`, but CI runs `--with-slow` :)

PR #1979, opened 2024-03-16
Linter: add patterns in config for classes and slots 

### LINKED ISSUE #1978: Linter: add pattern as addidional config in StandardNamingConfig
I am trying to use LinkML in the context of an already existing project where I don't have the liberty to alter the naming convention for slots and classes. In order to still be able to use the _linter_ I would need to have a more flexibility in the `standard_naming` rule.

The feature that I'd like to propose:
-  add `class_pattern` and `slot_pattern` as additional, optional configuration in the `standard_naming` linter rule

**How important is this feature?** 
• Medium - can do work without it; but it's important (without this change I won't be able to use the linter)

**When will use cases depending on this become relevant?** 
• Short-term - 2-4 weeks

linupi:
closes #1978

Proposes the following two features:

- add `class_pattern` and `slot_pattern` as additional, optional configuration in the `standard_naming` _linter_ rule
- enable patterns defined in `LinterRule` as well as regular expressions

here is an example of what the config for the _linter_ would/could look like

```
extends: recommended
rules:
  standard_naming:
    level: error
    slot_pattern: camel  
    class_pattern: "[A-Z][a-zA-Z0-9]+"
```

Would something in this direction be acceptable? I'd be happy to adopt, if necessary.

### DIFF
--- a/docs/schemas/linter.md
+++ b/docs/schemas/linter.md
@@ -105,6 +105,8 @@ Enforce standard naming conventions: CamelCase for classes, snake_case for slots
 
 **Additional Configuration**
 * `permissible_values_upper_case`: If `true`, permissible values will be checked for UPPER_SNAKE, otherwise snake_case. Default: `false`.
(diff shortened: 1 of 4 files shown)


linupi:
@cmungall : thanks for your comment, rebase should have fixed the docker issue...

Issue #1977, opened 2024-03-16
Docker build broken :(


sneakers-the-rat:
**Describe the bug**
All the docker builds aren't working :(
https://github.com/linkml/linkml/actions/workflows/docker-build.yaml

Comparing the last working one: https://github.com/linkml/linkml/actions/runs/8275166684
to the first broken one: https://github.com/linkml/linkml/actions/runs/8301994721

it looks like buildkit version went from 

 builder-86d20cc5-b3b9-4627-af4e-eeaa12ffee050: moby/buildkit:buildx-stable-1 => buildkitd github.com/moby/buildkit v0.12.5 bac3f2b673f3f9d33e79046008e7a38e856b3dc6

to

builder-d03649ed-a16e-4ff9-8de4-d96bd25661840: moby/buildkit:buildx-stable-1 => buildkitd github.com/moby/buildkit v0.13.0 2afc050d57d17983f3f662d5424c2725a35c60f4

am not a docker person, do not particularly care for it, don't know how to fix :'(

cmungall:
@Silvanoc do you happen to know offhand what is happening here? If not we'll have someone check this out ASAP monday

Silvanoc:
@sneakers-the-rat I don't know where you got the version information that you're providing, but:
1. there is no buildx v0.12.5. It jumps from v0.12.1 to v0.13.0
2. I don't see any difference in the buildx version between the [working pipeline](https://github.com/linkml/linkml/actions/runs/8275166684/job/22641685363#step:5:115) and the [failing pipeline](https://github.com/linkml/linkml/actions/runs/8301994721/job/22723316507#step:5:115)

Am I missing something?

Silvanoc:
I haven't been able to dig deep enough to get to the root-cause. Here what I've found out:
1. the [code reporting the error](https://github.com/moby/buildkit/blame/6c0b576ff334443369ad1b675fe59bba5f905d72/frontend/dockerfile/dockerfile2llb/convert.go#L1203) is 2 years old and the code around it is years (2-5) old, so the issue might be in any of the functions being called
2. I suspected that the called function [gitutil.ParseGitRef](https://github.com/moby/buildkit/blob/6c0b576ff334443369ad1b675fe59bba5f905d72/frontend/dockerfile/dockerfile2llb/convert.go#L1200) might be the origin of the error
3. wrong, nothing has changed in that function between both releases

Now I'm clueless 🤷

sneakers-the-rat:
[...]

PR #1980, opened 2024-03-17
[docker] Fix error from `COPY` .git directory by using `ADD`

sneakers-the-rat:
Fix: https://github.com/linkml/linkml/issues/1977

(Don't know if this will work yet, will update description if it does)


cmungall (approved):
🙏 🙏 🙏

Issue #1881, opened 2024-02-08
specifying `default_prefix:` prevents `gen-jsonld-context` from emitting classes


tgbugs:
**Describe the bug**
specifying `default_prefix:` prevents `gen-jsonld-context` from emitting classes

**To reproduce**
``` yaml
id: https://sparc.olympiangods.org/sparcur/schemas/1/sparc
title: sparc export schema
name: sparc-linkml
imports:
  - linkml:types

default_prefix: sparc  # uncomment this line to get default behavior

prefixes:
  linkml: https://w3id.org/linkml/
  sparc: http://uri.interlex.org/tgbugs/uris/readable/sparc/

classes:
  Thing:

slots:
  # manual
  id:
  label:
    slot_uri: rdfs:label

```
run `gen-jsonld-context` and `Thing` will not be included.

**Expected behavior**
Should generate jsonld context with `Thing`

vincentkelleher:
I'm having the same issue and the culprit is here 👉  https://github.com/linkml/linkml/blob/main/linkml/generators/jsonldcontextgen.py#L137

```python
def visit_class(self, cls: ClassDefinition) -> bool:
        class_def = {}
        cn = camelcase(cls.name)
        self.add_mappings(cls)
        cls_uri_prefix, cls_uri_suffix = self.namespaces.prefix_suffix(cls.class_uri)
        if not self.default_ns or not cls_uri_prefix or cls_uri_prefix != self.default_ns:
            class_def["@id"] = (cls_uri_prefix + ":" + cls_uri_suffix) if cls_uri_prefix else cls.class_uri
            if cls_uri_prefix:
                self.add_prefix(cls_uri_prefix)
        if class_def:
            self.slot_class_maps[cn] = class_def

        # We don't bother to visit class slots - just all slots
        return False
```

The only issue is that I don't exactly understand why this condition has been defined in the first place 🤷

vincentkelleher:
I've also noticed the same thing for slots when a `default_prefix` is provided : https://github.com/linkml/linkml/blob/main/linkml/generators/jsonldcontextgen.py#L173

PR #1981, opened 2024-03-18
JSON-LD Context Generator Fixes

vincentkelleher:
This is directly linked to issue https://github.com/linkml/linkml/issues/1881 which I stumbled upon after encountering class/slot generation issues with the JSON-LD Context generator.

### Dependecies

This pull request depends on https://github.com/linkml/linkml-runtime/pull/308 as the CURIE prefix/suffix extraction function is needed.

### Fixing Class/Slot Generation

The main fix of this pull request consists in making sure that classes and slots are correctly generated even when a default namespace prefix is defined in the source LinkML file, please see https://github.com/linkml/linkml/issues/1881 for more details.

This was done by changing the conditions in which those elements are generated involving `linkml-runtime`'s namespace functions to resolve the CURIE of each class/slot and determine the corresponding URI for each element.

### Disable Mixin Generation

I also noticed that the output JSON-LD Context contained all the expected classes but it also had mixin classes which should not be known of the end users as they are only relevant in a LinkML entity conception context. This is a straightforward fix but it has a lot of impact on tests.


cmungall:
> This pull request depends on https://github.com/linkml/linkml-runtime/pull/308 as the CURIE prefix/suffix extraction function is needed.

merged and available in
https://github.com/linkml/linkml-runtime/releases/tag/v1.7.3

linkml/generators/jsonldcontextgen.py
```diff
@@ -130,19 +130,19 @@ def end_schema(
             print(as_json(context))
 
     def visit_class(self, cls: ClassDefinition) -> bool:
+        if cls.mixin:
```
cmungall, line 133: this seems like a separate concern from the main intent of the PR?

cmungall (changes requested):
can you clarify the intent w.r.t. mixins? Otherwise looks great, thanks so much for this!

linkml/generators/jsonldcontextgen.py
vincentkelleher: I had the feeling it was linked to fixing class and slot generation by the JSON-LD context generator, this can be another PR if needed 😊

vincentkelleher:
> can you clarify the intent w.r.t. mixins? Otherwise looks great, thanks so much for this!

To my understanding, mixins are equivalent to abstract classes which cannot be directly manipulated by end users, they only allow the owner of the code to build concrete classes. Maybe my vision is wrong in the case of JSON-LD contexts ?

cmungall:
Can you move the mixin change to a new issue or PR? It's true that mixins should not be directly instantiated, but someone may still want to refer to a mixin in another schema, so we want to expose as a jsonld-context.

Also I'm afraid you have falled victim to the annoying `output` conflict problem, we hope to remove these folders soon...

vincentkelleher:
@cmungall I've moved the mixin generation part to another branch and I'll raise another PR when this one is merged 😊

cmungall (approved):

vincentkelleher:
Thank you for the merge @cmungall 🙏

PR #1976, opened 2024-03-16
[tests] Remove test outputs and GeneratorTestCase

sneakers-the-rat:
Supercedes: https://github.com/linkml/linkml/pull/1968
Fix: https://github.com/linkml/linkml/issues/1716
Fix: https://github.com/linkml/linkml/issues/1715
Fix: https://github.com/linkml/linkml/issues/1958

In https://github.com/linkml/linkml/pull/1968 @pkalita-lbl said we don't want to mix pytest behavior with unittest behavior, which is valid. That PR also didn't finish the job of restoring desired snapshots (so the tests were doing pointless comparisons of "does the thing we just generated equal the thing we just generated")

This PR extends that one by removing the `GeneratorTestCase` altogether, replacing it with the `snapshot` fixture. 

The remaining tests that use `TestEnvironmentTestCase` behave like snapshot in such a way that `snapshot` can be trivially substituted in most of them.

So this PR:
- preserves existing test rigor
- continues work switching unittest to pytest, knocking out two modules where that needed to be done (`test_base` and `test_biolink_model`)
- replaces expensive comparison logic with simple string comparison - which is all we need when we're comparing snapshots, since they're just tests for change
- removes `output` directories so that we don't have to wait until all test modules are switched to `pytest` to make developer experience not include deselecting a bunch of files in `output` directories on every commit.

I hate how huge this PR is too, but as with `1968` pretty much all of the changes are from moving snapshots from `output` to `__snapshots__`. There isn't really a way to decouple this afaik because if i made the PR without the snapshots (to add them later and just consider the substance of the PR) then all the tests would fail until then. 

If it were up to me, we would remove pretty much all of these tests except for the metamodel -> python generator because that's a functional test that indicates that we need to regenerate the models in linkml_runtime (and i would rewrite it to compare the generated python models to the models in locked version of linkml_runtime). The rest of them don't really tell us much, because every time we change the generators we should expect a change in the models they generate, and so these are basically redundant markers of that change.

My goal here is to remove a pretty annoying bit of linkML DX, and for that I make this humble offering of paying down technical debt while preserving existing testing behavior in the process. Not trying to mess anyone's day up with a one million line PR.

edit: the reason why there are more markdown files in the `test_base` `__snapshots__` folder are now they are each generated to a different directory, rather than all being in the same directory - ie. the tests are independent.

edit: here is a diff of all the non-output and non-snapshot changes:

<details>
<summary>expand/collapse diff</summary>

```diff
[...]

PR #1982, opened 2024-03-18
Cast rdflib `Namespace` instances to strings when used as `base` for contextgen 

sneakers-the-rat:
So what's the deal with the `@base` [being](https://github.com/sneakers-the-rat/linkml/blob/2e7fac785f59ecded70768b737be2e18a158dadb/tests/test_base/__snapshots__/types.context.jsonld#L13) `"https:/w3id.org/linkml"` in https://github.com/linkml/linkml/pull/1976 ?

That's no string! that's an `rdflib.Namespace` object! 

`Namespace` overrides `__contains__`: https://github.com/RDFLib/rdflib/blob/98346657aaea6bad23fe574b7e575d8fbc002da7/rdflib/namespace/__init__.py#L175 so the check `'://' in base` always fails. 

cast to a string and make the type annotations correct.

also override `serialize` and add `base` because otherwise there's no way to know it's an allowable parameter to `serialize`. would b cool to have a docstring but 🤷

### DIFF
--- a/linkml/generators/jsonldcontextgen.py
+++ b/linkml/generators/jsonldcontextgen.py
@@ -6,14 +6,14 @@ Generate JSON-LD contexts
 import os
 import re
 from dataclasses import dataclass, field
-from typing import Dict, Optional, Set
+from typing import Dict, Optional, Set, Union
 
 import click
 from jsonasobj2 import JsonObj, as_json
 from linkml_runtime.linkml_model.meta import ClassDefinition, SlotDefinition
 from linkml_runtime.linkml_model.types import SHEX
 from linkml_runtime.utils.formatutils import camelcase, underscore
-from rdflib import SKOS, XSD
+from rdflib import SKOS, XSD, Namespace
 
 from linkml._version import __version__
 from linkml.utils.generator import Generator, shared_arguments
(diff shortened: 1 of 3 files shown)


dalito:
Also change https://github.com/linkml/linkml/blob/16f4e621f944d35cfdee3eb1bcbd748a3eb416be/linkml/generators/prefixmapgen.py#L64-L66 ?

sneakers-the-rat:
> Also change
> 
> https://github.com/linkml/linkml/blob/16f4e621f944d35cfdee3eb1bcbd748a3eb416be/linkml/generators/prefixmapgen.py#L64-L66
> ?

not sure what you mean? those lines are unchanged. the failure comes from the `if "://" not in base` part, which always returns `False` because of the overridden `__contains__` method. that's fixed by casting to `str` without changing the lines. 

or are you saying those lines should also be changed somehow?

dalito:
I meant that there is the same code as in the end_schema-method as in linkml/generators/jsonldcontextgen.py which you just fixed by inserting `base = str(base)` so wondered if you need to make the change there too.

sneakers-the-rat:
oh gotcha. sure one sec

sneakers-the-rat:
i'll leave the tests for that to someone else. idk what that generator does really (and there are currently no tests for it)

cmungall (approved):

PR #1983, opened 2024-03-18
[contextgen] Honor parameterization in generator instance, expose parameters to `serialize`

sneakers-the-rat:
Builds on https://github.com/linkml/linkml/pull/1982

Noticed that the contextgen advertises parameters in its dataclass fields that it then ignores during serialization.

If those fields exist, they should do something!

set all args to default `None` and use instance attributes as defaults instead of method args as defaults - parameters can be set in the instance body, but overridden when calling `serialize`

### DIFF
--- a/linkml/generators/jsonldcontextgen.py
+++ b/linkml/generators/jsonldcontextgen.py
@@ -6,14 +6,14 @@ Generate JSON-LD contexts
 import os
 import re
 from dataclasses import dataclass, field
-from typing import Dict, Optional, Set
+from typing import Dict, Optional, Set, Union
 
 import click
 from jsonasobj2 import JsonObj, as_json
 from linkml_runtime.linkml_model.meta import ClassDefinition, SlotDefinition
 from linkml_runtime.linkml_model.types import SHEX
 from linkml_runtime.utils.formatutils import camelcase, underscore
-from rdflib import SKOS, XSD
+from rdflib import SKOS, XSD, Namespace
 
 from linkml._version import __version__
 from linkml.utils.generator import Generator, shared_arguments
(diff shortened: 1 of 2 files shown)


sneakers-the-rat:
looks like the failing tests are choices to make - the old behavior didn't honor the defaults set in the instance, so the tests confirm that. if we want that to be the case (ie. when you set a value in the generator instance it is actually used) then we need to change those tests probably. since this PR is unsolicited i'll wait until someone confirms this is the desired behavior before fixing

amc-corey-cox:
Hey @sneakers-the-rat sorry I misunderstood the comments above. It sounds like this may still be worthwhile, so I'm going to leave it open... and possibly revisit this myself. As you mentioned, we have had some PRs piling up. That is something I'm trying to improve and some of that is reducing our cognitive tax in having so many dead PRs open.

With that said, I don't want to lose meaningful work so I'll see what we can do about this one. Also, if you feel like anything is sitting please ping me on it and I'll try to move it along.

sneakers-the-rat:
This one is relatively simple - should I update the test cases to reflect the working state? Or should we remove those parameters?

amc-corey-cox:
@sneakers-the-rat I'm sorry my attention was pulled elsewhere for a while and I just got back to this one. Yes, this looks simple and I'd like to get it in. Can you resolve the conflict and I'll try to get it merged right away? It looks like a simple matter of accepting your change but I don't want to end up doing something foolish so I defer to you.

sneakers-the-rat:
alright, so maybe unsurprisingly, fixing this unmasked some other bugs.

So since the previous behavior was to incorrectly ignore the instance attributes on the context genera, and the library had come to expect it to be incorrect, when the contextgen started using its instance attributes a lot of the tests, particularly the snapshot tests which are not specific and always require changes whenver the code changes at all, we had to fix a few tests to explicitly specify the things they were expecting of the context generator.

Additionally, the non-context json-ld generator's CLI was broken, where not passing `--context` results in an empty `tuple()` rather than `None`, so the cli behavior differed from the programmatic behavior, and that was reflected in the tests. so in order to make that possible to pass without just weakening the tests,
- i fixed that problem and cast it to None when not passed
- added the ability to pass kwargs to the context generator from the cli and via the `serialize` method.
- adjusted tests that expected the previous incorrect contextgen behavior to use that.

amc-corey-cox (approved):
This looks good to me.

PR #1990, opened 2024-03-19
Include python version in CI cache key

sneakers-the-rat:
OK this is pretty funny.

I have been wondering for awhile why there would be some bugs related to python version incompatibility that would only sporadically show themselves in CI. 

Eg. https://github.com/linkml/linkml/pull/1989 should have been spotted in https://github.com/linkml/linkml/pull/1964 , but it didn't, what gives!?

Take a look at the python version that pytest reports in the (alleged) python 3.8 test: https://github.com/linkml/linkml/actions/runs/8267528787/job/22618158448#step:11:12

```
platform linux -- Python 3.10.13, pytest-7.4.4, pluggy-1.4.0
```

what in tarnation!??!!?

since we are caching *the whole venv* rather than just `site-packages`, we also restore the python binaries in that venv over whatever we had installed in the prior python install step.

PR adds python version to the cache key, so each python version should have its own cache - ie. its own version of python lmao. 

doh! 🤦

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -68,7 +68,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: .venv
-          key: venv-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}
+          key: venv-${{ matrix.python-version }}-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}
 
       #----------------------------------------------
       # install dependencies if cache does not exist


sneakers-the-rat:
https://github.com/linkml/linkml/actions/runs/8336894542/job/22814770506?pr=1990 now correctly fails (until https://github.com/linkml/linkml/pull/1989 is merged)

sneakers-the-rat:
any cause for delay here? this is gonna allow intermittent, invisible bugs to be introduced until it gets merged :).

doesn't change anything except makes separate caches during CI runs, should be totally function neutral. 

asking bc i am doin some stuff i want to make sure works across python versions and i dont' want to have to make a nox thing <3 not trying to hassle anybody

cmungall:
WOW! Amazing catch. Thank you.

PR #1991, opened 2024-03-19
Make docs dependencies optional and in a separate group

sneakers-the-rat:
in https://github.com/linkml/linkml/pull/1957#issuecomment-2000398930 @amc-corey-cox notes hesitancy adding matplotlib as a dev dependency, which is valid. when in doubt keep dependencies low.

this PR separates docs dependencies as optional so they are only ever installed when building docs, making the default dev environment smaller and also making it so we don't need to be so shy about dependencies for docs :)


cmungall (approved):

cmungall:
wondering if there was a quick github web interface drive by way to resolve the content-hash conflict then fell down a rabbit-hole... https://github.com/python-poetry/poetry/issues/496#issuecomment-738680177

sneakers-the-rat:
Haha yes I am sure there is some optimal way. I tend to think we should lock and relock frequently, and esp if we are changing the deps like this :)

pkalita-lbl:
I like these changes and was going to resolve the conflicts and merge, but I just wanted to confirm there should no longer be any need to do `poetry run pip install 'pydantic>=2'` in these doc workflows, right?

sneakers-the-rat:
Hopefully!?!?!?!?! Docs will fail if pydantic 1 makes its way back in, so at least we'll know lol

pkalita-lbl:
Just out of curiosity what's the connection between Pydantic and docs building that would cause it to fail with Pydantic v1?

sneakers-the-rat:
Check out https://github.com/linkml/linkml/blob/main/docs/schemas/arrays.md?plain=1

It adds a little brittleness to have real code examples in the docs, but I think it makes up for it by always being up to date/no stale examples. (I havent found a satisfying case where I would prefer true doctests vs having tests separated).

In particular here: https://github.com/linkml/linkml/blob/6ea755c417d903ee87b137c366abb85e1e4559f8/docs/schemas/arrays.md?plain=1#L91-L92

I render and compile pydantic 2 models explicitly so that they stay consistent instead of accidentally flipping forms. Its possible to render pydantic 1 and 2 from either 1 or 2, but not possible to compile and use them, which I do to show validation here: https://github.com/linkml/linkml/blob/6ea755c417d903ee87b137c366abb85e1e4559f8/docs/schemas/arrays.md?plain=1#L144-172

(In the future I want to build pyodide in-browser live examples for the whole docs like the proof of concept at https://hackers.piracy.solutions/linkml.html but baby steps ;)

pkalita-lbl:
Ah yes inline examples. That makes sense, thanks.

The only failed check was due to a failed codecov upload. Going to go ahead and merge.

PR #2001, opened 2024-03-21
[pythongen] Simplify slot generation iterators

sneakers-the-rat:
Noticed that most of the logic of the slot and postinit generation iteration logic in pythongen was just sorting, and also made some potentially expensive calls multiple time. unified a sorting method and trimmed down unused/duplicated code. generated kitchen sink and biolink and both were identical.

sorting usually a lil easier to maintain than truth tables <3.

marginal perf benefit, the lions share of time is still in `induced_slot` since ig the changes i made there got deployed but we haven't updated the lockfile here yet?

### DIFF
--- a/linkml/generators/pythongen.py
+++ b/linkml/generators/pythongen.py
@@ -480,7 +480,7 @@ dataclasses._init_fn = dataclasses_init_fn_with_kwargs
         :param cls: class containing variables to be rendered in inheritance hierarchy
         :return: variable declarations for target class and its ancestors
         """
-        domain_slots = self.sort_slots(self.domain_slots(cls))
+        domain_slots = self.sort_slots(self.class_slots(cls))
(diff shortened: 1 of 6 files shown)


sneakers-the-rat:
so i'm trying to make the postinit behavior be identical across classes, and i'm curious - why does most of the validation behavior happen before the `super()` call? 

I would assume that the classes would be structured one of two ways - 
- all slots are rolled down to children, who do all validation except for whatever top-level validation might happen in the root object. all defaults are set by dataclass fields, so are already set by the time the parent tests are called anyway, and so then we want to call all the bottom-level model field validation and coercion last
- no slots are rolled down to children unless they have something special about them, and so the parent classes handle all the validation of the inherited slots.

why would the super() come at the end? i'm sure i'm missing something

Issue #1749, opened 2023-11-24
We need to test before release. `linkml-runtime` release v1.6.1 is broken, and it's impacting the rest of the supply chain (e.g. `bmt` toolkit)


vemonet:
**Describe the bug**
A clear and concise description of what the bug is (including the error message that is printed, if any)

We are using `bmt` to get some ancestors. But since the release of `linkml-runtime` version `1.6.1` creating the `Toolkit()` fails due to error with wikidata URI inside `linkml-runtime`

**To reproduce**
1. `pip install "linkml-runtime ==1.6.1"`
2. Create python script with:
```python
from bmt import Toolkit

biolink = Toolkit()
```
3. Run the script to see the error

**Expected behavior**
Such a basic action should never fails in a released package. 

**Screenshots**

Here is the error trace:

```
File "/usr/local/lib/python3.10/dist-packages/trapi_predict_kit/trapi_parser.py", line 10, in <module>
    biolink = Toolkit()
  File "/usr/local/lib/python3.10/dist-packages/bmt/toolkit.py", line 52, in __init__
    self.view = SchemaView(schema)
  File "/usr/local/lib/python3.10/dist-packages/linkml_runtime/utils/schemaview.py", line 138, in __init__
    schema = load_schema_wrap(schema)
  File "/usr/local/lib/python3.10/dist-packages/linkml_runtime/utils/schemaview.py", line 70, in load_schema_wrap
    schema = yaml_loader.load(path, target_class=SchemaDefinition, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/linkml_runtime/loaders/loader_root.py", line 71, in load
    results = self.load_any(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/linkml_runtime/loaders/yaml_loader.py", line 41, in load_any
    return self._construct_target_class(data_as_dict, target_class)
  File "/usr/local/lib/python3.10/dist-packages/linkml_runtime/loaders/loader_root.py", line 132, in _construct_target_class
    return target_class(**data_as_dict)
  File "<string>", line 66, in __init__
  File "/usr/local/lib/python3.10/dist-packages/linkml_runtime/linkml_model/meta.py", line 591, in __post_init__
    self._normalize_inlined_as_dict(slot_name="slots", slot_type=SlotDefinition, key_name="name", keyed=True)
[...]

Issue #1641, opened 2023-09-20
Outdated lockfile hides test failures


sneakers-the-rat:
**Describe the bug**
The CI installs the package using poetry, which respects the poetry lockfile.
not all installation methods respect the lockfile, and will install the most recent versions of packages that are allowed by the version pinning. This causes the tests to be overconfident. 

eg. Pydantic:

- lockfile pins 1.10.8: https://github.com/linkml/linkml/blob/0d2f3f0d5d65e619a00c81de81a5d1879e3c6830/poetry.lock#L1672
- dependency spec allows '*':  https://github.com/linkml/linkml/blob/0d2f3f0d5d65e619a00c81de81a5d1879e3c6830/pyproject.toml#L115

so a relock currently fetches 2.3.0 which causes the tests to fail because it doesn't allow keyword initialization in class MROs as used by `ConfiguredBaseModel` in the pydanticgen and tests. 

**To reproduce**

```
poetry lock
poetry install
pytest tests/test_generators/test_pydanticgen.py
```

**Expected behavior**
Should tighten up dependency specifiers or add versions to the test matrix

having a slow lockfile is good for keeping production systems stable but i don't think it's as good for in-use code in a language environment where lockfiles are not universally supported, unfortunately :(

kevinschaper:
I wonder if we can use `pydantic = <2.0.0` as a dev dependency while keeping `pydantic = "*"` as a runtime dependency. I think that accurately describes our situation, but I don't know if it's supported.

sneakers-the-rat:
I am one of those "upper bounds are good" ppl, which i know is controversial, but i think pydantic v2 is a decent example of when that is good - it isn't strictly true that the package is compatible with pydantic version `'*'`, since we don't know what v3 will be like!

So my opinions aside, if y'all want to keep the wide open dependency spec, I think the best case would be to add some explicit versions to the test matrix to run the tests for v1 and v2. 

I think tox can handle this ( eg. https://stackoverflow.com/q/59377071/13113166 ), or else you could do it with the github action w/ environment variables+poetry dep specifications or manual pip overrides to install specific versions. I could make a PR for that fi y'all are interested.

Other things that might be good to avoid getting surprised by this problem in the future
- More frequently re-lock, either before every commit or as part of every PR
- Add an additional testing branch to test both installation via pip (which the majority of ppl will probably do in practice) as well as poetry

[...]

Issue #1567, opened 2023-08-04
rdflib v7 introduces breaking changes


cmungall:
When we migrated from v5 to v6 (https://github.com/linkml/linkml/issues/291) we left the `.decode()`s in place; while no longer necessary, they worked in both 5 and 6.

v7 seems to drop support for this altogether, which introduces breaking changes in owlgen

`AttributeError: 'str' object has no attribute 'decode'. Did you mean: 'encode'?`

The quick fix is for schema repos to pin to 6. but we should change the code to remove the decodes - we have already dropped support for rdflib v5 anyway

sneakers-the-rat:
is this literally as simple as removing the `decode` calls? because i'd be happy to make that PR because this is breaking https://github.com/linkml/linkml-project-cookiecutter

cmungall:
yes, that is all that is required, and yes, most welcome!

PR #2002, opened 2024-03-21
Update lockfile

sneakers-the-rat:
Noticed during https://github.com/linkml/linkml/pull/2001 that we're still on linkml-runtime 1.7.0 but 1.7.4 has been released. still not sure what da deal is with relocking but here's one. 

(imo ideally this should be a gh action cron task that relocks, runs tests, and commits if they pass. if we're using the lockfile during testing, we should not let it get too far out of date bc all users install with pip aka not with a lockfile, so we get progressively out of step with what everyone else sees and miss bugs we shouldn't <3)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -12,6 +12,20 @@ files = [
 ]
 
 [[package]]
+name = "annotated-types"
+version = "0.6.0"
+description = "Reusable constraint types to use with typing.Annotated"
+optional = false
+python-versions = ">=3.8"
+files = [
+    {file = "annotated_types-0.6.0-py3-none-any.whl", hash = "sha256:0641064de18ba7a25dee8f96403ebc39113d0cb953a01429249d5c7564666a43"},
+    {file = "annotated_types-0.6.0.tar.gz", hash = "sha256:563339e807e53ffd9c267e99fc6d9ea23eb8443c08f112651963e24e22f84a5d"},
+]
+
+[package.dependencies]
+typing-extensions = {version = ">=4.0.0", markers = "python_version < \"3.9\""}
+
+[[package]]
 name = "antlr4-python3-runtime"
 version = "4.9.3"
 description = "ANTLR 4.9.3 runtime for Python 3.7"
@@ -157,13 +171,13 @@ css = ["tinycss2 (>=1.1.0,<1.3)"]
 
 [[package]]
 name = "cachetools"
-version = "5.3.2"
+version = "5.3.3"
 description = "Extensible memoizing collections and decorators"
 optional = false
 python-versions = ">=3.7"
 files = [
-    {file = "cachetools-5.3.2-py3-none-any.whl", hash = "sha256:861f35a13a451f94e301ce2bec7cac63e881232ccce7ed67fab9b5df4d3beaa1"},
-    {file = "cachetools-5.3.2.tar.gz", hash = "sha256:086ee420196f7b2ab9ca2db2520aca326318b68fe5ba8bc4d49cca91add450f2"},
+    {file = "cachetools-5.3.3-py3-none-any.whl", hash = "sha256:0abad1021d3f8325b2fc1d2e9c8b9c9d57b04c3932657a72465447332c24d945"},
+    {file = "cachetools-5.3.3.tar.gz", hash = "sha256:ba29e2dfa0b8b556606f097407ed1aa62080ee108ab0dc5ec9d6a723a007d105"},
 ]
 
 [[package]]
@@ -416,13 +430,13 @@ files = [
 
 [[package]]
 name = "comm"
-version = "0.2.1"
+version = "0.2.2"
 description = "Jupyter Python Comm implementation, for usage in ipykernel, xeus-python etc."
 optional = false
 python-versions = ">=3.8"
 files = [
-    {file = "comm-0.2.1-py3-none-any.whl", hash = "sha256:87928485c0dfc0e7976fd89fc1e187023cf587e7c353e4a9b417555b44adf021"},
-    {file = "comm-0.2.1.tar.gz", hash = "sha256:0bc91edae1344d39d3661dcbc36937181fdaddb304790458f8b044dbc064b89a"},
+    {file = "comm-0.2.2-py3-none-any.whl", hash = "sha256:e6fb86cb70ff661ee8c9c14e7d36d6de3b4066f1441be4063df9c5009f0a64d3"},
+    {file = "comm-0.2.2.tar.gz", hash = "sha256:3fd7a84065306e07bea1773df6eb8282de51ba82f77c72f9c85716ab11fe980e"},
 ]
 
 [package.dependencies]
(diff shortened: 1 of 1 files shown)


sneakers-the-rat:
oop so ya by updating the lockfile we see a bug that someone installing from pip would experience but we were missing bc old lockfile

sneakers-the-rat:
oh boy i'm just seeing how much history there is i missed on this.

Previously i raised this:
- https://github.com/linkml/linkml/issues/1641
which was closed with this:
- https://github.com/linkml/linkml-runtime/pull/288
which references these other issues:
- https://github.com/linkml/linkml/issues/1749
- https://github.com/linkml/linkml/issues/1567

tagging @sierra-moxon and @pkalita-lbl as the two other ppl i see engaging here consistently.

This is what i said previously:
> with poetry/python most people install through pip or some variant and so don't see the lockfile. ... this specifically breaks non-development installs - it breaking `pip install linkml` and the development install `poetry install` not matching non-development conditions is the problem.

And i see @vemonet [said](https://github.com/linkml/linkml/issues/1749#issuecomment-1829663419) basically the same thing:
> When we publish a python package the poetry.lock is not published (it is only used in development), which means that:
> 
> - We test in an environment with locked dependencies version, not necessarily the latest one
> - When a user install our package, the dependency manager will always try to resolve to the latest available version for each dependency
> 
> For integration tests to be efficient, they need to be the closest possible to the experience the user will have using the package. In this case this means the tests needs to be run after resolving the latest version of each dependency (instead of using locked versions).

but i disagree with the argument that we should not version the `poetry.lock` file, more on that in one second, just trying to bring background here.

@pkalita [said](https://github.com/linkml/linkml-runtime/pull/288#pullrequestreview-1760712250):
> We have to consider the trade-offs between reproducibility (by including the lock file) and faster failures due to new dependency versions (by omitting the lock file). Given the limited resources we have to devote to LinkML, we've opted for reproducibility.

So i think there is some justified lack of clarity about "whats the point of the lockfile anyway"

Yes the purpose is to allow for someone to (fingers crossed) perfectly reproduce a python environment. But since lockfiles aren't a python standard, but a poetry novelty, we don't have the same kind of use for them that yarn/npm do where each package can exist in its own perfect reality - instead python packages are all installed in the same soupy environment. 

The [poetry docs](https://python-poetry.org/docs/basic-usage/#committing-your-poetrylock-file-to-version-control) do not help with clarity here, and i think actually make it worse: 

> A simple way to avoid such a scenario is to omit the poetry.lock file. However, by doing so, you sacrifice reproducibility and performance to a certain extent. Without a lockfile, it can be difficult to find the reason for failing tests, because in addition to obvious code changes an unnoticed library update might be the culprit. Further, Poetry will have to lock before installing a dependency if poetry.lock has been omitted. Depending on the number of dependencies, locking may take a significant amount of time.
>
> If you do not want to give up the reproducibility and performance benefits, consider a regular refresh of poetry.lock to stay up-to-date and reduce the risk of sudden breakage for users.

They seem to forget that **not only is it possible to install the package with pip, it is almost always how packages using poetry are installed.**

So the tradeoff is not between committing the lockfile or not, and it's not between reproducibility and not - the important thing is we need to know when the lockfile is useful. 

- testing off a lockfile is useful because you should get the same results on different machines (modulo OS and any other thing poetry doesn't lock) - that's good when we want someone to be able to submit a PR with a specific environment attached, etc.
- testing off a lockfile is *not* useful because it's disconnected from the POV of literally everyone else using the package - if a test fails because a new package was released, **that means that the test will fail during normal usage of the package.** It means that a) our version bounds need to be revised, or b) we have a bug to fix. 
 
so the point the poetry docs re "not knowing if a bug is caused by a package update or new code" is misleading by a mile imo - if there is a bug that's difficult to explain, one can **install the package from the lockfile afterwards and run the tests again** and if that fixes it, diagnose the source of the problem in one step. 

so tl;dr - 
- we should test using `pip install .` in regular CI
- we should *also* periodically (and imo automatically) relock and test.

---

as a side note, re: https://github.com/linkml/linkml-runtime/pull/288
we don't need to delete the lockfile during testing. `pip install .` does the job and is actually what we want here. we don't need to use poetry's dep resolver bc it might differ from pip and our tests being aligned with the experience of running the package is what we want. also that PR removes the lockfile after the package is installed anyway so it still would have used it when installing <3

pkalita-lbl:
Thanks for writing up your thoughts on this. I think we're totally in alignment on the scope of the problem. And I just wanted to clarify that I fully appreciate that the `poetry.lock` file is only part of the equation in a Poetry-managed development environment (i.e. my local machine, a GitHub runner if `poetry install` is used, etc). 

Let me also just elaborate on the scenarios I'm thinking about that I'd prefer to avoid:

* If we had an automated process that runs `poetry update` and unconditionally commits the resulting `poetry.lock` changes into `main`, that seems like a really easy way for CI to "go red" and never return. People see a failing test in their PR and see it's also failing in `main`. They say "ah, good, not my problem" and move on.
* If we removed `poetry.lock` from git or did `pip install .` in CI, tests may fail on your PR because of an updated dependency and not because of your changes. To be clear, _finding these issues earlier would be a good thing_. However. The downside is that it _may_ turn a small PR into a rabbit hole of understanding unrelated test failures. This _may_ be a turn off to outside contributors, and to be honest sometimes core LinkML developers don't have time to go down the rabbit hole.

What I think would make more sense for this project would be either:

1. Have an automated process that runs `poetry update` but only opens a PR with the changes. A member (or members) of the core development team would need to be assigned to the PR for review, and they would be responsible for debugging and resolving any test failures before merging the PR. OR...
2. I don't know exactly if/how the automation for 1 would work. If it's not feasible, have a rotation of core developers manually do the `poetry update`/fix (if necessary)/commit on a periodic basis.

Either way we're limiting the number of people who potentially have to go down the dependency-upgrade-test-failure rabbit hole and also allowing them to do it when it works with their schedule (hopefully).

I _believe_ (although the meeting notes aren't exactly backing me up) that we talked about this on a developers call and roughly agreed to keeping the CI process the same but updating the `poetry.lock` file more regularly by some mechanism. @sierra-moxon or @cmungall might be able to tell me if I'm just hallucinating that. If we're all still on board with that I think the next step would be investigating if the automation for 1 above is feasible and deciding on an appropriate schedule.

---

With all that being said since this PR just updates the `poetry.lock` file and all the tests pass I'm happy to approve and merge if you're ready, @sneakers-the-rat.

sneakers-the-rat:
I am ready! And I think the "PR with lockfile updates" makes sense! Maybe it automerges if tests pass, but otherwise doesnt? We also probably want to turn on branch protections to main requiring that a PR be up to date (I think you can currently merge if there are no conflicts, but forcing a merge/rebase might prevent weird problems from like accidentally backtracking)

Happy to PR that, ive seen it done a few times. Maybe it runs once a week or so? And always from the same branch so if an old PR is still open it can just get updated instead of stacking up.

sneakers-the-rat:
It might also be good to make a habit of putting the developer call discussions into the docs - if not as like a "when we make decisions we make sure theyre reflected in the developer docs" kinda thing, then copying whatever notes yall take into a changelog style section in the docs. Good to keep everyone on the same page by making stuff public and referenceable :)

pkalita-lbl:
> Maybe it automerges if tests pass, but otherwise doesnt?

If that's automatable that would be great.

> We also probably want to turn on branch protections to main requiring that a PR be up to date (I think you can currently merge if there are no conflicts, but forcing a merge/rebase might prevent weird problems from like accidentally backtracking)

I hadn't though about that, but on first glance yeah that might be a good idea

> Maybe it runs once a week or so? And always from the same branch so if an old PR is still open it can just get updated instead of stacking up.

I was thinking bi-weekly but obviously it wouldn't be set in stone or hard to change later so I don't have a strong preference. The same branch thing is a good idea, too.

> It might also be good to make a habit of putting the developer call discussions into the docs

I'll bring it up on the next developer call 😂

pkalita-lbl (approved):

rly:
How about this github action (see example weekly workflow): https://github.com/Apakottur/action-poetry-package-update 

I haven't tried it, but it might be as easy as adding this workflow.

sneakers-the-rat:
that one looks like it's gonna update the deps in `pyproject.toml` too, but yes something like that :)

PR #2004, opened 2024-03-21
Move sort classes from python generators to linkml-runtime

sneakers-the-rat:
Can only be merged after: 
- [x] https://github.com/linkml/linkml-runtime/pull/313 is merged
  - [x]  and deployed and
- [x] linkml's lockfile is updated to reflect that release

Starting to tidy up pydanticgen, noticed first off a `TODO` saying that the sort classes method should move to schemaview. definitely agreed. 

removed it and the identical one from the python generator and updated the places where it's called. 

Other changes: I also noticed i wasnt' getting type hints for `self.schema` from the parent Generator class, so i went looking for why, and in the process of adding the hint noticed that the init from schemaloader method had some redundancies and cleaned them up.


vladistan:
+1 for this.  I see this problem a lot with gendoc

sierra-moxon:
closing draft PR, see issue to continue work: https://github.com/linkml/linkml/issues/2702

PR #2011, opened 2024-03-22
Adding documentation for linkml expression language

turbomam:
and linked from docs/schemas/advanced.md

### DIFF
--- a/docs/schemas/advanced.md
+++ b/docs/schemas/advanced.md
@@ -160,5 +160,7 @@ values is None then the entire expression evaluates to None.
 You can insert missing values by using the `--infer` option when
 running `linkml-convert`.
 
+Documentation of the expression language is available [here](../schemas/expression-language).
+
 See the developer documentation on [inference](../developers/inference) for
 details of how to use this in code.
--- a/docs/schemas/expression-language.md
+++ b/docs/schemas/expression-language.md
@@ -1,0 +1,114 @@
+# Expression Language
+
+The Expression Language is a domain-specific language (DSL) that allows for the evaluation of expressions with a syntax similar to Python. It provides a restricted subset of Python's features to ensure safe and controlled execution.
+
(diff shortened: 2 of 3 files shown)


sneakers-the-rat:
omg this is amazing i didn't know this existed!!!! 

questions:
- where is the parser? can we add that to the API docs and then link to it from this page? (i wanna check it out!)
- where can one use these expressions?
- what variables are in scope for an expression? eg. what can i use as `x, y, z` within the expression?

cmungall (approved):

cmungall:
The main way in which you'd use them is (very loosely) documented here:

https://linkml.io/linkml/schemas/advanced.html#equals-expression

Here's an example of inference in Python:
https://linkml.io/linkml/developers/inference.html

I am not sure if you've taken a gander at the LinkML Transformer framework, but there's an example of how they can be used in schema mapping here:
https://github.com/linkml/linkml-transformer/blob/main/notebooks/Tutorial.ipynb

By using a safe declarative subset of python we can in theory map transformations to e.g SQL/DuckDB, SPARQL, ... even FOPL theorem provers

You can run the [metacircular interpreter](https://en.wikipedia.org/wiki/Meta-circular_evaluator) independently, see:

 - https://github.com/linkml/linkml-runtime/blob/main/linkml_runtime/utils/eval_utils.py) 

We don't expose this in the sphinx docs and in fact this could become an upstream module

sneakers-the-rat:
Ohhhhhh I like this a lot. I have seen equals_expression before but didnt know there was a full syntax. Im gonna implement it in pydanticgen as an `after` validator.

I have taken a brief look at transformers but not enough, I want to see how youre handling some of the tricker transforms I havent been able to do declaratively like level shifting/nesting, splits and merges with type casts, etc.

Thx for letting me butt in :)

Issue #2007, opened 2024-03-22
Docker build failing for release tags


pkalita-lbl:
When 1.7.6 was released the Docker build and push workflow failed: https://github.com/linkml/linkml/actions/runs/8368783078. Specifically, during checkout the following error occurred:

```
Fetching the repository
  /usr/bin/git -c protocol.version=2 fetch --prune --progress --no-recurse-submodules --depth=100 origin +4977f426cf4f3971cd66ea0dbae8fc0d039c7799:refs/tags/v1.7.6
  Error: fatal: Cannot fetch both 4977f426cf4f3971cd66ea0dbae8fc0d039c7799 and refs/tags/v1.7.6 to refs/tags/v1.7.6
```

I believe the culprit was the changes introduced in https://github.com/linkml/linkml/pull/1980. A brief glance at https://github.com/actions/checkout/issues/1467 indicates that there may be a problem with using both `fetch-depth` and `fetch-tags`, but I haven't read through that issue in great detail.

cc @sneakers-the-rat

sneakers-the-rat:
Oh boy you love to see issues like that with a zillion other issues referencing them. Got it in a bit

cmungall:
I am not sure if this is related:

https://twitter.com/chrismungall/status/1748530964605948336

(rarely post to X but was hoping to see a bit more engagement there)

sneakers-the-rat:
This is part of why I do not care for docker

PR #2017, opened 2024-03-23
[ci] only checkout with tags for non-tag runs

sneakers-the-rat:
Fix: https://github.com/linkml/linkml/issues/2007

@pkalita-lbl basically did all the work for this - 
- The cloned version of the repository needs to have at least one tag in the current branch's history in order for dynamic versioning to work
- https://github.com/linkml/linkml/pull/1980 made the docker build action do a depth=100 fetch instead of a depth=1 fetch to catch the most recent tagged version (hopefully, it's sort of a hack since there's no builtin method of "fetch up to the most recent tag in this branch").
- I wasn't sure if tags got fetched without the tags option ( since that's what it was introduced for here: https://github.com/actions/checkout/pull/579 ), and looking now it seems like they [don't](https://github.com/linkml/linkml/actions/runs/8319237799/job/22762230217?pr=1980#step:2:51)
- The combination of a tag-triggered build and with-tags specifically breaks the checkout action, eg for the same commit:
  - this fails: https://github.com/linkml/linkml/actions/runs/8368783078
  - this doesn't: https://github.com/linkml/linkml/actions/runs/8367682830
- That's because of a [malformed git fetch command](https://github.com/linkml/linkml/actions/runs/8368783078/job/22986576495#step:2:123): 
```
/usr/bin/git -c protocol.version=2 fetch --prune --progress --no-recurse-submodules --depth=100 origin +4977f426cf4f3971cd66ea0dbae8fc0d039c7799:refs/tags/v1.7.6

  Error: fatal: Cannot fetch both 4977f426cf4f3971cd66ea0dbae8fc0d039c7799 and refs/tags/v1.7.6 to refs/tags/v1.7.6
```

[compared to](https://github.com/linkml/linkml/actions/runs/8367682830/job/22910511296#step:2:51)

```
/usr/bin/git -c protocol.version=2 fetch --prune --progress --no-recurse-submodules --depth=100 origin +4977f426cf4f3971cd66ea0dbae8fc0d039c7799:refs/remotes/origin/main
```

I've never actually used `refspec` syntax like that before, but the docs are here: 
- https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt-ltrefspecgt
- https://git-scm.com/book/en/v2/Git-Internals-The-Refspec

so basically the second one works because it will fetch that commit and create a ref under the usual `refs/remotes/origin/main` and since the `--no-tags` flag isn't given, it will also grab the remote ref for the tag into `/refs/tags/v1.7.6` - two refs to two refs. the second one fails because it is fetching the commit and the tag into the same ref `/refs/tags/v1.7.6`. That's because a tag push is a separate event from a commit push (the linked examples are from the commit vs tag push for a given commit) and thus has the tag ref as the current `{{github.ref}}`.

since the purpose for the change in the first place was to make sure we have tags, and the tag-triggered actions by definition fetch the tag, introducing a switch that only fetches tags when the action is triggered by a commit. doing it this way vs. adding an additional `run` command bc of the extremely mild overhead of adding an extra step.

opening this PR not as a draft while still working on it so that CI runs bc it's a CI-based PR, sry for the notifs.

theoretically this should apply to all CI actions but docker is the only action that runs on a new tag. unsure why that is, just fixing it.

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -2,11 +2,10 @@ name: Build Docker Image
 
 on:
   push:
-#    branches:
-#      - "main"
+    branches:
+      - "main"
     tags:
-     # - "v*"
-      - "*"
+      - "v*"
   pull_request:
     branches:
       - "main"
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -18,7 +18,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
         with:
-          fetch-tags: true
           fetch-depth: 100
 
       - name: Docker metadata
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -19,6 +19,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 100
+          fetch-tags: ${{ github.event_name == 'push' && !contains(github.ref, 'refs/tags/') }}
 
       - name: Docker metadata
         uses: docker/metadata-action@v4
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -2,10 +2,11 @@ name: Build Docker Image
 
 on:
   push:
-    branches:
-      - "main"
+#    branches:
+#      - "main"
     tags:
-      - "v*"
+     # - "v*"
+      - "*"
   pull_request:
     branches:
       - "main"
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -19,7 +19,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 100
-          fetch-tags: ${{ github.event_name == 'push' && !contains(github.ref, 'refs/tags/') }}
+          fetch-tags: ${{ github.event_name == 'push' && ! contains(github.ref, 'refs/tags/') }}
 
       - name: Docker metadata
         uses: docker/metadata-action@v4


sneakers-the-rat:
OK so i don't want to commit a tag that looks like a version without asking first because that might trigger some unwanted deploy, but i do need to push a tag in order to trigger the bug.

so i made a temporary tag `test-tag-remove-me` and that solves the checkout problem: https://github.com/linkml/linkml/actions/runs/8399144254/job/23004877707?pr=2017

it introduces another problem, presumably because the tag name no longer is a semver, and that metadata extraction step looks like it needs the tag to be. that should not happen in the normal conditions that this action would be run in, but again i don't want to test that by myself rn.

So I am reverting the changes [here](https://github.com/linkml/linkml/pull/2017/commits/6d1bf8c039f2b8b1f8cd96536b0b1d18a1539904) that were needed to trigger the tag build outside of `main` and saying this PR is ready for review - whether y'all want to test it by pushing a semver tag or building the docker image without deploying it, up to y'all.

sneakers-the-rat:
Alright it looks like my conditional gh action test

```
fetch-tags: ${{ github.event_name == 'push' && ! contains(github.ref, 'refs/tags/') }}
```

which should be `false` if it's a tag push but `true` otherwise doesn't work. so we're back to the old behavior of fetch-tags being false for the docker build, which means that the python package inside of it won't have a meaningful linkml version, but it will if we only deploy the image on the tagged builds. 

that's about as much time as i have for solving a docker problem! up to y'all whether we want to leave it like that or just take that line out. won't be offended if you close this and just PR removing the line, idrc about the docker stuff!

pkalita-lbl:
Did you consider this?

```yaml
uses: actions/checkout@v4
with:
  fetch-depth: 0
```

According to [the docs](https://github.com/actions/checkout?tab=readme-ov-file#usage):

> 0 indicates all history for all branches and tags.

That's what we use on other projects I work on, and it seems to work fine. I was actually surprised to see we weren't already using it here. I believe that would make the dynamic versioning correct for both tag-based and non-tag-based builds. The downside is perhaps a slightly slower checkout? I'm not sure you'd even really notice in practice.

sneakers-the-rat:
Hell ya the only reason I didn't was bc it wasnt already like that and I didnt want to be the one to blast the full clone if we were trying to sip resources for some reason. Feel free to edit/close this PR and do that :)

pkalita-lbl:
I had trouble updating this branch (probably my fault) so I just created a PR on a new branch: https://github.com/linkml/linkml/pull/2022

PR #2018, opened 2024-03-23
Make compliance tests fail and skip

sneakers-the-rat:
I was trying to understand how the compliance tests work, and like in https://github.com/linkml/linkml-runtime/pull/312 i noticed there were a number of places that should be test failures or skips.

I went through and tried to change very little, just turning logging messages that say 'skipping testing' into skips and places where it seemed like there should be a test failure into a test failure, lmk where i was incorrect with that.

it looks like there are a number of masked failures, and marking tests as skips helps us keep track of which we still need to implement. 

|     | main | this pr | 
| -- | ---- | ------ |
| passed | 5533   | 1981 |
| skipped | 936  | 4379 |
| failed | 0  | 110 |

this PR is mostly to try and raise the question if this is expected or not, and ask more generally what the plan is for the compliance tests and how they're supposed to work? in a longer term sense, a lot of why i'm pitching in heavily at the moment on testing and CI is that i'm trying to lay a base for future work i want to build off of with linkml, and am also still trying to scope out how much work it would take to do that vs. build on my own and meet up later. I see the git blames for the compliance tests are more recent than elsewhere, so is this a new test suite that has a lot of skips because it's new? i think this kind of testing is ambitious and cool and if so i want to help out to try and make it robust and implementable. Trying to understand this thing, so let me know where i'm wrong!

I also see a reporting system started here, but I can't really tell how to read the outputs i'm getting - is that ongoing work on the system for reporting the capabilities? if so i think this change lets us fail/skip the tests faster and have better visiblity - we can hook into pytest's collection/teardown features to do that as well while making the report visible in the actual testing output.

### DIFF
--- a/tests/test_compliance/helper.py
+++ b/tests/test_compliance/helper.py
@@ -159,7 +159,7 @@ class FeatureSet(BaseModel):
     features: List[Feature]
 
 
-cached_generator_output: Dict[Tuple[SCHEMA_NAME, FRAMEWORK], Tuple[Generator, str]] = {}
+cached_generator_output: Dict[Tuple[SCHEMA_NAME, FRAMEWORK], Tuple[Generator, str, Path]] = {}
 """Cache generators and their outputs to avoid repeated computation."""
 
 all_test_results: List[DataCheck] = []
@@ -255,7 +255,7 @@ def _get_metamodel_elements(obj: Any) -> Iterator[str]:
         pass
 
 
-def _generate_framework_output(schema: Dict, framework: str, mappings: List = None) -> Tuple[Generator, str, str]:
+def _generate_framework_output(schema: Dict, framework: str, mappings: List = None) -> Tuple[Generator, str, Path]:
     """
     Compile a schema using a framework (e.g. jsonschema generation).
 
@@ -270,10 +270,6 @@ def _generate_framework_output(schema: Dict, framework: str, mappings: List = No
     if schema["name"] not in schema_name_to_metamodel_elements:
         schema_name_to_metamodel_elements[schema["name"]] = list(_get_metamodel_elements(schema))
     if pair not in cached_generator_output:
-        if mappings is None:
-            # this should only happen when executing individual pytest combos
-            mappings = []
-            logging.warning(f"No mappings for {pair} - called out of order?")
         gen_class = GENERATORS[framework]
         if isinstance(gen_class, tuple):
             gen_class, gen_args = gen_class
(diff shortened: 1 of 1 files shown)


cmungall:
Yes the goal is to produce detailed compliance reports as a website

See
https://github.com/orgs/linkml/discussions/1549

I think we may also have a recording of the dev meeting discussing this.

The priority is transparency for linkml users >> transparency for core devs
(though latter is obv important). This is also part of than plan for linkml
ports to other languages, extending validation strategies etc

There are likely many things that can be improved here but best to discuss
first before diving in with PRs, to make sure goals are aligned!

sneakers-the-rat:
Ah! thank you https://github.com/orgs/linkml/discussions/1549 does make it clearer :)

> The priority is transparency for linkml users >> transparency for core devs
> (though latter is obv important). This is also part of than plan for linkml
> ports to other languages, extending validation strategies etc

hell yes! I think we can do both. I see the pieces are in place for reporting, and a note in 1549 says we want to make markdown summaries from it. i would love to help with that bc i think this is a great idea!! Summary could be dumped to JSON during test runs and versioned, with a test to check if the compliance results have changed, and then the docs could read that and emit a markdown version of it when they are built. 

We could use pytest hooks to gather results so we don't need to worry about reaching the end of a helper function, so that way we can get both - clarity for devs by using `pytest.skip` and test failures as expected, and also human readable output of test results in markdown. 

the `pytest_terminal_summary` hook could be used to print the results there after a test run, `pytest_report_teststatus` also could be used for the actual report collection, and then `pytest_sessionfinish` to write the results out. 

> There are likely many things that can be improved here but best to discuss
> first before diving in with PRs, to make sure goals are aligned!

Yes! the reason I opened this one was to start conversation about that, as well as having a CI run on the books demonstrating what i'm talking about re: results - we can close this if we don't like it, but i raised this without discussion just bc it seemed like we probably won't want `assert`s hardcoded to `True`, and the logging messages indicating a skip probably should just be a `pytest.skip()`, and then if my read was right on what was supposed to be happening here i could demo how to pull in those skips/fails/etc. using the pytest hooks to report summaries of the compliance tests

some of the test failures seem substantive, like [dropping a value on a roundtrip](https://github.com/linkml/linkml/actions/runs/8400490003/job/23007823563#step:11:489) or [uri problems](https://github.com/linkml/linkml/actions/runs/8400490003/job/23007823563#step:11:1796) and whatnot, so i also figured it would be useful to document those since i noticed them in local testing. 

lmk if there is something else we want here, because i would love to help out finishing up the reporting stuff and documenting how these work and all that!

PR #1927, opened 2024-02-15
Pydanticgen - Modularize template

sneakers-the-rat:
Another cleanup step for the pydantic generator - modularizing the templates and coupling them with models. Typing quickly and opening a draft rn but my battery is about to die and so i'll edit this with more info once i find my charger.

pretty big PR but this one makes very few changes to the actual behavior of the generator. mostly a refactoring

goals:
- be able to generate each of the components of a schema separately for testing and also for runtime use
- be able to inspect what's going on in the template more easily
- have models for the endpoint so it is clear what the generator supports - eg. the attribute model supports `maximum_value` etc. but does not support literally all possible values for a slot definition. 
- make it clearer where one should hook into the generation process
- make future steps of cleaning up the pydantic generator easier by having less anonymous objects

todo:
- [x] clean up the code and templates - i have a hell of a time with jinja whitespace formatting so i just unindented everything, but it could be made more readable. Also there are some odd ordering and mix and match syntax i need to fix
- [x] make tests pass - i thought i had replicated the existing template pretty exactly, but i'm getting mysterious errors from inlining and unenforced constraints. will follow up tmrw
- [x] investigate kwargs unpacking in jinja templates because it's pretty brittle to need to hardcode all the arguments in the macro calls - not possible
- [x] test new functionality - imports `__add__` method most notably, rest is pretty testing neutral so far.

edit: ok returned

## template structure 

the strategy is pretty self explanatory i think, split each separate part of the template into its own sub-template. i used jinja macros here to keep the iteration in the top level template so each individual template could be used on its own. As is the top level `PydanticModule` class holds everything that should be given to the `module` template, and that holds for the rest of the subtemplates except for `attribute` and `validator`. 

### alternative

Another strategy I was thinking about was to give the `TemplateModel` metaclass a `build` method that subclasses could override that does the actual rendering for each piece, and then they get stitched together - that would make for nice clean build syntax where the relationship between the model and the template is super clear, but wouldn't be the single jinja template render call, idk what's good there.

### template values

i have thought it odd that there are two additional template properties for precomputed field values and class bases, so i moved both of those to the attribute and class models, respectively. 

I also changed the pydantic version parameter to an int because it is so it would be possible to make range comparisons in case godforbid there is a v3 debacle. at the moment i'm seeing some places where it's still a string and will fix. this change was made in this PR because pydantic version has to propagate down through the templates and i was getting iffy with it being a string

### imports

[...]

Issue #1094, opened 2022-11-03
pydanticgen doesn't support inlined dicts


kevinschaper:
It doesn't seem to be possible to get an inlined dictionary with a class range. 

Using this example schema:
```
classes:
  person:
    slots:
      - id
      - bikes
    attributes:
      - bikes:
          range: bike
          multivalued: true
          inlined: true
          required: true
  bike:
    slots:
      - name
      - color
    slot_usage:
      name:
        identifier: true
        required: true
slots:
  - id:
      required: true
  - name
  - color
  - bikes
```

With gen-typescript I'm getting: 

```
export interface Person  {
     id?: string,
    bikes?: {[index: BikeName]: Bike },
}
```

but gen-pydantic is producing:
```
class Person(ConfiguredBaseModel):    
    id: str = Field(None)
    bikes: List[Bike] = Field(default_factory=list)
```

I think I should get 

```
class Person(ConfiguredBaseModel):    
    id: str = Field(None)
    bikes: Dict[str, Bike] = Field(default_factory=dict)
```

PR #2019, opened 2024-03-24
[pydanticgen] Simplify pydantic generator serialization method

sneakers-the-rat:
Part of a continuing series of refactoring the pydanticgen.

Should not be merged until: 
- [ ] https://github.com/linkml/linkml/pull/1887 is merged
- [ ] Combine ClassResult with BuildResult in 1887, correctly propagating SlotResult imports

When last we left it in https://github.com/linkml/linkml/pull/1927 i had made a mess at the bottom of the serialization method casting the previous serialization method result into the new template models. 

This PR refactors the serialization method into a `generate_class` and `generate_slot` method, splitting up the logic that was all sorta piled up in the serialization method. 

I also started transforming some of the `get` methods into properties, since we want to minimized repeated calls to `induced_slots` as much as we possibly can, and those should need to only be computed once and have a static result (ie. be properties not methods).

Quite a ways to go yet, but another incremental step there :)

### DIFF
--- a/linkml/generators/pydanticgen/pydanticgen.py
+++ b/linkml/generators/pydanticgen/pydanticgen.py
@@ -2,7 +2,6 @@ import inspect
 import logging
 import os
 from collections import defaultdict
-from copy import deepcopy
 from dataclasses import dataclass
 from pathlib import Path
 from types import ModuleType
@@ -15,13 +14,13 @@ from jinja2 import ChoiceLoader, Environment, FileSystemLoader
 from linkml_runtime.linkml_model.meta import (
     Annotation,
     ClassDefinition,
-    SchemaDefinition,
     SlotDefinition,
     TypeDefinition,
 )
 from linkml_runtime.utils.compile_python import compile_python
 from linkml_runtime.utils.formatutils import camelcase, underscore
 from linkml_runtime.utils.schemaview import SchemaView
+from pydantic import BaseModel
 from pydantic.version import VERSION as PYDANTIC_VERSION
 
 from linkml._version import __version__
@@ -94,6 +93,15 @@ DEFAULT_IMPORTS = (
         ),
     )
 )
+
+
+class ClassResult(BaseModel):
+    """
+    TODO: Merge with BuildResult when array implementations is merged
+    """
+
+    cls: PydanticClass
+    imports: Optional[List[Import]] = None
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


sneakers-the-rat:
In fixing the failing test, i both noticed that the test was a string comparison test and that the test seems to be incorrect. 

Stems from https://github.com/linkml/linkml/issues/1094

The model in question has a slot `has_bikes` that is inlined as a dict and required. 

formerly, the test produced a field annotation like this:

```python
bikes: Dict[str, str] = Field(default_factory=dict)
```

but that would allow a model to be instantiated without `bikes`, ie `Person()` would incorrectly validate (`id` also generates a default of `None` even when it's required).

The annotation is also too general - it shouldn't be a dictionary of strings, but a dictionary of `Bike` instances. Looking through the `_inline_as_simple_dict_with_value` , it looks like the `str` annotation is from the situation where you have the `id` of the object, but the `Dict` would incorrectly fail to validate if it was a dict of `Bike` instances.

The correct annotation, i think, should be:
```python
bikes: Dict[str, Union[str, Bike]] = Field(...)
```

so I updated the code and the test to match, and also added a test for list inlining.

This should generally be true, right? 
- if a slot is `required`, i would expect there to be no default or default factory (unless there is an `if_absent` or other precomputed value), and for models to fail validation is no value is passed. 
- if a slot is not `required`, the type annotation should be wrapped in `Optional[]` and the default should be `None`. 
 
[`Optional` is an alias for `Union[X, None]`](https://docs.python.org/3/library/typing.html#typing.Optional), so there should be a three way definitional equality
`required: false === Optional[] === default = None`

sneakers-the-rat:
This PR also:
- modifies the array generators to create a `RangeResult` rather than a `SlotResult` because they're just generating ranges, and the terminology is more uniform that way (they are called `ArrayRangeGenerator`s). 
- implements a `SlotResult` as well as a `ClassResult` so that we can propagate side-effects of generating "lower level" objects up to the module, in this case imports and injected classes, in a uniform way. Still some work to do on that but another step towards a reasonable build system. 
- uses plain strings rather than `Annotation` objects, which were just a sort of perplexing carrier of a string anyway.
- creates a `_clean_injected_classes` method, trying to keep clutter out of the `render` method.

sneakers-the-rat:
tests pass - bump on reviewing this, this is a blocker for https://github.com/linkml/linkml/pull/2036 and other future planned improvements to pydantic generator. basically a tech debt PR that makes few if any changes to behavior, just cleaning up code, breaking it up into discrete sections, reducing the amount of implicit mutation in place, etc.

once this is merged i'll keep working on https://github.com/linkml/linkml/pull/2036 :)

cmungall (approved):

sneakers-the-rat:
This probably needs to be updated, but before i do that can i get a nod if we would merge this? working on pydanticgen again and would really love to have this cleanup :)

sneakers-the-rat:
merged in main, would be nice if this could get merged before the generator drifts again :)

pkalita-lbl (approved):

PR #2035, opened 2024-03-27
Ensure unique key constraints on foreign keys are mapped in SQL transform.

### LINKED ISSUE #2006: SQL schema generation fails for "complex" unique_key_slots
**Describe the bug**
SQL schema generation fails on linkml v1.7.5 and v1.7.6 if the range of a "unique_key_slot" is an array or an object.

**To reproduce**
1. Create a schema with a slot which range is an array and enter that slot into `unique_key_slots`.
2. Run `gen-sqltable` or `gen-sqlddl` on the schema.
3. See error `sqlalchemy.exc.ConstraintColumnNotFoundError: Can't create UniqueConstraint on table 'MyClass': no column named 'my_slot' is present.`

**Expected behavior**
Successful SQL schema generation.

**Additional context**

According a `git bisect` the issue appeared on commit c5b908fe931cbd81ba4e7b7e08d11599f92921b3.

Error can be reproduced with following schema:
```
id: https://examples.org/my-schema
prefixes:
  myschema: https://examples.org/my-schema
imports:
  - linkml:types
default_prefix: myschema
default_range: string

classes:
  MyClass:
    slots:
      - my_slot
    unique_keys:
      identifier:
        unique_key_slots:
          - my_slot

slots:
  my_slot:
    multivalued: true
    range: string
```

cmungall:
the SQLTableGenerator first pre-processes using RelationalModelTransformer,
including

 - mapping multivalueds to backrefs
 - translating ranges of classes to foreign keys, using (x_id) conventions

Previously the slots referred to in unique_keys were not mapped
forward, see #2006. This PR fixes this

- unique slots are mapped forward where possible
- where not possible (multivalueds) the constraint is dropped

Fixes #2006

### DIFF
--- a/linkml/generators/sqlalchemygen.py
+++ b/linkml/generators/sqlalchemygen.py
@@ -121,6 +121,7 @@ class SQLAlchemyGenerator(Generator):
             is_join_table=lambda c: any(tag for tag in c.annotations.keys() if tag == "linkml:derived_from"),
             classes=rel_schema_classes_ordered,
         )
+        logging.debug(f"# Generated code:\n{code}")
         return code
 
     def serialize(self, **kwargs) -> str:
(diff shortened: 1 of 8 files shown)


cmungall:
arbitrarily failing on some jobs.. I think this is the venv cache/shacl issue. Is there a way to trigger clear cache?

Issue #2005, opened 2024-03-21
adding metadata to pydantic fields


djarecka:
**Is your feature request related to a problem? Please describe.**

We are working with the pydantic (v.2) representation of the model, but there are important pieces of informations that are missing from the fields, e.g. `exact mapping` or `local_names` that we need to know to get the correct type of data from external sources.

**Describe the solution you'd like**

I think `exact_mapping`, `local_names` etc. could be added to `metadata` kwarg. of the pydantic `Field` class

**How important is this feature?** Select from the options below:
• Medium - can do work without it; but it's important (e.g. to save time or for convenience)

I would say that the information is critical, but we are thinking about way around: using linkml tools to view the yaml schema 

**When will use cases depending on this become relevant?** Select from the options below:
• Short-term - 2-4 weeks

**Additional context**

sneakers-the-rat:
gotcha https://github.com/linkml/linkml/pull/2036

Issue #1830, opened 2024-01-16
"LinkML class attributes" are lost in gen-python


kappe-c:
**Is your feature request related to a problem? Please describe.**

Working with a schema in Python, previously generated with the gen-python tool, I would very much like to have access to the "class attributes" defined in the LinkML schema for my classes, ideally with the Pythonic dot syntax.
With "LinkML class attributes" (in quotes) I mean the attributes a class can have according the meta model. In Python, class attributes would be the actual correct name (analogously to class methods).
Simple example. File schema.yaml:
```yaml
id: https://example.org/test-schema
name: testschema

imports: https://w3id.org/linkml/types

classes:
  
  MyClass:
    abstract: true
    attributes:
      my_attr:
        range: string
```

**Describe the solution you'd like**

Having used `gen-python schema.yaml > schema.py` I would expect to be able to do something like the following in a Python script.
```py
assert MyClass.abstract == True
# or e.g.
if hasattr(MyClass, 'abstract'):
  print(MyClass.abstract)
```
Similar for *comments*, *todos*, etc.
Honestly, I was very surprised that this expected behavior was not implemented. Did I miss anything? Is this intentional? I would not mind if this were kept in, say, `MyClass.class_attributes` or whatever, but I would very much like to access the stuff I have defined in my schema.

(This is not a feature request for having actual class attributes in LinkML that can be defined by the user (attributes for which each instance has the same shared value.) Even though I have missed this at times.) 

**How important is this feature?**
• Low to Medium

**When will use cases depending on this become relevant?**
• Short-term - 2-4 weeks (actually right now already in my project)

sneakers-the-rat:
related to https://github.com/orgs/linkml/discussions/1820

cmungall:
What is the best way to do this that maximizes compatibility across frameworks?

Should these be dunder variables (this minimizes potential clashes with user-defined slots)? Or should it be done in a more pydantic way?

Would it make sense to try and have e.g. javagen follow the same patterns?

dalito:
For Python, I like @sneakers-the-rat´s proposal to store the information as an attribute more than having lots of private / dunder variables.

PR #2036, opened 2024-03-28
[pydanticgen] Embed extra metadata in modules, classes, and fields

sneakers-the-rat:
Fix: https://github.com/linkml/linkml/issues/2005

Related to: 
- https://github.com/linkml/linkml/issues/1830
- https://github.com/orgs/linkml/discussions/1820
- https://github.com/linkml/linkml-runtime/pull/305#issuecomment-2022063311

I feel like this comes up a lot. with the new template system it was pretty easy to implement.

this PR adds all metadata that isn't explicitly excluded - either from being already represented by the template model, or by being present in their `meta_exclude` classvars - to a `linkml_meta` attribute in modules, classes, and fields.

opening this as a draft because i figure there is plenty of disagreement to be had about where to put them, what should be excluded, etc. but this general framework works.

Currently using `json_schema_extra` in linkml 2 to store it, because the `metadata` field isn't really for this, but we can also talk about where that should go - bonus of that is we get all the extra metadata in pydantic's generated json schema for free :)

but anyway here's an overview using `personinfo.yaml` as a sample:

Adds a `LinkMLMeta` class that is basically a subclass of dict:

```python

class LinkMLMeta(RootModel):
    root: Dict[str, Any] = {}

        return self.root[key]

        self.root[key] = value

```

then schema metadata looks like this:

```python

linkml_meta = LinkMLMeta(
    {
        "default_curi_maps": ["semweb_context"],
        "default_prefix": "personinfo",
        "default_range": "string",
        "description": "Information about people, based on "
        "[schema.org](http://schema.org)",
        "emit_prefixes": ["rdf", "rdfs", "xsd", "skos"],
        "id": "https://w3id.org/linkml/examples/personinfo",
        "license": "https://creativecommons.org/publicdomain/zero/1.0/",
        "name": "personinfo",
        "prefixes": {
            "CODE": {
                "prefix_prefix": "CODE",
                "prefix_reference": "http://example.org/code/",
            },
           "...": "..."
            },
        },
        "source_file": "examples/PersonSchema/personinfo.yaml",
        "subsets": {
            "basic_subset": {
                "description": "A subset of the schema that "
                "handles basic information",
                "from_schema": "https://w3id.org/linkml/examples/personinfo",
                "name": "basic_subset",
            }
        },
    }
)

```

class metadata is like this:

```python
class Person(HasAliases, NamedThing):
    """
    A person (alive, dead, undead, or fictional).
    """

    linkml_meta: ClassVar[LinkMLMeta] = LinkMLMeta(
        {
            "class_uri": "schema:Person",
            "from_schema": "https://w3id.org/linkml/examples/personinfo",
            "in_subset": ["basic_subset"],
            "mixins": ["HasAliases"],
            "slot_usage": {
                "age_in_years": {"name": "age_in_years", "recommended": True},
                "primary_email": {"name": "primary_email", "pattern": "^\\S+@[\\S+\\.]+\\S+"},
            },
        }
    )
```

attribute metadata is like this:

```python
started_at_time: Optional[date] = Field(
    None,
    json_schema_extra={
        "linkml_meta": {
            "alias": "started_at_time",
            "domain_of": ["Event", "Relationship"],
            "slot_uri": "prov:startedAtTime",
        }
    },
)
```

Access to metadata is simple and uniform, even if the attribute version is a little verbose

```python
# schema
module.linkml_meta
# class
Person.linkml_meta
# attribute
Person.model_fields['age_in_years'].json_schema_extra['linkml_meta']
```

we can do more sophisticated transformations of the embedded values like casting prefixes to a specific prefix class, filtering `"alias"` if it is identical to the name of the attribute, etc. too. That'll be easier once PRs like https://github.com/linkml/linkml/pull/2019 get merged and the rendering logic for each type of object is more separated - i'd prefer to wait on that so i can clean this up, i don't really like adding to the junk heap i made at the bottom of `serialize()` lol, but figured i was worth getting a draft on the books so we have something to point to when this comes up, as it often does.

no tests yet, wanted to wait for feedback before trying to finish it

### DIFF
--- a/linkml/generators/pydanticgen/black.py
+++ b/linkml/generators/pydanticgen/black.py
@@ -18,6 +18,7 @@ except ImportError:
 def _default_mode() -> "Mode":
     return Mode(
         target_versions={TargetVersion.PY311},
+        line_length=120
     )
 
 
--- a/linkml/generators/pydanticgen/black.py
+++ b/linkml/generators/pydanticgen/black.py
@@ -16,10 +16,7 @@ except ImportError:
 
 
 def _default_mode() -> "Mode":
-    return Mode(
-        target_versions={TargetVersion.PY311},
-        line_length=120
-    )
+    return Mode(target_versions={TargetVersion.PY311}, line_length=120)
 
 
 def format_black(code: str, mode: Optional["Mode"] = None) -> str:
--- a/linkml/generators/pydanticgen/includes.py
+++ b/linkml/generators/pydanticgen/includes.py
@@ -1,0 +1,46 @@
+"""
+Classes to inject in generated pydantic models
+"""
+from typing import Dict, Any
+from pydantic import BaseModel
+from pydantic.version import VERSION
+PYDANTIC_VERSION = int(VERSION[0])
+
+
+LinkMLMeta_v1 = """
+class LinkMLMeta(BaseModel):
+    __root__: Dict[str, Any] = {}
+    
+    def __getattr__(self, key:str):
+        return getattr(self.__root__, key)
+        
+    def __getitem__(self, key:str):
+        return self.__root__[key]
+    
+    def __setitem__(self, key:str, value):
+        self.__root__[key] = value
+   
+    class Config:
+        allow_mutation = False
+"""
+
+LinkMLMeta_v2 = """
+class LinkMLMeta(RootModel):
+    root: Dict[str, Any] = {}
+    model_config = ConfigDict(frozen=True)
+    
+    def __getattr__(self, key:str):
+        return getattr(self.root, key)
+        
+    def __getitem__(self, key:str):
+        return self.root[key]
+    
+    def __setitem__(self, key:str, value):
+        self.root[key] = value
+    
+"""
+
+# if PYDANTIC_VERSION >= 2:
+#     LinkMLMeta = eval(LinkMLMeta_v2)
+# else:
+#     LinkMLMeta = eval(LinkMLMeta_v1)
--- a/linkml/generators/pydanticgen/includes.py
+++ b/linkml/generators/pydanticgen/includes.py
@@ -1,9 +1,9 @@
 """
 Classes to inject in generated pydantic models
 """
-from typing import Dict, Any
-from pydantic import BaseModel
+
 from pydantic.version import VERSION
+
 PYDANTIC_VERSION = int(VERSION[0])
 
 
--- a/linkml/generators/pydanticgen/pydanticgen.py
+++ b/linkml/generators/pydanticgen/pydanticgen.py
@@ -3,7 +3,7 @@ import logging
 import os
 import textwrap
 from collections import defaultdict
-from copy import deepcopy
+from copy import copy, deepcopy
 from dataclasses import dataclass, field
 from pathlib import Path
 from types import ModuleType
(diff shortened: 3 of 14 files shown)


sneakers-the-rat:
alright, this is ready to check out -  again i have held off adding tests for this until i get a better idea if this is the kind of thing we want, but if i get a nod i'll go ahead and add them

cmungall:
How about adding some command line options to control

- whether metadata is included (off by default, to avoid surprises?)
- the extent of the metadata
- whether the base class is inlined in the module (default) or using a runtime import (future)

sneakers-the-rat:
> whether metadata is included (off by default, to avoid surprises?)
> the extent of the metadata

this is already in there, but not as a cli option. will add!

> whether the base class is inlined in the module (default) or using a runtime import (future)

want this in this PR or in one after we make the runtime import? :)

cmungall:
let's save the inlined vs runtime as a separate PR. Incremental is good! (as is preserving default behavior)

sneakers-the-rat:
After this PR, all metadata will be (optionally) included.

Until then, all fields that are in the template models have some representation in the generated pydantic models https://linkml.io/linkml/generators/pydantic.html#templates

markdoerr:
Thanks, @sneakers-the-rat, 
for the fast reaction :)
If I understand you correctly, all of these metadata will be supported after the PR : [CommonMetadata](https://linkml.io/linkml-model/latest/docs/CommonMetadata/).
How much effort would it be, just to mention this liink/information in the documentation of the pydanticgen documentation / README ?
(I am a big fan of explicit information ;).

Looking forward to the merge - do you have a rough time estimate, when it is scheduled ?

One last question: will URIs (like slot_uri) also be transferred to the pydandic output ?

sneakers-the-rat:
> all of these metadata will be supported after the PR

Yes, all metadata. It will be configurable depending on if you want literally every field in the source schema, only those fields not represented by the template models, or no metadata.

See an example in the first post in this issue.

> How much effort would it be, just to mention this liink/information in the documentation of the pydanticgen documentation / README ?

Behavior of the template classes is already documented. This PR will also be documented after we reach a final form for it.

> do you have a rough time estimate, when it is scheduled ?

Im AFK until next week. Sometime after that. Still need to write tests and docs and decide implementation details

> will URIs (like slot_uri) also be transferred to the pydandic output ?

All metadata

markdoerr:
Hi @sneakers-the-rat , 
I tested your extension by a simple example and it does, what I need :)
Hope that it will find it's way soon into the next release. Thanks :+1:

djarecka:
@sneakers-the-rat - thanks a lot for working on this! I was trying to test it, but I'm not completely sure how I can use `template` to ask for specific fields to be included, e.g. `aliases`

sneakers-the-rat:
Aha, what I think we'll do is expose that as a param instead of having to fiddle with the template classes, since that's likely to come up a lot. 

do you mean to override the default `meta_exclude`, or do you mean you want metadata inclusion to be "opt-in" and include *only* those fields you explicitly specify?

cmungall (approved):

sneakers-the-rat:
this was not quite done but i can follow on in another PR.

- Needs to be reconciled with https://github.com/linkml/linkml/pull/2019 - want to avoid adding to the pile at the end of the `render` method, and the split up class and slot generation methods will give us better control over the metadata embedding. 
- needs tests 
- needs CLI options
- needs docs

Issue #2037, opened 2024-03-28
Append "Shape" suffix to SHACL shape in shaclgen.py


robertschubert:
**Is your feature request related to a problem? Please describe.**
We use linkml to genereate OWL classes as well as SHACL shapes. Those are then written to a graph db.
And this is where our problem arises:
There is a name clash between the OWL class and the SHACL shape since they have the same name (according to the name specified in the linkml yaml file).

**Describe the solution you'd like**
My suggestion would be to append the keyword "Shape" to a SHACL shape. 
Concrete example:
```
...
classes:
  MyClass:
    title: 'title of my class'
...

```

Currently following SHACL shape is generated:
```
prefix:MyClass a sh:NodeShape ;
    sh:closed false ;
...
    sh:targetClass prefix:MyClass .
```

Suggestion for the future:
Append "Shape" to MyClass and point to targetClass MyClass (without "Shape"):
```
prefix:MyClassShape a sh:NodeShape ;
    sh:closed false ;
...
    sh:targetClass prefix:MyClass .
```

**How important is this feature?** Select from the options below:
• Medium - can do work without it; but it's important (e.g. to save time or for convenience)
--> we are able to patch the shaclgen.py with the proposed code in the PR.

**When will use cases depending on this become relevant?** Select from the options below:
• Short-term - 2-4 weeks

**Additional context**
I have already implemented a solution and will open a pull request for this feature.

cmungall:
Thanks! Having the option to append `Shape` will better adhere to shacl conventions.

Is there a good quick guide that explains the rationale for having separate URIs, and what kinds of problems people might expect in punning a shape and a class? IMO the way most people use semweb vocabs like schema.org, prov, etc "classes" are more like OO-classes/frames and less like OWL classes....

robertschubert:
> Thanks! Having the option to append `Shape` will better adhere to shacl conventions.
> 
> Is there a good quick guide that explains the rationale for having separate URIs, and what kinds of problems people might expect in punning a shape and a class? IMO the way most people use semweb vocabs like schema.org, prov, etc "classes" are more like OO-classes/frames and less like OWL classes....

No problem :) I agree that I mostly see a "Shape" suffix for SHACL shapes so this feature will really be useful.
I have not a good reference to a quick guide. In our case we use an owl class for the ontology definition and link the SHACL shape via targetClass to that owl class. The SHACL is then of cource used for schema validation.

robertschubert:
[...]

PR #2038, opened 2024-03-28
Issue 2037: extend shaclgen.py to add the suffix "Shape" to the generated SHACL shape

robertschubert:
This PR implements the feature requested in https://github.com/linkml/linkml/issues/2037.


cmungall:
Can you make it such that appending `Shape` or any other suffix is opt-in and the default behavior remains the same?

robertschubert:
> Can you make it such that appending `Shape` or any other suffix is opt-in and the default behavior remains the same?

I will try to find the mechanism to pass CLI parameters to the generator.

robertschubert:
Hi @cmungall 
thanks for your valuable feedback. I changed the implementation as you requested. The suffix is now a cli parameter and defaults to "".

But I am struggling with the checks:

- The "build and test documentation" states: "/home/runner/work/linkml/linkml/docs/generators/shacl.rst:18: CRITICAL: Unexpected section title or transition." 
  - There was no change in shacl.rst. The part where my changes come into the game is starting from the section "Command line" starting at line 87. You have a  hint?

- The "build and test linkml / test (ubungu 3.0 and 3.10) seem to have an infrastructural problem. 
  - They fail with "ModuleNotFoundError: No module named 'pyshacl'"
  - Other recent PRs seem to have the same problem with those three checks.
  - Do we need another issue to create?

sneakers-the-rat:
super weird. checking on this because i was the one who proposed making `pyshacl` an `extra` dependency here: https://github.com/linkml/linkml/pull/1992

so the cached environment for 3.9 is from here: https://github.com/linkml/linkml/actions/runs/8445452327/job/23132690282
key: `venv-3.9-Linux-2130111127bb0cfd765b175f385377faa156d7790932052a2ac8d8368c099e5b`

pyshacl is installed: https://github.com/linkml/linkml/actions/runs/8445452327/job/23132690282#step:6:169

it does not seem to be removed in the subsequent call to install pydantic.

these runs restore the same cache key: https://github.com/linkml/linkml/actions/runs/8473903961/job/23219220620?pr=2038#step:5:17

but it seems to say "the venv is broken" - https://github.com/linkml/linkml/actions/runs/8473903961/job/23219220620?pr=2038#step:7:9

no idea what that means.

pretty sure the fix is to do `poetry run pip install 'pydantic<2'` (not sure if you need `--force` rather than `poetry add` because that attempts to re-solve the environment rather than modifying a single package, which is what we want.

linkml/generators/shaclgen.py
```diff
@@ -55,13 +56,15 @@ def as_graph(self) -> Graph:
         for pfx in self.schema.prefixes.values():
             g.bind(str(pfx.prefix_prefix), pfx.prefix_reference)
 
+        print(f"# class suffix: {self.suffix}")
```
sneakers-the-rat, line 59: my guess on the docs failing would be this print statement being left in messing up the detection of headers and transitions when sphinx goes to do autoclass at the bottom. that is a pretty cryptic error, haven't found a better way to debug it.

sneakers-the-rat:
invalidated the caches and it looks like tests are passing now. ya we probs want to change that pydantic install command (or move ahead with deprecating pydantic 1!!!)

linkml/generators/shaclgen.py
robertschubert: Removed that print but it is still failing.
My next try would be moving @click.command above @click.option. I have read that command should encapsulate the option. Let's see.

robertschubert: Still failing. But now it fails in line 20 (it was 18 before):
/home/runner/work/linkml/linkml/docs/generators/shacl.rst:20: CRITICAL: Unexpected section title or transition.
Need to think about next steps.

robertschubert:
@sneakers-the-rat: Thanks for making the failing checks running! :)

cmungall:
The sphinx error messages are a bit opaque but I think what is happening is
that terminusdb was recently moved from a dev requirement to an extra, it
needs to be added to the list of extras to install for doc building.

Sorry you were caught in all this Robert, your PR has nothing to do with it!

Will try and get to this later today

robertschubert:
I'm in easter holiday now for a week but I will of course have a look at the progress here. Could be that I am not able to respond early ;) have a good time!

sneakers-the-rat:
figured it out - it was from having the empty string as the default and having "show default" set to `True`

can be resolved by removing `show_default` in the `suffix` click option or by making the default `None`

imo optional params should default to `None` so it's possible to tell when they aren't passed vs. when they should be filled with a default. so that would be something like 

```python
@click.option(
    "-s",
    "--suffix",
    help="Use --suffix to append given string to SHACL class name (e. g. --suffix Shape: Person becomes PersonShape).",
)
``` 

and then 

```python
if self.suffix is not None:
    class_url += suffix
```

but that's just a style thing i don't use shaclgen so up to you!!!!

robertschubert:
@sneakers-the-rat Thanks for the hint. Changed it and that step is fine now :)
After all checks are completed, I will set this PR to "ready".

robertschubert:
ready for final review

cmungall (approved):

paul121:
Thank you for adding this!

Issue #1955, opened 2024-03-05
Use Pydantic strict mode/strict types (at least for arrays)


sneakers-the-rat:
**Describe the bug**
In testing arrays for https://github.com/linkml/linkml/pull/1887 , I test an array declared as an `int` type using a pydantic `conlist` as the innermost type, and give a test case with a `float` array, and the test passes. 

By default, pydantic 1 will coerce:
https://docs.pydantic.dev/1.10/usage/validation_decorator/#coercion-and-strictness
and 

and requires using strict types to fail validation: https://docs.pydantic.dev/1.10/usage/types/#strict-types

and pydantic 2 will coerce, but has a strict mode at the model level:
https://docs.pydantic.dev/2.6/concepts/strict_mode/
which additionally differentiates between parsing from JSON vs. parsing from python when in strict mode, which is what we want

imo models should be as close to as strict as the linkml schema as they can be (without going over), and in this case we are less strict than the model (with the ability to be exactly as strict).

Propose: use strict types in pydanticgen 1, and strict mode in pydanticgen 2. If this is too restrictive, propose at least doing so for arrays, where dtype is much more impactful and coercion shouldn't happen (except from json, as described).

Adding this to the pile of "reasons why we should deprecate pydantic 1 to cut the complexity of the generator in half" re: https://github.com/linkml/linkml/issues/1925 . let's set that deprecation timeline!

cmungall:
Are there any performance implications for strict mode? Might some users
prefer to decouple some kinds of validation from object instantiation?

sneakers-the-rat:
Re: perf, I doubt it, I would think that coercion would be more expensive bc the strict types fail faster, but I can profile.

Re: validation vs. Instantiation, one can always do [model_construct](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_construct), or did you mean finer grain control than that?

cmungall:
I think this is sufficient, thx!

sneakers-the-rat:
For the sake of recordkeeping - here I added the `strict` mode to the inner types of the `AnyShapeArray` for pydantic 2 - https://github.com/linkml/linkml/pull/1887/commits/0d6926a0324e9aedba19aa47bcbcf4b2c42221df

core schemas are cool and transparent like this:
https://github.com/linkml/linkml/blob/0d6926a0324e9aedba19aa47bcbcf4b2c42221df/linkml/generators/pydanticgen/array.py#L50-L52

[...]

PR #2041, opened 2024-03-29
Fix testing infrasucture for Pydantic 2 default

### LINKED ISSUE #2040: Update `tox.ini` and github CI to make Pydantic 1 the special test scenario
Previously we had Pydantic 1.x.x specified in our `poetry.lock` file so that it was the default version for development and testing environments. We had a tox testenv for explicitly testing with Pydantic v2. The main GitHub testing workflow included `pydantic-version` in it's testing matrix, but excluded Pydantic 2 most of the time.

With https://github.com/linkml/linkml/pull/2002 and https://github.com/linkml/linkml/pull/2023, we've made the cut over the have Pydantic 2.x.x in our `poetry.lock` file. But we still need to:

* flip the tox testenv so that you can test with Pydantic v1 if needed
* exclude Pydantic v1 for most matrix combinations in the main GitHub test workflow.

pkalita-lbl:
Fixes #2040 

### Summary

* With Pydandic 2.x.x specified in our `poetry.lock` file, running tests directly via `pytest` or the `tox` `py3x` testenvs will use Pydantic v2. These changes replace the existing "special case" `pydantic2` testenv with a `pydantic1` testenv. The new testenv operates the same as the old one, except swapping Pydantic v1 for v2.
* Previously the `main.yaml` testing workflow had a `pydantic-version` variable in the configuration matrix. With the addition of Python 3.11 and 3.12 to the testing matrix (https://github.com/linkml/linkml/pull/1704), the matrix permutations were starting to get a little unruly. Here I'm splitting off Pydantic 1 testing into a separate workflow (`.github/workflows/test-pydantic1.yaml`). I then updated the `exclude` combinations in `main.yaml` so that what we end up running is:
  * All supported Python version with Pydantic 2 on Ubuntu (`main.yaml`)
  * The oldest and newest supported Python versions with Pydantic 2 on Windows (`main.yaml`)
  * The latest supported Python version with Pydantic 1 on Ubuntu (`test-pydantic1.yaml`)
* The updated testing workflows caught a few issues we had missed with our previous configuration matrix. @sneakers-the-rat fixed a few issues in the Pydantic generator that crop up when using older Python versions. I updated a few compliance tests to accurately reflect the actual behavior of Pydantic on older Python versions.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -20,16 +20,14 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest]
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
-        pydantic-version: ["1", "2"]
         exclude:
+          # Test on Windows with only the oldest and newest Python versions
           - os: windows-latest
-            python-version: "3.8"
+            python-version: "3.9"
           - os: windows-latest
-            pydantic-version: "2"
-          - python-version: "3.8"
-            pydantic-version: "2"
-          - python-version: "3.9"
-            pydantic-version: "2"
+            python-version: "3.10"
+          - os: windows-latest
+            python-version: "3.11"
 
     # See https://github.com/snok/install-poetry#running-on-windows
     defaults:
@@ -75,13 +73,7 @@ jobs:
       #----------------------------------------------
       - name: Install dependencies
         if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
-        run: poetry install --no-interaction --no-root -E tests
-
-      #----------------------------------------------
-      # install pydantic
-      #----------------------------------------------
-      - name: Install Pydantic
-        run: poetry add pydantic@^${{ matrix.pydantic-version }}
+        run: poetry install --no-interaction --no-root --all-extras
 
       #----------------------------------------------
       #         setup and install graphviz
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -45,6 +45,7 @@ jobs:
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v3
+        id: setup-python
         with:
           python-version: ${{ matrix.python-version }}
 
@@ -66,7 +67,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: .venv
-          key: venv-${{ matrix.python-version }}-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}
+          key: venv-${{ steps.setup-python.outputs.python-version }}-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}
 
       #----------------------------------------------
       # install dependencies if cache does not exist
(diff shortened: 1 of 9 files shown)


pkalita-lbl:
Notes to self: seems like we've never actually tested the output of the Pydantic generator, targeting v2, on Python 3.8 or 3.9. The 3.8 tests are failing with a lot of
```
ImportError: cannot import name 'Annotated' from 'typing'
```
Makes sense, because that was introduced in Python 3.9 (https://docs.python.org/3/library/typing.html#typing.Annotated)

On 3.9 the errors are mostly
```
AttributeError: '_SpecialForm' object has no attribute '__name__'
```
They also seem to be related to array test cases. Needs further investigation.

sneakers-the-rat:
oop how did i miss that. assigning myself bc these are my fault probs

sneakers-the-rat:
the remaining test failures look like they're from pydantic coercion, where a model like

```python
class C(BaseModel):
    s1: Optional[bool] = None
```

incorrectly validates with

```python
inst = C(s1=1.1)
```

as

```python
C(s1=True)
```

which is a pydantic coercion thing. idk what would be the desired behavior there - we could enable strict mode ( https://github.com/linkml/linkml/issues/1955 ) or we could mark the expected behavior as `COERCES`. i'll defer to y'all on what the fix should be. weird that it's happening only in older python versions though, don't know why that would be.

pkalita-lbl:
Much appreciated @sneakers-the-rat! I will take care of the last few failures.

pkalita-lbl:
I think our best move here is to just mark those narrows cases in the compliance tests as having coerce as the expected behavior. I didn't find any issues in the Pydantic repo that directly address what we're seeing on Python < 3.10 even though it directly contradicts what they say in their docs. 

The error is coming from Pydantic's Rust core. Because of that I'm not _totally_ surprised that the issue seems to only effect certain Python versions on Ubuntu and not Windows.

.github/workflows/main.yaml
```diff
@@ -20,16 +20,14 @@ jobs:
...
+            python-version: "3.9"
           - os: windows-latest
-            pydantic-version: "2"
-          - python-version: "3.8"
-            pydantic-version: "2"
-          - python-version: "3.9"
-            pydantic-version: "2"
+            python-version: "3.10"
+          - os: windows-latest
+            python-version: "3.11"
```
sierra-moxon, line 30: not 3.12 right?  (pydantic1 seems to be setting up a 3.12 environment)

docs/contributing/contributing.md
```diff
@@ -63,8 +63,8 @@ issues are with reference to the old [biolinkml](https://github.com/biolink/biol
 
 Use marks to run or exclude groups of tests:
 
-- `network` - (currently unused in CI) - tests marked as requiring network access/making network requests in order to succeed. 
-- `slow` - tests that are necessary to test technical correctness but are sufficiently long that it's worth excluding them during routine development/when running tests locally. 
+- `network` - (currently unused in CI) - tests marked as requiring network access/making network requests in order to succeed.
+- `slow` - tests that are necessary to test technical correctness but are sufficiently long that it's worth excluding them during routine development/when running tests locally.
```
sierra-moxon, line 67: thank you for this doc.

sierra-moxon (approved):

.github/workflows/main.yaml
sierra-moxon: I see, I missed the excludes part.

sneakers-the-rat:
Go team!!!

PR #2045, opened 2024-04-02
[perf] Finish putting file I/O behind `--with-output` flag in compliance tests (and other free perf)

sneakers-the-rat:
hey it's me again, profiling and fixing perf while i wait for my PR tests to finish running lol.

Previously in https://github.com/linkml/linkml/pull/1941 we had found out that making I/O optional behind a `--with-output` flag brought down testing time by 25%, and now i'm here to finish the job!

Profiling, we see that when running the whole `test_compliance` directory that `safe_dump` takes up 243s of 527 cumulative testing time.

Screenshot 2024-04-01 at 5 35 56 PM

There are a number of places in the compliance testing helpers that read and write schemas and output, but each of them also allows the in-memory schema or generated content to be used instead. The only exception to being moved behind `--with-output` are the import tests, which do require that the schema files be written to disk. Otherwise I tried to preserve existing behavior as much as possible.

While I was doing that I also noticed a few other low-hanging fruit perf optimizations:
- Make pydanticgen template environment a class attribute rather than a method since it actually ends up taking awhile to load the env every time - it's the same every time so don't keep loading it!
- Make the shacl validation plugin only load each graph for each schema once by caching it in a private dict by the hash of the stringified schema.
- in `Generator._initialize_using_schemaloader` Rather than roundtripping all the way to yaml and back when initializing from schemaloader, just dump to a dict, which seems to do the same thing according to the tests. 
- Make the pydantic and json schema validation context a single fixture function with a module-level scope also so we don't need to load the same schema anew for every file :)
- Use the CSafeDumper when possible.
- Remove one no-op dump and load ( i assume the ability to dump and load python generated models should be tested elsewhere )

Altogether for a run of the whole compliance test directory with cprofile enables, this makes for a perf boost:
- Current time: 527s
- This PR: 157s
- Change: -370s (70%)

### DIFF
--- a/linkml/generators/jsonldgen.py
+++ b/linkml/generators/jsonldgen.py
@@ -60,8 +60,8 @@ class JSONLDGenerator(Generator):
     """Path to a JSONLD context file"""
 
     def __post_init__(self) -> None:
-        super().__post_init__()
         self.original_schema = deepcopy(self.schema)
+        super().__post_init__()
 
     def _add_type(self, node: YAMLRoot) -> dict:
         if self.format == "jsonld":
--- a/linkml/generators/pydanticgen/pydanticgen.py
+++ b/linkml/generators/pydanticgen/pydanticgen.py
@@ -499,7 +499,7 @@ class PydanticGenerator(OOCodeGenerator):
         return None
 
     def _template_environment(self) -> Environment:
-        env = TemplateModel.environment()
+        env = TemplateModel.environment
         if self.template_dir is not None:
             loader = ChoiceLoader([FileSystemLoader(self.template_dir), env.loader])
             env.loader = loader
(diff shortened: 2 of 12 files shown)


sneakers-the-rat:
oof getting blindsided by test outputs still being used https://github.com/linkml/linkml/pull/1976

sneakers-the-rat:
Alright one change that might be substantive, but tests are passing so seems fine - 
the the `Generator` base class says that it was dumping and loading the incoming schema because there were needed modifications on load, but then the JSONLD generator was caching the copy *after* that loading was done while it seems to be counting on the values being those from *before* that modification happens. swapping the order of the deepcopy and the call to `__post_init__` fixed.

Similarly, `RDFGenerator` seems to have passed a copy of *itself* to the contextgenerator, rather than a copy of its schema. not sure how that works, but did something similar to what the context generator did and it fixed tests. not really sure if that's right, have a hard time following schemaloader stuff.

sneakers-the-rat:
whoop whoop going from 16m-27m min in `main` down to 13-17m in this PR. progress!

PR #2050, opened 2024-04-05
convert test_data to pytest

sneakers-the-rat:
Fix: https://github.com/linkml/linkml/issues/1717

additional notes:
- optional params should be `Optional`
- the `SQLStore` class has a lot of implicit logic and post-instantiation modification, i tried to do a little there but the ability to instantiate from both a module and a schema without using `@classmethod`s makes refactoring tricky.
- removed static models, generate them instead: the generators change, so the models they generate should too.
- make all outputs temporary files. there is a lot of I/O here that could be avoided by doing in-memory comparisons rather than roundtripping to and from yaml on disk, but i left that in place.
- since the pydantic models have `default_factory` for optional `List` args (for now), made the yaml comparison method allow for that.
- made model generation a module-level fixture so we wouldn't be compiling the same modules a bunch of times, but that still happens within the test functions without model compilation being split out into separate unit tests so idk!


pkalita-lbl (approved):
Looks great! Thanks for knocking another directory out.

> there is a lot of I/O here that could be avoided by doing in-memory comparisons rather than roundtripping to and from yaml on disk, but i left that in place.

Very true, but I'm happy to have this PR be focused on straightforward conversion! 

> made model generation a module-level fixture so we wouldn't be compiling the same modules a bunch of times, but that still happens within the test functions without model compilation being split out into separate unit tests so idk!

I like the pattern in general.

PR #2053, opened 2024-04-05
[pydanticgen] implement `string_serialization`

sneakers-the-rat:
depends on:
- [ ] merging https://github.com/linkml/linkml/pull/2036
  
  this PR builds on #2036, see the [diff between branches](https://github.com/sneakers-the-rat/linkml/compare/pydanticgen-metadata..pydanticgen-string-template) for what is unique to this PR

---

> *i have implemented*
> *the `string_serialization`*
> *that was in*
> *the metamodel*
> 
> *and which*
> *you were probably*
> *hoping*
> *i wouldnt*
> 
> *Forgive me*
> *the metaprogramming potential*
> *so fucked up*
> *and so good*

couldn't sleep, saw this earlier while doing https://github.com/linkml/linkml/pull/2052 and had a whacky idea

so like

```yaml
classes:
  MyClass:
    attributes:
      attr:
    annotations:
      meta:
        value: my special metadata annotation
    aliases:
    - alias1
    - alias2
    string_serialization: |
      ---
      annotations: {annotations[meta][value]}
      aliases: {aliases}
      attr: {attr}
``` 

becomes

```python
class MyModel(ConfiguredBaseModel):
    linkml_meta: ClassVar[LinkMLMeta] = LinkMLMeta({'aliases': ['alias1', 'alias2'],
         'annotations': {'meta': {'tag': 'meta',
                                  'value': 'a string value set as a linkml '
                                           'classdefinition metadata value'}},
         'from_schema': 'pydantic'})

    attr: Optional[str] = Field(None, json_schema_extra = { "linkml_meta": {'alias': 'attr', 'domain_of': ['StringSerialization']} })
    
    def __str__(self) -> str:
        if self.linkml_meta is not None:
            kwargs = self.linkml_meta
        else:
            kwargs = {}

        kwargs.update(self.model_dump())
        return '''
---
annotations: {annotations[meta][value]}
aliases: {aliases}
attr: {attr}
'''.format(**kwargs)

```

which of course works like

```python
>>> instance = MyModel(attr=attr_str_value)
>>> print(str(instance))

---
annotations: a string value set as a linkml classdefinition metadata value
aliases: ['alias1', 'alias2']
attr: a string value set as a class attribute
```

so one could imagine the metaprogramming possibilities that could be had if one were to eval (after, perhaps, safely escaping) a string that looked like another model or format and so on. 

also this is just a demonstration of 𝓉𝒽𝑒 𝒻𝓊𝓃𝓏𝒾𝑒𝓈 𝓉𝑜 𝒷𝑒 𝒽𝒶𝒹 with the pydanticgen templating system :)

@cmungall i also added a little sketch of a pytest-fixture based complement to the compliance framework - the teeniest, tiniest hookable version. obvs it doesn't implement all the semantics of the existing compliance framework, but i'm thinking about how it would be nice to be able to surface some of what happens in the `helper` module, and allow for both 'proper' compliance tests alongside the ability to allow tests that might not fit well there to declare that a certain test implements a part of the metamodel. just a quick idea :)

### DIFF
--- a/linkml/generators/pydanticgen/black.py
+++ b/linkml/generators/pydanticgen/black.py
@@ -18,6 +18,7 @@ except ImportError:
 def _default_mode() -> "Mode":
     return Mode(
         target_versions={TargetVersion.PY311},
+        line_length=120
     )
 
 
--- a/linkml/generators/pydanticgen/black.py
+++ b/linkml/generators/pydanticgen/black.py
@@ -16,10 +16,7 @@ except ImportError:
 
 
 def _default_mode() -> "Mode":
-    return Mode(
-        target_versions={TargetVersion.PY311},
-        line_length=120
-    )
+    return Mode(target_versions={TargetVersion.PY311}, line_length=120)
 
 
 def format_black(code: str, mode: Optional["Mode"] = None) -> str:
--- a/linkml/generators/pydanticgen/includes.py
+++ b/linkml/generators/pydanticgen/includes.py
@@ -1,0 +1,46 @@
+"""
+Classes to inject in generated pydantic models
+"""
+from typing import Dict, Any
+from pydantic import BaseModel
+from pydantic.version import VERSION
+PYDANTIC_VERSION = int(VERSION[0])
+
+
+LinkMLMeta_v1 = """
+class LinkMLMeta(BaseModel):
+    __root__: Dict[str, Any] = {}
+    
+    def __getattr__(self, key:str):
+        return getattr(self.__root__, key)
+        
+    def __getitem__(self, key:str):
+        return self.__root__[key]
+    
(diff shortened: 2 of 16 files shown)


sneakers-the-rat:
That *should* render recursively, so since the fields of a linkml pydantic model are occasionally also linkml models that can also have their own string serializations, we have a composable templating system. Since `range` has `any_of`,  and `rules` and other constructs allow for logic to be added to ranges and slot usage (TBD in pydanticgen), you have a programmable, composable templating system.

We will have a Turing complete self describing code generator soon at this rate lmao

sierra-moxon:
closing for now, branch remains in fork I believe, also captured details in an issue for someone to pick up and continue with here: https://github.com/linkml/linkml/issues/2701

PR #2055, opened 2024-04-05
Run code quality checks only once in testing workflow

### LINKED ISSUE #2054: Do not run code quality checks on every testing matrix permutation
Currently in our `main.yaml` testing workflow, we have one job (`test`) which has a matrix of Python versions and OSes. It performs code quality checks and runs tests. This is a bit of overkill because the results of the code quality checks shouldn't changed based on Python version or OS. We should split the code quality checks off into a separate job in the same workflow that only runs once.

pkalita-lbl:
Fixes #2054 

### Summary

* Moved code quality checks to a new job in `main.yaml`. This job only runs once and if it successfully completes the `test` job (with the same permutations as before) will run. 
* Added a call to `poetry check` to the code quality job. This will fail if our `pyproject.toml` file has formatting issues or is out of sync with `poetry.lock` (https://python-poetry.org/docs/cli/#check). This was inspired by the recent attempts to get automatic, periodic updates to the lock file working.
* Updated Poetry installation and virtual environment caching to be inline with the current best practices suggested by the `setup-python` action https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages

### Comments

* With the way the workflow is currently configured, the tests will not run if the code quality checks fail. This is consistent with the current behavior (except that the quality checks only happen once). But by having them as separate jobs, we now have control over that. If we removed the `needs` block in the `test` job, the code quality checks and tests could run in parallel. Something to consider.
* I really like that the `setup-python` action now has built-in support for Poetry virtual environment caching. It's nice to let someone else get that right for us! If it seems to work well here we should roll that out to workflows in our other repos.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -14,7 +14,27 @@ on:
   workflow_dispatch:
 
 jobs:
+  quality-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install poetry
+        run: pipx install poetry
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - name: Check pyproject.toml and poetry.lock
+        run: poetry check
+      - name: Install tox
+        run: python -m pip install tox
+      - name: Run codespell
+        run: tox -e codespell
+      - name: Run code quality checks
+        run: tox -e lint
+
   test:
+    needs:
+      - quality-checks
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


sierra-moxon (approved):

sierra-moxon:
> ### Summary
> * Moved code quality checks to a new job in `main.yaml`. This job only runs once and if it successfully completes the `test` job (with the same permutations as before) will run.
...
> * With the way the workflow is currently configured, the tests will not run if the code quality checks fail. This is consistent with the current behavior (except that the quality checks only happen once). But by having them as separate jobs, we now have control over that. If we removed the `needs` block in the `test` job, the code quality checks and tests could run in parallel. Something to consider.

and we want two `jobs` stanzas in one action file vs. two action files each with their own single job, to take advantage of the chained execution here?  (and/or share setup)

pkalita-lbl:
Right, having them as two jobs in one workflow allows you to run one conditionally based on the result of the other. If we didn't want/need that then there wouldn't be much of a difference between two jobs in one workflow and two workflows.

PR #2060, opened 2024-04-09
plantuml: hide circle and empty members, remove quotes

VladimirAlexiev:


VladimirAlexiev:
@Silvanoc

cmungall:
Thanks! The corresponding test should be fixed too - let us know if you want one of us to take over from here or if you're able to make the additional change yourself!

sujaypatil96:
Hi @VladimirAlexiev it looks like certain Actions are not going through because you haven't run the ruff linter on your code.

Could you run the below command using [ruff](https://github.com/astral-sh/ruff):

```bash
ruff check linkml/generators/plantumlgen.py
```

linkml/generators/plantumlgen.py
```diff
@@ -129,13 +133,11 @@ def add_class(self, cn: ClassDefinitionName) -> str:
                 if True or cn in slot.domain_of:
                     mod = self.prop_modifier(cls, slot)
                     slot_defs.append(
-                        '    {field} "'
```
Silvanoc, line 132: I tend to add quotation marks to generated names for robustness: it doesn't harm having them and it won't break if any unexpected whitespaces appear in the names.

What does it harm to have them? If it doesn't harm, why removing them?

I suppose that we can rely on `underscore()` to ensure that no whitespaces appear in the names. So removing them shouldn't suppose any risk.

Silvanoc:
LGTM

I have only what is probably a nitpick: https://github.com/linkml/linkml/pull/2060#discussion_r1577309102

linkml/generators/plantumlgen.py
sujaypatil96: Good questions @Silvanoc! I'd like to know the answer to the question you asked as well -- what are the pros of removing the quotes?

cmungall (approved):

cmungall:
lgtm. We can circle back on escaping issues

PR #2061, opened 2024-04-09
Pass strings rather than printing to `stdout` in generators

sneakers-the-rat:
Part of: https://github.com/linkml/linkml/issues/923

I was sorta like 'this can wait' until i tried to debug something in rdfgen and couldn't use a debugger because stdout was being captured.

design is p self explanatory, only thing that might be weird is that the top-level `Generator` class does simple string concatenation rather than joining with newlines because we assume that the lower level generators will handle newlining (and don't want to prevent them from not doing it if they don't want to).

also one bugfix: 
- previously, headers for protogen, shexgen, and i think one more were just printed into thin air because stdout wasn't being captured during `__post_init__` - restored that

p much all the changes here are removing extraneous linebreaks that were there bc there wasn't a good way to exlude the newline created by printing `None`. we can restore that if we want a smaller PR idrc :) all the substantive changes are in separate commits to the snapshot regeneration for easier diffing.

anyway merry christmas, a gift of labor 2 u


linkml/utils/generator.py
```diff
@@ -267,81 +267,97 @@ def serialize(self, **kwargs) -> str:
...
+                    sub_out = self.visit_class_slot(cls, self.aliased_slot_name(slot), slot)
+                    if sub_out is not None:
+                        out += sub_out
+                sub_out = self.end_class(cls)
+                if sub_out is not None:
+                    out += sub_out
+        sub_out = self.end_schema(**kwargs)
+        if sub_out is not None:
+            out += sub_out
+        return out
```
sierra-moxon, line 327: 🎉 🎉 🎉

linkml/generators/markdowngen.py
```diff
@@ -85,248 +84,263 @@ def visit_schema(
...
+                if slot.in_subset and subset.name in slot.in_subset:
+                    items.append(self.bullet(self.slot_link(slot, use_desc=True), 0))
+            items.append(self.header(3, "Types"))
+            for type in sorted(self.schema.types.values(), key=lambda s: s.name):
+                if type.in_subset and subset.name in type.in_subset:
+                    items.append(self.bullet(self.type_link(type, use_desc=True), 0))
+            items.append(self.header(3, "Enums"))
+            for enum in sorted(self.schema.enums.values(), key=lambda s: s.name):
+                if enum.in_subset and subset.name in enum.in_subset:
+                    items.append(self.bullet(self.enum_link(enum, use_desc=True), 0))
```
sneakers-the-rat, line 343: oh one more bugfix - `type` -> `enum`

cmungall:
Happy to merge but just want to check one thing. That's a lot of modified snapshots. Looks to be all spurious, I don't think the newline makes much of a difference. But do we anticipate merging this causing a lot of conflicts with any open PRs?

sneakers-the-rat:
> Looks to be all spurious

Yes I should have said I did look through to see and I did not see any substantive changes, and I only kept this change because it seemed like "correct" behavior to not have the extra newlines even if markdown doesnt care. But since markdowngen is destined for deprecation we might also want to minimize change here, so up to yall.

> But do we anticipate merging this causing a lot of conflicts with any open PRs?

I think any open PRs should be able to merge code and regenerate through and conflict in the snapshots? Or at least that seems like the goal of the snapshotting system. The PR touches a bit of "backbone" code in Generator that doesnt look v high velocity, and it also touches a decent amount of the markdowngen which is similarly low velocity, but in those cases the changes are almost all peripheral/wrapping changes with obvious paths to merge.

Any we're worried about in particular?

PR #2062, opened 2024-04-10
Use app installation token in poetry update workflow

pkalita-lbl:
Following [these instructions](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens), I created a new private GitHub App in the `linkml` org. I installed it in this repo and added the app's private key as a secret to this repo. 

With these changes the app's private key and ID will be used to get a short-lived access token in the workflow. That token is then passed to the `peter-evans/create-pull-request` action. I believe the result of this will be that:

* The PR will be open by the LinkML PR Bot app (as opposed to a specific user or the default [github-actions](https://github.com/apps/github-actions) account)
* The tests will run automatically when the PR is opened

I also added a few core developers to the reviewers list on the PR so that we get notified.

### DIFF
--- a/.github/workflows/poetry-update.yaml
+++ b/.github/workflows/poetry-update.yaml
@@ -47,22 +47,23 @@ jobs:
           echo "$UPDATE_MSG"
           echo EOF
         } >> "$GITHUB_ENV"
-
+    - uses: actions/create-github-app-token@v1
+      id: app-token
+      with:
+        app-id: ${{ vars.PR_BOT_APP_ID }}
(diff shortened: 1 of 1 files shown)


sneakers-the-rat:
Wanna try triggering the poetry lock action from this branch and see what happens?

pkalita-lbl:
Yeah, good call. I ran it manually, and I think it did what I would have expected it to do https://github.com/linkml/linkml/pull/2063

sneakers-the-rat (approved):
Test run looks good, changes look good, thats a big 10-4 from me good buddy

PR #2079, opened 2024-04-24
Add generator flag to include an import outside the normal import mechanism

sierra-moxon:
In particular, this has use in supporting a separate YAML file to collect deprecated elements that we want to serialize in some generators but not others (e.g. we want to serialize deprecated elements in docgen when we're pointing schema URIs to a documentation site so that URIs are persistent even after a deprecation event.  But we don't want to serialize elements in python data classes or jsonschema, nor include the deprecated elements in the main imports of the schema so that they are no longer used by mistake, etc.)


pkalita-lbl (approved):
Overall looks good to me. Two non-critical thoughts:

1. Would a name like `--extra-include` or `--additional-include` be helpful to make it more explicit that you're appending to, not replacing, your schema's imports?
2. Would it make sense to add `multiple=True` to the `@click.option` definitions and then loop over the resulting array in `Generator.__post_init__`? That way you could do something like:
    ```
    gen-doc --include deprecated1.yaml --include deprecated2.yaml ...
    ```
    Or would we consider it better practice to have only one CLI include and have `deprecated1.yaml` include `deprecated2.yaml` in the schema itself?

cmungall (approved):

Issue #2088, opened 2024-05-01
Validation error message is too vague to fix


paulmillar:
**Describe the bug**

I ran the command:

    linkml-validate -s DataModels/LinkML/services.yaml Distro/services.json 

to validate a particular JSON file matches the provided description.  There were a few validation errors reported.  Here is an example:

```
[ERROR] [Distro/services.json/0] {'abbreviation': 'VizieR', 'name': 'VizieR astronomical catalogue service', 'resourceOrganisation': 'cds', 'webpage': 'https:/
/vizier.cds.unistra.fr/', 'description': 'The VizieR archive consists of scientific data from refereed astronomy journal publications or data from ground based
 observatories/experiments or space missions. In VizieR terminology, a VizieR dataset is called a “catalogue”. Each "catalogue” can contain: \n\n- Tabular data
: Predominantly related to astronomical objects. \n\n- “Associated data”: non-tabular data such as spectra, images, data cubes etc. associated with journal pub
lications or with an observation catalogue provided by an observatory or a project\n\nThe VizieR service offers the functionalities for interactive and program
matic queries, including access via International Virtual Observatory Alliance (IVOA) protocols.', 'tagline': 'The VizieR astronomical catalogue service VizieR
 provides a library of published astronomical catalogues *tables and associated data* with verified and enriched data, accessible via multiple interfaces.', 'l
ogo': 'https://cds.unistra.fr/img/communication/vizier_420x245.png', 'scientificDomains': [{'scientificDomain': 'scientific_domain-natural_sciences', 'scientif
icSubdomain': 'scientific_subdomain-natural_sciences-physical_sciences'}], 'categories': [{'category': 'category-sharing_and_discovery-data', 'subcategory': 's
ubcategory-sharing_and_discovery-data-scientific_research_data'}], 'targetUsers': ['target_user-researchers', 'target_user-research_organisations', 'target_use
r-research_communities'], 'accessTypes': ['access_type-remote'], 'accessModes': ['access_mode-free'], 'tags': ['astronomy, astrophysics'], 'geographicalAva
ilabilities': ['WW'], 'languageAvailabilities': ['en'], 'resourceGeographicLocations': ['FR', 'US', 'IN', 'ZA', 'JP', 'CN'], 'publicContacts': [{'email': 'cds-ques
[...]

PR #2089, opened 2024-05-01
validator: support including validator context when reporting errors

paulmillar:
Motivation:

When validating JSON, the `linkml-validate` command builds a JSON Schema against which the JSON input is checked.  If there are validation problems, the command prints a single line for each such problem.

In some cases, the failing node is a composition (`allOf`, `anyOf`, `oneOf`) and the resulting error message is currently rather vague.

Modification:

Update the `ValidationResult` class to include a List of strings.  These are the validation errors that led to the composition failing.

Note, these contextual failures could (themselves) have further contextual failures, forming a tree.  In the interests of keeping the patch small, this change records only the next level; however, a future patch may address this issue.

The error output is updated to support providing the validation error's context as lines starting `[CONTEXT]`.

The command-line arguments are updated to support the `-v`/`--verbose` flag.  Specifying this flag results in the validator printing any validation error context.

Note, the flag is included to support backwards compatibility, but it may be acceptable to print the validation error's context always.

Result:

`linkml-validate` now supports the `-v`/`--verbose` option.  Specifying this option will result in certain validation error reports containing more information.

Closes: #2088

### DIFF
--- a/linkml/validator/cli.py
+++ b/linkml/validator/cli.py
@@ -108,6 +108,9 @@ DEPRECATED = "[DEPRECATED: only used in legacy mode]"
     help=f"{DEPRECATED} When handling range constraints, include all descendants of the range "
     "class instead of just the range class",
 )
+@click.option(
+    "--verbose", "-v", is_flag=True, default=False, help="Provide more verbose reporting of validation errors."
+)
 @click.argument("data_sources", nargs=-1, type=click.Path(exists=True))
 @click.version_option(__version__, "-V", "--version")
 @click.pass_context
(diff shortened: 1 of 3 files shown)


cmungall:
This is great!

How do you feel about using a different option name? Elsewhere —verbose is
used exclusively to control logging level and used for debugging code not
data. Perhaps —context?

cmungall (approved):

paulmillar:
Hi @cmungall
Sorry for not responding faster to your comment.
Yes, I agree that using a name involving `context` would be better than verbose.
Thanks for updating the pull-request: the new option name looks good to me.
Cheers,
Paul.

PR #2095, opened 2024-05-03
Allow control over which OWL types are used for enums. Fixes #2090

### LINKED ISSUE #2090: Allow options of mapping permissible values to non classes in OWL
Currently gen-owl translates PVs to classes. This is very convenient because we can take advantage of constructs like like UnionOf to represent an enumeration. It is also the default case since often PVs are classes taken from ontologies

But this is not always the case.

- authors may want to follow the [Rector partition model](https://www.w3.org/2001/sw/BestPractices/OEP/SpecifiedValues-20050405/) e.g. for LOW MEDIUM HIGH
- the PVs may represent actual individuals. E.g an enum `HomePlanet` with PVs for `{Earth, Venus, ...}`
- the PVs may be other OWL constructs like ObjectProperty, Datatype, even Ontology
- the author may prefer to be non-committal

IF the user decides that ALL the members of a PV are individuals then we can take advantage of the OWL `AnyOf` construct

IF the PV is a mix of classes AND inviduals, it's impossible to have a single OWL definition of the enum in OWL-DL, unless we pun. But punning creates a lot of confusion.

Of course some users may not care about having an enum definition. It may be sufficient to simply use annotation properties to link an enum to PVs, as we do for `gen-rdf`

This PR is to gather feedback from the community on preferred mappings for PVs in OWL translations

### LINKED ISSUE #2091: Allow customization of mapping to property types when converting schemas to OWL
OWL has the following property types

- OP ObjectProperty; range is a node
- DP DatatypeProperty; range is a literal
- AP AnnotationProperty; range is either, and the property has no logical semantics

Currently two behaviors are supported

1. Slots are mapped to OPs or DPs depending on range
2. All slots are mapped to OPs, and there is a single builtin DP that maps "data individuals" to their literal representation

1 is the default as it is what people expect. However, people can get into trouble if they want to use a range of `Any`, or use unions with ranges of both literals and classes. There is literally no way to directly map the schema to OWL-DL. the current behavior is to illegally pun. Suggestions welcome!

2 is guaranteed to avoid punning but the resulting OWL schema is not isomorphic to direct RDF translations of instances, and is somewhat non-idiomatic (although OBI uses this pattern)

Currently there is no way to map to APs. This should be supported. This would be controlled by stating

```
Person:
  attributes:
    name:
      slot_uri: rdfs:label
      implements:
         - owl:AnnotationProperty
```

Note: my personal preference is to avoid APs. You lose a lot of validation benefits. It is more LinkML-like to consider LinkML individuals to be in the domain of discourse even if they are also classes. In fact I think APs are a hack that were only necessary in OWL1, and continue to be necessary in OWL2 due to the fact that punning looks odd and confusing in many tools. But others disagree so we should support AP mapping!

cmungall:
Allow mapping of properties to annotation properties in OWL
Fixes #2091


samsontu:
The current implementation of enumerated values as OWL instances generates URIs of the form prefix:TypeName#EnumValue. So, for example, I get names like http://who.int/icd#IndexType#Narrower. Given that '#' has specific meaning in a URI, it would be good if another character is used to separate the type and value names.

PR #2100, opened 2024-05-07
Update slot cardinality meaning on main linkml docs site

sujaypatil96:
Update the current description shown here: https://linkml.io/linkml/schemas/slots.html#slot-cardinality to capture the latest description of cardinality according to the method definition here: https://github.com/linkml/linkml/blob/main/linkml/generators/docgen.py#L567-L623

### DIFF
--- a/docs/schemas/slots.md
+++ b/docs/schemas/slots.md
@@ -128,15 +128,20 @@ slots:
 
 See the [type-designators](type-designators.md) section of the docs for more details.
 
-## Slot cardinality
+## Cardinality
 
-The cardinality of a slot is indicative of two properties on a slot. It tells us about whether a slot is *required* or not, and also about how many values it is allowed to have, i.e., whether it is *single-valued* or *multi-valued*.
+The cardinality of a slot in UML class diagrams indicates the number of instances of one class 
+that can be linked to an instance of another class. Cardinality (sometimes also referred to as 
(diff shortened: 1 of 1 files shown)


docs/schemas/slots.md
```diff
@@ -128,15 +128,20 @@ slots:
...
-The following list summarizes the expansions of the various possible combinations of cardinalities that can be asserted on a slot:
-* `1..*` - slot is *required* and *multivalued*
-* `1..1` - slot is *required* but *not multivalued*
-* `0..*` - slot is *not required* but if provided it must be *mulitvalued*
-* `0..1` - slot is *not required* and *not multivalued*
+## Cardinality
+
+The cardinality of a slot in UML class diagrams indicates the number of instances of one class 
+that can be linked to an instance of another class. Cardinality (sometimes also referred to as 
+*multiplicity*) notations are placed near the end of an association.
```
cmungall, line 135: "...placed near the end of an association" - did this come from UML documentation. I think this is fine for UML documentation but you need to place it in more context here

cmungall (changes requested):

docs/schemas/slots.md
sujaypatil96: I pieced it together from some UML docs and mermaid docs. But yes, you're right it would be helpful to clarify what "association" means in this context.

cmungall (approved):

PR #2102, opened 2024-05-09
`genshacl`: fix the propagation of `types` to SHACL `sh:datatype` and `sh:nodeKind`

jsheunis:
In the previous state, the edited code was never reached. With this change, a linkml or custom type as slot range will propagate to the property's `sh:datatype` in SHACL. In addition, a `sh:nodekind` of `sh:Literal` is added for such properties (based on my interpretation of [this comment](https://github.com/linkml/linkml/pull/1300#issuecomment-1439412104))

This PR is intended to close https://github.com/linkml/linkml/issues/1299. It does pretty much the same as https://github.com/linkml/linkml/pull/1300, but that PR has been stagnant for over a year and is based on old state of `main`. So I though a fresh PR might be useful to get the fix merged. Happy to close again if the preference would be to rather update the existing PR.

TODO:
- test updates

### DIFF
--- a/linkml/generators/shaclgen.py
+++ b/linkml/generators/shaclgen.py
@@ -130,7 +130,7 @@ class ShaclGenerator(Generator):
 
                             self._add_class(cl_node_pv, r)
                             range_list.append(class_node)
-                        elif r in sv.all_types().values():
+                        elif r in sv.all_types():
                             t_node = BNode()
 
                             def t_node_pv(p, v):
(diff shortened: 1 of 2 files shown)


cmungall:
🙏 

let us know if you want help with the test output. This is failing on the compliance tests. It may be the case that we want to change the test itself. You may find it useful to run this test with `--with-output` and examine the results in detail

cmungall (approved):

Issue #1483, opened 2023-06-06
Clarify how any_of interacts with previously specified ranges


cmungall:
The semantics of `any_of` should be clear in the absence of other range specifications.

However, we need to clarify how this interacts with:

1. default_range
2. ranges inherited from either the slot definition or superclasses

range inheritance should always be monotonic - refine rather than override.

## default_range

default range should only be specified when there is no explicit range *or range expression*; i.e. if we have default range string, then this should not be applied if there is a any_of with ranges

This was recently fixed in pydanticgen

- #1272

## inheritance

An element of an any_of cannot be less restrictive than any inherited range constraint

## test case

Courtesy of @Silvanoc:

```yaml
id: https://w3id.org/linkml/examples/device
name: device
prefixes:
  linkml: https://w3id.org/linkml/
  device: https://w3id.org/linkml/examples/device
imports:
  - linkml:types
default_range: string
default_prefix: device

classes:
  ConnectionPoint:

  Device:
    slots:
      connection_points

  EthernetPort:
    is_a: ConnectionPoint
    slots:
      - mac_address
    defining_slots:
      - mac_address

  IpAddress:
    is_a: ConnectionPoint
    slots:
      - ip_address
    defining_slots:
      - ip_address

  EthernetDevice:
    is_a: Device
    slot_usage:
      connection_points:
        any_of:
          - range: EthernetPort
          - range: IpAddress

slots:
  connection_points:
    range: ConnectionPoint
    multivalued: true

  mac_address:

  ip_address:
```

data:

```yaml
connection_points:
  - mac_address: 00:11:22:33:44:55
  - ip_address: 192.168.178.2
```

Currently this fails to validate, for a couple of reasons:

- [ ] jsonschemagen issue where the range is interpreted as `ConnectionPoint AND (EthernetPort OR IpAddress)`. Logically this is equivalent to `(EthernetPort OR IpAddress)`. However, when this is written directly as jsonschema we need to unroll ConnectionPoint otherwise things fail
- [ ] currently the validation necessarily goes via pythongen and pythongen doesn't fully support any_of yet.

cmungall:
See https://github.com/orgs/linkml/discussions/2926

Silvanoc:
`default_range` and `any_of` are an explosive and unstable mix... It just exploded again in my face 🤕 Once again, I tend to forget it 🫣

PR #2108, opened 2024-05-14
Adding documentation on monotonicity.

cmungall:
Aligning monotonicity tests with jsonschema behavior

See also https://github.com/linkml/linkml/issues/1483

### DIFF
--- a/docs/schemas/advanced.md
+++ b/docs/schemas/advanced.md
@@ -96,7 +96,8 @@ enums:
 Note that the range of `vital_status` is declared as `Any`, which is further constrained by the `any_of` expression.
 
 Currently, it is important to always have a range declaration (even if it is `Any`), because LinkML constraint semantics are
-monotonic. If this were not stated, then the `default_range` of string would be applied. 
+monotonic (i.e. new constraints can be specified but additional ones cannot be overridden - see [slots](slots.md) for more on this). 
+If this range declaration were not explicitly stated, then the `default_range` of string would be applied. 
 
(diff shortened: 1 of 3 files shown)


docs/schemas/advanced.md
```diff
@@ -53,13 +93,14 @@ enums:
       UNDEAD:
 ```
 
-Note that these constructs may ignored by some generators in the linkml 1.1 series.
+Note that the range of `vital_status` is declared as `Any`, which is further constrained by the `any_of` expression.
+
+Currently, it is important to have this base range declaration, because LinkML constraint semantics are
```
dalito, line 98: What exactly is meant with "base range declaration", is it "range:any"?

docs/schemas/advanced.md
```diff
@@ -53,13 +93,14 @@ enums:
...
+monotonic. If this were not stated, then the `default_range` of string would be applied.
```
dalito, line 99: "monotonic" may be unclear for non-experts. Having an explanation would be helpful. I did not know about the non-math meaning of "monotonic" before you used it in the discussions some time ago.

dalito (commented):
Just some hints were I thing it the explanation could be even clearer.

sierra-moxon (approved):

docs/schemas/advanced.md
cmungall: Thx! Linked back to slots docs for this, and added a bit of qualification

Issue #1665, opened 2023-10-12
`jsonasobj2` causes inconsistent behavior on attr assignment vs. instantiation for linkml_model classes


sneakers-the-rat:
**Describe the bug**
When instantiating a `SchemaDefinition` object with `annotations` like

```python
schema = SchemaDefinition(
  # other stuff...
  annotations = [
    {'tag': 'my_annotation', 'value': 'true'}
  ]
)
```

one will get annotations like
```python
>>> schema.annotations
{'my_annotation': Annotation(
  tag='my_annotation', 
  value='true', 
  extensions={}, 
  annotations={}
)}
```

however if one were to assign annotations after instantiation like
```python
schema = SchemaDefinition("stuff in here")
schema.annotations = {
  'my_annotation': Annotation(
    tag='my_annotation', 
    value='true'
)}
```

then one gets annotations like
```python
>>> schema.annotations
JsonObj(is_namespace=Annotation(...))
```

because of the root `JsonObj`'s `__setattr__` method that converts dicts to JsonObjs: https://github.com/hsolbrig/jsonasobj2/blob/51d966a97bc123561f31f06480957ba8a14f71af/jsonasobj2/_jsonobj.py#L163

I just tested with annotations, but this should be true of any `dict` field on any linkml model class.

**To reproduce**
```python
from linkml_runtime.linkml_model import SchemaDefinition, Annotation

schema_1 = SchemaDefinition(
  id='schema_1', name='schema_1',
  annotations=[{'tag':'my_annotation', 'value': 'true'}]
)

assigned_annotation = {'my_annotation': Annotation(tag='my_annotation', value='true')}

schema_2 = SchemaDefinition(
  id='schema_2', name='schema_2'
)
schema_2.annotations = assigned_annotation

# this will pass
assert schema_1.annotations == assigned_annotation
# this won't
assert schema_1.annotations == schema_2.annotations
```

**Expected behavior**
should be possible to have same behavior on assignment vs. instantiation for dataclasses. It's fine that the dataclass inlines the dictionary given at instantiation, but currently it's impossible as far as I can tell to replicate the instantiation behavior on assignment because assignment always gets coerced to a `JsonObj` if it's a dict. 

**About your computer (if applicable, please complete the following information):**
 - `linkml_runtime` version: 1.5.7

jsheunis:
[...]

Issue #1618, opened 2023-09-13
Include custom annotations in generated schemas / code


simontaurus:
**Is your feature request related to a problem? Please describe.**
We are currently working on a link schema repo https://opensemantic.world/ combining json-schema and json-ld (=> https://github.com/OO-LD/schema), compareable to https://github.com/linkml/linkml/issues/474 
On the way we recognized that linkml (great work!) already addresses a lot of problems we have identified, especially the bridges to OWL and SHACL. 
However, since we focus both on programmatical and graphical interfaces would also need some custom annotations 
for custom vocabular e. g. used by https://github.com/json-editor/json-editor to generate html forms.

**Describe the solution you'd like**
Option to transfer 
* a) existing annotations (like `range` even when `inlined_as_list: false`)
* b) custom annotation like `hidden: true` for autogenerated fields in forms
into jsonschema and pydantic classes

```yml
  classes:
  Person:
  Container:
    attributes:
      id:
        option:
          hidden: true
      persons:
        multivalued: true
        inlined_as_list: true
        range: Person
```

```json
{
    "$defs": {
        "Container": {
            "description": "",
            "properties": {
                "id": {
                    "options": {
                        "hidden": "true"
                    },
                    "type": "string"
                "persons": {
                    "items": {
                        "type": "string"
                    },
                    "type": "array",
                    "range": "#/$defs/Person"
                }
            },
            "title": "Container",
            "type": "object"
        },
        "Person": {
        }
    }
}
```

```python
class Person:
    pass

class Container(ConfiguredBaseModel):
    id: str
    persons: Optional[List[str]] = Field(default_factory=list, range=Person, options={'hidden': True})
```
Note: the extra parameter for Field() is also generated by https://github.com/koxudaxi/datamodel-code-generator

Maybe utilizing linkml `annotations` and dump the value as dict in the json-schema property and the pydantic Field if the generator was called with `--include-annotations=True` would be a solution for b)

**How important is this feature?** Select from the options below:
• Important - it's a blocker and can't do work without it (but of course I understand if this is out-of-scope)

**When will use cases depending on this become relevant?** Select from the options below:
• Mid-term - 2-4 months

**Additional context**
Related: 
* https://github.com/json-schema-org/json-schema-vocabularies/issues/55

cmungall:
[...]

PR #2111, opened 2024-05-15
`shaclgen`: Add `--include-annotations` option to let annotations be part of shacl shapes

jsheunis:
This is for the SHACL generator in response to https://github.com/linkml/linkml/issues/1618.

Code is added to `shaclgen.py` to:
- allow users to specify the --include-annotations tag if they want annotations (on classes, slots, and types) to be included in the exported SHACL shapes
- determine the datatype of both annotation tag and value:
   - a CURIE is identified by searching for the ':' character in the tag or value
   - if not a CURIE, the annotation tag is assumed to be a literal with type string
   - if not a CURIE, the annotation value is a literal with datatype determined using python `type` and `extended_float`, `extended_int`, `extended_str`, etc
   - add the correct triples to the shacl output (to a nodeshape for classes, and to a property shape for slots and slots with types as ranges)

In addition, the `test_shaclgen` test is updated to test for annotations on classes, and the kitchen sink schema is updated to include annotations on the person class. Several snapshot data files needed to be updated as well because of the change to that schema, so that existing tests would still succeed.

Uncertainties:
- I was unsure how to test for annotations on types and slots, because the related rdf terms in the graph generated from the kitchen sink schema would be blank nodes with transient IDs, so these can't be captured in the "EXPECTED" triples.
- Are there other elements that can also have annotations that aren't included and should ideally form part of this PR?
- I am not sure if my approach to get the element type, and hence the XSD datatype, is sensible; happy to update this approach if there are LinkML-preferred ways of doing the same
- I currently have a workaround in the code for a `jsonasobj2` issue, mentioned here: https://github.com/linkml/linkml/issues/1665#issuecomment-2112481434
- UPDATE: conflicts can arise when a slot has annotations, and if that same slot has a type as a range, where the type also has annotations. If some of these annotations have the same `tag`, the current way that the code is implemented means that the values will be appended, not replaced. It seems to me that the desired behaviour will differ based on the use case. So the question is whether more user-defined controls should be implemented to direct such behaviour?

### DIFF
--- a/linkml/generators/shaclgen.py
+++ b/linkml/generators/shaclgen.py
@@ -222,6 +222,7 @@ class ShaclGenerator(Generator):
             logging.error(f"No URI for type {rt.name}")
 
     def _add_annotations(self, func: Callable, item) -> None:
+        # TODO: migrate some of this logic to SchemaView
         sv = self.schemaview
         annotations = item.annotations
         # item could be a class, slot or type
--- a/linkml/generators/shaclgen.py
+++ b/linkml/generators/shaclgen.py
@@ -4,12 +4,18 @@ from dataclasses import dataclass
 from typing import Callable
 
 import click
+from jsonasobj2 import as_dict, JsonObj
 from linkml_runtime.linkml_model.meta import ElementName
 from linkml_runtime.utils.formatutils import underscore
+from linkml_runtime.utils.yamlutils import (   
+    extended_float,
+    extended_int,
+    extended_str,
+)
 from linkml_runtime.utils.schemaview import SchemaView
 from rdflib import BNode, Graph, Literal, URIRef
 from rdflib.collection import Collection
-from rdflib.namespace import RDF, SH
+from rdflib.namespace import RDF, SH, XSD
 
 from linkml._version import __version__
 from linkml.generators.shacl.ifabsent_processor import IfAbsentProcessor
@@ -24,6 +30,8 @@ class ShaclGenerator(Generator):
     """True means add 'sh:closed=true' to all shapes, except of mixin shapes and shapes, that have parents"""
     suffix: str = None
     """parameterized suffix to be appended. No suffix per default."""
+    include_annotations: bool = False
+    """True means include all class / slot / type annotations in generated Node or Property shapes"""
     generatorname = os.path.basename(__file__)
     generatorversion = "0.0.1"
     valid_formats = ["ttl"]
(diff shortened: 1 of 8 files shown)


linkml/generators/shaclgen.py
```diff
@@ -206,9 +214,49 @@ def _add_type(self, func: Callable, r: ElementName) -> None:
...
         if rt.uri:
             func(SH.datatype, URIRef(sv.get_uri(rt, expand=True)))
+            if rt.pattern:
+                func(SH.pattern, Literal(rt.pattern))
+            if rt.annotations and self.include_annotations:
+                self._add_annotations(func, rt)
         else:
             logging.error(f"No URI for type {rt.name}")
 
+    def _add_annotations(self, func: Callable, item) -> None:
```
cmungall, line 224: This could be moved into schemaview (but doesn't need to happen for this PR)

jsheunis:
Regarding:

> - UPDATE: conflicts can arise when a slot has annotations, and if that same slot has a type as a range, where the type also has annotations. If some of these annotations have the same `tag`, the current way that the code is implemented means that the values will be appended, not replaced. It seems to me that the desired behaviour will differ based on the use case. So the question is whether more user-defined controls should be implemented to direct such behaviour?

I just came across a relevant example. Among other annotations, I want to assign `sh:order` to slots to define how they are represented in a user interface autogenerated from the exported SHACL. E.g.:

```
slot_usage:   
   distance:
      annotations:
         sh:group: mycurie:BasicPropertyGroup
         sh:order: 0
```

the output from `gen-shacl` will then be:

```ttl
[ sh:datatype xsd:string ;
     sh:description "Distance of a thing from source location." ;
     sh:group mycurie:BasicPropertyGroup ;
     sh:nodeKind sh:Literal ;
     sh:order 0,
        27 ;
     sh:path mycurie:distance ],
```

i.e. the `sh:order` has two comma-separated values. This is because the code also assigns order to each slot that is being processed: 
https://github.com/linkml/linkml/blob/a14e6ee104bb6983f429868f9b4da785956827d2/linkml/generators/shaclgen.py#L106-L107

So, for my use case, I would want my annotation to overwrite whatever the code does internally.

cmungall:
Thank you!

yes, a limit of the current rdf testing is that it's incomplete when blank nodes are present. 

Override is the correct behavior in the case of clashes. We'll want to implement this same behavior in other rdf-serializing generators.

It would be good to extend `test_annotation` in `test_annotation_compliance`. This wouldn't add anything that is not already in your test, but the compliance testing framework is becoming the de facto standard way to test one construct across multiple generators. I am happy for this to be done in a later PR though

cmungall (approved):

jsheunis:
Does it make sense to make the `SH.order`-related patch highlighted above part if the PR as well? For me it does, but for my purposes I could also just apply the patch separately whenever I need to use shaclgen for that specific use case.

> It would be good to extend test_annotation in test_annotation_compliance. This wouldn't add anything that is not already in your test, but the compliance testing framework is becoming the de facto standard way to test one construct across multiple generators. I am happy for this to be done in a later PR though

I can take a shot at this in a next PR 👍

Issue #2122, opened 2024-05-22
JSON-LD Context Generator Enums ("@vocab": "@null")


vincentkelleher:
Hi 👋 

When trying to parse a generated JSON-LD context using tools such as [JSON-LD Playground](https://json-ld.org/playground/#startTab=tab-expanded&json-ld=%7B%22%40context%22%3A%5B%22https%3A%2F%2Fwww.w3.org%2F2018%2Fcredentials%2Fv1%22%2C%22https%3A%2F%2Fw3id.org%2Fsecurity%2Fsuites%2Fjws-2020%2Fv1%22%2C%22https%3A%2F%2Fw3id.org%2Fgaia-x%2Fdevelopment%23%22%5D%2C%22type%22%3A%5B%22VerifiableCredential%22%5D%2C%22id%22%3A%22did%3Aweb%3Agaia-x.eu%23my-legal-registration-numbe%22%2C%22issuer%22%3A%22did%3Aweb%3Alocalhost%3A3000%3Amain%22%2C%22issuanceDate%22%3A%222024-05-22T11%3A04%3A05.111%2B02%3A01%22%2C%22credentialSubject%22%3A%7B%22id%22%3A%22did%3Aweb%3Agaia-x.eu%3AlegalRegistrationNumber.json%22%2C%22type%22%3A%22gx%3AlegalRegistrationNumber%22%2C%22gx%3AvatID%22%3A%22FR79537407926%22%2C%22gx%3AvatID-countryCode%22%3A%22FR%22%7D%2C%22evidence%22%3A%5B%7B%22gx%3AevidenceOf%22%3A%22VAT_ID%22%2C%22gx%3AevidenceURL%22%3A%22https%3A%2F%2Fec.europa.eu%2Ftaxation_customs%2Fvies%2Fservices%2FcheckVatService%22%2C%22gx%3AexecutionDate%22%3A%222024-05-22T11%3A04%3A05.110%2B02%3A03%22%7D%5D%7D) with `safe` mode set to `true`, I get a `relative @vocab reference` error.

I've noticed that LinkML uses [an `ENUM_CONTEXT` constant](https://github.com/linkml/linkml/blob/2407a2f2c629092c15da6e0295600d895d34a465/linkml/generators/jsonldcontextgen.py#L23) to supply a specific context for enums. This context uses `"@vocab": "@null"` which is causing the issue mentioned above, this issue has been mentioned by @hsolbrig in #173 too.

[...]

PR #2152, opened 2024-06-10
Remove @vocab from enums in JSON LD contexts

vincentkelleher:
This fixes the issue mentioned in #2122 😊 

I've also worked on the `Deprecation` class tests which were depending on the `linkml` package version meaning they would end up by failing with version upgrades, to fix this I've introduced a mock to control this value.


vincentkelleher:
There is an issue with some Github tokenless rate limiting 😅 

image

vincentkelleher:
This is fixed and ready for review 😄

vincentkelleher:
Pull request updated with the latest LinkML updates 😉

vincentkelleher:
@cmungall I noticed that you updated the branch with a merge, I've also rebased it so it is up-to-date 😇

cmungall (approved):

PR #2157, opened 2024-06-11
Genplantuml fixes improvements

jsnyder-csdisco:
Fixes for spacing and quoting PlantUML structures and adding support for Tooltip Text in PlantUML so that SVG can contain descriptions when you hover.


jsnyder-csdisco:
I don't know why these three extra files got added to this PR.

linkml/generators/plantumlgen.py
```diff
@@ -50,6 +51,11 @@ class PlantumlGenerator(Generator):
     directory: Optional[str] = None
     kroki_server: Optional[str] = "https://kroki.io"
     load_image: bool = True
+    tooltips_flag: bool = False
+
+    def __post_init__(self):
+        self.schemaview = SchemaView(self.schema)
+        super().__post_init__()
```
sujaypatil96, line 58: Thank you for the contribution @jsnyder-csdisco!! 

`plantumlgen.py` unfortunately is one of the generators in LinkML's suite of generators that follows the _old_ visitor design pattern as opposed to the _new_ SchemaView based approach that we are using to implement generators.

You're brining in SchemaView and using it in code, which is great, that's the direction we want to be heading towards, but I'm wondering if it would be best to include your changes after an entire overhaul/transition to SchemaView for `plantumlgen.py`.

How important/urgent is your fix?

sujaypatil96 (commented):
How important/urgent is your fix?

linkml/generators/plantumlgen.py
jsnyder-csdisco: We can skip the tooltips part but the mixins just break plantuml and there are other bugs lurking in there as well because of the spacing bugs.

jsnyder-csdisco: The tool tips depend on the schema view to get the description. If there is another way to get descriptions without schema views I am happy to adjust it.

jsnyder-csdisco:
Part of this fix is import because my models cannot properly generate diagrams with mixins and other. In some cases several objects are dropped from diagram generation.

sujaypatil96:
@jsnyder-csdisco not sure if you've looked at this PR from @VladimirAlexiev before: https://github.com/linkml/linkml/pull/2060?

The state in which you have your code right now is the state in which plantumlgen.py was when @Silvanoc originally wrote the generator in the library. After that it has been modified by @VladimirAlexiev to remove quotation marks. 

We can certainly add them back and i think your comment above answers the question posed by @Silvanoc here: https://github.com/linkml/linkml/pull/2060#discussion_r1577309102

We'll wait for some time to see if @VladimirAlexiev has any concerns/objections since he was the one who requested that we remove the quotation marks.

linkml/generators/plantumlgen.py
sujaypatil96: Yup, you should be able to get the description for the LinkML object, which in this case is 'class' by simply getting the value associated with the description property (using the dot syntax).

In your case it would be:

```python
cls.description
```

sujaypatil96:
Everything on this PR looks good @jsnyder-csdisco 😁 As discussed, no rush, but if you could limit the files on this PR to just the ones that are critical/relevant, we can get this PR merged in soon after!

Thank you for your work on this!!

jsnyder-csdisco (commented):
This needs to be closed out. It's broken because of a git rebase trying to remove unwanted files.

PR #2161, opened 2024-06-13
Fix plantUML syntax bugs to do spacing and add suppor for SVG tooltips

jsnyder-csdisco:
The Plantuml Generator doesn't support proper spacing and quotations that lead to broken diagram generation and SVG generation.
This also adds support for SVG tooltips via PlantUML tooltip syntax.

### DIFF
--- a/tests/test_generators/test_plantuml.py
+++ b/tests/test_generators/test_plantuml.py
@@ -16,14 +16,14 @@ MARKDOWN_FOOTER = """
 """
 
 PERSON = """
-class "Person" {
-    {field} id: string [req]
-    {field} name: string [opt]
-    {field} age_in_years: integer [opt]
-    {field} species_name: string [opt]
-    {field} stomach_count: integer [opt]
-    {field} is_living: LifeStatusEnum [opt]
-    {field} aliases: string [0..*]
+class "Person" [[{A person, living or dead}]]{
+    {field} "id" : "string" [req]
+    {field} "name" : "string" [opt]
+    {field} "age_in_years" : "integer" [opt]
+    {field} "species_name" : "string" [opt]
+    {field} "stomach_count" : "integer" [opt]
+    {field} "is_living" : "LifeStatusEnum" [opt]
+    {field} "aliases" : "string" [0..*]
 }
 """
 
--- a/tests/test_generators/test_plantuml.py
+++ b/tests/test_generators/test_plantuml.py
@@ -16,7 +16,7 @@ MARKDOWN_FOOTER = """
 """
 
 PERSON = """
-class "Person" [[{A person, living or dead}]]{
+class "Person" [[{A person, living or dead}]] {
     {field} "id" : "string" [req]
     {field} "name" : "string" [opt]
     {field} "age_in_years" : "integer" [opt]
--- a/tests/test_generators/test_plantuml.py
+++ b/tests/test_generators/test_plantuml.py
@@ -17,13 +17,13 @@ MARKDOWN_FOOTER = """
 
 PERSON = """
 class "Person" [[{A person, living or dead}]] {
-    {field} "id" : "string" [req]
-    {field} "name" : "string" [opt]
-    {field} "age_in_years" : "integer" [opt]
-    {field} "species_name" : "string" [opt]
-    {field} "stomach_count" : "integer" [opt]
-    {field} "is_living" : "LifeStatusEnum" [opt]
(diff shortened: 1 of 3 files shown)


sujaypatil96:
Could you fix the tests that are erroring out in `test_docgen.py`?

sujaypatil96:
Note: This PR has been created in favor of https://github.com/linkml/linkml/pull/2157

jsnyder-csdisco:
> Could you fix the tests that are erroring out in `test_docgen.py`?

I will check it and make sure I didn't break this.

jsnyder-csdisco:
I understand there is some concern with the PlantUML structure that is emitted from `gen-plantuml.py` -- meaning it is not "normative" to recommended examples. The examples at https://plantuml.com/class-diagram basically outline the same structures that get generated from LinkML. So, I would argue (after using plantuml for a long time) that what comes out IS ... PlantUML because the compiler accepts it and it is normative to the dialect of the examples.  If there are other definitions, I would be happy to look at them.

tests/test_generators/test_plantuml.py
```diff
@@ -16,14 +16,14 @@
...
-class "Person" {
-    {field} id: string [req]
-    {field} name: string [opt]
-    {field} age_in_years: integer [opt]
-    {field} species_name: string [opt]
-    {field} stomach_count: integer [opt]
-    {field} is_living: LifeStatusEnum [opt]
-    {field} aliases: string [0..*]
+class "Person" [[{A person, living or dead}]] {
+    {field} "id" : "string" [req]
```
cmungall, line 20: AFAICT plantuml will render this with the quotes - I don't think we want this?

cmungall (changes requested):

tests/test_generators/test_plantuml.py
jsnyder-csdisco: I see what you mean. Quoting the class name I think is required but the "fields" are not desireable.
```plantuml
@startuml

class "Person A" [[{A person, living or dead}]] {
    {field} id : string a [req]
    {field} name : string [opt]
    {field} age in years : integer [opt]
    {field} "species_name" : "string" [opt]
    {field} "stomach_count" : "integer" [opt]
    {field} "is_living" : "LifeStatusEnum" [opt]
    {field} "aliases" : "string" [0..*]
}
@enduml
```

image
Given this "mix" of quote and not quote .. I can remove the "field quote" because PlantUML is able to handle field names with spaces in the names correctly.
Is this acceptable?

sujaypatil96: Certain"fields" in the above image still have quotes around them? like "species_name", "stomach_count", etc.?

Also, I'm wondering if we want [req] and [opt]? that is syntax that we started using in the yuml generator and it stuck, i don't know if that is standard syntax anywhere else. Same for showing cardinality.

Do we want all of that information in the class diagram?

jsnyder-csdisco: I wanted to show the "differences" to be sure this is what people wanted and that the "spaces" are handled. So, making the quotes "all go away" is the goal.
On the "req vs opt" there is a bit of a dilemma. Since we are talking about a "Model" of fields, there is no "optional" notation in PlantUML but in some languages a "field" can be optional. This is _different_ than a "range" with zero or more "instances" (e.g. array , list, link, etc). I personally find it valuable to call this out given the limitations of PlantUML.
I am happy to make this work however everyone sees fit.
Quotes being removed is a "given".

jsnyder-csdisco: BTW, there are several threads on this "optional" problem in PlantUML in many forums.

jsnyder-csdisco:
@sujaypatil96 I fixed the quoting and [req] and [opt] to be removed from the PlantUML generation to be more compatible with the examples found on the PlantUML site.

jsnyder-csdisco:
@cmungall I fixed the structure for PlantUML generate to be conformant to the PlantUML examples and removed the quotes, etc. per your requested changes. Hopefully, this meets the expectations of the team. Would be good to get this merged because I have breaking generation with Mixins.

cmungall (approved):

stanleyj-edsn:
this PR does more than the title suggests :)

by removing  `return " [req]" if slot.required else " [opt]"` there is no indication anymore whether an attribute is optional or required, so this change made the output lose useful information.
May I propose: 
`return " " if slot.required else " [0..1]"`

PR #2166, opened 2024-06-21
Adding ifabsent handling to pydanticgen Fixes #2165

### LINKED ISSUE #1334: pydanticgen generates invalid code for ifabsent=class_curie
I think the root cause of this is the same as #1333, but it manifests differently.

given a schema:

```yaml
id: test
name: test

classes:
  ClassDefinition:
    attributes:
      class_uri:
        ifabsent: class_curie
```

pydantic gen makes:

```python
class ClassDefinition(ConfiguredBaseModel):

    class_uri: Optional[str] = Field(class_class_curie)
```

however, class_class_curie is not imported, so the code does not compile

even if it were imported, the behavior would be wrong, as per #1333.

I think the fix for 1333 will fix this but for full safety perhaps we should be especially carefully about reusing ifabsent_functions across dataclass and pydantic generation

### LINKED ISSUE #2165: add ifabsent handling to pydanticgen

cmungall:
Also fixes #1334

### DIFF
--- a/linkml/generators/pydanticgen/pydanticgen.py
+++ b/linkml/generators/pydanticgen/pydanticgen.py
@@ -71,6 +71,7 @@ DEFAULT_IMPORTS = (
     + Import(module="decimal", objects=[ObjectImport(name="Decimal")])
     + Import(module="enum", objects=[ObjectImport(name="Enum")])
     + Import(module="re")
+    + Import(module="uuid")
     + Import(module="sys")
     + Import(
         module="typing",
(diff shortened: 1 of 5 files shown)


sierra-moxon (approved):

sierra-moxon:
rebase not available here b/c of conflicts, updated via merge instead.

amc-corey-cox:
Closing this PR — its contributions have been incorporated across two targeted PRs:

- **#2366** resolved the ifabsent special cases (`class_curie`, `slot_curie`, `class_uri`, `slot_uri`, `default_range`) that were producing `None` values. This covered the Python, SHACL, and compliance test improvements from this PR.
- **#3238** (new) pulls the pydantic `bnode` fix from this PR into a clean implementation, fixing #3237 (and the originally reported #1334).

Together these fully supersede this PR. Thank you for the original work, @cmungall!

amc-corey-cox:
Closing in favor of #2366 and #3238 — see above comment for details.

Issue #355, opened 2021-09-21
Mixins aren't factored into post-init for pythongen classes


kevinschaper:
I'm hitting this problem with generated python classes for the biolink model.  It looks like it doesn't recognize that GeneOrGeneProduct is a mixin of Gene.

In the biolink model, Gene is defined with:

``` gene:
    description: >-
      A region (or regions) that includes all of the sequence elements
      necessary to encode a functional transcript. A gene locus may include
      regulatory regions, transcribed regions and/or other
      functional sequence regions.
    is_a: biological entity
    mixins:
      - gene or gene product
      - genomic entity
      - chemical entity or gene or gene product
      - physical essence
      - ontology class
    aliases: ['locus']
    slots:
      - symbol
      - synonym
      - xref ...
```

and GeneToPhenotypicFeatureAssociation is defined as:

```
gene to phenotypic feature association:
    is_a: association
    exact_mappings:
      - WBVocab:Gene-Phenotype-Association
    defining_slots:
      - subject
      - object
    mixins:
      - entity to phenotypic feature association mixin
      - gene to entity association mixin
    slot_usage:
      subject:
        range: gene or gene product
        description: "gene in which variation is correlated with the phenotypic feature"
        examples:
          - value: HGNC:2197
            description: "COL1A1 (Human)"
```

My code

```
gene = Gene(id='Xenbase:' + row['SUBJECT'], category="biolink:Gene")

phenotype = PhenotypicFeature(id=row['OBJECT'], category="biolink:PhenotypicFeature")

association = GeneToPhenotypicFeatureAssociation(
    category="biolink:GeneToPhenotypicFeatureAssociation",
    id="uuid:" + str(uuid.uuid1()),
    subject=gene,
    predicate="biolink:has_phenotype",
    object=phenotype,
    relation=row['RELATION'].replace('_', ':')
)```

This fails, it’s hitting this spot in the post init,
```if not isinstance(self.subject, GeneOrGeneProduct):
    self.subject = GeneOrGeneProduct(**as_dict(self.subject))
```
and giving this error
```  File "/Users/kschaper/Documents/Monarch/koza/examples/xenbase/gene-to-phenotype.py", line 16, in <module>
    association = GeneToPhenotypicFeatureAssociation(
  File "<string>", line 20, in __init__
  File "/Users/kschaper/Documents/Monarch/koza/koza/biolink/model.py", line 6376, in __post_init__
    self.subject = GeneOrGeneProduct(**as_dict(self.subject))
[...]

PR #2167, opened 2024-06-21
Update pythongen.py to include mixins in class inheritance.

Maranju:
Mixins declared in yaml file are added to python dataclass generation. Order of class definition in file adapted such that all mixed-in classed are already defined.


cmungall:
Thanks for this! We'll want to do some additional testing on this as the metamodel makes heavy use of mixins

cmungall:
Fixes #355

sierra-moxon:
As we need more tests for this PR, I've recreated the branch with the original change and authorship here: maranju-doc-patch and updated the linked issues accordingly.  closing in favor of a new PR with tests.

PR #2172, opened 2024-06-21
Mappings section should show on individual element pages and not just classes

sujaypatil96:
Prior to this PR it used to be the case that the `Mappings` section got rendered only on class documentation pages. This came in as a request from @wdduncan where he pointed that we should ideally be able to see `Mappings` on all element pages (class, slot, etc.) if present.


cmungall (approved):

wdduncan:
Thanks @sujaypatil96 !

PR #2175, opened 2024-06-26
Unify ifabsent processing for Shacl, Pydantic and Python

vincentkelleher:
Sorry for the pretty big pull-request there has been a lot of replacing in multiple files

## Unify The IfAbsentProcessor Approach

Within a previous pull-request, I introduced the `IfAbsentProcessor` class which was only used by the Shacl generator at that time. 

This processor has become an abstract class which is extended for Shacl (`ShaclIfAbsentProcessor`) and Python/Pydantic (`PythonIfAbsentProcessor`). It is now easy to maintain and evolve those processors to support new features or even new generator types.

Everything I've developed in those processors is a 100% tested and I made sure the changes were backwards compatible except for one case in `test_ifabsent_compliance.py` which isn't possible anymore with the new processors.

## Support Enum Default Values In Python/Pydantic Generators

The main goal of this pull-request was to support enum default values in the Python generator to be able to use snippets like:

```yaml
id: ifabsent_tests
name: ifabsent_tests

prefixes:
  ex: https://example.org/
default_prefix: ex

classes:
  Student:
    attributes:
      - name: presence
        range: PresenceEnum
        ifabsent: PresenceEnum(Missing)

enums:
  PresenceEnum:
    permissible_values:
      Present:
        description: It's there.
      Missing:
        description: It's not there.
```

The *Slot* documentation chapter has been updated to reflect this change. FYI, this is already used in the Shacl generator.


vincentkelleher:
This pull-request depends on the metadata changes in https://github.com/linkml/linkml-runtime/pull/325

linkml/generators/python/python_ifabsent_processor.py
```diff
@@ -0,0 +1,119 @@
...
+                raise ValueError(
+                    f"The ifabsent value `{slot.ifabsent}` of the `{slot.name}` slot does not match a valid boolean "
+                    f"value"
+                )
+
+        if slot.range in [Integer.type_name, Float.type_name, Double.type_name, Decimal.type_name]:
+            return ifabsent_default_value
+
+        if slot.range == Time.type_name:
+            match = re.match(r"^(\d{2}):(\d{2}):(\d{2}).*$", ifabsent_default_value)
```
sneakers-the-rat, line 61: just cruising through recent PRs and taking a look, don't mind me. 

it feels like these patterns should live in the metamodel, or at least in the parent class. ran into this just today where i was wondering if there was some kind of schema validation step on load, and this is another good example - this would be true for every generator, not just python, right?

so you could imagine the parent class being like

```python
class IfAbsentProcessor(ABC):
    """
    Processes value of ifabsent slot.

    See `<https://w3id.org/linkml/ifabsent>`_.
    """

    patterns = {
      Time.type_name: r"^(\d{2}):(\d{2}):(\d{2}).*$",
      Date.type_name: r"^(\d{4})-(\d{2})-(\d{2})$",
      # etc.
    }

    ifabsent_regex = re.compile("""(?:(?P<type>\w+)\()?[\"\']?(?P<default_value>[^\(\)\"\')]*)[\"\']?\)?""")

    def __init__(self, schema_view: SchemaView):
        self.schema_view = schema_view

    def process_slot(self, slot: SlotDefinition, cls: ClassDefinition) -> Optional[str]:
        if not slot.ifabsent:
            return None

        ifabsent_match = self.ifabsent_regex.search(slot.ifabsent)
        ifabsent_default_value = ifabsent_match.group("default_value")
        self.validate(slot)

        if slot.range == Date.type_name:
            return self._handle_date(slot)
        # and so on...

    def validate(slot):
        if not re.match(self.patterns[slot.range], slot.ifabsent):
            raise ValueError('some informative error message')

    @abs.abstractmethod
    def _handle_date(slot) -> str:
        # ...
```

where you could skip a lot of the boilerplate and make implementation for all the other generators a lot easier and more structured :)

vincentkelleher: Hi @sneakers-the-rat 👋 

I'm happy you brought that up because, while I was developing this, I was wondering if I should add a specific abstract method for each LinkML range/datatype to force (or at least inform) upcoming developers that there are many types to manage not just the ones they use in their use case.

Having the the common regular expressions in the parent class is a good idea indeed 👍  In a near future we can move relevant parts to the LinkML metamodel, runtime or other but in the mean time we can focus on moving things to the parent class.

What do you think ? 😊

sneakers-the-rat: I think up to you!!! In general I feel like the generators need to be overhauled to have lifecycle methods incl. Validation, transformation, and generation which is a much longer project, but I like what youve done here with this `common` package and making the python and shacl generators into packages as well. The generators have a lot of duplicated logic and wiggle room in implementation so having more specific handlers for specific parts of the metamodel is great - currently the compliance tests are the things measuring how compliant the generators are, but if we had structured subgenerators like this for everything we could just specify what needs to be implemented for each generator.

Like I said I was just doing something similar making a validator for `ArrayExpression` and I couldnt find any schema validation stuff: https://github.com/sneakers-the-rat/linkml/blob/a3cfb2b82d7519cf9c64d113250c1714db2b3f6e/linkml/generators/pydanticgen/array.py#L190

And previously this: https://github.com/sneakers-the-rat/linkml/blob/a3cfb2b82d7519cf9c64d113250c1714db2b3f6e/linkml/generators/pydanticgen/array.py#L308

Both of which I think should be made into `common` ABCs, so ill follow your lead on moving them there.

vincentkelleher: I've started the refactoring work which I will finish next week as a few `TODO`s are remaining 😊 

Thanks for the suggestions @sneakers-the-rat 🙏

vincentkelleher:
I rebased this on the latest `main` changes 😉

cmungall (approved):

vincentkelleher:
Thank you @cmungall for your multiple approvals and merges concerning this PR 🙏

sneakers-the-rat:
@vincentkelleher just want to say again that i love this!

vincentkelleher:
> @vincentkelleher just want to say again that i love this!

Thank you for the tips above and the positive feedback 🙏

PR #2178, opened 2024-06-28
Array Representation 2: `numpydantic`

sneakers-the-rat:
what up it is me again I have brought you this:

```yaml
MyModel:
  attributes:
    array:
      range: integer
      array:
        dimensions:
        - alias: x
          exact_cardinality: 3
        - alias: y
          exact_cardinality: 4
        - alias: z
```

which makes this

```python
from pydantic import BaseModel
from numpydantic import NDArray, Shape

class MyModel(BaseModel):
    array: NDArray[Shape["3 x, 4 y, * z"], int]
```

which you can use like this:

```python
import numpy as np
import dask.array as da
import zarr

# numpy
model = MyModel(array=np.zeros((3, 4, 5), dtype=int))
# dask
model = MyModel(array=da.zeros((3, 4, 5), dtype=int))
# hdf5 datasets
model = MyModel(array=('data.h5', '/nested/dataset'))
# zarr arrays
model = MyModel(array=zarr.zeros((3,4,5), dtype=int))
model = MyModel(array='data.zarr')
model = MyModel(array=('data.zarr', '/nested/dataset'))
# video files
model = MyModel(array="data.mp4")
```

more examples (or just see the rendered docs):

<details>
<summary>expand/collapse examples</summary>

## Any Shape

```python
class AnyType(ConfiguredBaseModel):
    array: Optional[NDArray] = Field(None)

class Typed(ConfiguredBaseModel):
    array: Optional[NDArray[Any, int]] = Field(None)
```

## Bounded

```python
class MinDimensions(ConfiguredBaseModel):
    array: Optional[NDArray[Shape["*, *, ..."], int]] = Field(None)
    
class MaxDimensions(ConfiguredBaseModel):
    array: Optional[
        Union[
            NDArray[Shape["*"], int],
            NDArray[Shape["*, *"], int],
            NDArray[Shape["*, *, *"], int],
            NDArray[Shape["*, *, *, *"], int],
            NDArray[Shape["*, *, *, *, *"], int],
        ]
    ] = Field(None)

class ExactDimensions(ConfiguredBaseModel):
    array: Optional[NDArray[Shape["*, *, *"], int]] = Field(None)

```

## Parameterized

```python
class ParameterizedArray(ConfiguredBaseModel):
    array: Optional[
        NDArray[Shape["2-* min, *-5 max, 2-5 range, 6 exact"], int]
    ] = Field(None)
```

## Complex

```python
class ComplexRangeShapeArray(ConfiguredBaseModel):
    array: Optional[
        Union[
            NDArray[Shape["*-5 max, 2-* min, 2-5 range, 6 exact, *"], int],
            NDArray[Shape["*-5 max, 2-* min, 2-5 range, 6 exact, *, *"], int],
            NDArray[Shape["*-5 max, 2-* min, 2-5 range, 6 exact, *, *, *"], int],
        ]
    ] = Field(None)
```

</details>

so anyway that should be pretty cool for some data formats and standards that yno like "abstraction between specification and implementation"

## Other stuff in here

- I made an `ArrayValidator` class to consolidate the checking of all the ways that it's technically possible but invalid to specify arrays. I noticed that the existing `validator` module looked like it was for validating instances of data against a schema, rather than validating the schema itself, but if there is already an existing system for doing that then I made it pretty decouply so it should be pretty easy to refactor into that.
- Correspondingly i made a module for custom exceptions in `utils` so i could have a `ValidationError` that can be handled specifically - i couldn't find an `exceptions` module anywhere else so lmk if there already is one and i can move mine there. 
- added a pytest mark `pydantic_npd` so that numpydantic tests could be run separately (and easily skipped)

Aside from that there really isn't much in here - numpydantic handles all the hard stuff separately, and the way that the docs and tests are structured gave them all to us for free :)

### DIFF
--- a/docs/schemas/arrays.md
+++ b/docs/schemas/arrays.md
@@ -113,12 +113,12 @@ def render_comparison(path, cls, string=False) -> str:
         class_def = {a_cls: class_def}
         class_strs.append(YAMLDumper().dumps(class_def))
         pydantic_strs.append(render_class(path, a_cls))
-        npd_strs.append(render_class(path, a_cls, representation='numpdantic'))
+        npd_strs.append(render_class(path, a_cls, representation='numpydantic'))
         
     class_str = "\n".join(class_strs)
     pydantic_str = "\n".join(pydantic_strs)
     npd_str = "\n".join(npd_strs)
-    md = COMPARISON.format(linkml=class_str, pydantic_lol=pydantic_str)
+    md = COMPARISON.format(linkml=class_str, pydantic_lol=pydantic_str, pydantic_npd=npd_str)
     if string:
         return md
     else:
@@ -173,6 +173,65 @@ try:
 except Exception as e:
     console.print(e)
 ```
+
+### Array Library Integration
+
+The most basic kind of array annotations that the pydantic generator can produce are "list of list" style arrays,
+but most practical uses of arrays require specialized array libraries. 
+
+The {class}`.PydanticGenerator` also supports generating array annotations with {mod}`numpydantic`, which allows
+you to use a single model with an extensible set of array libraries.
+
+To generate numpydantic-style array annotations:
+
+```python
+PydanticGenerator('my_schema.yaml', array_representations=['numpydantic'])
+```
+
+Which yields a model like this (and see the examples throughout the rest of this page)
+
+```python
+from pydantic import BaseModel
+from numpydantic import NDArray, Shape
+
+class MyModel(BaseModel):
+    array: NDArray[Shape["3 x, 4 y, * z"], int]
+```
+
+Then use it as you please:
+
+```python
+import numpy as np
+import dask.array as da
+import zarr
+
+# numpy
+model = MyModel(array=np.zeros((3, 4, 5), dtype=int))
+# dask
+model = MyModel(array=da.zeros((3, 4, 5), dtype=int))
+# hdf5 datasets
+model = MyModel(array=('data.h5', '/nested/dataset'))
+# zarr arrays
+model = MyModel(array=zarr.zeros((3,4,5), dtype=int))
+model = MyModel(array='data.zarr')
+model = MyModel(array=('data.zarr', '/nested/dataset'))
+# video files
+model = MyModel(array="data.mp4")
+```
+
+This makes it possible for your schema to be extremely *implementation general* - data formats and standards
+can support many array backends out of the box, reducing coupling between the abstract standard and its concrete implementation.
+More humble array users get powerful array modeling tools right from a yaml schema.
+
+```{tip}
+To use numpydantic arrays, you'll need to add it as a dependency in your project.
+
+Schema rendered with numpydantic arrays will include a module-level constant `NUMPYDANTIC_VERSION` which specifies the
+minimum version that it was designed for, but if you already depend on `linkml` in your project you can add the
+`numpydantic` extra (eg. `pip install linkml[numpydantic]`) which will give you a `numpydantic` dependency that matches
+the version supported by `linkml`.
+```
+
 
 ### Specification
 
--- a/docs/schemas/arrays.md
+++ b/docs/schemas/arrays.md
@@ -75,25 +75,25 @@ COMPARISON = """
 :::
 ::::
 ::::{{tab-item}} numpydantic
-:::{{admonition}} Coming Soon
-Direct support for array libraries like numpy, hdf5, dask, zarr, and xarray with pydantic validation <3
+:::{{code-block}} python
+{pydantic_npd}
 ::: 
 ::::
 :::::
 """
 
-def render_module(path):
-    generator = PydanticGenerator(str(path), array_representations=['list'])
+def render_module(path, representation='list'):
+    generator = PydanticGenerator(str(path), array_representations=[representation])
     module = generator.render()
     return module
     
-def compile_module(path):
-    generator = PydanticGenerator(str(path), array_representations=['list'])
+def compile_module(path, representation='list'):
+    generator = PydanticGenerator(str(path), array_representations=[representation])
     module = generator.compile_module()
     return module
 
-def render_class(path, cls) -> str:
-    module = render_module(path)
+def render_class(path, cls, representation='list') -> str:
+    module = render_module(path, representation)
     cls = module.classes[cls]
     code = cls.render(black=True)
     return code
@@ -106,15 +106,18 @@ def render_comparison(path, cls, string=False) -> str:
     sch = YAMLLoader().load_as_dict(path)
     class_strs = []
     pydantic_strs = []
+    npd_strs = []
     
     for a_cls in cls:
         class_def = sch['classes'][a_cls]
         class_def = {a_cls: class_def}
         class_strs.append(YAMLDumper().dumps(class_def))
         pydantic_strs.append(render_class(path, a_cls))
+        npd_strs.append(render_class(path, a_cls, representation='numpdantic'))
         
     class_str = "\n".join(class_strs)
     pydantic_str = "\n".join(pydantic_strs)
+    npd_str = "\n".join(npd_strs)
     md = COMPARISON.format(linkml=class_str, pydantic_lol=pydantic_str)
     if string:
         return md
@@ -284,8 +287,12 @@ class Typed(ConfiguredBaseModel):
 :::
 ::::
 ::::{tab-item} numpydantic
-:::{admonition} Coming Soon
-Direct support for array libraries like numpy, hdf5, dask, zarr, and xarray with pydantic validation <3
+:::{code-block} python
+class AnyType(ConfiguredBaseModel):
+    array: Optional[NDArray] = Field(None)
+
+class Typed(ConfiguredBaseModel):
+    array: Optional[NDArray[Any, int]] = Field(None)
 ::: 
 ::::
 :::::
@@ -475,10 +482,10 @@ render_comparison(sch, 'ComplexAnyShapeArray')
 At release, only pydanticgen supports arrays, but arrays will be implemented gradually for the rest of the generators.
 
 | generator                 | representation | anyshape | bounded | parameterized | complex |
-|---------------------------|----------------|----------|-----------|---------|-------|
-| [pydantic](pydanticgen)   | List of Lists  | Y        | Y         | Y       | Y     |
-| [pydantic](pydanticgen)   | Numpydantic    | X        | X         | X       | X     |
-| ... (the rest of em)      |                |          |           |         |       |
+|---------------------------|----------------|----------|---------|---------------|---------|
+| [pydantic](pydanticgen)   | List of Lists  | Y        | Y       | Y             | Y       |
+| [pydantic](pydanticgen)   | Numpydantic    | Y        | Y       | Y             | Y       |
+| ... (the rest of em)      |                |          |         |               |         |
 
 (array-representations)=
 ### Representations
@@ -487,16 +494,16 @@ Since arrays are unlike other data types in that they usually require some speci
 and many formats don't have a single canonical array type, generators may accommodate multiple array
 representations.
 
-The pydantic generator currently supports a "List of lists" style array representation to be able to use
+The basic representation supported by the pydantic generator is "List of lists" style array representation to be able to use
 arrays without any additional external dependencies beyond pydantic. 
 
-See each [generator](../generators/index.rst)'s documentation page for a summary of the array representations they support.
+Pydanticgen now *also* supports most common array libraries from a single annotation using {mod}`numpydantic` - 
+a single array specification generates a single pydantic model, but the numpydantic {class}`numpydantic.NDArray` 
+type abstracts the validation process for an extensible set of array libraries. Use whatever you want as an array,
+why stop at numpy arrays - currently it also supports hdf5, zarr, video files, and also allows custom array
+interfaces via subclassing. See the [numpydantic docs](https://numpydantic.readthedocs.io/en/latest/) for more.
 
-```{admonition} Numpy And Friends
-We are in the process of implementing an additional pydantic array representation that can
-directly work with numpy-style arrays, including lazy loading and passthrough array routines for
-several major array libraries as a standalone package. See {mod}`numpydantic` 
-```
+See each [generator](../generators/index.rst)'s documentation page for a summary of the array representations they support.
 
 ### Implementation Guidance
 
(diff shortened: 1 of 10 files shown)


.github/workflows/test-pydantic1.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
       # install pydantic
       #----------------------------------------------
       - name: Install Pydantic
-        run: poetry add pydantic@^1
+        run: poetry run pip install 'pydantic<2'
```
rly, line 54: Just curious - why use pip to install pydantic into the poetry env instead of poetry?

cmungall:
> I made an ArrayValidator class to consolidate the checking of all the ways that it's technically possible but invalid to specify arrays. I noticed that the existing validator module looked like it was for validating instances of data against a schema, rather than validating the schema itself, but if there is already an existing system for doing that then I made it pretty decouply so it should be pretty easy to refactor into that.

Currently there are limited supported mechanisms for semantic validation of schemas, but you do get this as a "side-effect" of some other frameworks (of course, for structural validation then just validate the schema against the metamodel)

- when translating a schema to OWL, you can use an OWL reasoner to detect if a schema is unsatisfiable
- older generators use schemaloader which performs additional *procedural* semantic checks.

We do have a mapping between schemas and logical formulae which can be reasoned over to check for unsatisfiabilities such as inconsistent min/max ranges etc.

https://github.com/linkml/linkml/blob/main/linkml/transformers/logical_model_transformer.py

.github/workflows/test-pydantic1.yaml
sneakers-the-rat: Bc the add fails here since numpydantic requires pydantic 2, the pip install just updates pydantic

sneakers-the-rat:
Would you like me to try and adapt what i wrote there for array validation to that? If we don't do semantic schema validation on load in general (ie in schemaview) then i'm not sure what advantage it might be, since each of the generators would have to call it themselves anyway, but i could move it to the new `linkml.generators.common` package (which i love)?

cmungall:
We’re on linkml 1.8 now and have dropped support for pyd 2 but may need to
update our toml to reflect!

sneakers-the-rat:
oo let me handle that :)

sneakers-the-rat:
This needs to be rebased, but I wonder if we can get a review on this sometime soon @rly ? 

Numpydantic is [getting pretty cool](https://numpydantic.readthedocs.io/en/dump_json/serialization.html) and it is maturing along with nwb-linkml to make for a pretty sweet data modeling tool when combined with linkml arrays models.

Am getting close to when I want to push a version of nwb-linkml to pypi but I cant do that until a few PRs are merged here and I can get rid of this: https://github.com/p2p-ld/nwb-linkml/blob/f94a144d75411c3c2d370d7171b03a0af7eb37c4/nwb_linkml/pyproject.toml#L18

linkml/generators/pydanticgen/array.py
```diff
@@ -20,11 +20,12 @@
 
 from linkml.generators.pydanticgen.build import SlotResult
 from linkml.generators.pydanticgen.template import ConditionalImport, Import, Imports, ObjectImport
+from linkml.utils.exceptions import ValidationError
 
 
 class ArrayRepresentation(Enum):
     LIST = "list"
-    NPARRAY = "nparray"  # numpy and nptyping must be installed to use this
+    NUMPYDANTIC = "numpydantic"  # numpy and nptyping must be installed to use this
```
rly, line 28:
```suggestion
    NUMPYDANTIC = "numpydantic"  # numpydantic must be installed to use this
```

linkml/generators/pydanticgen/array.py
```diff
@@ -175,6 +176,124 @@ def _validate(_v: Union[List[_T], list]):
...
+        cls.dimension_ordinal(dimension)
+
+    @staticmethod
+    def array_exact_dimensions(array: ArrayExpression):
+        """Arrays can have exact_number_dimensions OR min/max_number_dimensions, but not both"""
+        if array.exact_number_dimensions is not None and (
+            array.minimum_number_dimensions is not None or array.maximum_number_dimensions is not None
+        ):
+            raise ValidationError(
+                "Can only specify EITHER exact_number_dimensions OR minimum/maximum dimensions, " f"got: {array}"
```
rly, line 226:
```suggestion
                "Can only specify EITHER exact_number_dimensions OR minimum/maximum dimensions, "
                f"got: {array}"
```

linkml/generators/pydanticgen/array.py
```diff
@@ -175,6 +176,124 @@ def _validate(_v: Union[List[_T], list]):
...
+            )
+
+    @staticmethod
+    def dimension_exact_cardinality(dimension: DimensionExpression):
+        """Dimensions can only have exact_cardinality OR min/max_cardinality, but not both"""
+        if dimension.exact_cardinality is not None and (
+            dimension.minimum_cardinality is not None or dimension.maximum_cardinality is not None
+        ):
+            raise ValidationError(
+                "Can only specify EITHER exact_cardinality OR minimum/maximum cardinality, " f"got: {dimension}"
```
rly, line 283:
```suggestion
                "Can only specify EITHER exact_cardinality OR minimum/maximum cardinality, " 
                f"got: {dimension}"
```

linkml/generators/pydanticgen/array.py
```diff
@@ -394,30 +525,21 @@ def complex_dimensions(self, array: ArrayExpression) -> SlotResult:
                 # equivalent to labeled shape
                 return self.parameterized_dimensions(array)
             else:
-                raise ValueError(
-                    "if exact_number_dimensions is provided, it must be greater than the parameterized dimensions"
-                )
+                # invalid - raise exception
+                ArrayValidator.array_consistent_n_dimensions(array)
```
rly, line 529: Since `validate` is called at the beginning of `make`, this shouldn't trigger, right? I guess, unless someone calls `complex_dimensions` without calling `make` first. I don't mind having it here, but maybe we should note that or add `# pragma: no cover` since it should not be reachable.

linkml/generators/pydanticgen/array.py
```diff
@@ -446,13 +569,177 @@ def complex_dimensions(self, array: ArrayExpression) -> SlotResult:
...
+        ):
+            exact_dims = array.exact_number_dimensions or array.minimum_number_dimensions
+
+            return SlotResult(annotation=self.ndarray_annotation(["*"] * exact_dims, self.dtype))
+        elif not array.maximum_number_dimensions and (
+            array.minimum_number_dimensions is None or array.minimum_number_dimensions == 1
+        ):
+            return self.any_shape()
+        elif array.maximum_number_dimensions:
+            # e.g., if min = 2, max = 3, annotation = Union[List[List[dtype]], List[List[List[dtype]]]]
```
rly, line 671:
```suggestion
            # e.g., if min = 2, max = 3, annotation = Union[NDArray[Shape["*, *"], dtype], NDArray[Shape["*, *, *"], dtype]]
```

linkml/generators/pydanticgen/array.py
```diff
@@ -446,13 +569,177 @@ def complex_dimensions(self, array: ArrayExpression) -> SlotResult:
...
+            min_dims = array.minimum_number_dimensions if array.minimum_number_dimensions is not None else 1
+            annotations = [
+                self.ndarray_annotation(["*"] * i, self.dtype)
+                for i in range(min_dims, array.maximum_number_dimensions + 1)
+            ]
+            return SlotResult(annotation="Union[" + ", ".join(annotations) + "]")
+        else:
+            # min specified with no max
+            # e.g., if min = 3, annotation = List[List[AnyShapeArray[dtype]]]
```
rly, line 680:
```suggestion
            # e.g., if min = 3, annotation = NDArray[Shape[*, *, *, ...], dtype]
```

rly:
I suggested a few corrections to comments, but otherwise this looks good to me! Sorry for the delay in review.

cmungall (approved):

linkml/generators/pydanticgen/array.py
sneakers-the-rat: i think black will probably want to put this back on one line, but if it does i'll combine the strings

linkml/generators/pydanticgen/array.py
sneakers-the-rat: oh good catch. ya the validator was intended to remove the need for all these check cases. I guess it's here just for logical completeness. something feels wrong about just falling through but you're right it shouldn't be reachable. maybe one thing is that even though they're not part of the intended API via the `make` method, they are public methods, so it would be rare but possible?

if we took this out we fall through to the error message on L473, which is more generic.

so i'm thinking we either make these private methods and take this out, or another thing i could do is run each of the validation failure tests against the relevant method if we're gonna leave em as public. also fine with taking this out np.

linkml/generators/pydanticgen/array.py
```diff
@@ -316,28 +447,19 @@ def complex_dimensions(self, array: ArrayExpression) -> RangeResult:
...
                 res.range = self._list_of_lists(array.minimum_number_dimensions - len(array.dimensions), res.range)
 
-        elif array.minimum_number_dimensions and array.maximum_number_dimensions is None:
-            raise ValueError(
-                (
-                    "Cannot specify a minimum_number_dimensions while maximum is None while using labeled dimensions - "
-                    "either use exact_number_dimensions > len(dimensions) for extra parameterized dimensions or set "
-                    "maximum_number_dimensions explicitly to False for unbounded dimensions"
-                )
-            )
```
sneakers-the-rat, line 340: in line with above, would want to be consistent, so if we leave errors in then we might as well bring these back and use the validator to raise errors, or else remove them all and leave some holes in the switch statements

linkml/generators/pydanticgen/array.py
```diff
@@ -366,13 +489,177 @@ def complex_dimensions(self, array: ArrayExpression) -> RangeResult:
...
     """
-    Represent array range with nptyping, and serialization/loading with an ArrayProxy
+    Represent array range with :class:`numpydantic.NDArray` annotations,
+    allowing an abstract array specification to be used with many different array
+    libraries.
     """
 
-    REPR = ArrayRepresentation.NPARRAY
+    REPR = ArrayRepresentation.NUMPYDANTIC
+    NUMPYDANTIC_VERSION = "1.2.1"
```
sneakers-the-rat, line 500: semantics on this - should this be the 'target' version or the current numpydantic version if present and installed at generation time? the goal would be to hint at what version of numpydantic this model could be used with, but there's not really a good way of keeping this true.

maybe it should be "minimum numpydantic version," so if it gets updated to use newer version features that's when you would update this. currently we're at `v1.6.1` and that just added `Union` types which we would want to have, so i'll bump this

sneakers-the-rat: i'm planning on adding linkml tests into the numpydantic tests, that's what the `pydanticgen_npd` mark is for, just fyi, so i'll be testing for not breaking this downstream

linkml/generators/pydanticgen/array.py
sneakers-the-rat: thanks, whoops

tests/test_generators/test_pydanticgen.py
```diff
@@ -1487,15 +1495,18 @@ def expectation(self):
     [TestCase(type="pass", array=np.zeros((3, 4, 5, 6), dtype=dt)) for dt in (int, float, str)]
     + [TestCase(type="fail", array=a) for a in (4, 3.0, "three")],
 )
-@pytest.mark.parametrize("representation", [[ArrayRepresentation.LIST], [ArrayRepresentation.NPARRAY]])
+@pytest.mark.parametrize(
+    "representation",
+    [[ArrayRepresentation.LIST], pytest.param([ArrayRepresentation.NUMPYDANTIC], marks=pytest.mark.pydantic_npd)],
```
sneakers-the-rat, line 1500: this is very not DRY, i'm going to split out the array representation parameterization and reuse it or maybe make it into a parameterized fixture instead of copy pasting this a million times

linkml/generators/pydanticgen/array.py
sneakers-the-rat: huh weird it didn't commit in the batch and i think my comment got lost? i'll commit this but i think black will want it to be on one line, but if it does i'll combine the strings

tests/test_generators/test_pydanticgen.py
sneakers-the-rat: done in https://github.com/linkml/linkml/pull/2178/commits/7ff138e7d9ab25bc64f51312210c1a41f93162d0

docs/schemas/arrays.md
```diff
@@ -487,16 +553,16 @@ Since arrays are unlike other data types in that they usually require some speci
 and many formats don't have a single canonical array type, generators may accommodate multiple array
 representations.
 
-The pydantic generator currently supports a "List of lists" style array representation to be able to use
+The basic representation supported by the pydantic generator is "List of lists" style array representation to be able to use
 arrays without any additional external dependencies beyond pydantic.
```
rly, line 557:
```suggestion
The basic representation supported by the pydantic generator is the "List of lists" style array representation. This can be used without any additional dependencies beyond pydantic.
```

linkml/generators/pydanticgen/array.py
rly: I like making these private, taking out these errors, and maybe leaving a comment saying that the missing branch should not occur because validation happens first or something like that.

sneakers-the-rat: I cant think of a circumstance where it would be preferable to call the generation method over the `make` method. Lets make em private.

Im gonna assume the methods in the Validator class are clear enough as documentation for what legs of the if/else clauses should be impossible and just remove them entirely

sneakers-the-rat:
i also added some more dtype tests for both generators: unions, numpy types (via repr and adding an import), and class dtypes. in the process fixed the `AnyShapeArray`'s handling of unions and of non-builtin types.

sneakers-the-rat:
ok i think those are all the changes i want to make, good to go :)

sneakers-the-rat:
@cmungall @rly as promised, tests against linkml have been added to numpydantic: https://github.com/p2p-ld/numpydantic/pull/26

these use the `pydanticgen_npd` marker to only run the relevant subset of tests, so any behavior tested using that marker will not break from changes upstream. I set it to automatically be applied using the parameterized `array_representation` fixture, so any future tests that use that should have it correctly applied, but otherwise it would need to be applied manually. 

Here's a run where i purposely broke numpydantic to ensure that the CI test was using the PR version of numpydantic:
- commit: https://github.com/p2p-ld/numpydantic/pull/26/commits/5b59e9fc07d2525de707121df15b3c1a0be3ccf3
- failing run: https://github.com/p2p-ld/numpydantic/actions/runs/11043243729/job/30677172830

and a working run showing that the tests are run and pass: https://github.com/p2p-ld/numpydantic/actions/runs/11043282805

i tag u here to say this to say that i don't want to add a broken thing to linkml and take that seriously, so if you do choose to use it it won't break from negligence (though it may break for [other reasons](https://github.com/h5py/h5py/issues/2506) that are outside of my control) <3

PR #2183, opened 2024-06-29
Document contributing SOP for slow snapshot tests

cmungall:


cmungall:
can someone explain why this fails:

` poetry run pytest  --generate-shapshots --with-slow tests/test_base/test_metamodel.py`

but this succeeds:

`poetry run pytest --with-slow --generate-shapshots tests/test_base/test_metamodel.py`

sneakers-the-rat:
> can someone explain why this fails:

only guess is typo in:

> `--generate-shapshots`

but other than that deeply mysterious

Issue #1784, opened 2023-12-12
Allow core and modular installation of linkml?


rly:
**Is your feature request related to a problem? Please describe.**
I would like to include `linkml` (or `linkml-runtime`) as a dependency in the library `pynwb`. Installing `pynwb` in a fresh python environment results in relatively few packages (around 15 very common packages) being downloaded and installed. We want to use basic linkml / linkml-runtime functionality around reading/writing schema and generating pydantic classes. If we want to add `linkml` as a dependency of `pynwb`, then an additional ~50 packages would need to be resolved, downloaded, and installed, many of which are not relevant to our basic needs, e.g., arrow, click, sqlalchemy, webcolors. This would add a minor burden in the install process to our users, but would more critically increase the complexity and number of failure points of installing and using pynwb because it would then have so many dependencies that could break or become incompatible with each other or with existing libraries in a user's environment (e.g., they may require `urllib3 < 2` for some reason).

Is there a way for us to include basic linkml pydantic class generation functionality without including the rest?

**Describe the solution you'd like**
One method used by some packages is to allow users to install specific functionality using optional dependency flags. Perhaps linkml could adopt that strategy, e.g., `pip install linkml[pydantic]` to install just the core linkml features + pydantic generating code.

A downside is that this might increase the complexity of managing and testing linkml. 

**How important is this feature?** Select from the options below:
• Low - it's an enhancement but not crucial for work

**When will use cases depending on this become relevant?** Select from the options below:
• Long-term - 6 months - 1 year

sneakers-the-rat:
big +1 to this - it seems like each of the generators should be their own package that make use of the [plugin system](https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/) since each has their own very different dependencies. I also planned ahead to make a specific package that had the linkml dependency for this reason (so the other packages wouldn't need it).

Here's the current output of `poetry show --tree --only main`:
<details>
<summary>Expand/Collapse dependency tree</summary>

```
antlr4-python3-runtime 4.9.3 ANTLR 4.9.3 runtime for Python 3.7
click 8.1.7 Composable command line interface toolkit
└── colorama *
graphviz 0.20.1 Simple Python interface for Graphviz
hbreader 0.9.1 Honey Badger reader - a generic file/url/string open and read tool
isodate 0.6.1 An ISO 8601 date/time/duration parser and formatter
└── six *
jinja2 3.1.2 A very fast and expressive template engine.
└── markupsafe >=2.0
[...]

PR #2196, opened 2024-07-04
Make single-use generator dependencies optional

sneakers-the-rat:
Related to: https://github.com/linkml/linkml/issues/1784#issuecomment-2208025581

This is a draft PR to spread the modular install effort to make dependencies only used in specific generators optional. i've been using `poetry show --tree` to trim impactful deps, and it turns out some of them were just plain unused.

this is opened in a purposefully unfinished state as a starting point and pattern for how to do it, but proposals and edits are welcome to add/change/etc. whatever ones we want to make optional or don't. 

I have basically set it up so that the deps that are marked for deprecation are put in a form where when it comes time to deprecate them we swap the `optional = false` to `true`, remove place where the import error is raised, and call it a day. 

I set the removal version at 1.9.0 but that's also just a number to put down and can be whatever.

### DIFF
--- a/docs/intro/install.md
+++ b/docs/intro/install.md
@@ -42,6 +42,39 @@ linkml-ws --help
 gen-project --help
 ```
 
+## Modular Installation
+
+Starting in [`v1.9.0`](../code/deprecation.rst), to reduce the number of dependencies needed for the average installation,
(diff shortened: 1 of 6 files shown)


sierra-moxon:
closing draft PR, work is captured and ready for pick up here: https://github.com/linkml/linkml/issues/2703

Issue #1845, opened 2024-01-26
propagate pattern constraints from referenced classes to foreign key fields in jsonschemagen (partial referential integrity)


cmungall:
jsonschema doesn't support referential integrity: https://stackoverflow.com/questions/56429622/make-sure-object-field-has-value-existing-in-another-field-in-json-schema

However, we can implement *partial* RI, where we propagate pattern constraints on the identifier field for a referenced class to the reference field

From #1841
default validation (jsonschemagen) can check patterns on foreign keys by a primary key lookup even in the absence of ref integrity. I think this is a v quick enhancement

sierra-moxon:
I believe this is done.

PR #2204, opened 2024-07-11
Test the validation patterns of a slot with a non-in-lined class range and a pattern. Are both checked?

turbomam:
discussed with @pkalita-lbl 

this probably wont need review for now

may change to a personal gist


pyproject.toml
```diff
@@ -125,6 +125,7 @@ watchdog = ">=0.9.0"
 pyshacl = { version = "^0.25.0", optional = true }
 black = { version=">=24.0.0", optional = true }
 typing-extensions = { version=">=4.4.0", python="<3.9" }
+owlready2 = "^0.46"
```
cmungall, line 128: remove

tests/test_issues/test_linkml_issue_2202.py
```diff
@@ -0,0 +1,191 @@
...
+    # Serialize the graph to RDF/XML
+    rdf_xml_data = g.serialize(format="xml")
+
+    # write as rdf/xml
+    with tempfile.NamedTemporaryFile(delete=False, suffix=".owl", mode="w") as tmp:
+        tmp.write(rdf_xml_data)
+
+    # parse with owlready2
+    onto = get_ontology(tmp.name).load()
+    print(list(onto.classes()))
```
cmungall, line 180: can be done directly with rdflib which is already a deoendency

cmungall (changes requested):
I recall seeing some discussion about the use of owlready2 on slack, it would be good to summarize here before it gets lost in the slack memory hole (I'm on a plane and can't seem to access slack right now).

You seem to be using owlready2 to list the classes which is easy with plain rdflib, so perhaps the idea is to also check that

cmungall:
@turbomam can you give some more background on what this is intended to do?

Is this a test for this proposed functionality?

- #1845

If so we'd expect any test to fail since it's not yet implemented. When we do implement it, I think the test should be part of the compliance tests.

PR #2208, opened 2024-07-15
[pydanticgen] Split generation :)

sneakers-the-rat:
Depends on: https://github.com/linkml/linkml/pull/2019

Pulling some more of my changes from https://github.com/p2p-ld/nwb-linkml/ upstream :)

## Problems
- pydantic 2 is super fast when declaring classes compared to pydantic 1. it's still pretty slow when you have a module with a zillion classes in it from a gigantic schema. 
- many people author schemas in multiple files for organization and modularity, but the generated pydantic classes all get dumped in a single file, wrecking that organization
- complex multipart schemas with multiple versions for each dependent schema are currently very difficult to use with pydantic, as there is no formal version specification in imports. It would be great to be able to create many different versions of a given schema and import between them, rather than making an independent module class for each. 

All of these problems are pretty strong barriers to practical implementation by existing formats like NWB, both in terms of ergonomics and expressiveness.

## Solution

this PR allows us to generate pydantic models split by schema!

### Generator

The generator now takes a trio of params:

- `split`: boolean flag indicating split mode
- `split_pattern`: A jinja template string that takes the `SchemaDefinition` and..
- `split_context`: additional variables passed to the generation context.

The PydanticGenerator still only generates a single module, `split` mode just replaces the generated classes imported from other schemas with import statements:

eg. with a `split_pattern` of `"..{{ schema.version | replace('.', '_') }}.{{ schema.name }}"`

we get import statements like

```python
from ..v0_1_2.schema_1 import (
    S1,
    S1Any
)
from ..v2_3_4.schema_2 import (
    S2,
    S2Any
)
```

### `generate_split`

A wrapper function `generate_split` handles generating a whole collection of python modules. So for the set of test schemas in `tests/test_generators/input/split` and the above `split_pattern`, we get a directory structure like this:

```
.
└── test_module
    ├── __init__.py
    ├── v0_1_2
    │   ├── __init__.py
    │   └── schema_1.py
    ├── v1_2_3
    │   ├── __init__.py
    │   └── main.py
    └── v2_3_4
        ├── __init__.py
        └── schema_2.py

5 directories, 7 files
```

the `generate_split` function and its helpers makes sure that we have `__init__.py` files, and also handles converting the relative imports from the generator into paths (so your `split_pattern` can imply any kind of directory structure you want to, relative to the specified output path).

questions? lmk :). Tried to document and test what seemed like it needed to be tested but lmk if i need to add any more that are missing.

## PS

Sorry this PR is non-modular and builds off https://github.com/linkml/linkml/pull/2019 , consider this a gentle nudge to merge that one (this PR will be easier to read then), and also a demonstration of how that structure + the system of passing build results back up is powerful and clear <3

### DIFF
--- a/linkml/generators/pydanticgen/array.py
+++ b/linkml/generators/pydanticgen/array.py
@@ -12,7 +12,7 @@ from linkml.utils.deprecation import deprecation_warning
 if int(PYDANTIC_VERSION[0]) >= 2:
     from pydantic_core import core_schema
 else:
-    #  Support for having pydantic 1 installed in the same environment will be dropped in 1.9.0
+    # Support for having pydantic 1 installed in the same environment will be dropped in 1.9.0
     deprecation_warning("pydantic-v1")
 
 if TYPE_CHECKING:
--- a/linkml/generators/pydanticgen/array.py
+++ b/linkml/generators/pydanticgen/array.py
@@ -12,7 +12,7 @@ from linkml.utils.deprecation import deprecation_warning
 if int(PYDANTIC_VERSION[0]) >= 2:
     from pydantic_core import core_schema
 else:
-    # Support for having pydantic 1 installed in the same environment will be dropped in 1.9.0
+    #  Support for having pydantic 1 installed in the same environment will be dropped in 1.9.0
     deprecation_warning("pydantic-v1")
 
 if TYPE_CHECKING:
--- a/linkml/generators/pydanticgen/pydanticgen.py
+++ b/linkml/generators/pydanticgen/pydanticgen.py
@@ -22,7 +22,6 @@ from linkml_runtime.linkml_model.meta import (
 from linkml_runtime.utils.compile_python import compile_python
 from linkml_runtime.utils.formatutils import camelcase, remove_empty_items, underscore
 from linkml_runtime.utils.schemaview import SchemaView
-from pydantic import BaseModel, ConfigDict
 from pydantic.version import VERSION as PYDANTIC_VERSION
 
 from linkml._version import __version__
@@ -30,7 +29,7 @@ from linkml.generators.common.type_designators import get_accepted_type_designat
 from linkml.generators.oocodegen import OOCodeGenerator
 from linkml.generators.pydanticgen import includes
 from linkml.generators.pydanticgen.array import ArrayRangeGenerator, ArrayRepresentation
-from linkml.generators.pydanticgen.build import ClassResult, SlotResult
+from linkml.generators.pydanticgen.build import ClassResult, SlotResult, SplitResult
 from linkml.generators.pydanticgen.template import (
     Import,
     Imports,
@@ -944,18 +943,6 @@ def _subclasses(cls: Type):
 
 
 _TEMPLATE_NAMES = sorted(list(set([c.template for c in _subclasses(TemplateModel)])))
-
-
-class SplitResult(BaseModel):
-    """Build result when generating with :func:`.generate_split`"""
-
-    model_config = ConfigDict(arbitrary_types_allowed=True)
-
-    main: bool = False
-    source_schema: SchemaDefinition
-    path: Path
-    serialized_module: str
-    module_import: Optional[str] = None
 
 
 def generate_split(
--- a/linkml/generators/pydanticgen/pydanticgen.py
+++ b/linkml/generators/pydanticgen/pydanticgen.py
@@ -6,14 +6,15 @@ from collections import defaultdict
 from dataclasses import dataclass, field
 from enum import Enum
 from pathlib import Path
+import re
 from types import ModuleType
-from typing import Dict, List, Literal, Optional, Set, Type, TypeVar, Union, overload
+from typing import ClassVar, Dict, List, Literal, Optional, Set, Type, TypeVar, Union, overload
 
 import click
 from jinja2 import ChoiceLoader, Environment, FileSystemLoader, Template
 from linkml_runtime.linkml_model.meta import (
-    ElementName,
     ClassDefinition,
+    ElementName,
     SchemaDefinition,
     SlotDefinition,
     TypeDefinition,
@@ -250,7 +251,7 @@ class PydanticGenerator(OOCodeGenerator):
      
     Further modification is possible by using jinja filters.
     
-    After templating, the string is passed through a ``snake_case`` function
+    After templating, the string is passed through a :attr:`SNAKE_CASE` pattern
     to replace whitespace and other characters that can't be used in module names.
      
     Examples:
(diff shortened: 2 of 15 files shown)


sneakers-the-rat:
Added:
- Ability to generate in split mode but also include all imports from the other modules regardless of whether or not they are used. did this to be able to make the NWB namespace files, which are used throughout the standard. nwb_linkml doesn't use them really since the generator is smart enough to resolve the relative imports and importing huge amounts of models gets to be a perf chokepoint for complex schemas (the whole motivation for this PR). An example of when linkml generator flexibility is to preserve schema fidelity above and beyond functional equivalence <3
- Fixed missing check for mixins
- Ability to check for membership in `Imports` - mostly to make the tests more robust so i don't have to use regex which doesn't even test whether the imports are correctly formed and is vulnerable to ordering mismatches - but also because i imagine modifying imports is going to be one of the more common downstream modifications to generated models since its sort of the basis for any other customizations that would happen in the rest of the module...
- Ability to index imports by module name, nbd, same reason
- Did above changes to tests to make them more robust

linkml/generators/pydanticgen/array.py
```diff
@@ -24,7 +24,7 @@
 else:
     from typing import Annotated
 
-from linkml.generators.pydanticgen.build import SlotResult
+from linkml.generators.pydanticgen.build import RangeResult
```
sneakers-the-rat, line 27: this is the thing that was looking weird to me - i had made this change in 2019, but i guess the diff is just stale here because main also is `RangeResult`: https://github.com/linkml/linkml/blob/669fbff40a2e7683797d0c4e510bc21dee7815aa/linkml/generators/pydanticgen/array.py#L27

sneakers-the-rat:
OK @pkalita-lbl this is ready for review now :)

sneakers-the-rat:
actually now that i am going to use it, i am wanting the split generation method to be a classmethod because the lifecycle methods are based on subclassing and overriding, which i can't do when i can't change the Generator inside it.

linkml/generators/pydanticgen/pydanticgen.py
```diff
@@ -229,13 +247,71 @@ class PydanticGenerator(OOCodeGenerator):
...
+    pattern is used to generate the module portion of the import statement.
+    
+    These patterns should generally yield a relative module import, 
+    since functions like :func:`.generate_split` will generate and write files
+    relative to some base file, though this is not a requirement since custom
+    split generation logic is also allowed.
+    
+    The pattern is a jinja template string that is given the ``SchemaDefinition``
+    of the imported schema in the environment. Additional variables can be passed
+    into the jinja environment with the ``split_environment`` argument.
```
pkalita-lbl, line 271: Should `split_environment` be `split_context` instead?

linkml/generators/pydanticgen/pydanticgen.py
```diff
@@ -775,6 +1082,38 @@ def _subclasses(cls: Type):
...
+
+
+def _ensure_inits(paths: List[Path]):
+    """For a set of paths, find the common root and it and all the subdirectories have an __init__.py"""
+    # if there is only one file, there is no relative importing to be done
+    if len(paths) <= 1:
+        return
+    common_path = Path(os.path.commonpath(paths))
+
+    if not (ipath := (common_path / "__init__.py")).exists():
```
pkalita-lbl, line 1103: Wow! A walrus operator spotted in the wild!

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "linkml"
-version = "0.0.0"
+version = "1.7.5.post241.dev0+8fd9c529"
```
pkalita-lbl, line 3: We should leave this as `"0.0.0"` as indication that dynamic versioning is being used.

pyproject.toml
```diff
@@ -44,7 +44,7 @@ packages = [
 ]
 
 [tool.poetry-dynamic-versioning]
-enable = true
+enable = false
```
pkalita-lbl, line 47: I know your feelings about the dynamic versioning plugin, but uh... Probably just an oversight I'm sure 😁

pkalita-lbl (commented):
Overall this looks great. Fantastic docstrings that really helped me understand what's going on!

Just a couple of minor things to take a second look at if you don't mind.

linkml/generators/pydanticgen/pydanticgen.py
sneakers-the-rat: I WAS SO EXCITED

sneakers-the-rat: The walrus even respects the parens scope which blew my mind

linkml/generators/pydanticgen/pydanticgen.py
sneakers-the-rat: Oh yep!

pyproject.toml
sneakers-the-rat: Right my bad. I havent figured out what triggers it but sometimes that field automatically replaces itself

pyproject.toml
sneakers-the-rat: I think it is good! Just a lil rascally sometimes

linkml/generators/pydanticgen/template.py
```diff
@@ -499,8 +535,58 @@ def __iter__(self) -> Generator[Import, None, None]:
...
+        if isinstance(item, int):
+            return self.imports[item]
+        elif isinstance(item, str):
+            # the name of the module
+            an_import = [i for i in self.imports if i.module == item]
+            if len(an_import) == 0:
+                raise KeyError(f"No import with module {item} was found.\nWe have: {self.imports}")
+            return an_import[0]
+        else:
+            raise TypeError("Can only index with an int or a string as the name of the module," "\nGot: {type(item)}")
```
sneakers-the-rat, line 548: Oop missing an f string

sneakers-the-rat: done in https://github.com/linkml/linkml/pull/2208/commits/7fed49b3764d3f8c1ee84422444e675d0baec99c

linkml/generators/pydanticgen/pydanticgen.py
sneakers-the-rat: done in https://github.com/linkml/linkml/pull/2208/commits/7fed49b3764d3f8c1ee84422444e675d0baec99c

pyproject.toml
sneakers-the-rat: done in https://github.com/linkml/linkml/pull/2208/commits/7fed49b3764d3f8c1ee84422444e675d0baec99c

pyproject.toml
sneakers-the-rat: https://github.com/linkml/linkml/pull/2208/commits/7fed49b3764d3f8c1ee84422444e675d0baec99c

sneakers-the-rat:
addressed comments :)

pkalita-lbl (approved):

Issue #2217, opened 2024-07-23
Deprecate jsonasobj2?


sneakers-the-rat:
**Is your feature request related to a problem? Please describe.**

hello it is me the perf gremlin again.

when generating a pydantic module, effectively all the meaningful perf problems come from using jsonasobj2:

- within `induced_slot` 
  - (6.17% of total time) the `__bool__` method of jsonasobj
  - (12.85%) `__setattr__`
- within this dict comp: https://github.com/linkml/linkml/blob/bd49d48575d20e80875c2e1ef735d910f3a0d250/linkml/generators/pydanticgen/pydanticgen.py#L314
  - (21%) `as_dict`  

The rest is from parsing yaml, an inefficiency in schemaview's `__hash__` method (which i'm about to PR), and from black formatting. 

So perf is one thing, but also the `jsonasobj` root class in the generated python models give them a ~decent amount of surprising behavior (eg. [as i have noted before](https://github.com/linkml/linkml/issues/1665), which i now see [chris also noted previously](https://github.com/linkml/linkml/issues/971))

I am not quite sure what it buys us? the main thing seems to be using `obj.attr` instead of `obj['attr']` syntax? but that's also something we get with dataclasses, which are the main place they seem to be used.  

**Describe the solution you'd like**

I would like to just use dataclasses and dicts :).

I guess to make that happen (and to see if there is some functionality we want to preserve from `jsonasobj2`) i suppose i would need to have a better picture of why it is used/what features of it we like, but once i get that i'd be happy to PR.

**How important is this feature?** Select from the options below:
• Medium - strong perf impact which makes runtime use costly, hard to monkeypatch around.

**When will use cases depending on this become relevant?** Select from the options below:
• Short-term - 2-4 weeks

sneakers-the-rat:
Just testing out what it would take, and i am getting almost all of the way there by doing this

https://github.com/sneakers-the-rat/linkml-runtime/commit/bc687a4bc8f2a6ee635fb6004d7882e2552c4953

and a handful of changes in `linkml` that are similar, usually calling `dataclasses.asdict` before some thing that expects it to be json-able.

anyway i don't think this would be hard and i would be willing to do it if we think this is a good idea

cmungall:
I agree it would be good to remove dependence on jsonasobj2. You can see in this issue tracker a few confusing issues that have been caused by it over the years.

[...]

PR #2218, opened 2024-07-23
[perf] minor pydanticgen perf boosts - remove unnecessary object creation

sneakers-the-rat:
i was dismayed to see that the pydantic generator had bad runtime characteristics in two places 

Screenshot 2024-07-22 at 7 52 24 PM

- one dictionary comprehension that was expensive because of [casting to a dict](https://github.com/linkml/linkml/issues/2217) 
- the `merge` method that combined injected classes, but most of the time these are identical (from the default value).

edit: so i fixed them

### DIFF
--- a/linkml/generators/pydanticgen/build.py
+++ b/linkml/generators/pydanticgen/build.py
@@ -51,9 +51,8 @@ class BuildResult(BaseModel):
                 self_copy.imports = other.imports
         if other.injected_classes:
             if self_copy.injected_classes is not None:
-                if self_copy.injected_classes == other.injected_classes:
-                    pass
-                else:
+                # only combine and dedupe when injected_classes don't match
+                if self_copy.injected_classes != other.injected_classes:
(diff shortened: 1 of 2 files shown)


sneakers-the-rat:
just a heads up it looks like the codecov check is hung and has been running for >1d, hopefully that doesn't cost y'all anything but it might be tying up one of the test runners

sneakers-the-rat:
you'd think either github or codecov would have a timeout somewhere, and yet...

Screenshot 2024-07-25 at 10 40 55 PM

RIP to all the virtual machines out there doing pointless stuff out there indefinitely

sneakers-the-rat:
Screenshot 2024-09-25 at 2 08 10 PM

idk if y'all have some kind of compute time limit for this org account but this is sort of incredible

cmungall (approved):

cmungall:
it had a good innings but it was time for it to end

(and yes wow.. no, doesn't cost us anything but not a good use of someone's resources)

PR #2220, opened 2024-07-23
Add note admonition for abstract slots

sujaypatil96:
Similar to how we have a _Note_ banner that gets rendered on class documentation pages when a class is asserted to be `abstract`, we need to replicate the behaviour for slots as well.

Example: https://microbiomedata.github.io/nmdc-schema/NamedThing/

### DIFF
--- a/linkml/generators/docgen/slot.md.jinja2
+++ b/linkml/generators/docgen/slot.md.jinja2
@@ -34,6 +34,10 @@ _{{ element_description_line }}_
 {% endfor %}
 {% endif %}
 
+{% if element.abstract %}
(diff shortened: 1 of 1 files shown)


linkml/generators/docgen/slot.md.jinja2
```diff
@@ -34,6 +34,10 @@ _{{ element_description_line }}_
 {% endfor %}
 {% endif %}
 
+{% if element.abstract %}
+* __NOTE__: this is an abstract slot and should not be instantiated directly
```
turbomam, line 38: I would say something more like "should not be populated directly"

cmungall (approved):

PR #2225, opened 2024-07-24
Add cardinality constraints to Pydantic generation for list values 

daniel-ji:
Currently, when `minimum_cardinality`, `maximum_cardinality`, and `exact_cardinality` are provided as slots in a SlotDefinition, the Pydantic generator does not implement these constraints with Pydantic. This PR implements this feature and adds corresponding tests. 

This PR also fixes the behavior of multivalued, optional attributes. Instead of generating an empty list / dict, the default value is `None` instead. This fix goes hand-in-hand with ensuring logical / correct behavior of the cardinality constraints. (For example, if the field is not required, but has a minimum cardinality of 5, providing no value should not cause validation to fail, but if we provide a default value of an empty list, then validation will fail).

An example input:

Note not only the `min_length` and `max_length` requirements now in the `Field()` constructor, but also how the default value is `None` and there is no `default_factory`. For `minimum_cardinality_array`, `...` indicates that the field is required, as desired. 

```yaml
id: https://example.org/arrays
name: arrays-cardinality-example
title: Array Cardinality Example
description: |-
    Example LinkML schema to test array cardinality
license: MIT
default_prefix: example
imports:
    - linkml:types
classes:
    CardinalityArray:
        description: A class with arrays of different cardinality constraints
        attributes:
            minimum_cardinality_array:
                range: float
                multivalued: true
                required: true
                minimum_cardinality: 1
            maximum_cardinality_array:
                range: float
                multivalued: true
                maximum_cardinality: 10
            exact_cardinality_array:
                range: float
                multivalued: true
                exact_cardinality: 5
            min_max_cardinality_array:
                range: float
                multivalued: true
                minimum_cardinality: 0
                maximum_cardinality: 8
```

and the generated Pydantic:

```python
from datetime import (
    datetime,
    date
)
from decimal import Decimal 
from enum import Enum 
import re
import sys
from typing import (
    Any,
    ClassVar,
    List,
    Literal,
    Dict,
    Optional,
    Union
)
from pydantic import (
    BaseModel,
    ConfigDict,
    Field,
    RootModel,
    field_validator
)
metamodel_version = "None"
version = "None"


    model_config = ConfigDict(
        validate_assignment = True,
        validate_default = True,
        extra = "forbid",
        use_enum_values = True,
        strict = False,
    )
    pass




class LinkMLMeta(RootModel):
    root: Dict[str, Any] = {}

        return self.root[key]

        self.root[key] = value

    def __contains__(self, key:str) -> bool:
        return key in self.root


linkml_meta = LinkMLMeta({'default_prefix': 'example',
     'description': 'Example LinkML schema to test array cardinality',
     'id': 'https://example.org/arrays',
     'imports': ['linkml:types'],
     'license': 'MIT',
     'name': 'arrays-cardinality-example',
     'title': 'Array Cardinality Example'} )


class CardinalityArray(ConfiguredBaseModel):
    """
    A class with arrays of different cardinality constraints
    """
    linkml_meta: ClassVar[LinkMLMeta] = LinkMLMeta({'from_schema': 'https://example.org/arrays'})

    minimum_cardinality_array: List[float] = Field(..., min_length=1, json_schema_extra = { "linkml_meta": {'alias': 'minimum_cardinality_array', 'domain_of': ['CardinalityArray']} })
    maximum_cardinality_array: Optional[List[float]] = Field(default=None, max_length=10, json_schema_extra = { "linkml_meta": {'alias': 'maximum_cardinality_array', 'domain_of': ['CardinalityArray']} })
    exact_cardinality_array: Optional[List[float]] = Field(default=None, min_length=5, max_length=5, json_schema_extra = { "linkml_meta": {'alias': 'exact_cardinality_array', 'domain_of': ['CardinalityArray']} })
    min_max_cardinality_array: Optional[List[float]] = Field(default=None, min_length=0, max_length=8, json_schema_extra = { "linkml_meta": {'alias': 'min_max_cardinality_array', 'domain_of': ['CardinalityArray']} })


# Model rebuild
# see https://pydantic-docs.helpmanual.io/usage/models/#rebuilding-a-model
CardinalityArray.model_rebuild()
```


linkml/generators/pydanticgen/pydanticgen.py
daniel-ji, line 1: Just reformatting changes here (from running pre-commit)

linkml/generators/pydanticgen/templates/attribute.py.jinja
```diff
@@ -7,6 +7,12 @@
     {%- if minimum_value != None %}, ge={{minimum_value}}{% endif -%}
     {%- if maximum_value != None %}, le={{maximum_value}}{% endif -%}
 {%- endif -%}
+{%- if multivalued and exact_cardinality != None %}
+, min_length={{exact_cardinality}}, max_length={{exact_cardinality}}
+{%- elif multivalued %}
```
daniel-ji, line 12: If properly indented, the indentation would cause extra space in the generated Python code

sneakers-the-rat: See those `-` signs in the `{%-` tags? Thats how jinja manages whitespace while keeping the template readable :) 
https://jinja.palletsprojects.com/en/3.1.x/templates/#whitespace-control

linkml/generators/pydanticgen/pydanticgen.py
sneakers-the-rat: Ha thats probably my bad, I find pre commit annoying so I never have it installed if the CI doesnt care about it

tests/test_generators/test_pydanticgen.py
```diff
@@ -578,6 +578,59 @@ def test_pydantic_arrays():
     assert temperature_dataset.temperatures_in_K == temperatures
 
 
+def test_pydantic_array_cardinality():
```
sneakers-the-rat, line 581: Maybe just `test_pydantic_cardinality` because `array` is something else in linkml (though I know what you mean) :)

linkml/generators/pydanticgen/templates/attribute.py.jinja
daniel-ji: Fixed, thank you for that!

tests/test_generators/test_pydanticgen.py
```diff
@@ -578,6 +578,59 @@ def test_pydantic_arrays():
...
+    gen = PydanticGenerator(schema=unit_test_schema)
+    code = gen.serialize()
+
+    mod = compile_python(code)
+    assert mod.CardinalityArray.model_fields["minimum_cardinality_array"].annotation == List[float]
+    assert mod.CardinalityArray.model_fields["minimum_cardinality_array"].metadata[0].min_length == 1
+    assert mod.CardinalityArray.model_fields["maximum_cardinality_array"].metadata[0].max_length == 10
+    assert mod.CardinalityArray.model_fields["exact_cardinality_array"].metadata[0].min_length == 5
+    assert mod.CardinalityArray.model_fields["exact_cardinality_array"].metadata[1].max_length == 5
+
```
sneakers-the-rat, line 632: These look good, could we also get some correct/incorrect instantiation tests? eg. A valid list correctly instantiates, and an incorrect list or a scalar float raises a `ValidationError` (see `with pytest.raises` examples elsewhere in this module) if it gets too verbose/repetitive you could also use `pytest.mark.parameterize` but this is also totally fine since the test is p focused!

sneakers-the-rat:
Lgtm so far, ill run it when i get to my desk and do a more formal review. I think in general for PRs we like to see something in the description for recordkeeping purposes (even tho we talked about this separately) - maybe copy that example schema and the model it generates?

I also think @cmungall probably would want us to update the compliance tests here too, but i can handle that since theyre a little complicated :) (although now that I look I dont see compliance tests for minimum and maximum cardinality, so maybe not)

daniel-ji:
Hi @sneakers-the-rat 

If the field is specified as multivalued and has a minimum_cardinality / exact_cardinality but is not required, should we fail when they provide no value?

sneakers-the-rat:
ok that's a pretty funny bug. I think that we want to do what I mentioned in https://github.com/linkml/linkml/pull/2201 and make the default `None` for optional, multivalued fields. 

This works:

```python
from pydantic import BaseModel, Field
from typing import Optional, List

class MyModel(BaseModel):
    a_field: Optional[List[int]] = Field(None, min_length=3)
```

```python
>>> a = MyModel()
>>> a = MyModel(a_field=[])
pydantic_core._pydantic_core.ValidationError: 1 validation error for MyModel
a_field
  List should have at least 3 items after validation, not 0 [type=too_short, input_value=[], input_type=list]
    For further information visit https://errors.pydantic.dev/2.8/v/too_short
>>> a = MyModel(a_field=[1,2,3])
```

You're already familiar with that part of the pydantic generator from 2201 :) so I think we want it to always be the case that if a slot/attribute is a) not required, and b) doesn't have an `ifabsent` value, then the default should be None. That's semantically correct too imo, since the current behavior makes it ambiguous whether a value was provided but was just empty, or whether a value is absent. `None` representing absence of value is idiomatic for both python and pydantic.

linkml/generators/pydanticgen/template.py
```diff
@@ -212,6 +212,10 @@ class PydanticAttribute(TemplateModel):
     equals_number: Optional[Union[int, float]] = None
     minimum_value: Optional[Union[int, float]] = None
     maximum_value: Optional[Union[int, float]] = None
+    exact_cardinality: Optional[int] = None
+    minimum_cardinality: Optional[int] = None
+    maximum_cardinality: Optional[int] = None
+    multivalued: Optional[bool] = None
```
sneakers-the-rat, line 218: super subtle thing - since the fields in template models are considered to be "the things that the template is capable of representing" it's good to have multivalued in here (and i see why you did it in the template too), could we remove it from `meta_exclude` above which is "the things that are represented in other ways than explicitly as a field in the template model and should be excluded when embedding schema metadata in the generated pydantic field"

tests/test_generators/test_pydanticgen.py
```diff
@@ -578,6 +578,59 @@ def test_pydantic_arrays():
...
+imports:
+    - linkml:types
+
+classes:
+    CardinalityArray:
+        description: A class with arrays of different cardinality constraints
+        attributes:
+            minimum_cardinality_array:
+                implements:
+                    - linkml:elements
```
sneakers-the-rat, line 602: i don't think you need these in here but it also doesn't hurt :)

tests/test_generators/test_pydanticgen.py
```diff
@@ -578,6 +578,59 @@ def test_pydantic_arrays():
...
+                range: float
+                multivalued: true
+                required: true
```
sneakers-the-rat, line 605: given that there was some bug with required vs not here, I think we should just go ahead and make this a `parameterize`d test and make all the conditions

something like

```python

@pytest.mark.parametrize(
    'cardinality,multivalued,required,value,valid',
    (
        ({'minimum_cardinality': 3}, True, True, [1,2,3], True),
        ({'minimum_cardinality': 3}, True, True, [1,2], False),
        # ... the other conditions with max, exact, etc.
        ({'minimum_cardinality': 3}, True, False, None, True)
    )
)
def test_pydantic_cardinality(cardinality, multivalued, required, value, valid):
    
    schema = SchemaDefinition(
        # some template schema or whatever with this in there...
        # I'm sure there's some fixture for this somewhere
        classes = {
            'CardinalityArray': ClassDefinition(
                attributes={'field':
                    SlotDefinition(
                        name='field',
                        multivalued=multivalued,
                        required=required,
                        **cardinality
                    )
                }
            )
        }
    )
    gen = PydanticGenerator(schema=schema)
    code = gen.serialize()
    mod = compile_python(code)
    cls = mod.CardinalityArray
    field = cls.model_fields["array"]

    if multivalued:
        assert field.annotation == List[float]

    # whatever else...

    # then
    if valid:
        _ = cls(array=value)
    else:
        with pytest.raises(ValidationError):
            _ = cls(array=value)

```

sneakers-the-rat (commented):
Thanks for doing this! 

Already mentioned in other comments the need to change default to `None` to accommodate optional slots/attrs, so w/ that and these minor comments should be good to go.

I would feel a tiny bit better if we had some more tests for validation (i like it when we can knock things out once and forever and never worry about them again <3), and that one modification to `meta_exclude` would be nice, but otherwise thanks for this! one more step towards pydanticgen completeness!!!!!

I am not a core dev here so i don't usually give final approval/merge PRs, but my 2c since we have been talking about this :)

linkml/generators/pydanticgen/template.py
daniel-ji: Do we want to remove `range` as well then?

tests/test_generators/test_pydanticgen.py
daniel-ji: I'll go ahead and remove it then, for the sake of clarity / concision

daniel-ji:
Should we be testing weird cases where it's a bad minimum_cardinality / maximum_cardinality / exact_cardinality input? Like if they specify maximum_cardinality < minimum_cardinality or maximum_cardinality > exact_cardinality. Thanks!

sneakers-the-rat:
nah, imo schema validation is not the responsibility of the generator - if someone specifies an invalid schema then it's sorta on them when pydantic yells at them & since that error would be invalid for all generators the test for it shouldn't be in any one generator. 

adding test with invalid data is in general good obvi, but in this case it would sorta be a trivial result that is orthogonal to the goal of the unit test. pydantic tests the constraints separately rather than raising an exception when declaring a logically inconsistent field, which imo is the right EAFP behavior for python.

```python
from pydantic import BaseModel, Field

# no errors raised when defining....
class MyModel(BaseModel):
    my_field: list[int] = Field(min_items=4, max_items=3)
    
# but on instantiation....
>>> a = MyClass(my_field=[1,2,3])
my_field
  List should have at least 4 items after validation, not 3 [type=too_short, input_value=[1, 2, 3], input_type=list]
    For further information visit https://errors.pydantic.dev/2.6/v/too_short
    
>>> a = MyClass(my_field=[1,2,3,4])
my_field
  List should have at most 3 items after validation, not 4 [type=too_long, input_value=[1, 2, 3, 4], input_type=list]
    For further information visit https://errors.pydantic.dev/2.6/v/too_long
```

almost kafka-esque coplike behavior - what do you want from me pydantic!?!?

tl;dr don't worry about it

cmungall:
> nah, imo schema validation is not the responsibility of the generator - if someone specifies an invalid schema then it's sorta on them when pydantic yells at them & since that error would be invalid for all generators the test for it shouldn't be in any one generator.

Agreed, and in fact logical validation of the schema can be performed independent of pydantic:

- #2234

cmungall:
Looks like a few conflicts have crept in here...

daniel-ji:
@cmungall Resolved!

cmungall:
thx! Will you be able to take care of the comments from @sneakers-the-rat?

daniel-ji:
I think they should all be resolved now, just waiting for Jonny's review. I was thinking that I'd leave them open until he came by and looked through the code again, unless it's convention to resolve them now. Thanks!

sneakers-the-rat:
my bad for making a mess and then disappearing, will review now

linkml/generators/pydanticgen/template.py
sneakers-the-rat: yes, but no need to do it in this PR. i am planning on doing some updates here in how this is handled soon. for now it's just a redundant value which isn't the worst thing in the world.

tests/test_generators/test_pydanticgen.py
```diff
@@ -661,6 +661,169 @@ def test_pydantic_arrays():
     assert temperature_dataset.temperatures_in_K == temperatures
 
 
+@pytest.mark.parametrize(
```
sneakers-the-rat, line 664: thanks for this, super thorough!

tests/test_generators/test_pydanticgen.py
```diff
@@ -661,6 +661,169 @@ def test_pydantic_arrays():
...
+        assert minimium_cardinality == min_length[0]
+    if maximum_cardinality:
+        assert len(max_length) == 1
+        assert maximum_cardinality == max_length[0]
+
+    if valid:
+        cls(cardinality_array=value)
+    else:
+        with pytest.raises(ValidationError):
+            cls(cardinality_array=value)
```
sneakers-the-rat, line 770: nice - combination of logical and functional tests here is plenty :)

tests/test_generators/test_pydanticgen.py
```diff
@@ -661,6 +661,169 @@ def test_pydantic_arrays():
...
+                range: float
+                multivalued: true
+                minimum_cardinality: 0
+                maximum_cardinality: 8
+"""
+
+    gen = PydanticGenerator(schema=unit_test_schema)
+    code = gen.serialize()
+
+    print(code)
```
sneakers-the-rat, line 815: rm print statement :)

daniel-ji:
No worries at all, definitely not a mess!

sneakers-the-rat (approved):
Great! one print statement to remove (but since pytest hides these anyway it's not that big of a deal), thanks for writing that exhaustive set of conditions!!!! looks good to me. thanks also for bearing with me as we upstreamed some of your commits before this PR went in (which retained attribution).

daniel-ji:
@sneakers-the-rat Removed the print statement! Thanks so much for the help! And no worries about the other PRs, wasn't a problem at all! @cmungall Should be good now!

cmungall (approved):

Issue #2221, opened 2024-07-23
`linkml-ws` Command Not Found after `pip install linkml`


NeilGraham:
**Describe the bug:**

- After running `pip install linkml`, the command `linkml-ws` is not found.

**To reproduce:**
1. Install LinkML using `pip install linkml`
2. Running the command `linkml-ws` or `linkml-ws --help` results in the command not being found.

**Expected behavior:**

- I am unable to initialize a new LinkML project directory because the command `linkml-ws` is not found.

**About your computer:**
 - OS: Microsoft Windows 11 Home
 - Version 10.0.22631 Build 22631

**Additional context:**

- The command `gen-project` is found.
- Also tested on Ubuntu 24.04 LTS (WSL2) with same issue.

sierra-moxon:
Thanks for the issue @NeilGraham - can I ask how you tracked down the `linkml-ws` command in our documentation?  I think we are trying to encourage folks to use our `linkml-project-cookiecutter` for new-project setups: https://github.com/linkml/linkml-project-cookiecutter but we still may have documentation pointing to the older setup.

NeilGraham:
@sierra-moxon Thank you for clarifying.

I was following the Quick Installation documentation here: https://linkml.io/linkml/intro/install.html 
(Github file location: https://github.com/linkml/linkml/blob/main/docs/intro/install.md)

sneakers-the-rat:
got it https://github.com/linkml/linkml/pull/2226

PR #2227, opened 2024-07-25
Unified `linkml` CLI

sneakers-the-rat:
What up it's me again with a single `linkml` CLI entrypoint that wraps all the others.

Important things to note right up top: 
- this does not change any existing behavior, all the other entrypoints still exist (except for one that didn't exist, but that wasn't me)
- we don't even need to deprecate them if we like them
- the added `name` args to the click commands also don't change anything except how they're displayed when used as subcommands.

## Motivation

1) i think it's nice
2) @sierra-moxon describes [over here](https://github.com/orgs/linkml/discussions/2203) how there can be some confusion when first being oriented to the tool. single entrypoint makes all the things you should be able to do discoverable
3) this gives us room to grow with what i was also describing over there as a generic `linkml build` and `linkml watch` commands that might replace `linkml generate project` for making linkml very satisfying to work with
4) over [here](https://github.com/linkml/linkml/issues/2221) there was some difficulty with a missing cli command, and when i went to fix it there wasn't really an obvious cli entrypoint that would be good for like a "run this first" thing. 
5) there was a dead `gen-prolog` command for a generator that no longer exists. having this be a unified entrypoint means that we don't have to maintain the `pyproject.toml` alongside the code. if we remove a generator without removing the cli command, it will just fail and alert us that it is broken.

## Implementation

Super simple. i just took all the existing entrypoints and grouped them up and gave them nice aliases. I gave them what i think is an understandable verb-based naming system. `linkml fix`, `linkml generate python`, `linkml lint`, which is a nice perk of putting them all under the single entrypoint (no longer need to prefix things to make them unique).

So behold

```bash
$ linkml --help
Usage: linkml [OPTIONS] COMMAND [ARGS]...

  LinkML: A flexible linked data modeling language

Options:
  --help  Show this message and exit.

Commands:
  convert   Converts instance data to and from different LinkML Runtime...
  dev       Helper tools for linkml development
  examples  Process a folder of examples and a folder of counter examples.
  fix       Apply schema-fixer commands.
  generate  Generate formats from a LinkML schema
  lint      Run linter on SCHEMA.
  sqldb     Run the LinkML SQL CLI.
  validate  Validate a LinkML Schema
```

```bash
$ linkml generate --help
Usage: linkml generate [OPTIONS] COMMAND [ARGS]...

  Generate formats from a LinkML schema

Options:
  --help  Show this message and exit.

Commands:
  csv             Generate CSV/TSV file from LinkML model
  doc             Generate documentation folder from a LinkML YAML schema
  erdiagram       Generate a mermaid ER diagram from a schema.
  excel           Generate Excel representation of a LinkML model
  golang          Generate Golang types
  golr-views      Generate GOLR representation of a LinkML model
  graphql         Generate graphql representation of a LinkML model
  graphviz        Generate graphviz representations of the LinkML model
  java            Generate java classes to represent a LinkML model
  json-schema     Generate JSON Schema representation of a LinkML model
  jsonld          Generate JSONLD file from LinkML schema.
  jsonld-context  Generate jsonld @context definition from LinkML model
  linkml
  markdown        Generate markdown documentation of a LinkML model
  namespaces      Generate a namespace manager for all of the prefixes...
  owl             Generate an OWL representation of a LinkML model
  plantuml        Generate a UML representation of a LinkML model
  prefix-map      Generate jsonld @context definition from LinkML model
  project         Generate an entire project LinkML schema
  proto           Generate proto representation of LinkML model
  pydantic        Generate pydantic classes to represent a LinkML model
  python          Generate python classes to represent a LinkML model
  rdf             Generate an RDF representation of a LinkML model
  shacl           Generate SHACL turtle from a LinkML model
  shex            Generate a ShEx Schema for a LinkML model
  sparql          Generate SPARQL queries for validation
  sqla            Generate SQL DDL representation
  sqltables       Generate SQL DDL representation
  sssom           Generate SSSOM TSV to represent a LinkML model
  summary         Generate TSV summary files for viewing in Excel and the...
  terminusdb      Generate graphql representation of a LinkML model
  typescript      Generate typescript interfaces and types
  yaml            Validate input and produce fully resolved yaml equivalent
  yuml            Generate a UML representation of a LinkML model
```

It should otherwise work exactly the same as the existing entrypoints.

## Docs

Also while i was doing https://github.com/linkml/linkml/pull/2226 I noticed there wasn't a good place for CLI documentation. That's super important to have, because without going to the literal `pyproject.toml` file i wouldn't be able to figure out what the heck i could do with the program once i install it!

So we get a few pages:

Screenshot 2024-07-24 at 5 17 40 PM

Screenshot 2024-07-24 at 5 18 31 PM

Screenshot 2024-07-24 at 5 19 03 PM

## one more thing(s).

- (also change pydanticgen Imports.schema to .is_schema because it shadows the pydantic `schema` attr and ruins the cli experience)
- (also put the terminusdbgen warning in the __post_init__ method because i was tired of seeing it and it also ruined the cli experience)

## conclusion

anyway i wasn't sure what we would want for tests here so i didn't write any, and also I didn't finish making all the links and putting all the api docs up for all the generators that don't have it, so ya lmk what's good <3


cmungall:
This is really awesome

Will look in more detail later but two general comments

- we don't need to be completely beholden to past commands (e.g. `sqldb` doesn't really conform to clig.dev)
- we want this to be extensible to when other parts of the wider linkml ecosystem become core
    - schema-automator (existing commands: https://linkml.io/schema-automator/cli.html)
    - linkml-store (existing commands: https://linkml.io/linkml-store/tutorials/Command-Line-Tutorial.html)
    - linkml-map

More later...

linkml/cli/main.py
```diff
@@ -0,0 +1,141 @@
...
+generate.add_command(gen_sqltables, name="sqltables")
+generate.add_command(gen_summary, name="summary")
+generate.add_command(gen_project, name="project")
+generate.add_command(gen_excel, name="excel")
+generate.add_command(gen_sssom, name="sssom")
+generate.add_command(gen_linkml, name="linkml")
+
+# Validators
+validate.add_command(linkml_jsonschema_validate, name="jsonschema")
+validate.add_command(linkml_sparql_validate, name="sparql")
```
pkalita-lbl, line 138: These are considered legacy at this point. I'd be in favor of not including them and only providing access to the CLI defined in the `linkml.validator` package.

linkml/validator/cli.py
```diff
@@ -132,6 +132,9 @@ def cli(
     include_range_class_descendants: bool,
     include_context: bool,
 ):
+    """
+    Validate a LinkML Schema
```
pkalita-lbl, line 136:
```suggestion
    Validate data according to a LinkML Schema
```

pkalita-lbl (commented):
The general approach seems good to me!

I would just say we should _not_ include the legacy validators in this. I'll also let @cmungall chime in about anything he'd like to see left out of the unified CLI.

> we want this to be extensible to when other parts of the wider linkml ecosystem become core

There's kind of a feature-not-a-bug aspect to that. If I see the command `linkml-map` it's a nice indication that it's _not_ coming from the core `linkml` package since it is its own top-level command.

Now it would be nice if there was a way for "blessed" third party packages (like `linkml-map` some day in the future) to hook into the official top-level `linkml` CLI. I might spend a little time looking into how that might work. But I don't think that needs to hold up this PR.

linkml/cli/main.py
sneakers-the-rat: Gotcha, didnt know that. Should we put deprecation warnings on those and set a version for removal?

sneakers-the-rat:
It should be v possible to make additional commands hook in as plugins, eg. With any of these strategies: https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/

But yes maybe better as a follow on PR once we know how we want that to work :)

cmungall (approved):

linkml/cli/main.py
pkalita-lbl: Yes, I think so. I can take a crack at following the new (to me, anyway 😁) deprecation process docs for the old `linkml.valiators` CLIs in a separate PR if you just want to drop them from this PR.

sneakers-the-rat:
OK, removed the validators. The lag was super bad, taking like 2.5-4.5s to open the cli, and it turns out that adding the metamodel dataclasses in the build results classes was making imports take 5ever, so I added a quick type annotation to skip building the schema for the metamodel classes (because they validate themselves, pydantic doesn't need to validate them, and we don't serialize using the build result models).

now the latency is basically down to how long it takes to import the whole generators module. I tried wrapping those with delayed imports that would only happen when you call them, but each of the other main commands also ends up importing `generators` at some point, so we would nee dto wrap the whole thing, and then the structure of the cli gets super confusing and the `--help` doesn't work, so i think this is fine for now.

 
Screenshot 2024-08-05 at 11 53 12 AM

pkalita-lbl (approved):

PR #2228, opened 2024-07-25
[pydanticgen][minor] Implement `equals_string`, `equals_string_in`

sneakers-the-rat:
Somewhere along the line we lost the ability to do `equals_string`, and also we don't know how to do `equals_string_in` yet in pydanticgen. Restored that functionality so we use literals for that.

for the sake of concreteness...

```yaml
id: test_schema
name: test_info
imports:
  - linkml:types
classes:
  A:
    attributes:
      my_slot:
        range: string
        equals_string: THE TRUE STRING
        required: true
      my_slot_2:
        range: string
        equals_string_in:
        - ONE STRING
        - TWO STRING
        required: true
```

gets you

```python
class A(BaseModel):
    my_slot: Literal["THE TRUE STRING"] = Field(...)
    my_slot_2: Literal["ONE STRING", "TWO STRING"] = Field(...)
```

also `predefined_slot_value` was broken because it turns out there's a little inconsistency in the pydantic generator about when exactly we camelcase the class names. It missed predefined slot values for any class name that changes when it's camelcased - so i fixed that and also changed the test class name to one of those to prevent regressions.

edit: actually i moved that last part to #2229


cmungall:
Thanks!

I'd also like to add this in the compliance tests. Currently we do have tests for equals_string and equals_string_in, but they are bundled in with boolean tests in test_boolean_slot_compliance. We should probably have an atomic test for this.

Which would you prefer

1. merge this PR now and we can do a separate compliance test later (fastest)
2. you include in this PR (next fastest)
3. I'll contribute to your PR and add this (slowest :-)

sneakers-the-rat:
I got it, am just sitting down to my keyboard now :)

sneakers-the-rat:
OK I think i did it, i honestly find the compliance tests sort of hard to think about - i really like the idea but each test requires so much special casing in the body of the test function, and i almost did it incorrectly bc i didn't realize the 'key' for the generation step needs to be unique or else i was reusing old models. what would be really nice to have is a set of fixtures, each of which is parameterized with all the test cases that currently go into each compliance test, and then to declare that a given framework implements that metamodel feature i request the relevant fixture and can then handle each of the framework-specific things in a separate test function. side bonus of then we need to run way fewer tests bc currently we are effectively skipping many tests via the `INCOMPLETE` implementation but still taking time to generate framework output for them.

anyway a conversation for another day, i think what i did is correct there but not really sure.

PS. I also implemented `equals_string_in` for the json schema generator bc it seemed easy enough to do, but not really familiar with that generator so lmk if i did that wrong.

cmungall:
Thx! Fully agree, made a new issue for the compliance suite comments

The SchemaLoader issue is unexpected, need to look into that more...

Issue #2219, opened 2024-07-23
`induced_slot` mutates values and does a hidden `deepcopy` and generators have become reliant on that


sneakers-the-rat:
**Describe the bug**

this is a pretty whacky one!!!! has taken me about 8 hours to wrap my head around...

So i was out perfbusting, right. and `induced_slot` was as usual, the culprit. 

## Act 1: `required`

So first I noticed that attributes are set on the wrong object here, on the `slot` class rather than the `induced_slot` - that's bad right? we're mutating the parent slot that we copied from, defeating the purpose of the copy:

https://github.com/linkml/linkml-runtime/blob/984e4b2657f681dfa50e4cbd692bdb946ea56976/linkml_runtime/utils/schemaview.py#L1380-L1383

So i swapped that out with `induced_slot` in both places and got an error in this test: https://github.com/linkml/linkml/blob/bd49d48575d20e80875c2e1ef735d910f3a0d250/tests/test_compliance/test_core_compliance.py#L505

for the SQL generator, the `id` slot is no longer required. This fails in a few different tests for a few generators - the SQL generators and the SHACL generators. 

i was like "oop, that's definitely a bug," because we shouldn't be depending on accidental mutation of the parent `slot` object, and started opening this issue.

But i wanted to have a working version of the method in hand, so i kept working and things got stranger...

## Act 2: `slot_usage`

The main thing i wanted to tackle `induced_slot` doing this iteration over `(n fields in SlotDefinition) * (n parent classes)` here: **https://github.com/linkml/linkml-runtime/blob/984e4b2657f681dfa50e4cbd692bdb946ea56976/linkml_runtime/utils/schemaview.py#L1352-L1374**

I figured "hey all that's doing is grabbing `slot_usage` and updating from it, so why don't get just iterate over `class_ancestors` and only do stuff if `slot_usage` is declared for that ancestor and cut out like 99% of the iteration time." 

So I refactored it to do so and then got a bug that i couldn't explain at all - in `test_slot_usage` ( https://github.com/linkml/linkml/blob/bd49d48575d20e80875c2e1ef735d910f3a0d250/tests/test_compliance/test_inheritance_compliance.py#L456 ) the SQL generator was making an SQLA model like...

(in pseudocode bc i can't remember how i got it to display the generated SQLA as i'm writing this)

```python
class D():
    s1_id: 'Y' = ForeignKey('X.id')
    relations = Relation('Y')
```

[...]

PR #2229, opened 2024-07-25
[pydanticgen] make default for `Optional[<T>]` always be `None`

sneakers-the-rat:
What @daniel-ji was working on over in https://github.com/linkml/linkml/pull/2201 intersected with what i was working on in https://github.com/linkml/linkml/pull/2228. Also related to https://github.com/linkml/linkml/pull/2019#issuecomment-2019224575 and recent problems in https://github.com/linkml/linkml/issues/2219 via the pattern of multivalued optional slots causing problems in the python generator.

to quote that:

> This should generally be true, right? 
>  - if a slot is `required`, i would expect there to be no default or default factory (unless there is an `if_absent` or other precomputed value), and for models to fail validation is no value is passed. 
> - if a slot is not `required`, the type annotation should be wrapped in `Optional[]` and the default should be `None`. 
> 
> [`Optional` is an alias for `Union[X, None]`](https://docs.python.org/3/library/typing.html#typing.Optional), so there should be a three way definitional equality
`required: false === Optional[] === default = None`

previously in 2019 i had changed the test, but only fixed the behavior by accident by breaking all predefined values for class names that change when they are camelcased - my bad. 

when there is no predefined value and a slot isn't required, the template class `PydanticAttribute` handles setting the default in the `Field` to `None`. 

this builds off 2201 so that should be merged first i think


cmungall (approved):

glyg:
Hi,

Yesterday I re-generated my pydantic classes from the linkml schema after not touching it for a few months.

I understand the rationale behind this change, but my code happens to rely a lot on the previous behaviour, as I assumed throughout that my multivalued class attributes were initialised as empty lists I could `append` to. Now I get `object of type None have no append method` errors.


What would be the best strategy for me to restore the `default_factory` behaviour?

I can always do 

```python
if my_obj.my_multivalued_slot is None:
    my_obj.my_multivalued_slot = []
my_class.my_multivalued_slot.append(var)
```

but this has to be done in a lot of places, as I tend to instantiate minimal objects and populate them later.

Thanks for any hint on how to solve that :)


Best
Guillaume

sneakers-the-rat:
probably the easiest thing to do would be to override the `attribute` template so e.g. `{% if multivalued and not required %}default_factory=list{% endif %}`

https://linkml.io/linkml/generators/pydantic.html#cmdoption-gen-pydantic-template-dir

glyg:
Thanks a lot for the super fast answer, and the rest of your worked :)
The fix works

cmungall:
I don't think the template solution is best in the long term. I don't love combinatorial explosion of flags but having some kind of flag here (at the pydanticgen level, potentially later in the schema).

There is a lot of code in the wild that is written to use the generated dataclasses. It would be great if it could just be swapped out without lots of null checks before appending..

sneakers-the-rat:
I'm looking at the linked PR, why is it unclear when the values should exist? I.e. why couldn't the models be initiated so they have a known null state and are then populated by assignment, or have the values computed and passed at instantiation, rather than being appended in a for loop?

sierra-moxon:
https://github.com/linkml/linkml-map/pull/87

sneakers-the-rat:
yes that is the above template workaround, it looks like that was the successor to https://github.com/linkml/linkml-map/pull/86 
i can see why an option might be useful, and that would be relatively straightforward to propagate through the template models. i think some of the awkwardness there could be handled by inverting the order of instantiation -> population to population -> instantiation and making types that handle the list -> dict[name, item] coercion (you'll want to do that with pydantic models anyway, since pydantic validation/coercion doesn't work when you mutate a collection instance's members instead of passing them on instantiation). less boilerplate overall.

for general use i still think the default should be for "no value present" to be distinguishable from "empty collection present," but i don't see a problem with having an option for collapsing them. want me to write that?

PR #2256, opened 2024-08-06
Support rdfs:subClassOf with the SHACL generator

vincentkelleher:
A quick update for the SHACL generator that adds support for `rdfs:subClassOf` which allows [more advanced class targeting as described in the SHACL specification](https://www.w3.org/TR/shacl/#targetClass).

I've assumed that the LinkML `is_a` attribute is a perfect fit for this feature as it expresses inheritance.


vincentkelleher:
There seems to be an error in a dependency referencing `debugpy:1.8.3` (which doesn't seem to exist according to their repository) 🤔 

image

cmungall:
Here's the issue:

- #2259             

You could merge in #2260 or update poetry.lock yourself...

vincentkelleher:
@cmungall I wasn't aware of the issue and I didn't want to break anything so I brought it up in this PR 😇 

I rebased my branch and ran `poetry update debugpy` to update the `poetry.lock` file and things are looking fine 🥳

cmungall (approved):

vincentkelleher:
Thanks @cmungall 🙏

PR #2263, opened 2024-08-09
[cli] `linkml build`

sneakers-the-rat:
Continuing: https://github.com/orgs/linkml/discussions/2203

i only had a lil bit of time to work on this today, so it's in a pretty hairy state, and similarly i have to run now so will fill in full details and clean up later. opening this as a draft to show the group what i gone and done and get feedback.

## want

I really wanted this today when i was working with my lab on some schemas - we're gonna want to make a set of schemas and models, including potentially some of our own custom generators, and then use those in a bunch of different packages in different languages (as is the whole idea), and what i wanted was:
- while we were working live it would be great to have a `linkml watch` command so that we could display the built versions of the schema as it was updated.... so this is a step towards that
- I want a command I can use in CI like `linkml build --check` where i can check if all the models are currently up to date in my project, as well as `linkml build --diff` to be able to see what would change on build. to make that reasonable what i need is some durable mapping between source and output formats along with their configuration
- i'm going to want to be able to do:

```yaml
imports:
- name: miniscope_schema
  from:
    pypi:
      package: miniscope-schema
      version: 1.0.1
```

and do that what i need to do is be able to locate where a schema file is within a python package from a name.
- it would be really nice to have a standard way of declaring that a given python package or git repo has some linkml schemas in it. we can also use the same models to eg. load from a top-level `linkml.yaml` file for non-python projects, the config file finding logic is pretty simple. That will also allow us to do stuff like the above structure import syntax for git repositories, but instead of needing to hardcode the location of the schema within the repo we can just refer to the schema name so the schema authors don't have to worry about moving them around and refactoring and whatnot.
- the current project generator works, but i think we will probably want some "nice to have" config ergonomics to be able to set global settings and customize the build process. it also feels like this is not necessarily a generator per se. but we can make this work with the project generator if we prefer that.

## did

Added a `linkml build` command that reads a build configuration from `pyproject.toml` like this:

```toml
[tool.linkml]
# inject a schema map into the top-level __init__.py file to discover
# schema sources and yaml files.
include_schemas = true
# or, for namespace packages... 
# include_schemas = [
#   package_name
# ] 

[tool.linkml.schema]
# define a mapping between short IDs and paths
# ideally these would match the `id` of the schema, but whatever.
my_schema = "path/from/root.yaml"
other_schema = "different/path.yaml"
# should also be able to support eg including from a url
# linkml = "https://example.com/wherever/the/actual/schema/is.yaml

[tool.linkml.generate]
# settings for individual generator for all schema
pythongen.enabled = true
owlgen.enabled = true

[tool.linkml.generate.global]
# global settings that apply to all generators (if applicable)

# can use jinja template strings, and the context has
# ``schema`` : the SchemaDefinition
# ``generator`` : the instantiated generator class
output = "path/{{ schema.version }}/{{ schema.name }}.{{ generator.file_extension }} "

[tool.linkml.generate.pydantic]
# equivalent to `pythongen.enabled = true` syntax above but expanded...
enable = false
# set additional defaults whether globally enabled for the `build` command or not,
# defaults are used when doing `linkml generate {generator}`
meta = "full"
# override global defaults
output = "path/otherpath/pydantic/{{ schema.name }}.py"

[tool.linkml.build.my_schema]
# enable generators with default settings
excelgen.enable = true
csvgen.enable = true

[tool.linkml.build.my_schema.pydanticgen]
# customize specific generator for a specific schema
enable = true
output = "specific/path.py"

# can use pre-build and post-build hooks.
# pre-build hooks receive the schema and the build config model if they
# request them in the function signature
pre_build = "mypackage.module:function"
# by default "type" is "module", but can also call shell scripts
post_build = {type = "shell", value="black {{ config.output }}" }
# or python code
# post_build = { type = "python", value='''
# with open({{ config.output }}, 'r') as f:
#     print(f.read())
# ''' }
```

- `tool.linkml.schema` - mapping from schema name to path
- `tool.linkml.generate` - configuration settings for generators, i split these out from `tool.linkml.build` so that a) `tool.linkml.build` only has schema names for keys (and we wouldn't have some limitation like not being able to name schemas "global" or something), and b) because i figure it might be nice to also configure generators per project so when i do `linkml generate pydantic my_schema.yaml` i read these defaults and don't have to declare them twice.
- `tool.linkml.build.{schema_name}` defaults for a given schema for all generators
- `tool.linkml.build.{schema_name}.{generator}` config for a schema and generator.

this format deviates from the other thing i had in the discussion, most notably in that it assumes a single build configuration per schema per generator, which may or may not be desirable. 

So, in the spirit of "self-defining schema language" i made the build models be generated by `linkml build`. try it out :)

## todo

- [ ] `include_schema` option: include all schemas in a `__linkml__` directory and add a mapping to `__init__.py` to get the source schema as well as the build artifacts for that schema.
- [ ] tests
  - [ ] test that the example config in the docs works
  - [ ] ... everything else, will expand l8r
- [ ] pre/post build hooks - need these bc eg. in `nwb_linkml` we have to do a separate pre-build stage to generate the linkml models from nwb schema


dalito:
So do I understand it correctly that the build tool should do the following:
```mermaid
flowchart
    A[Multiple versioned schemas and models] -->|pre-build hooks| B(combined *build* LinkML model)
    B -->|post-build hooks| C(custom generators)
    C -->D[packages in configured languages]
```
So it combines all schemas before it runs generators and lets me configure all involved steps? Sounds great!

sneakers-the-rat:
sort of, maybe something like this:

```mermaid
flowchart TD
    subgraph schema_declaration
        schema_a --> schema_a.yaml
        schema_b --> schema_b.yaml
    end


    subgraph global_config
        output
        option_a
    end

    subgraph generator_a_config
        output_gena(output)
        output --> output_gena
        option_b
    end

    subgraph generator_b_config
        output_genb(output)
        output --> output_genb
        option_c
    end

    subgraph generator_a_schema_a
        output_gena_scha(output)
        option_b_scha(option_b)
        option_a_scha(option_a)
        output_gena --> output_gena_scha
        option_a --> option_a_scha
        option_b --> option_b_scha
    end

    subgraph generator_b_schema_a
        output_genb_scha(output)
        option_c_scha(option_c)
        output_genb --> output_genb_scha
        option_c --> option_c_scha
    end

    generator_a_schema_a --> build_a_schema_a
    generator_b_schema_a --> build_b_schema_a
    schema_a --> build_a_schema_a
    schema_a --> build_b_schema_a

    subgraph build_a_schema_a
        direction LR
        pre_build_a(pre_build)
        generate_a(generate)
        customization_a(customizations)
        post_build_a(post_build)
        output_build_a(output)
        pre_build_a --> generate_a
        customization_a --> generate_a
        generate_a --> post_build_a
        post_build_a --> output_build_a

    end

    subgraph build_b_schema_a
        direction LR
        pre_build_b(pre_build)
        generate_b(generate)
        customization_b(customizations)
        post_build_b(post_build)
        output_build_b(output)
        pre_build_b --> generate_b
        customization_b --> generate_b
        generate_b --> post_build_b
        post_build_b --> output_build_b

    end
```

where the schema declarations are independent from the build config -- you are just declaring that "there exists some schema `schema_name` that is located at `path/relative/to/root.yaml`" 

then for each of those declared schemas you can configure some build that happens when you call `linkml build` with predictable outputs. so i'm not sure what you mean by 'combines schemas before it runs generators,' but all the schema logic will be unchanged (eg. schemas can import from one another, so in that way they might be combined), and each schema can have an independent set of outputs. So using this PR as an example, we might have multiple schemas in multiple locations (maybe that's desirable, maybe that's not, idk) and they're used for different things - some things are used in the package itself, so it needs python classes and nothing else. other things are used in different ways and so need a different set of outputs with different configurations, and so on.

the multiple layers of global --> generator global --> schema global --> schema * generator configuration are just for ease of use and not having to add one million lines to your `pyproject.toml`. You ultimately could configure everything at the schema * generator level and not worry about all that.

so to summarize, the two goals of this PR are:
- make a standard way of declaring that "this python package has linkml schema, and here is where you can find them" (and that same model could be reused with a root`linkml.yaml` file for non-python packages)
- make a generalized build config system similar to projectgen but without the assumption of 1 schema per config s.t. there are predictable, repeatable outputs (eg. for a CI action that tests if models need to be updated, or for live rebuilding while working) that can be easily inferred by static analysis of the config (ie. I can find all the built artifacts from each schema)

dalito:
Thanks, Jonny. Your additional explanation makes it even clearer. I like the `linkml.yaml` for declaring _"in this package is a LinkML model and you find it here"_.

Currently, linkML model and generated code are (always?) distributed as one big package. I wonder if distributing the model separately from the language artifacts would be better. It has the advantage that the language-specific packaging tooling could be used and that the problem with versioning (code vs. model) would disappear. This would require a different cookiecutter/copier template (or or a `linkml scaffold` command) that produces one repo for the schema and one more for each generated language. Package authors could still include a copy of the model that the specific package version is based on.

In this case of pure-model repos the `linkml.yaml` could also include references to the derived packages to allow discovery without a linkml-specific package index.

sneakers-the-rat:
>  linkml.yaml 

yep! intended target is to be able to do any of the supported formats, `pyproject.toml`, `linkml.yaml`, etc. because many people don't want any more dotfile-like things in the repo root (myself included) and move to package config into combined config files.

> that produces one repo for the schema and one more for each generated language. 

idk about the producing many repos part, but yes this would be complementary to (and support) the ability to do any pattern of generation that you wanted to. we would want to guard against generating above the root of the config (directory traversal can get confusing fast), but eg. if someone made a parent directory/repo and had a bunch of child repos/submodules then one could handle generating all the language artifacts to them (as well as generate/update metadata with a `post_build` hook). 

i'm in general more focused on making it easier to integrate linkml in existing projects rather than making a bunch of new ones, but i suppose that's why i'm more interested in this approach than hacking on the cookiecutter, so again, complementary approaches that i don't think conflict with one another :)

sneakers-the-rat:
ok getting back to this because i have needed it too many times since i started it and hacked around instead lol. i'm reorganizing the schema because the build config makes location arbitrary but that doesn't mean we need to be messy. 

In working on nwb-linkml and a few other projects, i realized pretty quick that the limitation of only having one schema * generator output per config was a pretty bad limitation and we need to transform the bottom level flattened build job to be a toml array of tables like

```toml
[tool.linkml.schema]
my_schema="some/path.yaml"

[tool.linkml.generator.my_generator]
# global settings for a generator, when used with `linkml generate` as well as `linkml build`
python_class_id="mymodule.submodule:GeneratorClass"
parent="pydanticgen"


[tool.linkml.build]
# global config for all builds
includemeta=true
output="path/{{schema.version}}/{{schema.name}}_{{generator.name}}.{{generator.file_extension}}"

[tool.linkml.build.schema]
# config for all builds of the named schema
my_schema.enabled = true # make an output with this schema for all globally configured generators
my_schema.naming_mode="strict"
other_schema.naming_mode="any"

[tool.linkml.build.generator]
# config for all builds with this generator
pydanticgen.enabled = true
pydanticgen.split = true

[[tool.linkml.build.artifact]]
# an individual build 
generator="pydanticgen"
schema="my_schema"
output="some/folder.py"

[[tool.linkml.build.artifact]]
generator="pydanticgen"
schema="my_schema"
another_option="idk something you'd want to change"
output="some/other/folder.py"

[[tool.linkml.build.artifact]]
generator="my_generator"
schema="my_schema"
output="some/third/folder.py"

```

still WIP but anyway we'll get there

sierra-moxon:
closing as a PR, but branch remains for future work in the forked repo.

PR #2265, opened 2024-08-09
simplify bug report template

### LINKED ISSUE #2250: Change bug report template
I have no idea why we ask people for their web browser in our github issue template.

And why screenshots? Most people will be using an IDE or command line or environment in which precise formatted text can be pasted

personally and perhaps ironically when I submit issues I don't like to be constrained by templates or schemas.. we should only ask the most important information. If we do want to gather info, I would say the most relevant things to ask in order are:

- have you provided both a schema (and if applicable, a data file) that replicates the issue
- version of linkml

then maybe

- how did you install linkml
- what version of python (if you know)

nlharris:
fixes #2250

btw I don't understand why the chunk that starts "name: Bug report" is repeated twice verbatim, but leaving it that way in case it's needed.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -22,4 +22,6 @@ A clear and concise description of what the bug is (including the error message 
 **Version of LinkML you are using**
 
 **Please provide a schema (and if applicable, a data file) that replicates the issue**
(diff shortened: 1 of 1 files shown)


cmungall (approved):

nlharris:
I'm not sure what codecov is objecting to here.

cmungall:
> I'm not sure what codecov is objecting to here.

spurious timeouts

PR #2266, opened 2024-08-09
simplify feature request template

nlharris:
#2265 didn't specifically mention also simplifying the feature request template but I thought we might as well do that as well


cmungall (commented):
actually for this one I think getting a sense of priority is useful? I defer to you and @sierra-moxon.

But definitely not BOTH importance AND timeline. And a qualitiative answer is fine?

nlharris:
how about if we keep "How important is this feature" but delete the timeframe question?

cmungall:
perfect!

nlharris:
Great, can you approve so this can be merged?

cmungall (approved):

PR #2273, opened 2024-08-15
fix autogen'd pydantic regex pattern validator

daniel-ji:
With the current generated Pydantic validator, a nice thing about it is if you have a union of types (generated by an `any_of` field with different `range`s), then the validation will only check the appropriate types to see if they match the pattern. So if you have a field that accepts an int and a string, it will only see if strings will pass the regex pattern check and leave the ints alone. However, this behavior is not consistent when it is now a multivalued slot. This PR makes the behavior consistent.

### DIFF
--- a/linkml/generators/pydanticgen/templates/validator.py.jinja
+++ b/linkml/generators/pydanticgen/templates/validator.py.jinja
@@ -3,7 +3,7 @@
         pattern=re.compile(r"{{pattern}}")
         if isinstance(v,list):
             for element in v:
-                if not pattern.match(element):
+                if isinstance(v, str) and not pattern.match(element):
                     raise ValueError(f"Invalid {{name}} format: {element}")
         elif isinstance(v,str):
             if not pattern.match(v):


daniel-ji:
@sneakers-the-rat If you could take a look at this, that would be great. Thank you!

sneakers-the-rat:
this looks good to me but for future PRs we might want to consider swapping it out with this

```python
class MyModel(ConfiguredBaseModel):
    a_field: Union[int, List[Annotated[str, Field(pattern="a_regex_pattern"]]
```

bc that shows up in the json schema and also is probably faster, but for this PR this looks good

linkml/generators/pydanticgen/templates/validator.py.jinja
```diff
@@ -3,9 +3,9 @@
...
             for element in v:
-                if not pattern.match(element):
+                if isinstance(v, str) and not pattern.match(element):
                     raise ValueError(f"Invalid {{name}} format: {element}")
         elif isinstance(v,str):
             if not pattern.match(v):
                 raise ValueError(f"Invalid {{name}} format: {v}")
-        return v
\ No newline at end of file
+        return v
```
sneakers-the-rat, line 11: it's not really a problem since i figure most people will be using black formatting if given the option, so don't worry about changing it, but ffr but the absence of newline at the end of template is intentional since it's a template <3

sneakers-the-rat (approved):

cmungall (approved):

Issue #2279, opened 2024-08-21
IfAbsentProcessor too strict with matching type with custom types.


sneakers-the-rat:
**Describe the bug**

The beloved new IfAbsentProcessor makes a series of checks using `<T>.type_name`: https://github.com/linkml/linkml/blob/2127636dbe178f13f314455d04a8aa5b93f77b4e/linkml/generators/common/ifabsent_processor.py#L64

the intent of the tests is to match against linkml base types. however when one is using custom types it fails because the type no longer has the matching type_name, for obvious reasons. 

MWE:

```yaml
id: name
imports:
- linkml:types
types:
  text:
    name: text
    typeof: string

classes:
  MyClass:
    attributes:
      value:
        range: text
        ifabsent: string(hey)
```

```shell
$ linkml generate pydantic test.yml                                                              
ValueError: The ifabsent value `string(hey)` of the `value` slot could not be processed
```

we are looking for something like

```python
typ = sv.induced_type(type_name)
typ.repr == 'str' or 'string' in sv.type_ancestors(type_name)
```

**Version of LinkML you are using**

`1.8.1.post78.dev0+be8dbf21` (but true in current `main` branch too)

PR #2281, opened 2024-08-21
fix 2279 - use base types in IfAbsentProcessor

sneakers-the-rat:
fix: https://github.com/linkml/linkml/issues/2279

hopefully self explanatory. i am seepy but can return to write more tmrw.

some of these i couldn't figure out how to actually test, but feel free to request changes and lmk on those.

edit: more complete description - we try and match the base type! I couldn't find a method for this in schemaview, where it probably belongs. matching with string names is implicit and incomplete, we want to actually get the relevant type so that we can handle custom types.
- first check to see if the range is already one of the base types, return if it is
- then we return early if it isn't a type at all (eg. for enums)
- then we try and find a matching type ancestor, returning if we do.
- and finally, as a last resort, we try and match the base type using the literal `mro` of the base types.

also fixed one minor warning that kept coming up because the regex pattern wasn't an `r''` string.

### DIFF
--- a/linkml/generators/common/ifabsent_processor.py
+++ b/linkml/generators/common/ifabsent_processor.py
@@ -1,7 +1,13 @@
 import abc
 import re
+import sys
 from abc import ABC
-from typing import Any, Optional
+from typing import Any, Optional, Type, Union
+
+if sys.version_info.minor < 10:
+    from typing_extensions import TypeAlias
+else:
+    from typing import TypeAlias
 
 from linkml_runtime import SchemaView
 from linkml_runtime.linkml_model import (
@@ -18,6 +24,7 @@ from linkml_runtime.linkml_model import (
     String,
     Time,
     Uri,
+    types,
 )
 from linkml_runtime.linkml_model.types import (
     Curie,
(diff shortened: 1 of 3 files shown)


sneakers-the-rat:
tests pass, update the description, this is ready 4 review - relatively high priority bug since it breaks all custom types for pydanticgen and probably shaclgen too

sneakers-the-rat:
ping on this - custom types still broken for pythongen, pydantic, and shaclgen, and the ifabsent processor is becoming a bit of a moving target lately so would be nice to merge this before having to rebase again :)

linkml/generators/common/ifabsent_processor.py
```diff
@@ -1,7 +1,13 @@
 import abc
 import re
+import sys
 from abc import ABC
-from typing import Any, Optional
+from typing import Any, Optional, Type, Union
+
+if sys.version_info.minor < 10:
```
vincentkelleher, line 7: While looking up some information about the changes in Python 3.10's typings, I found that it could be interesting to also check the major version to avoid "issues" in the future if Python 4 comes out as Python 2 only goes to minor version 7 😇 

```suggestion
if sys.version_info < (3, 10):
```

I promise that my nitpicking self was under control but it's getting harder and harder 🤣

vincentkelleher: (You can ignore this if you want, it's not critical for the next 5 years or so 😉 )

vincentkelleher:
Learned a lot through this review, great contribution @sneakers-the-rat as always 👍 

The `_base_type(..)` method has its place in `SchemaView` for sure but that can be done later on when it has been battle tested.

I'm approving this although I have one non-critical comment to avoid blocking it.

vincentkelleher (approved):

linkml/generators/common/ifabsent_processor.py
sneakers-the-rat: reasonable ;)

linkml/generators/common/ifabsent_processor.py
```diff
@@ -31,6 +38,34 @@
...
+    Type[Time],
+    Type[Uri],
+    Type[Uriorcurie],
+]
+
+TYPES = [
+    t
+    for t in types.__dict__.values()
+    if isinstance(t, type) and t.__module__ == types.__name__ and hasattr(t, "type_name")
+]
```
sneakers-the-rat, line 67: i also think that this kind of thing would be great to have in a `linkml-meta` repository which just has the metamodel and metamodel accessories in it. couldn't find a way to get "all the base linkml types" anywhere

sierra-moxon: testing resolving this comment @sneakers-the-rat - just so I can see if merging is a possibility

sierra-moxon (approved):

sneakers-the-rat:
Thanks @sierra-moxon !!

PR #2285, opened 2024-08-26
Add some additional information to documentation generated for deprecated classes, slots, and enums

jat255:
I was surprised to see there was no information about deprecation in the default docgen templates. Looking at https://biolink.github.io/biolink-model/genetic_association/, I was inspired to make some changes to our local templates that maybe are of use upstream.

The styles/colors are hard-coded in, which isn't great, but I wasn't sure if there was a way to make that match a particular theme, or a better way to do it. I think my implementation is also non-exhaustive, but perhaps it's a start. Here's an example of what it looks like:

Deprecated classes and slots are strikethrough'ed and a "deprecated" note is added:
image

Headers clearly indicate when an item is deprecated or abstract, with a link to the exact replacement if present:
image

image

I also added some basic tests following the example of the other docgen tests.

### DIFF
--- a/linkml/generators/docgen/class.md.jinja2
+++ b/linkml/generators/docgen/class.md.jinja2
@@ -8,6 +8,17 @@
     {%- endif -%}
 {%- endif -%}
 
+{%- if element.deprecated %}
+    {%- set title = '~~' ~ title ~ '~~' -%}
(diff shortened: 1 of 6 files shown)


dalito:
Showing the information was addressed in #2583. Do you think that your PR is still needed?

sierra-moxon:
functionality added; closing PR.

Issue #2027, opened 2024-03-26
Convert tutorial markdown to jupyter notebooks and publish using nbsphinx


cmungall:
Currently the tutorial is hard to maintain - it uses a homegrown format to run all embedded scripts and check results against what is there.

Instead we could just author the tutorial as notebooks, in the docs/ folder, enable [nbsphinx](https://docs.readthedocs.io/en/stable/guides/jupyter.html) which will render them (statically) as part of the docs. It's then very easy to extend the tutorial or add new examples or change examples in existing pages.

We currently do this in the linkml-transformer repo, here is an example
https://linkml.io/linkml-transformer/#examples/Tutorial/
But note this is a mkdocs repo, we use `mknotebooks` which may look different

Some potential downsides

1. Our tutorials focus on command line usage rather than python usage. The `!`s to run a script in bash may be distracting
2. The documentation may end up looking like a notebook, e.g. with numbered cells (todo: investigate if nbsphinx allows disabling this; I like how it looks in mknotebooks

Issue #1540, opened 2023-07-16
Executing tutorials doesn't pass


rly:
**Describe the bug**
Executing the tutorial code results in various (minor) errors, such as mismatches between the output of `linkml-validate` and the expected output.

This occurs for tutorials 1, 2, 4, 8, 9, and 10. An example output for tutorial 01 is below.

**To reproduce**
1. Clone repo
2. Run `poetry run python -m linkml.utils.execute_tutorial -d /tmp/tutorial docs/intro/tutorial01.md` or in the branch for #1539, run `make test-tutorial-01`
3. Observe errors:
```
INFO:root:INPUT=docs/intro/tutorial01.md
FILE=personinfo.yaml
FILE=Outputs
FILE=data.yaml
FILE=Validate
FILE=bad-data.yaml
FILE=data.json
## 13 Blocks
INFO:root:Executing in: /tmp/tutorial/tutorial01
+++ # Part 1: Creating your first LinkML schema
+++ 
+++ We assume that you already have LinkML [installed](install)
+++ 
+++ For the purposes of this tutorial, the simplest setup is to use a virtual environment, and then install linkml:
+++ 
+++ 
INFO:root:# Block: bash None
+++ 
+++ You can check the install worked:
+++ 
INFO:root:# Block: bash None
INFO:root:Executing: ['linkml-convert', '--help']
INFO:root:OUT [sample] = Usage: linkml-convert [OPTIONS
INFO:root:Success!
+++ 
+++ As always, you can consult the [FAQ](../faq/index) if you have issues.
+++ 
+++ ## Your first schema
+++ 
+++ Our first schema consists of a single *class* `Person`, with a number of *slots*:
+++ 
+++ First create a file
+++ 
+++ 
INFO:root:# Block: yaml personinfo.yaml
+++ 
+++ (note that all files are available in the [examples/tutorial](https://github.com/linkml/linkml/tree/main/examples/tutorial) folder of this repository)
+++ 
+++ ## Converting to JSON-Schema
+++ 
+++ Now run the following command on the file you just created:
+++ 
INFO:root:# Block: bash None
INFO:root:Executing: ['gen-json-schema', 'personinfo.yaml']
INFO:root:OUT [sample] = {
    "$defs": {
        "Pers
[...]

PR #2289, opened 2024-08-28
Low hanging coverage fruit

sneakers-the-rat:
bored while waiting for tests to run and i went spelunking around in some parts of the package i haven't checked out before, seeing if i could find low-hanging fruit to boost coverage, found 3 things:

- `utils/execute_tutorial.py` - looks like it hasn't really been touched substantively since early 2022, it isn't used elsewhere in the repo, and it seems like an internal doctest tool rather than something anyone might be depending on. if we need to do something like this now we can just use doctests :)
- workspaces datamodel - this also isn't used anywhere, i'm still not really sure what workspaces are, but since the datamodel is autogenerated, i don't think we need to include it in coverage
- `sqldb` roundtrip - the `cli` part of the sqlutils module is untested, but it was easy enough to add a simple roundtrip test that mimics the other tests in the `test_data` module. 

i am like an aquarium shrimp that roves around the tank and eats the little bits of debris


cmungall:
I am not sure the execute-tutorial use case is completely covered by doctest. However, it is ad-hoc and we do need a replacement for it

Some discussion here:

- #1540              

Perhaps we should just move everything to notebooks:

- #2027

sneakers-the-rat:
It it ever run? I couldnt find it in the tests or CI. Definitely in favor of using notebooks, or the myst code block style as used in the arrays page which is a nice blend of notebooks and markdown that doesnt require diffing a bunch of JSON, and also gives you more control over style. I added a directive to run shell commands and show output to the docs a few weeks ago, and so if we want that to also be tested rather than just rendered that would be something we can use Sybil for.

sierra-moxon:
@sneakers-the-rat - would you like to close this PR in favor of us converting to myst code block style?  I came here hoping to resolve conflicts in case that helps push this along on the review side, but I do not have perms to resolve them and noticed the discussion going on :)

PR #2306, opened 2024-09-09
Introduce a docgen subfolder type separation argument

vincentkelleher:
This pull request originates from a concrete use case in the Gaia-X project.

### Before this change

Initially the Gaia-X ontology documentation website used to look something like this:

image

Classes, enums, types, slots and others were all hosted in the same folder and differentiated by a title prefix representing their type as you can see in the left navigation bar.

### Tackling the issue

To fix this issue, I propose to introduce a `--subfolder-type-separation` argument to the `gen-doc` generator. Once argument is provided, the generator will create a subfolder per type (classes, types, enums, slots and subsets) and generate the corresponding pages in those subfolders.

The result is the following on the [Gaia-X ontology documentation website](https://w3id.org/gaia-x):

image

image

image

The result can be seen at https://w3id.org/gaia-x.

### Caveats 

It is important to state that the Gaia-X ontology documentation website uses custom templates to remove the type prefix and some other elements that are to our liking.

I chose not to implement those changes in the LinkML project as templates should be tweaked by users in my opinion.

### DIFF
--- a/linkml/generators/docgen.py
+++ b/linkml/generators/docgen.py
@@ -50,6 +50,13 @@ DIALECT = MarkdownDialect
 MAX_CHARS_IN_TABLE = 80
 MAX_RANK = 1000
 
+SCHEMA_SUBFOLDER = "schemas"
+CLASS_SUBFOLDER = "classes"
+SLOT_SUBFOLDER = "slots"
+ENUM_SUBFOLDER = "enums"
+TYPE_SUBFOLDER = "types"
+SUBSET_SUBFOLDER = "subsets"
+
 
 def enshorten(input):
     """
@@ -144,6 +151,9 @@ class DocGenerator(Generator):
     include_top_level_diagram: bool = False
     """Whether the index page should include a schema diagram"""
 
+    subfolder_type_separation: bool = False
+    """Whether each type (class, slot, etc.) should be put in separate subfolder for navigation purposes"""
+
     example_directory: Optional[str] = None
     example_runner: ExampleRunner = field(default_factory=lambda: ExampleRunner())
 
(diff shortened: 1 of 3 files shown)


cmungall:
Looks great!

How do you deal with the fact that logical IRIs for elements are automatically assigned to the same top-level? This could easily be resolved with some redirects but AFAIW redirects are not easy in mkdocs.

This also can work as a fix for #1151 for people who want to deploy from a mac

vincentkelleher:
I've added a commit to use a local Kroki container through [TestContainers](https://testcontainers.com/) in the PlantUML generator tests to reduce the risk of timeout from the free trial Kroki.io server.

This only works on the `linux/amd64` and `linux/arm64` Github Action hosts as the Kroki server image is not built for other architectures 😇

vincentkelleher:
> Looks great!
> 
> How do you deal with the fact that logical IRIs for elements are automatically assigned to the same top-level? This could easily be resolved with some redirects but AFAIW redirects are not easy in mkdocs.
> 
> This also can work as a fix for #1151 for people who want to deploy from a mac

Hi @cmungall,

Great positive feedback as usual 🙌 

Are you talking about a case like this highlighted link in the Gaia-X ontology documentation ?

image

If so, I didn't manage this case at the moment as it really depends on how you publish your documentation: at the same URL as the ontology, on a different server, with specific redirect strategies, etc.

cmungall (approved):

cmungall:
Does this need to have #2318 merged/rebased in?

vincentkelleher:
@cmungall I've just rebased it on `main` so it's ready for a final review and merge 🥳

PR #2318, opened 2024-09-16
SHACL generator ignore subclass properties

vincentkelleher:
### Issue with my previous changes

I previously tried to manage class hierarchy in SHACL shapes with [this commit](https://github.com/linkml/linkml/commit/2ac21d3bfb8163f0862ac71b5fdfad63b59d1392) which introduced the use of `rdfs:subClassOf` in the SHACL shape description.

This was wrong as it was pointed out to me in [this issue](https://github.com/zazuko/rdf-validate-shacl/issues/144). The `sh:subClassOf` predicate should only be in the data/ontology graph not in the shapes graph, SHACL shapes do not have inheritance and only check that data matches the shape.

### Ignoring properties from children classes

I've tried specifying the class hierarchy in the data graph with `rdfs:subClassOf` but when I try to validate a case such as the one in [the previously mentioned issue](https://github.com/zazuko/rdf-validate-shacl/issues/144), it fails because it tries to apply the closed parent shape to the child instance which has extra properties hence failing the validation phase.

In this pull request, I propose to collect all the properties (slots, mixins, etc.) from child entities recursively to as them to the `sh:ignoredProperties` predicate of a parent.

### Example

If one has the following LinkML file:

```yaml
id: https://w3id.org/linkml/examples/ignoredProperties
name: ignoredProperties
prefixes:
  ex: https://w3id.org/linkml/examples/ignoredProperties/
imports:
  - linkml:types
default_range: string
default_prefix: ex

classes:
  Parent:
    attributes:
      name:
        range: string

  Child:
    is_a: Parent
    attributes:
      anotherAttribute:
        range: string

  GrandChild:
    is_a: Child
    attributes:
      anotherAttributeAgain:
        range: string
```

The current SHACL generator would generate the following:

```turtle
ex:ParentShape a sh:NodeShape ;
    sh:closed true ;
    sh:ignoredProperties ( rdf:type ) ;
    # ...

ex:ChildShape a sh:NodeShape ;
    sh:closed true ;
    sh:ignoredProperties ( rdf:type ) ;
    # ...

ex:GrandChildShape a sh:NodeShape ;
    sh:closed true ;
    sh:ignoredProperties ( rdf:type ) ;
    # ...
```

After this pull request, the SHACL generator will generate the following:

```turtle
ex:ParentShape a sh:NodeShape ;
    sh:closed true ;
    sh:ignoredProperties ( rdf:type ex:anotherAttribute ex:anotherAttributeAgain ) ;
    # ...

ex:ChildShape a sh:NodeShape ;
    sh:closed true ;
    sh:ignoredProperties ( rdf:type ex:anotherAttributeAgain ) ;
    # ...

ex:GrandChildShape a sh:NodeShape ;
    sh:closed true ;
    sh:ignoredProperties ( rdf:type ) ;
    # ...
```

This will allow the validator to ignore the presence of `ex:anotherAttribute` when applying the `ex:ParentShape` to an `ex:Child` class instance, or `ex:anotherAttribute` and `ex:anotherAttributeAgain` when applying the `ex:ParentShape` to an `ex:GrandChild` class instance.


vincentkelleher:
I've cherry-picked a commit from https://github.com/linkml/linkml/pull/2306 to use a local Kroki instance avoiding random check stage failures.

pyproject.toml
```diff
@@ -125,6 +125,7 @@ watchdog = ">=0.9.0"
 pyshacl = { version = "^0.25.0", optional = true }
 black = { version=">=24.0.0", optional = true }
 typing-extensions = { version=">=4.4.0", python="<3.9" }
+testcontainers = "3.7.1"
```
cmungall, line 128: I think this should be a test dependency only?

cmungall (changes requested):
check dependencies

pyproject.toml
vincentkelleher: Indeed, I'm going to update that right now 👍

vincentkelleher:
@cmungall I've fixed the `testcontainers` dependency position in the `pyproject.toml` file 😇

cmungall (approved):

vincentkelleher:
Thanks @cmungall 🙏

Issue #2297, opened 2024-08-30
Use dedicated namespace for the logger, not global one


yarikoptic:
ATM (v1.8.3-10-g4a2e6f7a) logging facilities are primarily used by invoking `logging.` instead of a dedicated logger:

```shell
❯ git grep 'logging\.\(info\|error\|debug\|warning\)' | wc -l
122
❯ git grep 'logger\.\(info\|error\|debug\|warning\)' | wc -l
48
```

and in runtime it is "similarish"

```shell
/home/yoh/proj/misc/linkml/linkml-runtime
❯ git grep 'logging\.\(info\|error\|debug\|warning\)' | wc -l
97
❯ git grep 'logger\.\(info\|error\|debug\|warning\)' | wc -l
1
```

It should be quite an easy refactoring via adopting similar to already existing convention to just define logger at the level of the module's name, e.g.

```shell
❯ git grep 'logger ='
linkml/generators/docgen.py:        self.logger = logging.getLogger(__name__)
linkml/generators/excelgen.py:        self.logger = logging.getLogger(__name__)
linkml/generators/linkmlgen.py:logger = logging.getLogger(__name__)
linkml/transformers/logical_model_transformer.py:logger = logging.getLogger(__name__)
...
```

NB brought attention to logging setup in linkml since @candleindark seems to observe our tool logging being "disabled" upon calling out to linkml's validator.

nlharris:
> It should be quite an easy refactoring via adopting similar to already existing convention to just define logger at the level of the module's name

Hi @yarikoptic, thanks for the suggestion. If you want to make a PR to address this, someone on the LinkML team will check and approve your PR. There's a huge backlog of requests right now, so the core team is unlikely to have time to implement this themselves.

yarikoptic:
Ok, will do, may be with the help from @candleindark ;-)

candleindark:
Changes to the linkml repo are proposed in https://github.com/linkml/linkml/pull/2323.

candleindark:
Changes to the [linkml-runtime](https://github.com/linkml/linkml-runtime) repo are proposed in https://github.com/linkml/linkml-runtime/pull/340.

dalito:
The PRs have been merged so I close this.

PR #2323, opened 2024-09-19
Improve logging by replacing the use of the root logger by module-level loggers

candleindark:
This PR improve logging by replacing the use of the root logger by module-level loggers. I.e., it addresses #2297.

Notes:
1. The changes touch many lines that are not tested previously. Thus, the low patch coverage.

### DIFF
--- a/linkml/generators/golanggen.py
+++ b/linkml/generators/golanggen.py
@@ -11,6 +11,8 @@ from linkml_runtime.utils.formatutils import camelcase, underscore
 from linkml._version import __version__
 from linkml.utils.generator import Generator, shared_arguments
 
+logger = logging.getLogger(__name__)
+
 type_map = {
     "str": "string",
     "int": "int",
@@ -175,7 +177,7 @@ class GolangGenerator(Generator):
                 if t.base and t.base in type_map:
                     return type_map[t.base]
                 else:
-                    logging.warning(f"Unknown type.base: {t.name}")
+                    logger.warning(f"Unknown type.base: {t.name}")
             return "string"
 
     @staticmethod
--- a/linkml/generators/jsonschemagen.py
+++ b/linkml/generators/jsonschemagen.py
@@ -24,6 +24,8 @@ from linkml._version import __version__
 from linkml.generators.common.type_designators import get_type_designator_value
 from linkml.utils.generator import Generator, shared_arguments
 
+logger = logging.getLogger(__name__)
+
 # Map from underlying python data type to json equivalent
 # Note: The underlying types are a union of any built-in python datatype + any type defined in
 #       linkml-runtime/utils/metamodelcore.py
@@ -222,14 +224,14 @@ class JsonSchemaGenerator(Generator):
 
     def __post_init__(self):
         if self.topClass:
-            logging.warning("topClass is deprecated - use top_class")
+            logger.warning("topClass is deprecated - use top_class")
             self.top_class = self.topClass
 
         super().__post_init__()
 
         if self.top_class:
             if self.schemaview.get_class(self.top_class) is None:
-                logging.warning(f"No class in schema named {self.top_class}")
+                logger.warning(f"No class in schema named {self.top_class}")
 
     def start_schema(self, inline: bool = False) -> JsonSchema:
         self.inline = inline
--- a/linkml/generators/owlgen.py
+++ b/linkml/generators/owlgen.py
@@ -40,6 +40,8 @@ from linkml import METAMODEL_NAMESPACE_NAME
 from linkml._version import __version__
 from linkml.utils.generator import Generator, shared_arguments
 
+logger = logging.getLogger(__name__)
+
 OWL_TYPE = URIRef  ## RDFS.Literal or OWL.Thing
 
 SWRL = rdflib.Namespace("http://www.w3.org/2003/11/swrl#")
@@ -270,7 +272,7 @@ class OwlSchemaGenerator(Generator):
                     # if isinstance(v, str):
                     #    obj = URIRef(msv.expand_curie(v))
                     # else:
-                    #    logging.debug(f"Skipping {uri} {metaslot_uri} => {v}")
+                    #    logger.debug(f"Skipping {uri} {metaslot_uri} => {v}")
                 else:
                     obj = Literal(v)
                 self.graph.add((uri, metaslot_uri, obj))
@@ -438,7 +440,7 @@ class OwlSchemaGenerator(Generator):
                 if slot:
                     own_slots.append(slot)
                 else:
-                    logging.warning(f"Unknown top-level slot {slot_name}")
+                    logger.warning(f"Unknown top-level slot {slot_name}")
         else:
             own_slots = []
         own_slots.extend(cls.slot_conditions.values())
@@ -695,7 +697,7 @@ class OwlSchemaGenerator(Generator):
         if element.equals_string is not None:
             equals_string = element.equals_string
             if is_literal is None:
-                logging.warning(f"ignoring equals_string={equals_string} as unable to tell if literal")
+                logger.warning(f"ignoring equals_string={equals_string} as unable to tell if literal")
             elif is_literal:
                 constraints[XSD.pattern] = equals_string
             else:
@@ -704,7 +706,7 @@ class OwlSchemaGenerator(Generator):
         if element.equals_string_in:
             equals_string_in = element.equals_string_in
             if is_literal is None:
-                logging.warning(f"ignoring equals_string={equals_string_in} as unable to tell if literal")
+                logger.warning(f"ignoring equals_string={equals_string_in} as unable to tell if literal")
             elif is_literal:
                 dt_exprs = [
                     self._datatype_restriction(XSD.string, [self._facet(XSD.pattern, s)]) for s in equals_string_in
@@ -770,7 +772,7 @@ class OwlSchemaGenerator(Generator):
                     if slot_uri == URIRef(att_uri):
                         n += 1
             if n > 1:
-                logging.warning(f"Ambiguous attribute: {slot.name} {slot_uri}")
+                logger.warning(f"Ambiguous attribute: {slot.name} {slot_uri}")
                 return
 
         self.add_metadata(slot, slot_uri)
@@ -906,7 +908,7 @@ class OwlSchemaGenerator(Generator):
             if pv_owl_type == RDFS.Literal:
                 pv_node = Literal(pv.text)
                 if pv.meaning:
-                    logging.warning(f"Meaning on literal {pv.text} in {e.name} is ignored")
+                    logger.warning(f"Meaning on literal {pv.text} in {e.name} is ignored")
             else:
                 pv_node = self._permissible_value_uri(pv, enum_uri, e)
             pv_uris.append(pv_node)
@@ -967,7 +969,7 @@ class OwlSchemaGenerator(Generator):
     def _add_rule(self, subject: Union[URIRef, BNode], rule: ClassRule, cls: ClassDefinition):
         if not self.use_swrl:
             return
-        logging.warning("SWRL support is experimental and incomplete")
+        logger.warning("SWRL support is experimental and incomplete")
         head = []
         body = []
         for pre in rule.preconditions:
@@ -1014,7 +1016,7 @@ class OwlSchemaGenerator(Generator):
                 owltypes.update(x_owltypes)
         owltypes.update(current)
         if len(owltypes) > 1:
-            logging.warning(f"Multiple owl types {owltypes}")
+            logger.warning(f"Multiple owl types {owltypes}")
             # if self.target_profile == OWLProfile.dl:
         return owltypes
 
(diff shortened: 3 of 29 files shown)


candleindark:
The "Build and test documentation / test_docs (pull_request)" is failing. Can it be a CI problem? I see the same failure in https://github.com/linkml/linkml/actions/runs/10931366172/job/30346202619?pr=2322.

sneakers-the-rat:
I love this.

Can I make a pitch for adding an init_logger function, and initializing a `logger` attribute on parent classes instead of spawning module level loggers?

Here's a module ive been carting around in various projects and tweaking: https://github.com/Aharoni-Lab/miniscope-io/blob/main/miniscope_io/logging.py

We don't need to add the `rich` part (though it does make for some very satisfying logs), but the parts that are good:

- logging formatting - add timestamps, log levels to logging messages
- both stream and file-based logging - sometimes you want both, sometimes you want one, sometimes you want none
- configurable log levels from env variables - this is the super useful one, its very nice being able to do this in a consistent way across the program without needing to get the nested module logging naming thing just right. Usually you want stdout/stderr log levels to be `warning`-level, but file-level logger is async and out of the way and so doesnt slow down program or clog stream and might want to be left lower to gather historical logs. Then sometimes you just want them to shut up entirely and it's good to have a knob for that.

Then we add a call like `self.logger = init_logger(self)` (or `init_logger(__name__)`, I guess I moved the object based name stuff out of that one) to the parent classes like Generator and Validator and whatnot and call it a day

candleindark:
@sneakers-the-rat Thanks for the input. I agree that setting up logging according to a centralized configuration is preferable to the setup in this PR. Is there a centralized configuration object in this project that I can hook the setup of the logging system to?

Here are my opinions/answers to the other suggestions.

> Can I make a pitch for adding an init_logger function, and initializing a `logger` attribute on parent classes instead of spawning module level loggers?

I think we can (this will allow setting formatters and handlers according to a centralized configuration as you illustrated) but only use this function to create the top logger for a package. For example, add the line `logger = init_logger(__name__)` in the `linkml/linkml/__init__.py` file where `__name__` evaluates to `"linkml"`. The `init_logger` function needs to return a logger with [`propagate`](https://docs.python.org/3.9/library/logging.html#logging.Logger.propagate) set to `False` though (more on this later). When a logger is needed in other places in the project, we should get one by calling `logger = logging.getLogger(__name__)` (or `logging.getLogger(f"{__name__}.{type(self).__name__}"` if a class specific logger is desirable). Note: We don't need to configure these lower level loggers. These loggers will use the [handlers](https://docs.python.org/3.9/library/logging.html#logging.Logger.propagate) and [log level](https://docs.python.org/3.9/library/logging.html#logging.Logger.setLevel) set up for the top logger.

These is a problem in using the `init_logger` in https://github.com/Aharoni-Lab/miniscope-io/blob/main/miniscope_io/logging.py to create multiple loggers within an hierarchy of loggers, caused by the behavior that it adds at least one handler to each logger it creates.

The following script prints `'This is a warning'` to `stderr` twice because log events logged to logger `"a.b"` [are passed to the handlers of all of its ancestor loggers](https://docs.python.org/3.9/library/logging.html#logging.Logger.propagate).

```python
import logging

logger_a = logging.getLogger('a')
logger_a_b = logging.getLogger('a.b')

handler_a = logging.StreamHandler()
handler_a_b = logging.StreamHandler()

logger_a.addHandler(handler_a)
logger_a_b.addHandler(handler_a_b)

logger_a_b.warning('This is a warning')
```

The problem is more pronounced if the code involved is being used as a package, and the application using it runs `logging.basicConfig()` which adds a handler to the root logger. The code below prints `'This is a warning'` to `stderr` three times.

```python
import logging

logging.basicConfig()

logger_a = logging.getLogger('a')
logger_a_b = logging.getLogger('a.b')

handler_a = logging.StreamHandler()
handler_a_b = logging.StreamHandler()

logger_a.addHandler(handler_a)
logger_a_b.addHandler(handler_a_b)

logger_a_b.warning('This is a warning')
```

To prevent logged events to be passed to the root logger, we need to set the `propagate` attribute of the top logger to `False` as mentioned earlier. The code below prints `'This is a warning'` to `stderr` two times.

```python
import logging

logging.basicConfig()

logger_a = logging.getLogger('a')
logger_a_b = logging.getLogger('a.b')

handler_a = logging.StreamHandler()
handler_a_b = logging.StreamHandler()

logger_a.addHandler(handler_a)
logger_a_b.addHandler(handler_a_b)

logger_a.propagate = False  # Stop passing the event to the root logger

logger_a_b.warning('This is a warning')
```

I have to say that repeating the statement, `logger = logging.getLogger(__name__)`, whenever a logger is needed in a module seems unseemly, but it makes the resulting the logger hierarchy [analogous](https://docs.python.org/3.9/library/logging.html#logger-objects), even identical, to the Python package hierarchy, which is great for locating the origin of a log message.

sneakers-the-rat:
ah yes, that was stuff i had done in the fuller version of this (which is much messier): https://github.com/auto-pi-lot/autopilot/blob/main/autopilot/utils/loggers.py

I have a mild personal preference for "one log file per python module" - so one logger per module with a `FileHandler` but no `StreamHandler`, and then only the top `linkml` logger has a `StreamHandler` and `propagate = False`, and then in my case i had child loggers for each class within the module, but we don't need to do that here if we don't want to. If we wanted only one logfile for all linkml, then we could just make the top level `linkml` logger have the handlers (and `propagate = False`), and then all the child loggers just propagate up to that.

> Is there a centralized configuration object in this project that I can hook the setup of the logging system to?

I don't think so, but i think we should. I am partial to the `pydantic-settings` package which can read from env vars as well as a .env file without too much trouble, but we could easily rig something like that up ourselves if we didnt' want to add the dep

candleindark:
> ah yes, that was stuff i had done in the fuller version of this (which is much messier): https://github.com/auto-pi-lot/autopilot/blob/main/autopilot/utils/loggers.py

This implementation has a similar problem as the previous one. A stream handler, in the form of `RichHandler`, and a `RotatingFileHandler` handler, are added to each module-level logger created. As you mentioned, a file handler for each module is a matter of preference. However, the stream handlers are problematic. With this `init_logger` function, loggers can be potentially created for package `a`, pacakge `a.b`, and module `a.b.c`. An event logged to logger `a.b.c` will be printed three times to `stdout`, the stream associated with `RichHandler`s, through three different instances of `RichHandler` in the three module-level loggers respectively.

> I have a mild personal preference for "one log file per python module" - so one logger per module with a `FileHandler` but no `StreamHandler`, and then only the top `linkml` logger has a `StreamHandler` and `propagate = False`, and then in my case i had child loggers for each class within the module, but we don't need to do that here if we don't want to. 

Having one `FileHandler` for each module level logger can be expensive. Again, consider the situation where loggers can be potentially created for package `a`, pacakge `a.b`, and module `a.b.c`. An event logged to logger `a.b.c` will be written to log files associated with package `a`, pacakge `a.b`, and module `a.b.c` respectively. Totaling three file write operations on three different files instead of just writing once. The situation can even be worse than exhibited in this example if the Python package hierarchy of project is deeper. Additionally, with this setup, you will potentially keep open the number of log files equal to the number of packages/modules in the project. If the goal is to separate log messages according to the modules they originate, separating log messages into different files is unnecessary. We can achieve the same goal with one log file and view it with a log viewer (I found https://github.com/tmoreno/open-log-viewer and https://github.com/tmoreno/open-log-viewer).

> If we wanted only one logfile for all linkml, then we could just make the top level `linkml` logger have the handlers (and `propagate = False`), and then all the child loggers just propagate up to that.

We can have optionally one file handler and optionally one stream handler for the logger `linkml` depending on a centralized config, and this handler should have `propagate = False`. We can also add a file handler to the loggers of selective modules in the project specified by the centralized config (no stream handler for loggers other than `linkml`). The `init_logger` function is only needed to create loggers that have a handler. To create a logger that doesn't have a handler, just `logging.getLogger(__name__)` for a module or `logging.getLogger(f"{__name__}.{type(self).__name__}")` if a class specific logger is desired (I think this is excessive, a module logger is specific enough unless the class logger is provided by the user/caller).

> > Is there a centralized configuration object in this project that I can hook the setup of the logging system to?
> 
> I don't think so, but i think we should. I am partial to the `pydantic-settings` package which can read from env vars as well as a .env file without too much trouble, but we could easily rig something like that up ourselves if we didnt' want to add the dep

I like `pydantic-settings` as well. I used the Pydantic V1 version before it spun off to a separate package.

I feel that the implementation of the centralize config and the `init_logger` function can be done in different PRs and require  consensus from the LinkML team, particularly the centralize config. The `init_logger` function can be built on the changes in this PR after it is merged for this PR only sets up loggers without any handlers.

Without the changes in this PR (possibly https://github.com/linkml/linkml-runtime/pull/340 as well), calling `linkml.validator.Validator.validate()` disables logging to `stderr` completely though I haven't found out the exact culprit.

sneakers-the-rat:
> An event logged to logger a.b.c will be printed three times to stdout, the stream associated with RichHandlers, through three different instances of RichHandler in the three module-level loggers respectively.

> An event logged to logger a.b.c will be written to log files associated with package a, pacakge a.b, and module a.b.c respectively. 

it doesn't ;) i have actually run this code before and would hopefully have noticed.

Screenshot 2024-09-26 at 7 37 24 PM

Screenshot 2024-09-26 at 7 38 31 PM

that's because loggers are only ever created in leaf modules (ie. there is no code in `__init__.py` files except for abstract classes that are never instantiated directly) - the message from `gui.widgets.terminal` doesn't propagate to `gui.widgets` and `gui` because there are no loggers for those modules. 

ime file logging is super inexpensive, especially for a non-perf-forward package like linkml, and writing the same number of log messages to multiple files has not been meaningfully more expensive than writing them to one file for normal logging rates. stream logging is super expensive tho, so ya it would be good to have a global config that could have separate log levels for file and stream loggers.

i brought it up just as an example, we can do whatever we want here. this logger was for a very different context, where you had ~dozens of instances of classes from a single module running simultaneously over long time periods in multiple processes (where you need separate files because they want to clobber each other), so probably not exactly what we want for linkml. i was more suggesting the idea of having a function that wraps logger initiation that does `_whatever we want_`. no need to go so hard on the poor autopilot logger.

> The init_logger function is only needed to create loggers that have a handler. To create a logger that doesn't have a handler, just logging.getLogger(__name__)

sure. that's basically what this function does. it only makes new handlers if a logger hasn't already been made for the given module, it just gives a single entrypoint to that. so you just do `_blammo_`: https://github.com/auto-pi-lot/autopilot/blob/6e2c543e74e627cf7061589a6543c7c26e54b38e/autopilot/hardware/__init__.py#L111

one call in a superclass and it infers what logger to get from the passed `self` parameter instead of needing to call `getLogger(__name__)` every time.

> I feel that the implementation of the centralize config and the init_logger function can be done in different PRs and require consensus from the LinkML team, particularly the centralize config. The init_logger function can be built on the changes in this PR after it is merged for this PR only sets up loggers without any handlers.

sure, i was just pitching an idea, i'm not a core maintainer so i certainly wouldn't hold up this (or any) PR with anything but suggestions that can be ignored.

candleindark:
> > An event logged to logger a.b.c will be printed three times to stdout, the stream associated with RichHandlers, through three different instances of RichHandler in the three module-level loggers respectively.
> 
> > An event logged to logger a.b.c will be written to log files associated with package a, pacakge a.b, and module a.b.c respectively.
> 
> it doesn't ;) i have actually run this code before and would hopefully have noticed.
> 
> Screenshot 2024-09-26 at 7 37 24 PM Screenshot 2024-09-26 at 7 38 31 PM
> that's because loggers are only ever created in leaf modules (ie. there is no code in `__init__.py` files except for abstract classes that are never instantiated directly) - the message from `gui.widgets.terminal` doesn't propagate to `gui.widgets` and `gui` because there are no loggers for those modules.

We actually agree on this point. The function shouldn't be called in a module and its parent module. If we are going to implement a similar function here, may be we can come up with a function name that's more informative of that behavior (but this can be just me being fastidious).

> ime file logging is super inexpensive, especially for a non-perf-forward package like linkml, and writing the same number of log messages to multiple files has not been meaningfully more expensive than writing them to one file for normal logging rates. 

I didn't know about this. Good to know. 🙂

> i brought it up just as an example, we can do whatever we want here. this logger was for a very different context, where you had ~dozens of instances of classes from a single module running simultaneously over long time periods in multiple processes (where you need separate files because they want to clobber each other), so probably not exactly what we want for linkml. i was more suggesting the idea of having a function that wraps logger initiation that does `_whatever we want_`. no need to go so hard on the poor autopilot logger.

Sorry that I misunderstood what you were suggesting.

> > I feel that the implementation of the centralize config and the init_logger function can be done in different PRs and require consensus from the LinkML team, particularly the centralize config. The init_logger function can be built on the changes in this PR after it is merged for this PR only sets up loggers without any handlers.
> 
> sure, i was just pitching an idea, i'm not a core maintainer so i certainly wouldn't hold up this (or any) PR with anything but suggestions that can be ignored.

Your suggestions were never ignored. I learned a lot from them. Thank you for your input to my other postings/questions as well. However, I do want to have the changes in this PR merged though if possible, and I don't know who to ping. Let's continue the discussion regarding a centralized config and logger initializer in https://github.com/linkml/linkml/issues/2331. Meanwhile, let's enjoy the best time of the year in LA after that terrible heat wave earlier this month 😊.

cmungall (approved):

cmungall:
We'll merge this now as it's massively better than the current state, thank you all!

We can definitely carry on the discussion about adding more configurability as a separate discussion/comments

PR #2329, opened 2024-09-27
fix: generated python with ifabsent

### LINKED ISSUE #2284: Validation fails on "ifabsent" for enumeration
**Describe the bug**

Validation of a schema with an [enumeration permissible value](https://linkml.io/linkml-model/latest/docs/PermissibleValue/) as default ([`ifabsent`](https://linkml.io/linkml-model/latest/docs/ifabsent/)) fails.

I run the validation as part of the Python generation: `gen-python --validate <schema-path>`.

The error that I get is consistently the same:
```shell
ERROR:root:Error compiling generated python code: name '<EnumClass>' is not defined
NameError: name '<EnumClass>' is not defined. Did you mean: 'PermissibleValue'?
```

The generated Python code looks fine.

**Version of LinkML you are using** v1.8.3

Reproduced with
* one of the test schemas: [kitchen_sink_ifabsent.yaml](https://github.com/linkml/linkml/blob/main/tests/test_generators/input/kitchen_sink_ifabsent.yaml)
* following very simple schema:
```yaml
id: https://example.org/test-ifabsent
name: Test-ifabsent
prefixes:
imports:
  - linkml:types
default_range: string

classes:
  MyClass:
    attributes:
      my_slot:
        range: PossibleValues
        ifabsent: PossibleValues(first_option)

enums:
  PossibleValues:
    permissible_values:
      first_option:
      second_option:
```

Silvanoc:
Generated Python code when `ifabsent` specifies an enumeration permissible value was erroneous. This patch fixes it.

Fixes #2284

### DIFF
--- a/linkml/generators/python/python_ifabsent_processor.py
+++ b/linkml/generators/python/python_ifabsent_processor.py
@@ -10,6 +10,8 @@ from linkml.generators.common.ifabsent_processor import IfAbsentProcessor
 class PythonIfAbsentProcessor(IfAbsentProcessor):
     UNIMPLEMENTED_DEFAULT_VALUES = ["class_curie", "class_uri", "slot_uri", "slot_curie", "default_range", "default_ns"]
 
+    enum_forward_references = {}
+
     def map_custom_default_values(self, default_value: str, slot: SlotDefinition, cls: ClassDefinition) -> (bool, str):
         if default_value in self.UNIMPLEMENTED_DEFAULT_VALUES:
             return True, None
@@ -68,10 +70,22 @@ class PythonIfAbsentProcessor(IfAbsentProcessor):
     def map_uri_default_value(self, default_value: str, slot: SlotDefinition, cls: ClassDefinition):
         return self._uri_for(default_value)
 
+    def enum_forward_reference(self, cls_name: str, slot_name: str):
+        if (
+            cls_name in self.enum_forward_references.keys()
+            and slot_name in self.enum_forward_references[cls_name].keys()
+        ):
+            return self.enum_forward_references[cls_name][slot_name]
+        else:
+            return None
+
     def map_enum_default_value(
         self, enum_name: EnumDefinitionName, permissible_value_name: str, slot: SlotDefinition, cls: ClassDefinition
     ):
-        return f"{enum_name}.{permissible_value_name}"
+        if cls.name not in self.enum_forward_references.keys():
+            self.enum_forward_references[cls.name] = {}
+        self.enum_forward_references[cls.name][slot.name] = f"{enum_name}.{permissible_value_name}"
+        return "field(default=None)"
 
     def map_nc_name_default_value(self, default_value: str, slot: SlotDefinition, cls: ClassDefinition):
         raise NotImplementedError()
--- a/linkml/generators/python/python_ifabsent_processor.py
+++ b/linkml/generators/python/python_ifabsent_processor.py
@@ -10,8 +10,6 @@ from linkml.generators.common.ifabsent_processor import IfAbsentProcessor
 class PythonIfAbsentProcessor(IfAbsentProcessor):
     UNIMPLEMENTED_DEFAULT_VALUES = ["class_curie", "class_uri", "slot_uri", "slot_curie", "default_range", "default_ns"]
 
-    enum_forward_references = {}
-
     def map_custom_default_values(self, default_value: str, slot: SlotDefinition, cls: ClassDefinition) -> (bool, str):
         if default_value in self.UNIMPLEMENTED_DEFAULT_VALUES:
             return True, None
@@ -70,22 +68,10 @@ class PythonIfAbsentProcessor(IfAbsentProcessor):
     def map_uri_default_value(self, default_value: str, slot: SlotDefinition, cls: ClassDefinition):
         return self._uri_for(default_value)
 
-    def enum_forward_reference(self, cls_name: str, slot_name: str):
-        if (
-            cls_name in self.enum_forward_references.keys()
-            and slot_name in self.enum_forward_references[cls_name].keys()
-        ):
-            return self.enum_forward_references[cls_name][slot_name]
-        else:
-            return None
-
     def map_enum_default_value(
         self, enum_name: EnumDefinitionName, permissible_value_name: str, slot: SlotDefinition, cls: ClassDefinition
     ):
-        if cls.name not in self.enum_forward_references.keys():
-            self.enum_forward_references[cls.name] = {}
-        self.enum_forward_references[cls.name][slot.name] = f"{enum_name}.{permissible_value_name}"
-        return "field(default=None)"
+        return f"'{enum_name}.{permissible_value_name}'"
 
     def map_nc_name_default_value(self, default_value: str, slot: SlotDefinition, cls: ClassDefinition):
         raise NotImplementedError()
--- a/linkml/generators/python/python_ifabsent_processor.py
+++ b/linkml/generators/python/python_ifabsent_processor.py
@@ -71,7 +71,7 @@ class PythonIfAbsentProcessor(IfAbsentProcessor):
     def map_enum_default_value(
         self, enum_name: EnumDefinitionName, permissible_value_name: str, slot: SlotDefinition, cls: ClassDefinition
     ):
-        return f"'{enum_name}.{permissible_value_name}'"
+        return f"'{permissible_value_name}'"
 
     def map_nc_name_default_value(self, default_value: str, slot: SlotDefinition, cls: ClassDefinition):
         raise NotImplementedError()
--- a/linkml/generators/pythongen.py
+++ b/linkml/generators/pythongen.py
@@ -29,6 +29,7 @@ from rdflib import URIRef
 
 import linkml
 from linkml._version import __version__
+from linkml.generators.pydanticgen.template import Import, Imports, ObjectImport
 from linkml.generators.python.python_ifabsent_processor import PythonIfAbsentProcessor
 from linkml.utils.generator import Generator, shared_arguments
 
@@ -127,13 +128,120 @@ class PythonGenerator(Generator):
                 self.emit_prefixes.add(type_prefix)
 
     def gen_schema(self) -> str:
-        # The metamodel uses Enumerations to define itself, so don't import if we are generating the metamodel
-        enumimports = (
-            ""
-            if self.genmeta
-            else "from linkml_runtime.linkml_model.meta import EnumDefinition, PermissibleValue, PvFormulaOptions\n"
+        all_imports = Imports()
+        # generic imports
+        all_imports = (
+            all_imports
+            + Import(module="dataclasses")
+            + Import(module="re")
+            + Import(
+                module="jsonasobj2",
+                objects=[
+                    ObjectImport(name="JsonObj"),
+                    ObjectImport(name="as_dict"),
+                ],
+            )
+            + Import(
+                module="typing",
+                objects=[
+                    ObjectImport(name="Optional"),
+                    ObjectImport(name="List"),
+                    ObjectImport(name="Union"),
+                    ObjectImport(name="Dict"),
+                    ObjectImport(name="ClassVar"),
+                    ObjectImport(name="Any"),
+                ],
+            )
+            + Import(
+                module="dataclasses",
+                objects=[
+                    ObjectImport(name="dataclass"),
+                ],
+            )
+            + Import(
+                module="datetime",
+                objects=[
+                    ObjectImport(name="date"),
+                    ObjectImport(name="datetime"),
+                    ObjectImport(name="time"),
+                ],
+            )
         )
-        handlerimport = "from linkml_runtime.utils.enumerations import EnumDefinitionImpl"
+
+        # The metamodel uses Enumerations to define itself, so don't import if we are generating the metamodel
+        if not self.genmeta:
+            all_imports = all_imports + Import(
+                module="linkml_runtime.linkml_model.meta",
+                objects=[
+                    ObjectImport(name="EnumDefinition"),
+                    ObjectImport(name="PermissibleValue"),
+                    ObjectImport(name="PvFormulaOptions"),
+                ],
+            )
+        # linkml imports
+        all_imports = (
+            all_imports
+            + Import(
+                module="linkml_runtime.utils.slot",
+                objects=[
+                    ObjectImport(name="Slot"),
+                ],
+            )
+            + Import(
+                module="linkml_runtime.utils.metamodelcore",
+                objects=[
+                    ObjectImport(name="empty_list"),
+                    ObjectImport(name="empty_dict"),
+                    ObjectImport(name="bnode"),
+                ],
+            )
+            + Import(
+                module="linkml_runtime.utils.yamlutils",
+                objects=[
+                    ObjectImport(name="YAMLRoot"),
+                    ObjectImport(name="extended_str"),
+                    ObjectImport(name="extended_float"),
+                    ObjectImport(name="extended_int"),
+                ],
+            )
+            + Import(
+                module="linkml_runtime.utils.dataclass_extensions_376",
+                objects=[
+                    ObjectImport(name="dataclasses_init_fn_with_kwargs"),
+                ],
+            )
+            + Import(
+                module="linkml_runtime.utils.formatutils",
+                objects=[
+                    ObjectImport(name="camelcase"),
+                    ObjectImport(name="underscore"),
+                    ObjectImport(name="sfx"),
+                ],
+            )
+        )
+
+        # handler import
+        all_imports = all_imports + Import(
+            module="linkml_runtime.utils.enumerations", objects=[ObjectImport(name="EnumDefinitionImpl")]
+        )
+        # other imports
+        all_imports = (
+            all_imports
+            + Import(
+                module="rdflib",
+                objects=[
+                    ObjectImport(name="Namespace"),
+                    ObjectImport(name="URIRef"),
+                ],
+            )
+            + Import(
+                module="linkml_runtime.utils.curienamespace",
+                objects=[
+                    ObjectImport(name="CurieNamespace"),
+                ],
+            )
+        )
+
         split_description = ""
         if self.schema.description:
             split_description = "\n#   ".join(d for d in self.schema.description.split("\n") if d is not None)
@@ -151,21 +259,7 @@ class PythonGenerator(Generator):
 # description: {split_description}
 # license: {be(self.schema.license)}
 
-import dataclasses
-import re
-from jsonasobj2 import JsonObj, as_dict
-from typing import Optional, List, Union, Dict, ClassVar, Any
-from dataclasses import dataclass
-from datetime import date, datetime, time
-{enumimports}
-from linkml_runtime.utils.slot import Slot
-from linkml_runtime.utils.metamodelcore import empty_list, empty_dict, bnode
-from linkml_runtime.utils.yamlutils import YAMLRoot, extended_str, extended_float, extended_int
-from linkml_runtime.utils.dataclass_extensions_376 import dataclasses_init_fn_with_kwargs
-from linkml_runtime.utils.formatutils import camelcase, underscore, sfx
-{handlerimport}
-from rdflib import Namespace, URIRef
-from linkml_runtime.utils.curienamespace import CurieNamespace
+{all_imports.render()}
 {self.gen_imports()}
 
 metamodel_version = "{self.schema.metamodel_version}"
--- a/linkml/generators/pythongen.py
+++ b/linkml/generators/pythongen.py
@@ -936,9 +936,8 @@ dataclasses._init_fn = dataclasses_init_fn_with_kwargs
                     # mutable. `python_ifabsent_processor.py` can specify
                     # the default as string and here that string gets
                     # converted into an object attribute invocation
-                    rlines.append(f"\tenum_name, pv_name = self.{aliased_slot_name}.split('.')")
-                    rlines.append("\tenum_cls = globals().get(enum_name)")
-                    rlines.append(f"\tself.{aliased_slot_name} = getattr(enum_cls, pv_name)")
+                    # TODO: fix according https://github.com/linkml/linkml/pull/2329#discussion_r1797534588
+                    rlines.append(f"\tself.{aliased_slot_name} = getattr({slot.range}, self.{aliased_slot_name})")
                 elif (
                     (self.class_identifier(slot.range) and not slot.inlined)
                     or slot.range in self.schema.types
--- a/linkml/generators/pythongen.py
+++ b/linkml/generators/pythongen.py
@@ -127,17 +127,6 @@ class PythonGenerator(Generator):
             if type_prefix:
                 self.emit_prefixes.add(type_prefix)
 
-    def _import_field(self):
-        enum_pvs = []
-        for enum_name, enum_def in self.schema.enums.items():
-            for pv in enum_def.permissible_values:
-                enum_pvs.append(f"{enum_name}({pv})")
-        for _, class_def in self.schema.classes.items():
-            for slot in class_def.slots:
-                if self.schema.slots[slot].ifabsent in enum_pvs:
-                    return True
-        return False
-
     def gen_schema(self) -> str:
         all_imports = Imports()
         # generic imports
@@ -252,9 +241,6 @@ class PythonGenerator(Generator):
                 ],
             )
         )
-
-        if self._import_field():
-            all_imports = all_imports + Import(module="dataclasses", objects=[ObjectImport(name="field")])
 
         split_description = ""
         if self.schema.description:
@@ -917,16 +903,10 @@ dataclasses._init_fn = dataclasses_init_fn_with_kwargs
         # Generate the type co-ercion for the various types.
         # NOTE: if you set this to true, we will cast all types.   This may be what we really want
         if not slot.multivalued:
-            iap = self.ifabsent_processor
             if slot.designates_type:
                 pass
             elif slot.required:
                 rlines.append(f"if not isinstance(self.{aliased_slot_name}, {base_type_name}):")
-            elif slot.range in self.schema.enums and iap.enum_forward_reference(cls.name, slot.name):
-                rlines.append(
-                    f"if self.{aliased_slot_name} is not field(default=None) and "
-                    f"not isinstance(self.{aliased_slot_name}, {base_type_name}):"
-                )
             else:
                 rlines.append(
                     f"if self.{aliased_slot_name} is not None and "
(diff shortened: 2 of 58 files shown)


Silvanoc:
Regression test added reproducing the issue with commit d576720b65cdc55bc659428c9655ff63850ea546 see [failing run](https://github.com/linkml/linkml/actions/runs/11071303849) before pushing the fix.

Silvanoc:
@pkalita-lbl I am leaving this PR as Draft because I am not 100% sure that the fix is correct.
Although the tests are passing and I get the expected output. Please take a detailed look at it and let me know if you find the fix is correct (to mark the PR as ready) or not (in this case please help me to fix it).

linkml/generators/pythongen.py
```diff
@@ -131,6 +131,14 @@ def gen_schema(self) -> str:
...
             else "from linkml_runtime.linkml_model.meta import EnumDefinition, PermissibleValue, PvFormulaOptions\n"
         )
+        enum_pvs = []
+        for enum_name, enum_def in self.schema.enums.items():
+            for pv in enum_def.permissible_values:
+                enum_pvs.append(f"{enum_name}({pv})")
+        for class_name, class_def in self.schema.classes.items():
+            for slot in class_def.slots:
+                if self.schema.slots[slot].ifabsent in enum_pvs:
+                    enumimports = enumimports + "from dataclasses import field\n"
```
pkalita-lbl, line 141: If multiple classes/slots satisfy this condition, I think this would result in multiple identical imports?

linkml/generators/python/python_ifabsent_processor.py
```diff
@@ -71,7 +71,7 @@ def map_uri_default_value(self, default_value: str, slot: SlotDefinition, cls: C
     def map_enum_default_value(
         self, enum_name: EnumDefinitionName, permissible_value_name: str, slot: SlotDefinition, cls: ClassDefinition
     ):
-        return f"{enum_name}.{permissible_value_name}"
+        return f"field(default_factory={enum_name}.{permissible_value_name})"
```
pkalita-lbl, line 74: I'm not quite sure I understand the need for using `field` with `default_factory` here. Can you elaborate?

pkalita-lbl (commented):
Thanks @Silvanoc! I think having the enum definitions before the class definitions makes sense. A couple questions about the other changes.

linkml/generators/pythongen.py
sneakers-the-rat: pydanticgen `Import` and `Imports` classes handle deduplication and sorting of imports (and are not pydantic specific) https://github.com/linkml/linkml/blob/68a880a458708c20cde6337b3493ee93a85c3ba8/linkml/generators/pydanticgen/template.py#L263

linkml/generators/python/python_ifabsent_processor.py
sneakers-the-rat: this would break pydanticgen (uses `Field` not `field`) and throws an error

```python
from dataclasses import dataclass, field
from enum import Enum
class MyEnum(Enum):
    value = 'value'
    
@dataclass
class MyClass():
    afield: MyEnum = field(default_factory=MyEnum.value)
    
MyClass()
  File "/Users/jonny/Library/Caches/pypoetry/virtualenvs/linkml-G-1YJs6r-py3.11/lib/python3.11/site-packages/IPython/core/interactiveshell.py", line 3508, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-14-a39b5d23b52a>", line 1, in <module>
    MyClass()
  File "<string>", line 3, in __init__
TypeError: 'MyEnum' object is not callable
```

Silvanoc: I am not an expert in python dataclasses, therefore I haven't found a better fix for following error that I get if `field(default_factory())` is not used.
``` shell
ValueError: mutable default <class 'linkml_runtime.linkml_model.meta.PermissibleValue'> for field ifabsent_not_literal is not allowed: use default_factory
```

You might know a better way to fix that issue without breaking pydanticgen.

BTW @sneakers-the-rat tests are passing, are we missing some tests on pydanticgen? Or how did you catch it?

linkml/generators/pythongen.py
Silvanoc: I have used the `Import` and `Imports` classes as proposed by @sneakers-the-rat. They are really cool!

I have not changed the rest of the imports to keep this PR scoped. But IMO a future PR should take care of using those same classes all the imports in the generated python code.

Silvanoc: BTW @pkalita-lbl good catch!

Silvanoc:
I am trying to declare a "derived" meta-model (one derived from the LinkML meta-model) and I am stumbling upon the same issue in `shaclgen`.

I don't really understand why the range of the slot `range` is `enum_definition`, in any case I am getting an error again because `ifabsent` is failing on resolving the default value. I have not tried yet, if my patch resolves `shaclgen` too.

Silvanoc:
> I am trying to declare a "derived" meta-model (one derived from the LinkML meta-model) and I am stumbling upon the same issue in `shaclgen`.
> 
> I don't really understand why the range of the slot `range` is `enum_definition`, in any case I am getting an error again because `ifabsent` is failing on resolving the default value. I have not tried yet, if my patch resolves `shaclgen` too.

This issue can be also easily reproduced installing the latest LinkML version (v1.8.4), getting the latest LinkML meta-model and running `gen-shacl linkml_model/model/schema/meta.yaml`.

This is the error:
``` shell
ValueError: The ifabsent value `default_range` of the `range` slot could not be processed
```

I have run it with the debugger and the issue has the same cause: `ifabsent` failing on enumerations.

I have also applied my patch of this branch and it does not seem to fix it 😞 We probably need a better analysis of someone with deeper knowledge than me on the LinkML codebase and Python programming. @vincentkelleher since you wrote the new ifabsent code, you might be the right guy for this...

IMO this is a "hidden" issue that might have deeper consequences than thought!

Silvanoc:
I have taken my time to go the issue to its root-cause and I think I got it.

I do not know how test the generated Pydantic code like @sneakers-the-rat did for the 1st try. I have tried against the [kitchen_sink_ifabsent.yaml](https://github.com/linkml/linkml/blob/main/tests/test_generators/input/kitchen_sink_ifabsent.yaml) schema. But then I have realized that I get exactly the same error that you get with the current `main` branch. So it is not an issue I have introduced 😄 

I have also replaced the generation of the imports to use the same module being used by pydanticgen. The result would make `isort` proud...

I would recommend reviewing on the individual commits.

sneakers-the-rat:
Oh boy if you swapped the import system ill gladly review.

sierra-moxon:
Thank you @Silvanoc!  And for bringing this one to our attention on slack.  It is indeed breaking other model distributions as well.

linkml/generators/pythongen.py
```diff
@@ -126,14 +127,135 @@ def visit_type(self, typ: TypeDefinition) -> None:
...
+                ],
+            )
+        )
+
         # The metamodel uses Enumerations to define itself, so don't import if we are generating the metamodel
-        enumimports = (
-            ""
-            if self.genmeta
-            else "from linkml_runtime.linkml_model.meta import EnumDefinition, PermissibleValue, PvFormulaOptions\n"
+        if not self.genmeta:
```
sneakers-the-rat, line 183: eg. could do this

at a module level...

```python
DEFAULT_IMPORTS = Imports(imports=[
    Import(module="dataclasses", objects=[ObjectImport(name="dataclass")]),
    # ... the rest of the default imports
])

METAMODEL_IMPORTS = Imports(imports=[
    Import(module="linkml_runtime.linkml_model.meta", objects=[ObjectImport(name="EnumDefinition")]
    # ... etc.
])

# and any other import groups
```

and then in the generator you could do

```python
def gen_schema(self):
    imports = DEFAULT_IMPORTS.model_copy()
    if not self.genmeta:
        imports += METAMODEL_IMPORTS
```

linkml/generators/pythongen.py
```diff
@@ -126,14 +127,135 @@ def visit_type(self, typ: TypeDefinition) -> None:
...
+                module="linkml_runtime.utils.curienamespace",
+                objects=[
+                    ObjectImport(name="CurieNamespace"),
+                ],
+            )
         )
-        handlerimport = "from linkml_runtime.utils.enumerations import EnumDefinitionImpl"
+
+        if self._import_field():
+            all_imports = all_imports + Import(module="dataclasses", objects=[ObjectImport(name="field")])
```
sneakers-the-rat, line 257: might as well just always import this, since importing one object from a module means executing the whole module, this wouldn't be a perf hit. pydanticgen uses a system of build results to handle imports that are needed in specific cases and not others, but retrofitting that into the pythongen would be a lot of work.

linkml/generators/pythongen.py
```diff
@@ -841,7 +949,13 @@ def gen_postinit(self, cls: ClassDefinition, slot: SlotDefinition) -> Optional[s
                     or slot.range in self.schema.types
                     or slot.range in self.schema.enums
                 ):
-                    rlines.append(f"\tself.{aliased_slot_name} = {base_type_name}(self.{aliased_slot_name})")
+                    enum_forward_ref = self.ifabsent_processor.enum_forward_references
+                    if cls.name in enum_forward_ref.keys() and slot.range in enum_forward_ref[cls.name].keys():
+                        rlines.append(
```
sneakers-the-rat, line 954: sort of having a hard time telling what's happening here, but i'll see if running it and inspecting the output clears things up

sneakers-the-rat (commented):
So I am not sure of the status of this PR, if it's done or not, but i can't see a test for the underlying problem (an `ifabsent` value that is a value of an enum) here? the tests i do see check for the existence of something in the `enum_forward_references` dict, but not that the default result is correctly set in an instance of that model? i might be missing something. 

I don't mean to throw cold water on all the work that you've done here (i love converting the import system), but I think that the easiest fix here probably would have been to just put enums before classes in the rendered module, since as far as i know enums can't refer to other objects that need to be defined first but classes can.

tests/test_generators/test_pythongen.py
```diff
@@ -68,6 +68,10 @@ def test_multiline_stuff(input_path):
     )
 
 
+def test_ifabsent_stuff(input_path):
```
sneakers-the-rat, line 71: sorry should have added this to the previous review - what does this do?

tests/test_generators/python/test_python_ifabsent_processor.py
```diff
@@ -403,9 +403,13 @@ def test_enum_default_value():
             schema_view.all_slots()[SlotDefinitionName("presence")],
             schema_view.all_classes()[ClassDefinitionName("Student")],
         )
-        == "PresenceEnum.Missing"
+        == "field(default=None)"
     )
 
+    print(processor.enum_forward_references)
```
sneakers-the-rat, line 409: no print statements plz :)

sneakers-the-rat (commented):
sorry for the double review, pressed submit before i was done. not sure if there's a way to add more comments to a previous review but here we are

sneakers-the-rat:
> This is the error:
> 
> ValueError: The ifabsent value `default_range` of the `range` slot could not be processed
> 
> I have run it with the debugger and the issue has the same cause: ifabsent failing on enumerations.

I think this is a separate issue than the order of definition problem re: enums and classes in the pythongen, basically `default_range` is a special case in `ifabsent` that is currently [explicitly skipped](https://github.com/linkml/linkml/blame/c4379e5a121732b6adda21284bb34dde590743b1/linkml/generators/python/python_ifabsent_processor.py#L14) by the python ifabsent processor and unhandled by the shaclgen. would recommend we split that off into a separate issue/pr

sneakers-the-rat:
here i just went and fixed the other issue with `ifabsent` for the special cases over here: https://github.com/linkml/linkml/pull/2366

linkml/generators/pythongen.py
```diff
@@ -126,14 +127,135 @@ def visit_type(self, typ: TypeDefinition) -> None:
...
+            for pv in enum_def.permissible_values:
+                enum_pvs.append(f"{enum_name}({pv})")
+        for _, class_def in self.schema.classes.items():
+            for slot in class_def.slots:
+                if self.schema.slots[slot].ifabsent in enum_pvs:
+                    return True
+        return False
+
     def gen_schema(self) -> str:
+        all_imports = Imports()
```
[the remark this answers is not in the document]
Silvanoc: Sure, but we would need a different "default", since generated Python code don't need to import Pydantic (for example).

I would like to focus here on fixing #2284 and address this proposal in a separate PR, since it is only a refactoring (no change on external behavior).

linkml/generators/pythongen.py
Silvanoc: Fully agree here too. But refactoring that I would rather address in a separate PR to bring this one to an end.

linkml/generators/pythongen.py
Silvanoc: I agree that the complexity of `_import_field` might not be worth. You will only get linter warnings on the generated code, but you get anyway so many other warnings 😄 I will change it as you propose.

linkml/generators/pythongen.py
Silvanoc: This is probably the trickiest part of the whole patch. It is probably worth a more extensive comment explaining it.

tests/test_generators/test_pythongen.py
Silvanoc: It simply tries to generate the Python code corresponding to the schema and it fails without the patch. It is a sort of "poor-man" regression test.

tests/test_generators/python/test_python_ifabsent_processor.py
Silvanoc: Ups, late-hours slip. Will remove it. Thanks for the good catch!

linkml/generators/pythongen.py
vincentkelleher: I like @Silvanoc's approach on generating only the necessary imports but each point of view has its pros and cons 😇

linkml/generators/pythongen.py
vincentkelleher: From what I understand, in the case of a forward reference for an enum it will return something like:

```python
self.my_slot = MyType.MyEnum(MyPermissibleValue)
```

And in the case of a basic type field, it will return:

```python
self.my_slot = MyType(self.my_slot)
```

Did I understand that correctly ?

linkml/generators/pythongen.py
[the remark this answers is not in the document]
sneakers-the-rat: well, sure, of course it would be different lol. i'm just saying cut and paste this block outside of the method and put it at the module level, but whatever you do u

linkml/generators/pythongen.py
sneakers-the-rat: i think it would take approximately 10 seconds, but i don't use the python generator so it doesn't really matter that much to me

linkml/generators/pythongen.py
sneakers-the-rat: i mean yeah i designed a whole system of build result models with the primary purpose of only including necessary imports for pydanticgen, so of course i prefer doing that, just saying that gating with private methods just doesn't scale that well

linkml/generators/pythongen.py
Silvanoc: My experience is that you know when you start refactoring, but you never know when you will end 😉

And I have the feeling that my PRs tend to get much too big and extensive over time to be digestible (see also #2303, with 104 comments!). That is why I am trying to close this one before it gets into another [Berlin Brandenburg Airport](https://en.wikipedia.org/wiki/Construction_of_Berlin_Brandenburg_Airport).

linkml/generators/pythongen.py
Silvanoc: Although I appreciate @vincentkelleher support on my approach. I have to agree with @sneakers-the-rat that saving that `field` import is not worth the added complexity. So I will go for the always import `field` approach.

linkml/generators/pythongen.py
sneakers-the-rat: sure sure, do whatever, like i said i don't use the python generator so idrc <3

linkml/generators/pythongen.py
sneakers-the-rat: don't mean to be a downer, also am indifferent on this change, defer to the author :)

linkml/generators/pythongen.py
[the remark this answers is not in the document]
Silvanoc: Created #2367 to address this one.

linkml/generators/pythongen.py
Silvanoc: This should be also addresses on #2367.

linkml/generators/pythongen.py
Silvanoc: @vincentkelleher you are right, it should be `MyEnum.MyPermissibleValue`, I will check if it returns it or something wrong (like you seem to expect from my code).

Silvanoc: The code was in fact buggy, but not as @vincentkelleher thought.

And this confirms that @sneakers-the-rat was right to "complain" about the lack of a specific test
> i can't see a test for the underlying problem (an ifabsent value that is a value of an enum) here

My poor-man regression test `test_ifabsent_stuff` was only asserting that the module could be generated and compiled, not that the resulting code was the expected one... I am almost done with the bug fixing and the corresponding test for `ifabsent` on enumerations.

linkml/generators/pythongen.py
Silvanoc: I have revised my decision. Although I am not happy with the complexity added with `_import_field`, importing always `dataclasses.field` results in a cascade of snapshots needing updating. I have decided to leave `_import_field` there... At least for the time being. I will consider afterwards removing it.

linkml/generators/pythongen.py
Silvanoc: Now `test_ifabsent_stuff` also tests the correctness of the `ifabsent` in the case of enumerations.

Silvanoc:
I think I have addressed all open points and since this PR seems to be [breaking other schemas](https://github.com/linkml/linkml/pull/2329#issuecomment-2406185641), I am triggering all previous reviewers to have a last look at it.

I have also assigned this PR (apparently that is not the normal workflow in the project) just to signal that I assume that @pkalita-lbl will merge or reassign to someone "permitted" to merge once happy with the reached status.

Silvanoc:
I have been able to simplify the code even more by using a string as default for the field instead of using `field(default=None)`. I can also convey the information about which is the permissible value to be taken by default in that string. The `__post_init__` only needs to convert that string into a call to the real permissible value.

tests/test_generators/python/test_python_ifabsent_processor.py
```diff
@@ -374,6 +374,44 @@ def test_uri_default_value():
     )
 
 
+# If this function does not get fixture scoped to the module the
+# PythonIfAbsentProcessor instance suddenly has some attributes
+# coming from tests using the kitchen-sink example! Invalidating that
+# way the results of the test.
+@pytest.fixture(scope="module")
```
sneakers-the-rat, line 381: this is a test, right? so it shouldn't be a fixture at all? I don't see how it interacts with the kitchen sink example here.

in any case it being a fixture meant it is not run as a test, note its absence (i did check out this branch and am looking at the code being there and it's in editable mode so i can cause tests to fail by throwing a spurious exception, so it's not my environment, just how fixtures work.

```
collected 15 items                                                                                            

tests/test_generators/python/test_python_ifabsent_processor.py::test_unimplemented_default_value[class_curie] PASSED [  6%]
tests/test_generators/python/test_python_ifabsent_processor.py::test_unimplemented_default_value[class_uri] PASSED [ 13%]
tests/test_generators/python/test_python_ifabsent_processor.py::test_unimplemented_default_value[slot_uri] PASSED [ 20%]
tests/test_generators/python/test_python_ifabsent_processor.py::test_unimplemented_default_value[slot_curie] PASSED [ 26%]
tests/test_generators/python/test_python_ifabsent_processor.py::test_unimplemented_default_value[default_range] PASSED [ 33%]
tests/test_generators/python/test_python_ifabsent_processor.py::test_unimplemented_default_value[default_ns] PASSED [ 40%]
tests/test_generators/python/test_python_ifabsent_processor.py::test_string_default_value PASSED        [ 46%]
tests/test_generators/python/test_python_ifabsent_processor.py::test_boolean_default_value PASSED       [ 53%]
tests/test_generators/python/test_python_ifabsent_processor.py::test_numeric_default_value PASSED       [ 60%]
tests/test_generators/python/test_python_ifabsent_processor.py::test_time_default_value PASSED          [ 66%]
tests/test_generators/python/test_python_ifabsent_processor.py::test_date_default_value PASSED          [ 73%]
tests/test_generators/python/test_python_ifabsent_processor.py::test_datetime_default_value PASSED      [ 80%]
tests/test_generators/python/test_python_ifabsent_processor.py::test_uri_default_value PASSED           [ 86%]
tests/test_generators/python/test_python_ifabsent_processor.py::test_enum_default_value PASSED          [ 93%]
tests/test_generators/python/test_python_ifabsent_processor.py::test_bnode_default_value PASSED         [100%]

============================================= 15 passed in 0.77s ==============================================
```

sneakers-the-rat: If i add a `pdb.set_trace()` call below the assignment and print it, i get what would be expected:

```yaml

id: ifabsent_tests
name: ifabsent_tests

prefixes:
  ex: https://example.org/
default_prefix: ex

classes:
  Student:
    attributes:

      - name: presence
        range: PresenceEnum
      - name: invalidPresence
        range: PresenceEnum

enums:
  PresenceEnum:
    permissible_values:
      Present:
        description: It's there.
      Missing:
        description: It's not there.
```

tests/test_generators/python/test_python_ifabsent_processor.py
```diff
@@ -403,7 +441,7 @@ def test_enum_default_value():
             schema_view.all_slots()[SlotDefinitionName("presence")],
             schema_view.all_classes()[ClassDefinitionName("Student")],
         )
-        == "PresenceEnum.Missing"
+        == "'PresenceEnum.Missing'"
```
sneakers-the-rat, line 444: Great! I like the string approach more than putting `None` there :) and testing this out it looks like it's working.

tests/test_generators/test_pythongen.py
```diff
@@ -68,6 +68,11 @@ def test_multiline_stuff(input_path):
...
```
sneakers-the-rat, line 71: also great! this is the correct test. Ideally it would be named something that is a bit more informative about what it tests, like `test_enum_permissiblevalue_ifabsent` , but whatever, it is in fact testing `ifabsent_stuff`.

linkml/generators/pythongen.py
```diff
@@ -836,7 +930,16 @@ def gen_postinit(self, cls: ClassDefinition, slot: SlotDefinition) -> Optional[s
             ):
                 rlines.append(f"\tself.{aliased_slot_name} = {base_type_name}()")
             else:
-                if (
+                if slot.range in self.schema.enums and slot.ifabsent:
+                    # `ifabsent` for an enumeration cannot be assigned to
```
sneakers-the-rat, line 934: appreciate this documentation

tests/test_generators/python/test_python_ifabsent_processor.py
```diff
@@ -374,6 +374,44 @@ def test_uri_default_value():
...
+
+    assert (
+        processor.process_slot(
+            schema_view.all_slots()[SlotDefinitionName("presence")],
+            schema_view.all_classes()[ClassDefinitionName("Student")],
+        )
+        is None
+    )
+
+    assert processor.enum_forward_references == {}
```
sneakers-the-rat, line 412: the test fails here because the processor doesn't have `enum_forward_references` - i think you removed that in the latest update?

tests/test_generators/python/test_python_ifabsent_processor.py
sneakers-the-rat, line 381:
```suggestion
```

Seems to work to me except for the assert on the forward references dict! and shouldn't be a fixture anyway :)

linkml/generators/pythongen.py
```diff
@@ -836,7 +930,16 @@ def gen_postinit(self, cls: ClassDefinition, slot: SlotDefinition) -> Optional[s
...
+                    # the dataclass field default, because it would be a
+                    # mutable. `python_ifabsent_processor.py` can specify
+                    # the default as string and here that string gets
+                    # converted into an object attribute invocation
+                    rlines.append(f"\tenum_name, pv_name = self.{aliased_slot_name}.split('.')")
+                    rlines.append("\tenum_cls = globals().get(enum_name)")
+                    rlines.append(f"\tself.{aliased_slot_name} = getattr(enum_cls, pv_name)")
```
sneakers-the-rat, line 941: OK i like the string approach, and as usual am truly not trying to be a nitpicker here, but here's an example that fails because names on enums aren't sanitized:

```yaml
CordialnessEnum:
  permissible_values:
    # ...
    ch.aotic:
      description: doesn't obey conventions

# ...

IfAbsent:
  attributes:
    ifabsent_dotted_literal:
      range: CordialnessEnum
      ifabsent: CordialnessEnum(ch.aotic)
  
```

Then we get 

```python
@dataclass(repr=False)
class IfAbsent(YAMLRoot):
    # ....
    ifabsent_dotted_literal: Optional[Union[str, "CordialnessEnum"]] = 'CordialnessEnum.ch.aotic'

    def __post_init__(self):
        # ...
        if self.ifabsent_dotted_literal is not None and not isinstance(self.ifabsent_dotted_literal, CordialnessEnum):
            enum_name, pv_name = self.ifabsent_dotted_literal.split('.')
            enum_cls = globals().get(enum_name)
            self.ifabsent_dotted_literal = getattr(enum_cls, pv_name)
```

which fails with the relatively uninformative error message: 

```
(Pdb) ksm.IfAbsent()
*** ValueError: too many values to unpack (expected 2)
```

I think that the easiest thing to do here would be to mimic the behavior of [`StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum) and just put the name of the enum value rather than the Enum and the value

So then that simplifies to 


```python
@dataclass(repr=False)
class IfAbsent(YAMLRoot):
    # ....
    ifabsent_dotted_literal: Optional[Union[str, "CordialnessEnum"]] = 'ch.aotic'

    def __post_init__(self):
        # ...
        if self.ifabsent_dotted_literal is not None and not isinstance(self.ifabsent_dotted_literal, CordialnessEnum):
            self.ifabsent_dotted_literal = CordialnessEnum(self.ifabsent_dotted_literal)
```

which, remarkably, works.

```python
(Pdb) ksm.CordialnessEnum('ch.aotic')
CordialnessEnum(text='ch.aotic', description="doesn't obey convention")
```

sneakers-the-rat (commented):
Great! I like the string approach a lot.

There's one change that i think should be made (removing the `@pytest.fixture` from the test in the ifabsent tests and fixing the check on the missing forward references dict), and I think you could simplify it further by just using the name of the enum value rather than the whole `Enum.value` and parsing the string, but other than that i like what ya done and think it is good.

tests/test_generators/test_pythongen.py
Silvanoc: It tests 2 things: that the resulting Python code (otherwise `make_python` would fail) can be compiled and test_enum_permissiblevalue_ifabsent. That is why I left that name. I will add some comments and rename it.

tests/test_generators/python/test_python_ifabsent_processor.py
Silvanoc: Late hour error, will fix it.

tests/test_generators/python/test_python_ifabsent_processor.py
Silvanoc: This was a desperate late-hours try to fix an issue that was driving me nuts without a logical explanation I could find.  Apparently my laptop only needed a nap, since I cannot reproduce the issue anymore 🤷

Silvanoc: Removing it, since the issue somehow magically has disappeared... And the fix was non-sense nevertheless.

linkml/generators/pythongen.py
Silvanoc: Good catch! You are fully right, we do not need the enum name, since the range already constraints it. So only passing the permissible value suffice.

Silvanoc: BTW ensuring that corner cases are properly covered is IMO not nitpicking, but high quality code 😉

vincentkelleher (approved):

vincentkelleher:
LGTM 👍 

@Silvanoc you put in a lot of effort on this one, thank you very much 🙏 

I see that you have the same Kroki.io issues as I have... 😮‍💨 I saw you've added a discussion about it too.

image

Silvanoc:
I suppose that kroki.io is limiting the calls somehow and some of the tests triggered after rebasing on current `main` are now failing 🤦

sneakers-the-rat (approved):
hell ya sorry i missed that you did this three whole days ago this rocks thanks for putting up with my garbage this is great!!!!!!!!!!!

Silvanoc:
If I am not supposed to merge #2373 straight away to get rid of the RCOD (Red Cross of Death)[^1] in this PR, could please someone change the permissions so that I can't?! The temptation is killing me slowly...

[^1]: The small brother of the [Windows BSOD](https://www.techtarget.com/searchwindowsserver/definition/blue-screen-of-death-BSOD).

Silvanoc:
Now it should be ready for merging!

PR #2335, opened 2024-10-01
Global Config

### LINKED ISSUE #2331: A centralized config for LinkML
Can we have a centralized config for LinkML? Particularly, we are interested in having logging configs specified, as a component, in this centralized config so that logging are done consistently across different components of LinkML. Currently, it seems that different components of LinkML configure logging independently as:

```console
▶ git -P grep basicConfig
linkml/utils/cli_utils.py:        logging.basicConfig(level=log_level_int)
linkml/utils/execute_tutorial.py:    logging.basicConfig(level=logging.INFO)
linkml/utils/generator.py:            logging.basicConfig(level=logging.DEBUG, force=True)
linkml/utils/generator.py:            logging.basicConfig(level=logging.INFO, force=True)
linkml/utils/sqlutils.py:        logging.basicConfig(level=logging.DEBUG)
linkml/utils/sqlutils.py:        logging.basicConfig(level=logging.INFO)
linkml/utils/sqlutils.py:        logging.basicConfig(level=logging.WARNING)
linkml/utils/sqlutils.py:        logging.basicConfig(level=logging.ERROR)
tests/test_utils/test_generator.py:        logging.basicConfig()
tests/test_utils/test_generator.py:        logging.basicConfig()
tests/test_utils/test_schemaloader.py:    logging.basicConfig()
tests/utils/test_environment.py:        logging.basicConfig()
```

With a centralized logging config. We can have formatters and handlers specified with default while taking in command line input to override the configs in the centralized config. This config should also allow addition of components other than the config component for logging.

• Medium - can do work without it; but it's important (saves time, more consistent thus predictable)

**Additional context**

Please following the posting starting at https://github.com/linkml/linkml/pull/2323#issuecomment-2375284558 to show how this feature request originated. Note: The use of "pydantic-settings" for implementing to feature is preferred.

sneakers-the-rat:
Fix: https://github.com/linkml/linkml/issues/2331

alright here's a global config.

At the moement it's not hooked up into anything, but if we like this then i can continue the lovely work in https://github.com/linkml/linkml/pull/2323 and make it control log levels and whatnot as the config object suggests should be possible.

It should be all spelled out in the docs, but just for the good of the order here's the impl:

## Motivation

We want to be able to configure logging at first, but generally have a tidy way to specify behavior that might need to be customized in a per-system, per-session, and per-project way.

This is probably extremely overengineered, but i'm also laying groundwork for [`linkml build`](https://github.com/linkml/linkml/pull/2263) with a project-level config that can get sourced from one of potentially several places in several formats, so that's why. In the future i would anticipate additional global config stuff including whether we should try to resolve schema via url, location of cache directories, etc. so this sets us up for any of that.

## Implementation

Use `pydantic-settings`. 

Specifically, create a `GlobalConfig` class that sources from the following locations, in order of priority from high to low, where higher priority sources override lower priority

* in the arguments passed to the class constructor (not user configurable)
* in environment variables like ``export LINKML_LOG_DIR=~/``
* in a ``.env`` file in the working directory
* in a ``linkml_config.yaml`` file in the working directory
* in the ``tool.linkml.config`` table in a ``pyproject.toml`` file in the working directory
* in the global ``linkml_config.yaml`` file in the platform-specific data directory
  (use ``linkml config get config_file`` to find its location)
* the default values in the `GlobalConfig` model

There is an example of each kind of configuration in the docs in a tab group.

We can use nested models here, and imo we should do that to keep config reasonable instead of turning into a junk drawer, so I added an initial set of config options for handling stream and file logging to demo that. Descriptions of how to handle nested models are also in the docs.

I also added some cli commands to get and set global config values. values are printed as yaml so they can be re-parsed if needed (i did when writing tests)

```shell
$ linkml config get

user_dir: /Users/jonny/Library/Application Support/linkml
config_file: /Users/jonny/Library/Application Support/linkml/linkml_config.yaml
log:
  dir: /Users/jonny/Library/Logs/linkml
  file_name: linkml.log
  level: WARNING
  # ...
```

```shell
$ linkml config get log.level
WARNING
```

Setting values handles `None` to delete keys, nested models, and does model validation as well. We shouldn't be trying to set nonscalar/anything that doesn't have an unambiguous representation in a shell in the config, so hopefully that's as complex as we need it to be.

```shell
$ linkml config set log.level WARNING
Updated linkml config:
key: log.level
value: WARNING
config file: /Users/jonny/Library/Application Support/linkml/linkml_config.yaml
```

## Caveats

There's bound to be a bit of stickiness where people want to specify where the global config file lives. There's a sort of chicken and egg problem here, where we want the global config to be the lowest priority source of config, but then we should have read it and applied it before we reach the shell/cwd-based config options that might want to override it. That might need more consideration in the future, but for now i'm assuming that people will just use the local config options rather than trying to reposition the global config

There's also probably going to be some confusion like "what the heck i called `linkml config set <key> <value>`" and it didn't set my config! when people have some conflicting config file in the cwd. I didn't want to rig up a whole `linkml config set global <key> <value>` like eg. `pyenv` does at the moment, but that might be what we need to do if that starts happening. Tried to flag that clearly in the docs but we'll see.

## Re: Deps

I added two required deps:
- `pydantic-settings`
- `platformdirs`

`pydantic-settings` has one additional non-pydantic dep, `python-dotenv`, which is pretty light, and `platformdirs` has no additional deps and is extremely common (it's probably already in our dep tree somewhere).

imo it's worth it - since we're already installing pydantic, `pydantic-settings` doesn't add a lot more.

I also moved some docs deps that were hanging out in the `dev` group down into the `docs` group and added a package to show pydantic models cleanly (because the config models looked like total garbage without it)

## Tests

Added the basic tests that i thought we might want, but lmk if i missed something.

what ya think?

### DIFF
--- a/docs/cli/config.md
+++ b/docs/cli/config.md
@@ -1,0 +1,24 @@
+{#linkml-config-cli}
+# `linkml config`
+
+```{eval-rst} 
+.. click:: linkml.cli.config:config
+    :prog: linkml config
+    :nested: short
+```
+
+## `linkml config get`
+
+```{eval-rst} 
+.. click:: linkml.cli.config:get
+    :prog: linkml config get
+    :nested: short
+```  
+
+## `linkml config set`
+
+```{eval-rst} 
+.. click:: linkml.cli.config:set
+    :prog: linkml config set
+    :nested: short
+```  
--- a/docs/cli/config.md
+++ b/docs/cli/config.md
@@ -7,6 +7,7 @@
     :nested: short
 ```
 
+{#linkml-config-get-cli}
 ## `linkml config get`
 
 ```{eval-rst} 
@@ -15,6 +16,7 @@
     :nested: short
 ```  
 
+{#linkml-config-set-cli}
 ## `linkml config set`
 
 ```{eval-rst} 
--- a/docs/cli/index.md
+++ b/docs/cli/index.md
@@ -2,6 +2,7 @@
 
 The CLI interface to LinkML is currently divided into two styles:
 - One combined cli underneath [`linkml`](./linkml.md)
+  - [`linkml config`](./config.md) 
   - [`linkml generate`](./generate.md) 
   - [`linkml lint`](./lint.md)
   - [`linkml validate`](./validate.md)
@@ -12,6 +13,7 @@ The CLI interface to LinkML is currently divided into two styles:
 maxdepth: 1
 ---
 linkml
+config
 generate
 lint
 validate
--- a/docs/cli/linkml.md
+++ b/docs/cli/linkml.md
@@ -11,6 +11,7 @@
 
 See subpages for further details on commands:
 
+- [`linkml config`](./config.md)
 - [`linkml generate`](./generate.md)
 - [`linkml lint`](./lint.md)
 - [`linkml validate`](./validate.md)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -49,6 +49,7 @@ extensions = [
     'sphinx_design',
     'matplotlib.sphinxext.plot_directive',
     'sphinx_jinja',
+    'sphinxcontrib.autodoc_pydantic'
 ]
 
 # The suffix(es) of source filenames.
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -14,8 +14,8 @@ priorities override lower priorities)
 * in a `linkml_config.yaml` file in the working directory
 * in the `tool.linkml.config` table in a `pyproject.toml` file in the working directory
 * in the global `linkml_config.yaml` file in the platform-specific data directory
-  (use `linkml config get global_config` to find its location)
-* the default values in the :class:`.Config` model
+  (use `linkml config get config_file` to find its location)
+* the default values in the {class}`~linkml.utils.config.GlobalConfig` model
 
 Parent directories are _not_ checked.
 
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -1,0 +1,151 @@
+# Configuration
+
+Global configuration in linkml uses [`pydantic-settings`](https://docs.pydantic.dev/latest/concepts/pydantic_settings/)
+
+
+## Sources
+
+Config values can be set (in order of priority from high to low, where higher
+priorities override lower priorities)
+
+* in the arguments passed to the class constructor (not user configurable)
+* in environment variables like `export LINKML_LOG_DIR=~/`
+* in a `.env` file in the working directory
+* in a `linkml_config.yaml` file in the working directory
+* in the `tool.linkml.config` table in a `pyproject.toml` file in the working directory
+* in the global `linkml_config.yaml` file in the platform-specific data directory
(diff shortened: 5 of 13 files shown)


sneakers-the-rat:
@candleindark for some reason it's not letting me request you as a reviewer, but since you were the one who kicked this whole show off would love to have your review :)

candleindark:
> @candleindark for some reason it's not letting me request you as a reviewer, but since you were the one who kicked this whole show off would love to have your review :)

You are so quick. I will see if will have time. I will if I can I do it before the LinkML team approves it. Thanks for the work.

sneakers-the-rat:
omfg ok sorry for the commit spray re: py38 compatibility. the toxfile isn't succeeding in making a python 3.8 environment (probably a poetry thing, another reason i have forsaken it), so I added [FA102](https://docs.astral.sh/ruff/rules/future-required-type-annotation/) to the ruff config to check for it - when we deprecate `3.8` and `3.9` (3.9 reaches EoL this saturday, 10/5) we just need to increment the `target-version` and that will cease to be checked for

pkalita-lbl (commented):
As always, I appreciate the thoughtful and high quality changes here. I guess I just have two comments, one narrow and one broad.

My narrow comment is I wonder about the `set` subcommand's behavior when a `value` isn't provided. It seems like having a command like `linkml config set log` that _unsets_ all of your logging settings could lead to unhappy accidents. I wonder if we should take inspiration from the `git config` [command](https://git-scm.com/docs/git-config) that has an explicit `unset` command and in turn require a `value` be provided to `set`.

My broader comment, though, is that I wonder if having this infrastructure in place before said infrastructure is used by anything could be confusing to users. If I'm allowed to say `linkml config set log.level_stream ERROR` but that doesn't actually affect any logging output, that could reasonably be seen as a bug.

sneakers-the-rat:
> I wonder if we should take inspiration from the git config [command](https://git-scm.com/docs/git-config) that has an explicit unset command and in turn require a value be provided to set.

sure no problem :)

>  If I'm allowed to say linkml config set log.level_stream ERROR but that doesn't actually affect any logging output, that could reasonably be seen as a bug.

i was just holding off because i didn't know if this would be acceptable, and there was also another PR for logging at the time i opened this where we were changing logging around but not making a centralized logging initializer, and i didn't want to rewrite that as soon as it got merged, but ya i can definitely add the changes to logging to make it work in this PR if that's what we'd like :)

pkalita-lbl:
> i was just holding off because i didn't know if this would be acceptable

That's fair. And I'm hedging because I'm not sure I fully grok the big picture surrounding these changes. I think we have a bit of chicken-egg scenario 😁

matentzn:
I am converting this PR to draft (many conflicts now, quite dated PR); @sneakers-the-rat @pkalita-lbl I will leave it to you to revive!

sneakers-the-rat:
I could def freshen this up if desired, there are lots of things that could be exposed as config in linkml :)

sneakers-the-rat:
superceded by #3782

PR #2336, opened 2024-10-01
schema validator plugin.

cmungall:
This is currently incomplete, and subject to change.
The plugin mechanism may not be the best place for this.

Ultimately the schema validator can include logical checks,
e.g. testing for *satisiability* of schemas.

This is also required for testing metamodel *implementation*
slots (but the two are not yet hooked up). Normally a schema is
hardwired to be an instance of a metamodel schema. Using implementation
slots allows the schema author to provide their own metaclasses
effetcively further constraining the base metamodel


sierra-moxon:
fixed linting from upstream merge and lint-fix changes.

pkalita-lbl:
I'm going to close this PR because I think the checks being done here (e.g. is the `multivalued` attribute of a slot in a subclass consistent with the parent class?) are more naturally suited to being [linter rules](https://linkml.io/linkml/schemas/linter.html). I think that introducing a validator plugin that _only_ operates on schemas could be a little confusing, as well.

PR #2341, opened 2024-10-01
Make materialization (attributes and patterns) default behaviour in `gen-linkml`

sujaypatil96:
This PR seeks to make materialization of both induced slots (as attributes) and structured patterns (as patterns) universal/true by default in `gen-yaml`.

There are two optional arguments called `--materialize-attributes/no-materialize-attributes` and `materialize-patterns/no-materialize-patterns` which we are combining into a single optional argument called `--materialize/no-materialize`.

<s>Let me know if you think I need to add any `CliRunner` tests to test the CLI option. If not, this is ready for review.</s> ✅

The two options `--materialize` and `--no-materialize` can be used for "master" control over the granular `--materialize-attributes/no-materialize-attributes`  and `materialize-patterns/no-materialize-patterns`. If `--materialize` has been set to True, then attributes and patterns will be materialized regardless of the granular parameters, and vice versa.


sierra-moxon (approved):

turbomam (commented):
I don't feel strongly about the default state of the materialization option(s), but I do I prefer keeping the attribute and structured pattern separate. The schema a file that NMDC people and software interact with the most is a merged and pattern materialized version. I really don't want to start explaining attributes to the users and modify software to access attributes instead of slots.

But I won't object to this if somebody can help me come up with a plan for taking more of an attribute-first view of the schema.

cmungall (changes requested):
rather than remove options, add an uber-option `--materialize` that sets all `--materialize-X` options.

sierra-moxon (approved):

cmungall (approved):

PR #2354, opened 2024-10-04
Remove the use of `jsonschema.exceptions.best_match()`

candleindark:
Calling this function on a list of a single error always return the error, so the current uses in the code are redundant.

The documentation of `best_match` is

```py
def best_match(errors, key=relevance):
    """
    Try to find an error that appears to be the best match among given errors. ...
    """
```


candleindark:
Never mind, best_match() can actually return a different error when the list of errors passed in contains only a single error.

Issue #2357, opened 2024-10-07
[chore] Drop python 3.8 support, add deprecation warning for 3.9, add 3.13 support


sneakers-the-rat:
**What is your feature request?**
Python 3.9 is now EoL: https://devguide.python.org/versions/

Python 3.13 is now released.

We still have python 3.8 as our lower bound, which has been EoL for a year. If we want to have one EoL version left in allowable versions, we should add a deprecation warning to 3.9 now so we give ample notice to upgrade before dropping.

**How important is this feature to you?** Select from the options below:
• Low - it's an enhancement but not crucial for work

dalito:
Not that it matters much but 3.8 EOL is today (not a year ago).

sneakers-the-rat:
ope you're right i got too excited.

sierra-moxon:
from dev call:
- happy to remove 3.8 support 
- let us know when you're ready to merge and we will!

sneakers-the-rat:
I think we need to have the linkml-runtime PR merged first? Otherwise the main PR fails and it feels bad to merge failing code :)

sneakers-the-rat:
OK this is ready to go, changes to `linkml-runtime` need to be merged first: https://github.com/linkml/linkml-runtime/pull/345
then the main lib can be merged: https://github.com/linkml/linkml/pull/2358

it should be uncontroversial updating of syntax except for https://github.com/linkml/linkml/issues/2370 and the removal of the `dataclass-extensions_376` (where does that name come from anyway?) module in `linkml-runtime`. If we want to keep that behavior then someone else be my guest to re-implement it, but i think it's in general not a good idea to monkeypatch builtins and the desired behavior (reporting source location of errors in loaded yaml files) is better handled at yaml loading time rather than model instantiation time (i think i've written this exact sentence 4 times now so forgive me if i'm being repetitive, trying to avoid context collapse across issues).

Otherwise, it would be GREAT if we could merge these ASAP since they touch many places in the code and i would love to not have to rebase this 20 times.

hoijui:
I installed linkml with `pipx` right now, and `linkml-convert` fails to execute because it uses `typing.re`, which according to <https://github.com/python/cpython/issues/92871>, was removed in python 3.12. Would it be possible/ok to mention the required python version range in the installation instructions?
If you tell me the file, I could make an PR for it.

sierra-moxon:
[...]

PR #2358, opened 2024-10-07
Add python 3.13, Remove python 3.8

sneakers-the-rat:
Fix: https://github.com/linkml/linkml/issues/2357

This is a big one!!!!! we can change all the `Dict[str,str]` annotations to `dict[str,str]` finally. 

I'm going to use ruff and `pyupdate` rules to make that tractable, will do the shift in several commits so the changes can be inspected. 

opening as a draft to run on CI as i go

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -39,15 +39,15 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
         exclude:
           # Test on Windows with only the oldest and newest Python versions
-          - os: windows-latest
-            python-version: "3.9"
           - os: windows-latest
             python-version: "3.10"
           - os: windows-latest
             python-version: "3.11"
+          - os: windows-latest
+            python-version: "3.12"
 
     # See https://github.com/snok/install-poetry#running-on-windows
     defaults:
--- a/.github/workflows/check_external_links.yaml
+++ b/.github/workflows/check_external_links.yaml
@@ -17,7 +17,7 @@ jobs:
       - name: Set up Python 3.
         uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: "3.12"
 
       - name: Install Poetry
         uses: snok/install-poetry@v1
--- a/.github/workflows/doc_pages.yaml
+++ b/.github/workflows/doc_pages.yaml
@@ -15,7 +15,7 @@ jobs:
       - name: Set up Python 3.
         uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: "3.12"
 
       - name: Install Poetry
         uses: snok/install-poetry@v1
--- a/.github/workflows/docs-test.yaml
+++ b/.github/workflows/docs-test.yaml
@@ -20,7 +20,7 @@ jobs:
       - name: Set up python
         uses: actions/setup-python@v3
         with:
-          python-version: "3.11"
+          python-version: "3.12"
 
       - name: Install Poetry
         uses: snok/install-poetry@v1
--- a/poetry.lock
+++ b/poetry.lock
@@ -2,13 +2,13 @@
 
 [[package]]
 name = "alabaster"
-version = "0.7.13"
-description = "A configurable sidebar-enabled Sphinx theme"
+version = "0.7.16"
+description = "A light, configurable Sphinx theme"
 optional = false
-python-versions = ">=3.6"
+python-versions = ">=3.9"
 files = [
-    {file = "alabaster-0.7.13-py3-none-any.whl", hash = "sha256:1ee19aca801bbabb5ba3f5f258e4422dfa86f82f3e9cefb0859b283cdd7f62a3"},
-    {file = "alabaster-0.7.13.tar.gz", hash = "sha256:a27a4a084d5e690e16e01e03ad2b2e552c61a65469419b907243193de1a84ae2"},
+    {file = "alabaster-0.7.16-py3-none-any.whl", hash = "sha256:b46733c07dce03ae4e150330b975c75737fa60f0a7c591b6c8bf4928a28e2c92"},
+    {file = "alabaster-0.7.16.tar.gz", hash = "sha256:75a8b99c28a5dad50dd7f8ccdd447a121ddb3892da9e53d1ca5cca3106d58d65"},
 ]
 
 [[package]]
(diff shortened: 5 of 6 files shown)


.github/workflows/main.yaml
```diff
@@ -39,15 +39,15 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
         exclude:
           # Test on Windows with only the oldest and newest Python versions
-          - os: windows-latest
-            python-version: "3.9"
```
sierra-moxon, line 46: this should stay, right (and add 13)?

sneakers-the-rat: yes yes, just bump up the versions - these are the intermediate versions excluded from the windows test matrix, so the lowest version we're testing on windows is 3.9 (i.e. no longer exclude it), and yes we also don't exclude 3.13 so are also testing on it (when `3.13` is added to the supported versions by the setup-python action)

sneakers-the-rat:
Alright looks like `3.13` isn't enabled in the github actions `setup-python` yet, so this will remain a draft until that happens :)

sneakers-the-rat:
alright, stepped away for most of the day to do something else, but that looks like tests passing on my machine. 

I fixed pydanticgen and pythongen to also use `list` and `dict` rather than List and Dict, and then went through and found all the places where pythongen was failing and added `pytest.mark`s and regenerated snapshots that way (`python -m pytest -m pythongen --regenerate-snapshots --with-slow`) so i wouldn't get the usual spurious OWL/RDF ordering diffs (and also because it's really nice to be able to run tests limited by a specific generator so you aren't just waiting around all day when you know you didn't change anything except in one generator!)

I'm sort of (?) alarmed (?) that sqlalchemygen didn't cause any errors. I'm not an sqlalchemy or sql expert, but i thought that it would be possible to do `list[T]` in sqlalchemy to declare a one to many relationship. if not then we're good. if that is the case then sqlalchemygen is missing tests

vincentkelleher:
Am I the only one getting an issue with the `typing.io` import from 'pyshexc/parser/ShExDocLexer.py` ?

image

sneakers-the-rat:
> Am I the only one getting an issue with the `typing.io` import from 'pyshexc/parser/ShExDocLexer.py` ?

Ya I fixed that but an updated package was never released: https://github.com/shexSpec/grammar-python-antlr/pull/15

sierra-moxon:
@sneakers-the-rat - can you please fix the failing tests?

sierra-moxon:
@ptgolden - if you have some cycles, getting these tests fixed and merged so that the downstream PR can be completed would be wonderful.

ptgolden:
Happy to give it a shot, but it seems the logs from the failing tests have expired. I'm not a GitHub Actions expert-- is there a way to manually retrigger them?

ptgolden:
I also have a hunch this might be a chicken and egg problem. If `linkml-runtime` does not support 3.13 until https://github.com/linkml/linkml-runtime/pull/345 is merged, then these tests will fail until that PR is merged. But that PR relies on this one being merged. (Unless I'm missing something)

sierra-moxon:
Lets try to get the conflicts fixed and if we can isolate the test failures to just the circular dependency issue, then we can go from there on how to get this done.

sierra-moxon:
@sneakers-the-rat - next time, would this kind of change that may require multiple developer contributions warrant a branch instead of a fork in your opinion?

sneakers-the-rat:
> I also have a hunch this might be a chicken and egg problem. If linkml-runtime does not support 3.13 until https://github.com/linkml/linkml-runtime/pull/345 is merged, then these tests will fail until that PR is merged. But that PR relies on this one being merged. (Unless I'm missing something)

That's correct. I've been rebasing both since October waiting on review. They need to be merged and deployed at the same time. I can do it again tho

sneakers-the-rat:
so we really need a new release of `pyshexc` to be cut after https://github.com/shexSpec/grammar-python-antlr/pull/15 , that's another dependency that needs to be updated before 3.13 can work. that's one of ours right?

sneakers-the-rat:
ok checked that remaining test failures are from the removed `dataclasses` [monkeypatch](https://github.com/linkml/linkml-runtime/blob/main/linkml_runtime/utils/dataclass_extensions_376.py) that formerly allowed all `**kwargs`, and since the current `linkml-runtime` imports that monkeypatch (or it is imported at some point during the tests) the tests fail even if this PR removes the application of the monkeypatch in generated python modules. sort of illustrates why runtime side effects, and especially monkeypatching stdlib modules using private stdlib methods is fragile.

confirmed locally that this PR passes tests against https://github.com/linkml/linkml-runtime/pull/345/commits/0846eb0ebc054cab254f4cbca5d51c0de2fea9dc , and will work on a way of running the `linkml-runtime` upstream tests against a specific branch for cases like this where we need to do paired PRs

amc-corey-cox:
@sneakers-the-rat Is this just stuck on https://github.com/linkml/linkml-runtime/pull/345 ?

sneakers-the-rat:
@amc-corey-cox the blocker at this point is really https://github.com/shexSpec/grammar-python-antlr/pull/15 not having a new release that includes the fix that is already merged for the `typing.re` import. i'm just going to add an exception handler to that import and say "shex generation is not supported in python 3.13 take it up with the shexc repo"

sneakers-the-rat:
ok we yet again have battled through the mysterious wilds of our dependency tree to reach a point where things are running again. would be great to be able to merge this before things shift again!

explanation of last commits:
- got tired of waiting for `pyshexc` to release a new version, just monkeypatched `typing.io` instead. this appears to only be used in tests (which is news to me, if i knew that i would have done this from the beginning). I added a reminder in the monkeypatching function that will let us know to remove that monkeypatch if `pyshexc` ever receives and update along with instructions.
- marked a notebook test to skip until `pyshexc` receives and update. we can't monkeypatch the env that the notebook runs in without ruining the notebook, and since those are user-facing that seems bad to do. same as above, instructions in the error say to remove that skip condition as well.
- an old matplotlib version in the lockfile was pinning `numpy<2` which was causing [this to happen](https://github.com/linkml/linkml-runtime/actions/runs/14096197716/job/39483924392). I am formally retired from dealing with poetry problems in favor of using a less obnoxious package manager, but it seems like we want to update the poetry CI action to do `poetry update` rather than just `poetry lock`

sierra-moxon:
I pulled this PR into a branch: https://github.com/linkml/linkml/pull/2589 to trivially fix the linting issue.

amc-corey-cox:
I wanted to give a quick update on this PR. Sierra made a branch to fix that trivial linting issue and we are making a new release of linkml-runtime. We want to wait on the linkml-runtime release until after the weekend so we'll be holding other merges until after the weekend as well. As soon as we have the new release, we'll bump the linkml-runtime version and get these changes merged in as well.

Thank you to everyone who has put in the effort on this especially all of the work by @sneakers-the-rat. We'll get this merged early next week.

sneakers-the-rat:
kool. merged that in here as well. either/or, they're both identical and #2589 links back to this one. i am pretty sure i turned 'allow edits by maintainers' on? sorry you needed to make another pr to fix that

dalito:
Merged in #2605

Issue #2362, opened 2024-10-09
Attach originating error or object to `linkml.validator.report.ValidationResult`


candleindark:
**What is your feature request?**
The originating error or the object used to create an instance of `linkml.validator.report.ValidationResult` is not captured in the `ValidationResult` object. Some important information in the originating object is lost. It would be nice, essential in some context, to have the originating object attached to the `ValidationResult` for diagnostic purposes. 

**How important is this feature to you?** Select from the options below:
• Important - it's a blocker and can't do work without it (Without this, debugging against great number of data instances is very difficult)

https://github.com/linkml/linkml/pull/2363 offers a possible solution.

PR #2363, opened 2024-10-09
Bundle originating error/object to `linkml.validator.report.ValidationResult`

candleindark:
The change in this PR is a possible solution to https://github.com/linkml/linkml/issues/2362. It allows an object of any type to be attached to a `linkml.validator.report.ValidationResult` at the `source` field and excludes the `source` field from the serialization of a `ValidationResult` object. 

This PR utilizes the `source` field in `JsonschemaValidationPlugin` only. If this PR is approved, other validation plugins should use the source field to attach the originating object to the `ValidationResult` as well.

If we want to enforce the utilization of the `source` field for all validation plugins. We should modify the following line

```py
source: Any = Field(None, description="The source of this validation result", exclude=True)
```

to

```py
source: Any = Field(description="The source of this validation result", exclude=True)
```

### DIFF
--- a/linkml/validator/plugins/jsonschema_validation_plugin.py
+++ b/linkml/validator/plugins/jsonschema_validation_plugin.py
@@ -55,4 +55,5 @@ class JsonschemaValidationPlugin(ValidationPlugin):
                 instantiates=context.target_class,
                 message=f"{best_error.message} in /{'/'.join(str(p) for p in best_error.absolute_path)}",
                 context=error_context,
+                source=best_error,
             )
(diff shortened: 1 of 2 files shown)


cmungall (approved):

candleindark:
@cmungall Don't you think it is better to make the `source` field required using

```py
source: Any = Field(description="The source of this validation result", exclude=True)
```

That will require all the validation plugins to provide a value for it. However, the code will be more predictable.

cmungall:
It’s already hard to predict what to do with it as it’s an Any. Perhaps we
could look a bit further and see what kind of diagnostic info is intended
to go there and harmonize that further?

PR #2366, opened 2024-10-11
handle special ifabsent cases in python ifabsent processor

sneakers-the-rat:
Related to: https://github.com/linkml/linkml/pull/2329
Fixes one of the problems in : https://github.com/linkml/linkml/issues/2284

Basic issue is that `ifabsent` has several special cases:

```yaml
ifabsent:
  domain: slot_definition
  range: string
  inherited: true
  description: >-
    function that provides a default value for the slot.
      * [Tt]rue -- boolean True
      * [Ff]alse -- boolean False
      * bnode -- blank node identifier
      * class_curie -- CURIE for the containing class
      * class_uri -- URI for the containing class
      * default_ns -- schema default namespace
      * default_range -- schema default range
      * int(value) -- integer value
      * slot_uri -- URI for the slot
      * slot_curie -- CURIE for the slot
      * string(value) -- string value
      * EnumName(PermissibleValue) -- enum value
  close_mappings:
    - sh:defaultValue
  in_subset:
    - SpecificationSubset
  see_also:
    - linkml:equals_expression

```

This PR fixes
- `class_curie`
- `class_uri`
- `default_range`
- `slot_uri`
- `slot_curie`

for the python ifabsent generator. I didn't handle `default_ns` because i don't know what that is.

Previously the processor would just produce a `None` for each of these, which is incorrect.

Now you can observe via the tests that they are handled correctly.

I think that `ifabsent: default_range` is a special case that really only makes sense to use in exactly one place in the metamodel: https://github.com/linkml/linkml-model/blob/76bc3279981f00f9f6590caf7fbe5a1f15520f85/linkml_model/model/schema/meta.yaml#L1381

but hey might as well make the processor spec compliant even if it only applies in exactly one case. 

Leaving as a draft for now because i figure this will require some more massaging of the tests/regenerating snapshots but i hate running regenerate snapshots on my local machine unless i have to because it still makes a mess

### DIFF
--- a/linkml/generators/common/ifabsent_processor.py
+++ b/linkml/generators/common/ifabsent_processor.py
@@ -1,7 +1,7 @@
 import abc
 import re
 from abc import ABC
-from typing import Any, Optional
+from typing import Any, Optional, Tuple, Union
 
 from linkml_runtime import SchemaView
 from linkml_runtime.linkml_model import (
@@ -48,7 +48,6 @@ class IfAbsentProcessor(ABC):
         if slot.ifabsent:
             ifabsent_match = self.ifabsent_regex.search(slot.ifabsent)
             ifabsent_default_value = ifabsent_match.group("default_value")
-
             return self._map_to_default_value(slot, ifabsent_default_value, cls)
 
         return None
@@ -174,7 +173,9 @@ class IfAbsentProcessor(ABC):
         raise ValueError(f"The ifabsent value `{slot.ifabsent}` of the `{slot.name}` slot could not be processed")
 
     @abc.abstractmethod
-    def map_custom_default_values(self, default_value: str, slot: SlotDefinition, cls: ClassDefinition) -> (bool, str):
+    def map_custom_default_values(
+        self, default_value: str, slot: SlotDefinition, cls: ClassDefinition
+    ) -> Tuple[bool, Union[str, None]]:
         """
         Maps custom default values that aren't generic behaviours.
 
--- a/linkml/generators/python/python_ifabsent_processor.py
+++ b/linkml/generators/python/python_ifabsent_processor.py
@@ -1,3 +1,5 @@
+from typing import Tuple, Union
+
 from linkml_runtime.linkml_model import (
     ClassDefinition,
     EnumDefinitionName,
@@ -8,16 +10,23 @@ from linkml.generators.common.ifabsent_processor import IfAbsentProcessor
 
 
 class PythonIfAbsentProcessor(IfAbsentProcessor):
-    UNIMPLEMENTED_DEFAULT_VALUES = ["class_curie", "class_uri", "slot_uri", "slot_curie", "default_range", "default_ns"]
 
-    def map_custom_default_values(self, default_value: str, slot: SlotDefinition, cls: ClassDefinition) -> (bool, str):
+    URI_SPECIAL_CASES = ("class_uri", "slot_uri")
+    CURIE_SPECIAL_CASES = ("class_curie", "slot_curie")
+    OTHER_SPECIAL_CASES = ("default_range",)
+    UNIMPLEMENTED_DEFAULT_VALUES = ("default_ns",)
+
+    def map_custom_default_values(
+        self, default_value: str, slot: SlotDefinition, cls: ClassDefinition
+    ) -> Tuple[bool, Union[str, None]]:
         if default_value in self.UNIMPLEMENTED_DEFAULT_VALUES:
             return True, None
-
-        if default_value == "bnode":
+        elif default_value in self.OTHER_SPECIAL_CASES:
+            return True, self._map_other_special_case(default_value, slot, cls)
+        elif default_value == "bnode":
             return True, "bnode()"
-
-        return False, None
+        else:
+            return False, None
 
     def map_string_default_value(self, default_value: str, slot: SlotDefinition, cls: ClassDefinition):
         return self._strval(default_value)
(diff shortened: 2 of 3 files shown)


cmungall (approved):

linkml/generators/common/ifabsent_processor.py
```diff
@@ -284,3 +293,27 @@ def _uri_for(self, s: str) -> str:
...
+        elif default_value == "slot_curie":
+            return f'"{self.schema_view.get_uri(slot, expand=False)}"'
+        else:
+            raise ValueError(
+                f"Default value must be one of the curie special cases: {self.CURIE_SPECIAL_CASES}. "
+                f"Got: {default_value}"
+            )
+
+    def _map_default_range_special_case(self, default_value: str, slot: SlotDefinition, cls: ClassDefinition) -> str:
+        return f'"{self.schema_view.schema.default_range}"'
```
Silvanoc, line 319: What is the reason for having so many useless arguments in this method? Signature consistency? It is not an abstract method that might get different functionality that requires those arguments.

Why not simply?
```suggestion
    def _map_default_range_special_case(self) -> str:
        return f'"{self.schema_view.schema.default_range}"'
```

linkml/generators/python/python_ifabsent_processor.py
```diff
@@ -60,12 +62,25 @@ def map_datetime_default_value(
         return f"datetime({int(year)}, {int(month)}, {int(day)}, " f"{int(hour)}, {int(minutes)}, {int(seconds)})"
 
     def map_uri_or_curie_default_value(self, default_value: str, slot: SlotDefinition, cls: ClassDefinition):
+        if default_value in self.URI_SPECIAL_CASES:
+            return self._map_uri_special_case(default_value, slot, cls)
+        if default_value in self.CURIE_SPECIAL_CASES:
+            return self._map_curie_special_case(default_value, slot, cls)
+
         return self._uri_for(default_value)
```
Silvanoc, line 70:
```suggestion
        elif default_value in self.CURIE_SPECIAL_CASES:
            return self._map_curie_special_case(default_value, slot, cls)
        else:
            return self._uri_for(default_value)
```

The result is the same, but the logic (only one among the three options) is clearer. This way the structure is the same as in `map_custom_default_values`.

Although it is then different from that of `map_curie_default_value` and `map_uri_default_value`, where there is no `else`.

I would try to keep the structure consistent between all of them in this module and also in `linkml/generators/shacl/shacl_ifabsent_processor.py`.

Either like a switch
```
if ...
    return
elif ...
    return
else
    return
```
Or
```
if ...
    return
elif ...
    return
return
```

And never
```
if ...
    return
if ...
    return
return
```

linkml/generators/shacl/shacl_ifabsent_processor.py
```diff
@@ -62,12 +62,29 @@ def map_datetime_default_value(
...
     def map_uri_or_curie_default_value(self, default_value: str, slot: SlotDefinition, cls: ClassDefinition):
-        raise NotImplementedError()
+        is_curie = False
+        if default_value in self.CURIE_SPECIAL_CASES:
+            is_curie = True
+            default_value = self._map_curie_special_case(default_value, slot, cls)
+        elif default_value in self.URI_SPECIAL_CASES:
+            default_value = self._map_uri_special_case(default_value, slot, cls)
+        else:
+            if ":" in default_value and "://" not in default_value and len(default_value.split(":")) == 2:
```
Silvanoc, line 72: "Been there, seen that" 😄 

I introduced a module for much more robust URI/CURIE validation some time ago: https://github.com/linkml/linkml-runtime/blob/main/linkml_runtime/utils/uri_validator.py You can simply use `validate_curie`.

Although, on the other side, if you can simply assume that the CURIEs/URIs are valid, your logic is simpler.

linkml/generators/shacl/shacl_ifabsent_processor.py
```diff
@@ -87,3 +104,9 @@ def map_json_path_default_value(self, default_value: str, slot: SlotDefinition,
 
     def map_sparql_path_default_value(self, default_value: str, slot: SlotDefinition, cls: ClassDefinition):
         raise NotImplementedError()
+
+    def _map_default_range_special_case(self, default_value: str, slot: SlotDefinition, cls: ClassDefinition) -> str:
```
Silvanoc, line 108: Why not?
```suggestion
    def _map_default_range_special_case(self) -> str:
```

Silvanoc (approved):
Nitpicks (style, no functional impact). Otherwise LGTM

linkml/generators/common/ifabsent_processor.py
sneakers-the-rat: Api consistency is the answer - some other generator may need the slot and class to know how to render that value, even if it's not very likely since it should only change on a schema level. So eg. In the shacl generator how it needs to wrap the value in some RDF literal, that could depend on the range in the slot, and who knows maybe the class has some additional rules that condition it.

linkml/generators/shacl/shacl_ifabsent_processor.py
sneakers-the-rat: Oh great, I was looking for this and didnt know where to find it

linkml/generators/shacl/shacl_ifabsent_processor.py
sneakers-the-rat: Same reason as above

linkml/generators/common/ifabsent_processor.py
```diff
@@ -61,6 +65,9 @@ def _map_to_default_value(
         if mapped:
             return custom_default_value
 
+        if ifabsent_default_value == self.DEFAULT_RANGE_SPECIAL_CASE:
```
vincentkelleher, line 68: I'm wondering if it would be better to also have an `if` statement for the cases managed by `_map_uri_special_case(..)` and `_map_curie_special_case(..)` ? 🤔 

What I mean is something like this:

```python
        if slot.range == Uri.type_name:
          if default_value in URI_SPECIAL_CASES:
            return self.map_uri_special_case(ifabsent_default_value, slot, cls)
          return self.map_uri_default_value(ifabsent_default_value, slot, cls)

        if slot.range == Curie.type_name:
          if default_value in CURIE_SPECIAL_CASES:
            return self.map_curie_special_case(ifabsent_default_value, slot, cls)
          return self.map_curie_default_value(ifabsent_default_value, slot, cls)

        if slot.range == Uriorcurie.type_name:
            if default_value in URI_SPECIAL_CASES:
              return self.map_uri_special_case(ifabsent_default_value, slot, cls)
            if default_value in CURIE_SPECIAL_CASES:
              return self.map_curie_special_case(ifabsent_default_value, slot, cls)            
            return self.map_uri_or_curie_default_value(ifabsent_default_value, slot, cls)
```

What do you think ?

vincentkelleher: (I forgot to publish this message last week, it was lying there in my browser 😬 )

sneakers-the-rat: fair proposal - the idea here was that we want to limit the switching logic that happens in this method as much as possible. basically this is already so stacked up that if any subclass needed to override it the entire purpose of the class tree would be lost. there is already in these two cases of the python and shacl handlers need to handle these switches in different ways, so i am deferring that responsibility to the sub-generators and providing these convenience methods to them if they need it.

sneakers-the-rat: oop forgot to add - the reason that the `DEFAULT_RANGE_SPECIAL_CASE` gets added here (and is explicitly labeled as a special case to prominently) is that it's a type-less parameter, where the others are based on detection of the type of the range. it's used exactly once in the metamodel, and as far as i can tell that's the only time you would use the value of `default_range` as the `ifabsent` value.

sneakers-the-rat: as Gödel demands, no self-describing schema language can exist without a few axioms taken for granted (aka special cases)

linkml/generators/python/python_ifabsent_processor.py
sneakers-the-rat: oop this was a copy paste error my b will fix

linkml/generators/common/ifabsent_processor.py
vincentkelleher: Ok I understand your logic 👍 
That's good to me so 😊

Issue #2339, opened 2024-10-01
Additional keys for UI schema definitions


Duanoc:
**What is your feature request?**
I am currently testing linkml to define the forms for our electronic lab notebook (ELN) to collect structured metadata.
To generate the forms in the ELN a json schema could be used directly in some cases. But to use the fully functionality additional parameters for the form definition are recommended.
To describe the forms a configuration file is needed and my intention was to exend the json schema with additional configurational propoerties like in json forms (https://jsonforms.io/).
They are needed to describe e.g. the input type in the form, additional handling for displaying within a template, etc.
How can I add additional key value pairs in the json schema? 

One workaround is to miss use the description key and define a json object as value.
I need the additional keys on class and on slot level.
The key 'notes' or 'comments' are not exported by 'gen-json-schema'.
Better would be using annotations. Within annotations I can use any key value pairs.

Another workaround would be to write a script which extracts the annotations and extends the json schema (e.g. using patchjson) or write a matching UI schema file. Theoretically the annotation keys can be validated. But following the example written in the [documentation](https://linkml.io/linkml/schemas/annotations.html#annotations) I could not create a annotation key validation, any key was accepted, when running `gen-json-schema Annotation_test.yaml`. Can you provide a working example?

<details>
  <summary>Annotation_test.yaml</summary>
  
  ```yaml
id: https://example.org/Annotation_test/
name: Annotation_test
prefixes:
  linkml: https://w3id.org/linkml/
imports:
  - linkml:types

classes:
  Person:
    instantiates:
      - mymetamodel:Reviewable
    annotations:
      review: A very useful class that is well defined
      input_type: combobox
    attributes:
      name:
        description: the name of a person
        instantiates: Reviewable
        annotations:
          review3: A very useful attribute that is well defined
  Reviewable:
    class_uri: mymetamodel:Reviewable
    attributes:
      review:
        description: an expert review of a schema element
        range: string
  ```
</details>

What is your suggestion how to solve my task if it is possible with linkml?
Or should a new generator be implemented to create a ui-schema.json file?

**How important is this feature to you?** Select from the options below:
• Medium - can do work without it; but it's important (e.g. to save time or for convenience)

**Additional context**
jsonforms.io is just one example. There might be other standards to describe the ui schema.

cmungall:
Hi @Duanoc, I like the idea of passing additional UI-hint metadata in the schema to jsonforms and similar tools (see also #1882).

So I think there are two asks here:

[...]

PR #2369, opened 2024-10-14
[jsonschemagen] Add lifecycle methods to jsonschemagen

sneakers-the-rat:
Related to: https://github.com/linkml/linkml/issues/2339

Lifecycle methods in generators are a great way to allow people to modify the generation process in ways that might not have a place in the main in-library generators while not having to override methods that would cause subclasses to get out of date with changes in in-library generators. 

In this case the request was to add special terms for a json schema dialect for UI elements. Adding lifecycle hooks allows them to experiment with making that work - and if it is something that we might want to integrate upstream then we have a clear, packaged way of doing that that cleanly divides the additional parts into separate methods rather than intercalating them in the other parts of the generator.

As usual, i could only implement some of the lifecycle hooks (eg. the `after` methods for the collective `classes` methods can't be done because the results aren't returned, but directly added to a shared dict), but i added which to the docstring. 

Added `before/after_generate_class_slot` because those seem different enough from generating a slot by itself to be worth a separate method that also passes the class that the slot is being built in the context of. will save swapping pydanticgen over to those for another PR.

Build results classes here are trivial since jsonschemagen doesn't use the builder pattern, used `model_construct` since there's no actual validation to be done and i wasn't sure if the pydantic validator would take extra time with a custom `dict` subclass.

### DIFF
--- a/linkml/generators/common/build.py
+++ b/linkml/generators/common/build.py
@@ -5,7 +5,6 @@ Models for intermediate build results
 """
 
 import dataclasses
-from abc import abstractmethod
 from typing import Any, TypeVar
 
 try:
@@ -57,11 +56,11 @@ class BuildResult(BaseModel):
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
-    @abstractmethod
     def merge(self, other: T) -> T:
         """
         Build results should have some means of merging results of a like kind
         """
+        raise NotImplementedError("This build result doesn't know how to merge!")
 
 
 class SchemaResult(BuildResult):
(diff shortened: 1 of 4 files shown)


sierra-moxon:
@sneakers-the-rat - just an update: we ran out of time in the dev call (you are always more than welcome to join) today to discuss this.  It is next on my list; please let us know if this is a blocker for you?

sneakers-the-rat:
Not a blocker for me, just helping out with https://github.com/linkml/linkml/issues/2339 :)

pkalita-lbl (approved):

Issue #2342, opened 2024-10-02
plantuml tests now fail without docker server


sneakers-the-rat:
**Describe the bug**
I don't keep a docker server running on my local machine and i would prefer it if running one was not required for the tests

When I run tests locally, i get this

```
ERROR tests/test_generators/test_plantuml.py::test_serialize_selected[Person-\nclass "Person" [[{A person, living or dead}]] {\n    {field} id : string  \n    {field} name : string  \n    {field} age_in_years : integer  \n    {field} species_name : string  \n    {field} stomach_count : integer  \n    {field} is_living : LifeStatusEnum  \n    {field} aliases : string  [0..*]\n}\n] - docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such fil...
ERROR tests/test_generators/test_plantuml.py::test_serialize_selected[Dataset-\n"Dataset" *--> "0..*" "Person" : "persons"\n] - docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such fil...
ERROR tests/test_generators/test_plantuml.py::test_serialize_selected[MedicalEvent-\n"Person" *--> "0..*" "MedicalEvent" : "has medical history"\n] - docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such fil...
ERROR tests/test_generators/test_plantuml.py::test_serialize_selected[FamilialRelationship-\n"FamilialRelationship" --> "1" "Person" : "related to"\n] - docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such fil...
ERROR tests/test_generators/test_plantuml.py::test_serialize - docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such fil...
ERROR tests/test_generators/test_plantuml.py::test_generate_svg - docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such fil...
```

Looks like PR https://github.com/linkml/linkml/pull/2318 introduced a docker fixture without proper exception handling. Tests that use a docker container should have a `docker` mark added, and during test collection if docker is not present then those tests should be marked `skip`.

Docker doesn't seem to imagine a world where one might not be running a docker server, so i can't find a function to test that directly, but this seems to do the trick

```python
import docker

def docker_server_running() -> bool:
[...]

Issue #1917, opened 2024-02-13
yuml generator (and all network tests) - intermittent test failure - replace live API calls with recorded requests.


sneakers-the-rat:
**Describe the bug**
Tests will sporadically fail after hanging for 60 seconds when making calls to live APIs, particularly in the yuml generator, but this is likely to be true for any test that does real network requests.

In this case, it was this test that failed for me: https://github.com/linkml/linkml/blob/55cc87f8d8e62eb5456f4d3431180c0d1bd98d0b/tests/test_scripts/test_gen_yuml.py#L38

in the first case (but not the second two).

with the error:
```
    def test_metamodel_output_directory(arguments, snapshot_dir, snapshot, tmp_path):
        runner = CliRunner()
        result = runner.invoke(cli, arguments + ["-d", str(tmp_path), LOCAL_METAMODEL_YAML_FILE])
>       assert result.exit_code == 0
E       AssertionError: assert 1 == 0
E        +  where 1 = <Result URLError(TimeoutError(60, 'Operation timed out'))>.exit_code
```

**To reproduce**
Its sorta hard to reproduce, which is the problem! run the tests until one of them fails!

**Expected behavior**
We should use a tool like `pytest-recording` which uses `VCR.py` to record and replay network requests. This lets us isolate tests for code correctness from tests that check API consistency (which should probably only run before eg. a release)

https://pypi.org/project/pytest-recording/

PR #2373, opened 2024-10-15
[plantuml] mark and skip plantuml docker tests when docker not present

sneakers-the-rat:
fix: https://github.com/linkml/linkml/issues/2342
fix: https://github.com/linkml/linkml/issues/1956
fix: https://github.com/linkml/linkml/issues/1917

let us be done with the plantuml tests being flaky. the tests run in ci where docker is present, but are skipped when running locally where it is not. we are no longer dependent on web requests to remote servers, so If Those For Whom PlantUML Is A Needed Feature May Adopt It, Allow Us To Delegate Its Health To Them.

we have some tests remaining that rely on network access, it is true, but we have cached most of them. the last things to do here would be to fully cache VCR requests and swap out hbreader for `requests` and then we'll have a hermetically sealed test system

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,6 +7,7 @@ from importlib.metadata import version
 from pathlib import Path
 from typing import Callable, List, Optional, Union
 
+import docker
 import pytest
 import requests_cache
 from _pytest.assertion.util import _diff_text
@@ -212,6 +213,13 @@ def pytest_collection_modifyitems(config, items: List[pytest.Item]):
             if item.get_closest_marker("pydanticgen_npd"):
                 item.add_marker(skip_npd)
 
+    # skip docker tests when docker server not present on the system
+    if not _docker_server_running():
+        skip_docker = pytest.mark.skip(reason="Docker server not running on host machine")
+        for item in items:
+            if item.get_closest_marker("docker"):
+                item.add_marker(skip_docker)
+
     # the fixture that mocks black import failures should always come all the way last
     # see: https://github.com/linkml/linkml/pull/2209#issuecomment-2231548078
     # this causes really hard to diagnose errors, but we can't fail a test run if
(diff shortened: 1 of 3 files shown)


sneakers-the-rat:
I feel justified in adding the condition to skip plantuml tests if the docker container doesn't work after seeing https://github.com/linkml/linkml/pull/2329#issuecomment-2413278697 RIP getting stuck on a red X after all that work @Silvanoc

tests/conftest.py
```diff
@@ -212,6 +213,13 @@ def pytest_collection_modifyitems(config, items: List[pytest.Item]):
             if item.get_closest_marker("pydanticgen_npd"):
                 item.add_marker(skip_npd)
 
+    # skip docker tests when docker server not present on the system
+    if not _docker_server_running():
+        skip_docker = pytest.mark.skip(reason="Docker server not running on host machine")
+        for item in items:
+            if item.get_closest_marker("docker"):
+                item.add_marker(skip_docker)
+
```
Silvanoc, line 222: With this configuration, we would not be testing against Kroki on CI Windows machines... what is probably a very good idea. But just to be sure that we are aware of the consequences.

sneakers-the-rat: into it!

sneakers-the-rat: imo the goal is to be like "by all means plantuml ppl implement as rigorous of tests as you want but if the tests are flaky please can we avoid that impeding other work?"

sneakers-the-rat: (it's why i did this, because i didn't want to be the next source of "oh dang it's breaking again", i think that's a good precedent to set https://github.com/linkml/linkml/pull/2178#issuecomment-2375514360 )

Silvanoc:
Tested on my system locally, works like a charm.

Silvanoc (approved):

vincentkelleher (approved):

vincentkelleher:
This is indeed a good solution for the time being before either changing the tests to check the request content or setup a Kroki server instance 👍

PR #2376, opened 2024-10-21
feat: allow enum IRI separator override in owlgen.py

egavard:
This feature allows to override the default # that separates an enum class and its value in the owl generator. 

 Useful if for example, your namespace already contains a #


docs/generators/owl.rst
```diff
@@ -74,6 +74,7 @@ Enums and PermissibleValues are treated as *classes* by default. An enum has a *
 an OWL ``UnionOf`` construct.
 
 * Use ``--default-permissible-value-type`` to change the default type
+* Use ``--enum_iri_separator`` to change the separator character in the enum IRI which is ``#`` by default
```
vincentkelleher, line 77: I imagine you meant :

```suggestion
* Use ``--enum-iri-separator`` to change the separator character in the enum IRI which is ``#`` by default
```

vincentkelleher (approved):

sierra-moxon (approved):

PR #2379, opened 2024-10-24
gen-typescript print to stdout if output not specified

kevinschaper:
Restores the print to stdout behavior, but only when output isn't specified. I also added tests to ensure that print is called when output is None and that print isn't called when the output file is specified.


amc-corey-cox (approved):
Ah, looks like we forgot to implement something when no output is specified. This looks like a good solution to me. I also appreciate this brings us more into alignment with standard Unix tools.

Issue #2384, opened 2024-10-28
Deprecate `linkml.validators` and `linkml.utils.validation`


sneakers-the-rat:
**Describe the bug**

Dead code is buggy code!

Validation was substantially reworked in 2023: https://github.com/linkml/linkml/issues/1494

But `linkml.validators` and `linkml.utils.validation` were never removed, and they are causing bugs, eg. https://github.com/linkml/linkml/issues/2382

We should deprecate and remove these modules/packages!

**Version of LinkML you are using**

`HEAD`

Issue #2382, opened 2024-10-25
Serialization of `PermissibleValue` is wrong


Silvanoc:
**Describe the bug**

The serialization of a `PermissibleValue` like `PermissibleValue('infant')` should be `'infant'`, instead it is `PermissibleValue({'text': 'infant'})`.

This issue is probably the root-cause of issue #2380 

**Version of LinkML you are using** v1.8.5

**Please provide a schema (and if applicable, a data file) that replicates the issue**
No schema, but some code snippet to reproduce it:

``` python
#!/usr/bin/env python3

from linkml_runtime.linkml_model.meta import PermissibleValue

value = PermissibleValue('infant')
print(value)
```

It prints `PermissibleValue({'text': 'infant'})` and not `'infant'` (what I would expect).

Silvanoc:
@sneakers-the-rat you have been active lately in that part of the code. Could you please confirm the issue?

sneakers-the-rat:
aha, yeah well when i wrote that I didn't know that we were using those models for literals like enum values. should be simple enough to fix by just removing the `__str__` method here: https://github.com/linkml/linkml-runtime/blob/ad74966db72be214a38de19f361df88489286e83/linkml_runtime/utils/yamlutils.py#L286

sneakers-the-rat:
> It is unclear why no tests (neither in linkml nor in linkml-runtime) have raised this issue

this is the bigger problem, i don't know much about the json schema generator but i don't think that any generator should be doing blind type coercions like `str(NonstandardType)`. Having the `__str__` method be overgeneral is definitely my bad, but i'm concerned that the compliance tests didn't catch this. it seems like this test should have?

 https://github.com/linkml/linkml/blob/08959cb73d8b09d66bbad730bb5f4d63f0195b08/tests/test_compliance/test_enum_compliance.py#L44
 
actually now that i look i'm not sure what that's happening actually, because i don't see a `str` call in the generator, i see a https://github.com/linkml/linkml/blob/08959cb73d8b09d66bbad730bb5f4d63f0195b08/linkml/generators/jsonschemagen.py#L383

which should handle the value correctly
 
[...]

PR #2385, opened 2024-10-28
Use `linkml.validator` in `linkml.utils.converter`

sneakers-the-rat:
Related to:
- https://github.com/linkml/linkml/issues/2382
- https://github.com/linkml/linkml/issues/2384

Replace the old `linkml.utils.validation` call in converter, should be self explanatory. 

Only thing i'm not sure about is what to do with the `results` from the validation - it seems like everywhere that's used (only in tests?) there is manual handling of the results rather than having a `report_results` method that appropriately handles those/raises/displays errors. so opening as a draft in case i'm missing something.


pkalita-lbl:
> Only thing i'm not sure about is what to do with the results from the validation ... so opening as a draft in case i'm missing something.

No, you're not missing anything. The `validate` function (and the underlying `Validator` class) is pretty unopinionated and leaves is up to the client to decide what to do with the results (although if you have alternate ideas I'm all ears). 

The call to `validate_object` that's being replaced here would have raised a `jsonschema.exceptions.ValidationError` in the case of an invalid instance. If you want to maintain rough parity with that you could check if the `ValidationReport` returned by `validate` has any `messages`, and if so raise an exception.

sneakers-the-rat:
>  The validate function (and the underlying Validator class) is pretty unopinionated and leaves is up to the client to decide what to do with the results

That makes sense. the think i would add is just a companion function that wraps what's done in a few places elsewhere, raise an exception and display the errors if they are there and otherwise pass, so the call would be something like

```python
validator.handle_messages(validator.validate(...))
```

or probably more ideally

```python
validator.validate(..., raise=True)
```

pkalita-lbl:
I like it! I'll just make one more comment that you can take or leave. What about like `raise_for_results` or something along those lines, as opposed to just `raise_`, for the parameter and method names? I'm sort of thinking along the lines of `Response.raise_for_status` in [requests](https://docs.python-requests.org/en/latest/api/#requests.Response.raise_for_status) and [httpx](https://www.python-httpx.org/api/#response), in addition to generally being more explicit.

matentzn:
@sneakers-the-rat given how long we made you wait on this, would you like one of is to deconflict / see it through or do it yourself?

sneakers-the-rat:
I went to try and rebase but it looks like this might have already been done? or at least the converter (now in `.converter.cli` rather than `utils.converter`, always love to see utils getting depopulated) uses `linkml.validator` now. the thing that this PR does differently than current impl now is that it restructures the exceptions a bit to differentiate schema validation errors from data validation errors and does a nicer print of exceptions in the validator class rather than in the thing that calls it, so i'll make a version of just the unique changes here and we can see how that looks? not sure if it's worth it/the underlying issue has been resolved and this would be sorta "bonus edition"

sneakers-the-rat:
ok there's that. i could take it or leave it, not attached to this one.

on the one hand i think this kind of thing is a normal feature of validation code: "create a validation report object that can be used, but also provide a standard way of raising and handling exceptions" since exception handling and validation are kissing cousins, and it's a pain to a) rewrite pprint code anywhere that the validation result is consumed, and b) have that become inconsistent - i see i noted above that the cli output from the validator was different from the programmatic call.

on the other hand it's a minor change and random code jitter with no clear objective is bad for teams and incurs keep-up burden.

so as i said - not attached to this one, feel free to close this and i won't be offended, or feel free to merge it and i'll feel the mild satisfaction that comes from other people merging your code.

> What about like raise_for_results or something along those lines, as opposed to just raise_, for the parameter and method names?

no opinion - please feel free to rename to anything.

edit: actually opinion, i made this change

sneakers-the-rat:
actually if i were to make a change to this it would be to remove the `raise_` parameter and instead do

```diff
--- run_validation(obj_dict, schema, target_class, raise_=True)
+++ report = run_validation(obj_dict, schema, target_class)
+++ report.raise_for_results()
```

since i have grown more allergic to param spray since i wrote this. but it's basically the same thing as the material changes made here, just cleaner calling convention.

edit: no ya this version ([42bd438](https://github.com/linkml/linkml/pull/2385/commits/42bd43892e2d13a44b2f6833fd59c6b906ffc154)) is way better than this version ([faf6d89](https://github.com/linkml/linkml/pull/2385/commits/faf6d89447d79aa180709823df77a562db904e0f))

packages/linkml/src/linkml/utils/exceptions.py
```diff
@@ -7,5 +7,13 @@ class SchemaError(ValueError):
...
-class ValidationError(SchemaError):
+class DataError(ValueError):
+    """Base class for errors relating to linkml instance data"""
+
+
+class SchemaValidationError(SchemaError):
     """Schema is invalid!"""
+
+
+class ValidationError(DataError, SchemaError):
```
amc-corey-cox: Was the dual inheritance from both `DataError` and `SchemaError` intentional, or left over from the refactor? The old `ValidationError` inherited from `SchemaError` because it meant "schema is invalid," but now that it means "data is invalid relative to a schema" I'd expect it to inherit only from `DataError`.

Nothing in-tree catches `SchemaError` so it wouldn't break anything either way — just a question of whether the hierarchy should match the new semantics.

amc-corey-cox (approved):
Clean, small PR — the raise_for_results() pattern is a nice addition and the schema vs data error distinction makes sense. One minor question inline.

I'll plan to merge this next week if you don't ask me to wait.

packages/linkml/src/linkml/utils/exceptions.py
sneakers-the-rat: Good point, that does make the meaning of `SchemaError` ambiguous. I *think* what i was thinking is that a data validation error "is an error in the data as it relates to a schema", but i think that is captured by just the existence of `ValidationError` (validation is always only meaningful relative to a schema, duh), and a data validation error doesnt mean that anything is wrong with the schema (which is what schemaerror means/probably should mean).

tldr yes youre right ill remove

sneakers-the-rat: If you make a suggestion ill commit it so you're credited as co-author

sneakers-the-rat: Actually i need to amend prior commits to add @pkalita-lbl as co-author because the `raise_for_status` was his idea, so ill just do that manually, one sec.

amc-corey-cox: @sneakers-the-rat - I appreciate your thoughtfulness and you're welcome to attribute this to me if you don't want to own it but I have plenty of recognition in this repo and this work so I don't need you to go to extra effort to get me in there. Thanks for this great work!

amc-corey-cox:
@sneakers-the-rat - I don't think I see the commit you appear to have mentioned above. Do you want me to wait? I'll merge this in a few days if we don't hear from you. I don't think this needs to wait on that minor nit.

sneakers-the-rat:
oh crap sorry i totally forgot about that friday. will do now

packages/linkml/src/linkml/utils/exceptions.py
sneakers-the-rat: partially a credit thing, partially a memory thing: whose idea was this/who was thinking about this/who can explain it in the future if there are questions. Bus factor and all that

PR #2388, opened 2024-10-29
[perf][tests] 300% free perf by comparing RDF graph hashes

sneakers-the-rat:
Comparing RDF graphs is slow because of godforsaken bnodes. 

Thankfully, in our snapshot tests, we don't actually need to compare full graphs -- since we are just comparing whether or not they are the same, we can just compare the hash of a canonicalized graph to a stored hash.

The difference is like this when running just the slow tests (`pytest -m slow --with-slow`, and the profiler adds a deterministic amount of overhead):

Before 771s (443s in compare_rdf):
Screenshot 2024-10-28 at 5 20 19 PM

After 344s (134s in compare_rdf):
Screenshot 2024-10-28 at 5 20 27 PM

From these numbers it looks like the graphs are actually getting canonicalized thrice in the existing impl, so something is wrong there. 

The approach here falls back to the old graph comparison method if a hash isn't available or if the hashes don't match so we can see the graph diff. 

This doesn't fix the annoying problem of [unsorted ttl output](https://github.com/RDFLib/rdflib/pull/1978) that makes every PR that touches an RDF generator one bazillion lines because I figured we owuld want to still have the snapshots to see the diff, but it at least helps with testing time.

Opening as a draft because ideally we would still only canonicalize the generated graph one time, and if hash comparison fails we would use that already-canonicalized graph again instead of recomputing it, but saving that for if ppl like this idea. if u do, lmk and i'll finish implementing.

edit: aw dang doesn't seem to have as much of an extreme benefit in CI


sneakers-the-rat:
that's genuinely weird - i made it raise exceptions if the hash was wrong or missing rather than falling through to the default graph comparison behavior, and all the ubuntu tests fail with two errors: one from the hash not matching, the other from the hash file not having been found. 

however the windows tests pass????????????

sneakers-the-rat:
OK i think that one of those means that rdflib's graph canonicalization algo is in trouble, and the other one of those indicates platform differences that i don't really want to debug, so unless there is any interest in this i'm going to close this one as "too much effort"

PR #2389, opened 2024-10-29
[tests][ci] Run slow tests separately

sneakers-the-rat:
Another entry in "trying to make the tests run faster" - seeing how much faster we are when we split the biolink tests off into a separate job


sneakers-the-rat:
dats a bit better, 22m -> 15m. we could take it further by splitting off all the `slow` tests so we basically have two parallel testing jobs running if we want, but i figured just splitting biolink was different enough from the metamodel tests that it justified the reduced testing matrix where i don't feel the same about metamodel tests. i'll test that now just so we can see times

sneakers-the-rat:
ok so ya splitting all the slow tests off gets us down to 12-15m, but there's a substantial bit of jitter in the test times from startup time and also restoring the cache. python 3.12 is systematically really slow for some reason. 

now if we really want to make a dent we can split all slow tests and *also* flip on xdist with `-n auto`. Then that takes us from 22m down to like 6m. some of the ubuntu tests almost finish before the windows tests finish unzipping the cached environment lol. what we lose with xdist is all the results being all lined up by their filename, but we still get all the same output that's important - test failures and tracebacks - so i think it's worth it

to make that work i just had to make the temporary directories safer by using the full node name in the tests rather than just the function name, and there might be some random test combos that break when they're run in parallel, but hopefully not since the tests should all be independent except for those temporary directories.

sierra-moxon:
@sneakers-the-rat, this is terrific, as usual!! (I want to approve it, but I should probably run this change locally to see the output. You have direct access to make branches, etc., right?  -- I see its a fork'd branch)

sneakers-the-rat:
> I should probably run this change locally to see the output. 

To run locally, one would run the nonslow and slow tests in separate commands like this :)

```shell
# nonslow
pytest -n auto

# slow
pytest --with-slow --with-biolink -m slow -n auto
```

I could go either way with having a separate `--with-biolink` flag: to CI it makes no difference, since that flag is just fixed in the command, locally it adds an additional flag to remember. shouldn't add to maintenance burden because the biolink tests shouldn't change.

I think if we like this "splitting all the slow tests off into a separate job" strategy, which seems to be a relatively equal split between regular/slow tests, then we should probably mimic the whole test matrix (include the missing python versions) because there isn't a principled thing that makes a test "slow" like there is a thing that makes a test a "biolink test."

>  You have direct access to make branches, etc., right? 

I think so, i just usually pull from a fork because i always forget to delete the branch and it lets me treat the fork as a 'junk drawer' with branches lol. 

to check this branch out locally you could do

```shell
git remote add sneakers https://github.com/sneakers-the-rat/linkml
git fetch sneakers
git switch -c tests-split-biolink sneakers/tests-split-biolink
```

Silvanoc:
> now if we really want to make a dent we can split all slow tests and _also_ flip on xdist with `-n auto`. Then that takes us from 22m down to like 6m. some of the ubuntu tests almost finish before the windows tests finish unzipping the cached environment lol. what we lose with xdist is all the results being all lined up by their filename, but we still get all the same output that's important - test failures and tracebacks - so i think it's worth it

I was also struggling about it when thinking on speed-ups. On the theoretical analysis I was like "what you see on screen gets so ugly compared to the nicely lined-up reporting, I cannot propose that". Now that I "feel in my hand" what it brings, I am like "who cares for esthetics, if you are done in 1/2, 1/3 or even 1/4 of the time! 🙂

Silvanoc:
> to make that work i just had to make the temporary directories safer by using the full node name in the tests rather than just the function name, and there might be some random test combos that break when they're run in parallel, but hopefully not since the tests should all be independent except for those temporary directories.

I have also very rarely observed those race-conditions, that should not happen. IMO we should not stop this improvement because of that, but rather switch it on and then fix those issues as they appear.

Silvanoc:
@sneakers-the-rat you were right on Slack, I love it! ❤️

Unfortunately I am getting an error trying to run it locally:

``` shell
________ ERROR collecting docs/_build/_downloads/76cf7ccbaa949d8efed48ecf06d010de/test_excelgen.py _________
ImportError while importing test module '[...]/linkml/linkml/docs/_build/_downloads/76cf7ccbaa949d8efed48ecf06d010de/test_excelgen.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
[...]/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
docs/_build/_downloads/76cf7ccbaa949d8efed48ecf06d010de/test_excelgen.py:7: in <module>
    from tests.test_generators.environment import env
E   ModuleNotFoundError: No module named 'tests.test_generators.environment'
```

tests/test_biolink_model/test_biolink_model.py
```diff
@@ -19,6 +19,8 @@
 
 BIOLINK_NS = Namespace("https://w3id.org/biolink/vocab/")
 
+pytestmark = pytest.mark.biolink
+
 
 @pytest.mark.slow
```
vincentkelleher, line 25: Why not use the annotation directly ? 😇 

```suggestion


@pytest.mark.biolink
@pytest.mark.slow
```

vincentkelleher:
Great job on this one @sneakers-the-rat 👏   
This will greatly improve the CI pipeline !

Just tried it locally and things are looking good 👌 

```bash
pytest --with-slow --with-biolink -m biolink -v
```

image

sierra-moxon:
CQ checks have been fixed upstream if you want to pull them into your fork to get the tests on this PR passing.

vincentkelleher (approved):

sierra-moxon (approved):

PR #2390, opened 2024-10-29
Pass default values to pydantic via keyword argument

### LINKED ISSUE #2377: Unintuitive type errors in Pydantic classes with optional fields
**Describe the bug**
When code instantiating Pydantic models built through LinkML is type-checked with Pyright, optional fields not included in kwargs to the constructor are marked as missing. I have not tested, but I imagine this is the case with schemas with default values as well. The cause of this is described [here](https://github.com/pydantic/pydantic/issues/3617) and [here](https://github.com/pydantic/pydantic/issues/3753).

1.8.4

Given this schema:
```
id: http://example.com/thing
imports:
  - linkml:types

classes:
  Thing:
    attributes:
      id:
        identifier: true
      label:
        required: true
        range: string
      tag:
        range: string
```

And this example code, not including optional `tag`:
```
from model import Thing

thing = Thing(id="id", label="label")
```

Output from pyright:
```
$ pyright main.py 

/home/ptgolden/Code/tmp/linkml-pydantic/main.py
  /home/ptgolden/Code/tmp/linkml-pydantic/main.py:3:9 - error: Argument missing for parameter "tag" (reportCallIssue)
1 error, 0 warnings, 0 informations
```

---

This error is fixed by explicitly setting default values via keyword in generated Pydantic definitions. That is, changing the generated code from this:
```
class Thing(ConfiguredBaseModel):
    linkml_meta: ClassVar[LinkMLMeta] = LinkMLMeta({'from_schema': 'http://example.com/thing'})

    id: str = Field(..., json_schema_extra = { "linkml_meta": {'alias': 'id', 'domain_of': ['Thing']} })
    label: str = Field(..., json_schema_extra = { "linkml_meta": {'alias': 'label', 'domain_of': ['Thing']} })
    tag: Optional[str] = Field(None, json_schema_extra = { "linkml_meta": {'alias': 'tag', 'domain_of': ['Thing']} })

```
to this
```
class Thing(ConfiguredBaseModel):
    linkml_meta: ClassVar[LinkMLMeta] = LinkMLMeta({'from_schema': 'http://example.com/thing'})

    id: str = Field(..., json_schema_extra = { "linkml_meta": {'alias': 'id', 'domain_of': ['Thing']} })
    label: str = Field(..., json_schema_extra = { "linkml_meta": {'alias': 'label', 'domain_of': ['Thing']} })
    tag: Optional[str] = Field(default=None, json_schema_extra = { "linkml_meta": {'alias': 'tag', 'domain_of': ['Thing']} })

```

(Note the `default=` addition)

Passing the default keyword via a keyword should not be a change requiring much work, since it is the only allowed positional argument in `Field`.

ptgolden:
This is necessary to inform pyright/pylance that optional fields are, indeed, optional.

<https://docs.pydantic.dev/latest/integrations/visual_studio_code/#adding-a-default-with-field>

Fixes #2377


kevinschaper (approved):

linkml/generators/pydanticgen/template.py
```diff
@@ -195,11 +195,13 @@ class PydanticAttribute(PydanticTemplateModel):
...
-            return self.predefined
+            value = self.predefined
         elif self.required or self.identifier or self.key:
-            return "..."
+            value = "..."
         else:
-            return "None"
+            value = "None"
+
+        return f"default={value}"
```
sneakers-the-rat, line 204: I would prefer that this happens in the template, rather than the template model - the template model is intended as computer-readable data that can be serialized using a template, where all the information about the representation as a concrete pydantic module is in the templates, and the template models are just data containers. In general I think the amount of string manipulation that happens in pydanticgen and other generators makes them fragile and opaque, so it's my preference that we keep it at a minimum :). Promise i am not just being nitpicky.

See how the rest of the kwargs are specified in the template here: https://github.com/linkml/linkml/blob/08959cb73d8b09d66bbad730bb5f4d63f0195b08/linkml/generators/pydanticgen/templates/attribute.py.jinja#L1

sneakers-the-rat (changes requested):
I get where the need for this change comes from, if i am being honest i find it a bit ugly, and i am also not a fan of how apparently this is Microsoft being inflexible and exerting their infrastructural monopolies, but since it's functionally equivalent, resolves a bug for you, i have no substantive objections :).

Looks good to me except that I would prefer it if we kept details about serialization in the templates themselves and avoid string manipulation in the template models as much as possible and keep the template models as close to neutral carriers of data as possible. This is a division of labor that is particular to pydantic after seeing some of the problems with other generators - small change, just moving the `default=` into the template.

linkml/generators/pydanticgen/template.py
kevinschaper: Oh good catch! I’m still getting used to the template model, but I agree about keeping the value as a pure value and formatting in the Jinja.

sneakers-the-rat: one of those "seems obvious in retrospect but the value of modeling both the source and the target domain was only made clear after a lot of pain and wasted time (mostly pre-linkml)" lol

ptgolden:
> i am also not a fan of how apparently this is Microsoft being inflexible and exerting their infrastructural monopolies

I hear you. I will be more than happy to move on from pyright once another good python type checking LSP is released. Unfortunately, that will likely come from a [VC-backed company](https://astral.sh/), which is its own can of worms. Until then, a PEP is a PEP...

I'll make the changes. It felt icky to do string manipulation in the Python code, but I couldn't quite tell where `{{ field }}` was coming from in the Jinja template. I see it now-- I'll make that change.

sneakers-the-rat:
> It felt icky to do string manipulation in the Python code

haha i feel that so much. if only `ast` was a little easier to work with...

> I hear you. 

thanks for being patient with my tangential-to-this-PR-really aside.

sierra-moxon:
from dev call
- @ptgolden - can you migrate the change into the jinja template as per review and then we can get this merged :) thank you!

ptgolden:
Done :+1:

rly (approved):

sneakers-the-rat (approved):

PR #2416, opened 2024-11-14
feat(generators): option to use CURIEs instead of names in JSON-Schema and JSON-LD context

### LINKED ISSUE #3801: Add --use-curies option to emit CURIE keys in JSON-Schema and JSON-LD context generators

Add an option to the `gen-json-schema` and `gen-jsonld-context` generators to emit **CURIEs** (compact IRIs such as `schema:location`) as JSON-Schema property/`$defs` keys and JSON-LD `@context` keys, instead of element *names* (`location`).

CURIEs are derived from each element's `class_uri` / `slot_uri`. Note: CURIEs are compact IRIs, not full URIs/IRIs — full URIs are never emitted as keys; they are compressed to CURIEs via the schema's registered prefixes.

Motivation: CURIE keys enable JSON-LD [compact-IRI](https://www.w3.org/TR/json-ld11/#compact-iris) form, which resolves term clashes automatically when mixing vocabularies (e.g. `ical:location` vs `schema:location`) while keeping the compacted JSON readable. A JSON-Schema generated with the same CURIE keys can then validate that compacted JSON-LD.

Example — with a clash between `ical:location` and `schema:location`, a CURIE-keyed context + schema produces:

```json
{
  "ical:dstart": "2011-04-09T20:00:00Z",
  "ical:location": "New Orleans Arena, New Orleans, Louisiana, USA",
  "ical:summary": "Lady Gaga Concert",
  "schema:location": "Lady Gaga Concert"
}
```

Both keys coexist unambiguously and the JSON-Schema validates the compacted document.

This feature is also intended to **pave the path toward the broader "namespacing" topic** discussed in [#1739](https://github.com/orgs/linkml/discussions/1739). That discussion weighs a dedicated import/aliasing syntax (`ns.Object`) against leaning on LinkML's existing CURIE/URI machinery as the *single* namespacing mechanism, resolving name clashes URI-based (URIs globally unique, prefixes local to a schema). Emitting CURIE-keyed outputs is a concrete first step in the CURIE-based direction: it demonstrates URI-driven clash resolution in generated artifacts without introducing any new schema syntax.

**Additional context**

- Proposed CLI: `--use-curies` / `--not-use-curies` (default: off) on both generators.
- Elements **without** a declared URI would get a native CURIE synthesized from the schema `default_prefix` + normalized name (`prefix:CamelCase` classes, `prefix:snake_case` slots).
- Open design question: in schemas where some elements declare a URI and others don't, keys mix real CURIEs (`schema:Event`) with synthesized `<default_prefix>:<name>` — acceptable, or should it warn/normalize?
- Relates to namespacing discussion [#1739](https://github.com/orgs/linkml/discussions/1739).

Silvanoc:
## Summary

Fixes #3801

Adds a `--use-curies` flag to `gen-json-schema` and `gen-jsonld-context`. When enabled, generated JSON-Schema property/`$defs` keys and JSON-LD `@context` keys are **CURIEs** (compact IRIs such as `schema:location`) derived from each element's `class_uri` / `slot_uri`, instead of the element *names* (`location`).

CURIEs are compact IRIs, not full URIs/IRIs — the earlier "use uris" title/wording was imprecise. Full URIs are never emitted as keys; they are always compressed to CURIEs via the schema's registered prefixes.

**Why:** CURIE keys enable JSON-LD [compact-IRI](https://www.w3.org/TR/json-ld11/#compact-iris) form, which resolves term clashes automatically when mixing vocabularies (e.g. `ical:location` vs `schema:location`) while keeping the compacted JSON readable. A JSON-Schema generated with the same CURIE keys can then validate that compacted JSON-LD.

**Behavior:**

- `--use-curies` / `--not-use-curies` (default: off) on both generators.
- Keys resolved via `SchemaView.get_curie()`: `get_uri()` → `compress_uri()`.
- Elements **with** `class_uri`/`slot_uri`: declared URI compressed to CURIE.
- Elements **without** declared URI: native CURIE synthesized from `default_prefix` + normalized name (`prefix:CamelCase` classes, `prefix:snake_case` slots).
- Full URI with no matching prefix → `ValueError`; URL-like value with no prefix → warning.

This PR introduces the strict `compress_uri()` contract (raise on an unregistered prefix or an invalid URI/CURIE). Applying the same level of strictness to the inverse operation, `expand_curie()`, is intentionally left out of scope here and will be addressed in a separate follow-up PR, since it requires fixing several callers that currently rely on its lenient pass-through behavior.

## How was this tested?

- New tests: `test_jsonschemagen.py` and `test_jsonldcontextgen.py` cover the `--use-curies` option (incl. multi-ontology input `multiple-ontologies.yaml`).
- Full fast + slow suite for `linkml` and `linkml-runtime`.

## Areas of uncertainty

- **Mixed key styles:** in schemas where some elements declare a URI and others don't, output keys mix real CURIEs (`schema:Event`) with synthesized `<default_prefix>:<name>`. Is this acceptable, or should it warn/normalize?
- Warning-vs-error policy for URL-like values that can't be compressed.

## Checklist

## AI Assistance

### DIFF
--- a/packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
+++ b/packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import logging
 import os
+import re
 import sys
 import uuid
 import warnings
@@ -43,6 +44,7 @@ from linkml_runtime.utils.context_utils import map_import, parse_import_map
 from linkml_runtime.utils.formatutils import camelcase, is_empty, sfx, underscore
 from linkml_runtime.utils.namespaces import Namespaces
 from linkml_runtime.utils.pattern import PatternResolver
+from linkml_runtime.utils.uri_validator import validate_curie, validate_uri
 
 if TYPE_CHECKING:
     from collections.abc import Callable, Mapping
@@ -1454,6 +1456,34 @@ class SchemaView:
             return self.expand_curie(uri)
         return uri
 
+    def get_curie(
+        self,
+        element: ElementName | Element,
+        imports: bool = True,
+    ) -> str:
+        """Return the CURIE for a schema element, compressing full URIs where possible.
+
+        This is the CURIE-returning companion to :meth:`get_uri`.  It first resolves
+        the element's declared URI via :meth:`get_uri`, then compresses it to a CURIE
+        using the schema's registered prefixes via :meth:`compress_uri`.
+
+        When the declared URI is already a CURIE (the common case) it is returned
+        unchanged.  When no URI is declared, :meth:`get_uri` constructs the fallback
+        ``"<default_prefix>:<element_name>"`` CURIE, which is then returned as-is.
+        When the URI is a full URI that cannot be compressed, a :exc:`ValueError` is
+        raised.
+
+        :param element: Name of schema element or element object.
+        :param imports: Include imports closure when resolving the element.
+        :return: A CURIE string such as ``"schema:Event"`` or ``"ex:something_else"``.
+        :raises ValueError: When the element's URI is a fully-qualified URI with no
+            matching prefix registered in the schema, or when it is not a valid URI or
+            CURIE.
+        """
+        e = self.get_element(element, imports=imports)
+        uri = self.get_uri(e, imports=imports)
+        return self.compress_uri(uri)
+
     def expand_curie(self, uri: str) -> str:
         """Expand a URI or CURIE to a full URI.
 
@@ -1468,6 +1498,43 @@ class SchemaView:
                 if pfx in ns:
                     return ns[pfx] + local_id
         return uri
+
+    def compress_uri(self, uri: str) -> str:
+        """Compress a URI or CURIE to a CURIE string using the schema's registered prefixes.
+
+        This is the inverse of :meth:`expand_curie`.
+
+        Resolution order:
+
+        1. Try to compress *uri* with
+           :meth:`~linkml_runtime.utils.namespaces.Namespaces.curie_for`.
+        2. If compression fails but *uri* is already a valid CURIE, return it as-is
+           (emitting a warning when it looks like a plain ``http(s)://`` URL to alert
+           callers that disambiguation is impossible).
+        3. If *uri* is a fully-qualified URI with no matching prefix, raise
+           :exc:`ValueError`.
+        4. If *uri* is neither a valid URI nor a valid CURIE, raise :exc:`ValueError`.
+
+        :param uri: A URI or CURIE string.
+        :return: A CURIE string such as ``"schema:Event"`` or ``"ex:something_else"``.
+        :raises ValueError: When *uri* is a fully-qualified URI whose namespace is not
+            registered, or when it is not a syntactically valid URI or CURIE.
+        """
+        ns = self.namespaces()
+        curie = ns.curie_for(uri) if validate_uri(uri) else None
+        if curie:
+            return curie
+        if validate_curie(uri):
+            if re.match("https?://.*", uri):
+                logger.warning(
+                    f"'{uri}' looks like a URL but no registered prefix matches it; "
+                    "it cannot be unambiguously compressed to a CURIE."
+                )
+            return uri
+        elif validate_uri(uri):
+            raise ValueError(f"'{uri}' cannot be converted to a CURIE, corresponding prefix not defined")
+        else:
+            raise ValueError(f"'{uri}' does not seem to be either a valid URI or a valid CURIE")
 
     @lru_cache(CACHE_SIZE)
     def get_elements_applicable_by_identifier(self, identifier: str) -> list[str]:
--- a/packages/linkml/src/linkml/generators/jsonldcontextgen.py
+++ b/packages/linkml/src/linkml/generators/jsonldcontextgen.py
@@ -76,6 +76,9 @@ class ContextGenerator(Generator):
     _local_slots: set | None = field(default=None, repr=False)
     _external_classes: set | None = field(default=None, repr=False)
     _external_slots: set | None = field(default=None, repr=False)
+    use_curies: bool = False
+    """If true, use class_uri/slot_uri CURIEs as context keys instead of element names."""
+
     xsd_anyuri_as_iri: bool = False
     """Map xsd:anyURI-typed ranges (uri, uriorcurie) to ``@type: @id`` instead of ``@type: xsd:anyURI``.
 
@@ -238,6 +241,16 @@ class ContextGenerator(Generator):
 
         return str(as_json(context)) + "\n"
 
+    def _curie(self, element: ClassDefinition | SlotDefinition) -> str:
+        if self.schemaview is None:
+            source = self.schema.source_file or self.schema
+            if isinstance(source, str) and self.base_dir and not Path(source).is_absolute():
+                source = str(Path(self.base_dir) / source)
+            sv = SchemaView(source, importmap=self.importmap, base_dir=self.base_dir)
+        else:
+            sv = self.schemaview
+        return sv.get_curie(element)
+
     def visit_class(self, cls: ClassDefinition) -> bool:
         if self.exclude_imports and cls.name not in self._local_classes:
             return False
@@ -245,7 +258,10 @@ class ContextGenerator(Generator):
             return False
 
         class_def = {}
-        cn = camelcase(cls.name)
+        if self.use_curies:
+            cn = self._curie(cls)
+        else:
+            cn = camelcase(cls.name)
         self.add_mappings(cls)
 
         self._build_element_id(class_def, cls.class_uri)
@@ -412,7 +428,10 @@ class ContextGenerator(Generator):
                 self._build_element_id(slot_def, slot.slot_uri)
                 self.add_mappings(slot)
         if slot_def:
-            key = underscore(aliased_slot_name)
+            if self.use_curies:
+                key = self._curie(slot)
+            else:
+                key = underscore(aliased_slot_name)
             self.context_body[key] = slot_def
 
             # collect @embed only for object-valued slots (range is a class)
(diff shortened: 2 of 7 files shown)


hadjian:
To add to the PR description, this new mode would enable the [compact IRI](https://www.w3.org/TR/json-ld11/#example-32-using-vocabularies) feature of JSON-LD. Also see the "event" example in the [JSON-LD playground](https://json-ld.org/playground/).

## Use Case: Current Situation

A compacted JSON-LD is easier to process for developers, as it shortens key names to simple strings instead of IRIs. On the other hand, IRIs enable the mixing of vocabularies by avoiding name clashes. When compacting to simple strings, name clashes need to be resolved for those simple strings. Take this JSON-LD document as an example:

```
  {
    "http://www.w3.org/2002/12/cal/ical#dstart": [
      {
        "@type": "http://www.w3.org/2001/XMLSchema#dateTime",
        "@value": "2011-04-09T20:00:00Z"
      }
    ],
    "http://www.w3.org/2002/12/cal/ical#location": [
      {
        "@value": "New Orleans Arena, New Orleans, Louisiana, USA"
      }
    ],
    "http://www.w3.org/2002/12/cal/ical#summary": [
      {
        "@value": "Lady Gaga Concert"
      }
    ]
  }
```

Now we apply the following context:
```
{
  "@context": {
    "ical": "http://www.w3.org/2002/12/cal/ical#",
    "summary": {
      "@id": "ical:summary"
    },
    "location": {
      "@id": "ical:location"
    },
    "dtstart": {
      "@type": "xsd:dateTime",
      "@id": "ical:dstart"
    }
  }
}
```
Run the JSON-LD compaction algorithm and get this nice JSON document (omitting the context on purpose):
```
{
  "summary": "Lady Gaga Concert",
  "location": "New Orleans Arena, New Orleans, Louisiana, USA",
  "dtstart": "2011-04-09T20:00:00Z"
}
```
I can use ```gen-json-schema``` to create a schema that can validate this JSON. Nice!

## Problem

We will run into problems, if two terms clash. For the sake of demonstration, we add the location from schema.org to our JSON-LD document:

```
  {
    "http://www.w3.org/2002/12/cal/ical#dstart": [
      {
        "@type": "http://www.w3.org/2001/XMLSchema#dateTime",
        "@value": "2011-04-09T20:00:00Z"
      }
    ],
    "http://www.w3.org/2002/12/cal/ical#location": [
      {
        "@value": "New Orleans Arena, New Orleans, Louisiana, USA"
      }
    ],
    "https://schema.org/location": [
       {
         "@value": "New Orleans Arena, New Orleans, Louisiana, USA"
       }
    ],
    "http://www.w3.org/2002/12/cal/ical#summary": [
      {
        "@value": "Lady Gaga Concert"
      }
    ]
  }

```

Now the context has to resolve this conflict, by assigning a different term w.r.t. the ical location:
```
{
  "@context": {
    "ical": "http://www.w3.org/2002/12/cal/ical#",
    "summary": {
      "@id": "ical:summary"
    },
    "location": {
      "@id": "ical:location"
    },
   "slocation": "https://schema.org/location",
    "dtstart": {
      "@type": "xsd:dateTime",
      "@id": "ical:dstart"
    }
  }
}
```
Which results in the compacted form:
```
{
  "summary": "Lady Gaga Concert",
  "location": "New Orleans Arena, New Orleans, Louisiana, USA",
  "slocation": "New Orleans Arena, New Orleans, Louisiana, USA",
  "dtstart": "2011-04-09T20:00:00Z"
}
```
## Suggestion

Add a flag to ```gen-json-schema``` and ```gen-jsonld-context``` to use compact IRIs. This will resolve clashes automatically and still make the compacted JSON readable. The JSON-LD context will then look as follows:

```
{
  "@context": {
    "ical": "http://www.w3.org/2002/12/cal/ical#",
    "schema": "http://schema.org/",
    "ical:dstart": {
      "@type": "xsd:dateTime"
    },
    "ical:location": {
      "@id": "ical:location"
    },
    "ical:summary": {
      "@id": "ical:summary"
    },
    "schema:location": {
      "@id": "schema:location"
    }
  }
}
```
and the corresponding compacted JSON as follows:
```
{
  "ical:dstart": "2011-04-09T20:00:00Z",
  "ical:location": "New Orleans Arena, New Orleans, Louisiana, USA",
  "ical:summary": "Lady Gaga Concert",
  "schema:location": "Lady Gaga Concert"
}
```
If the generated json schema now also contains these compact IRIs, we can validate the JSON and use the context for expansion afterwards.

Readable and name clashes automatically resolved.

Silvanoc:
According the information provided by @hadjian in [this comment](https://github.com/linkml/linkml/pull/2416#issuecomment-2477659462), we also need to extend the JSON-LD Context generation in order to get the same IDs using CURIEs.

Silvanoc:
This PR is still at an early phase.

TODOs:

- [x] Rename option from `--prefer-uris` to `--use-curies`.
- [x] Convert local names to CURIEs (simply appending the default prefix), whenever the use of CURIEs is requested.
- [x] Ensure use of CURIEs (not URIs), converting URIs to CURIEs whenever possible
- [x] Warn of URLs without a known prefix.
- [x] Adapt also JSON-LD context generation to use "namespaced" CURIEs instead of relying on the [default vocabulary](https://www.w3.org/TR/json-ld11/#default-vocabulary).

Silvanoc:
Although this PR would be ready for merging on the JSON-Schema generator, it is not of much use without the needed changes in the JSON-LD Context generator, according [this comment](https://github.com/linkml/linkml/pull/2416#issuecomment-2478205844).

Due to lack of time I am now completing it and therefore it remains as _Draft_, but @hadjian or me later on might take it over.

Silvanoc:
Within the scope of this PR bugs #3749 and #3751 have been discovered, PR #3750 fixes them and therefore this PR will depend on that merge.

packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
```diff
@@ -1469,6 +1499,43 @@ def expand_curie(self, uri: str) -> str:
...
+           (emitting a warning when it looks like a plain ``http(s)://`` URL to alert
+           callers that disambiguation is impossible).
+        3. If *uri* is a fully-qualified URI with no matching prefix, raise
+           :exc:`ValueError`.
+        4. If *uri* is neither a valid URI nor a valid CURIE, raise :exc:`ValueError`.
+
+        :param uri: A URI or CURIE string.
+        :return: A CURIE string such as ``"schema:Event"`` or ``"ex:something_else"``.
+        :raises ValueError: When *uri* is a fully-qualified URI whose namespace is not
+            registered, or when it is not a syntactically valid URI or CURIE.
```
kevinschaper: The behaviour described here doesn't match what the code does, and it took me a while to work out which one was wrong. It's the docs — the code is deliberate and your tests pin it.

`validate_curie` (existing, in `uri_validator.py`, not part of this PR) accepts a plain URL, because a URL satisfies the CURIE grammar — `https` parses as the prefix and `//host/path` as the reference:

```
'schema:Event'                        validate_curie=True
'https://unregistered.example/Thing'  validate_curie=True    <-- 
'urn:uuid:1234'                       validate_curie=False
```

So an unregistered `http(s)://` URI is claimed by branch 2 and returned unchanged with a warning. Branch 3 as written can only ever fire for non-URL URIs like `urn:` — measured:

```
'https://unregistered.example/Thing'  -> returned unchanged (+ warning)
'urn:uuid:1234'                       -> ValueError
```

I initially assumed this was a bug and tried reordering the checks so URLs take the error path. That breaks `test_use_uris[True]` in both generators, because `multiple-ontologies.yaml` deliberately includes `slot_uri: https://saref.etsi.org/saref4auto/RendezvousLocation` with no `saref4auto` prefix, and the test explicitly asserts the warning is emitted. So warn-and-pass-through is intended — it's only the documentation that promises an error.

Suggestion below just makes the docstring describe the actual behaviour. The same correction is needed in the PR description, which says *"Full URI with no matching prefix → ValueError; URL-like value with no prefix → warning"* — those are the same case, and it's always the warning.

One thing worth deciding separately, since it's the practical consequence: with `--use-curies` that slot's key comes out as a full URL alongside real CURIEs —

```
s4city:Event properties: ['ex:something_else',
                          'https://saref.etsi.org/saref4auto/RendezvousLocation',
                          's4ehaw:hasLocation']
```

That's a third key style beyond the two you raise under "mixed key styles" in the description. Emitting a non-CURIE key from a `--use-curies` run may still be the right call (failing the whole generation over one unregistered prefix is harsh), but it's worth saying so explicitly in the flag's help text so users aren't surprised.

```suggestion
        1. Try to compress *uri* with
           :meth:`~linkml_runtime.utils.namespaces.Namespaces.curie_for`.
        2. If compression fails but *uri* is already a valid CURIE, return it as-is.
           Note that a plain ``http(s)://`` URL satisfies the CURIE grammar (``https``
           parses as the prefix), so an unregistered URL lands here rather than in the
           error case below: it is returned unchanged and a warning is emitted, and
           callers may therefore receive a full URL where a CURIE was requested.
        3. If *uri* is a non-URL URI (e.g. ``urn:``) with no matching prefix, raise
           :exc:`ValueError`.
        4. If *uri* is neither a valid URI nor a valid CURIE, raise :exc:`ValueError`.

        :param uri: A URI or CURIE string.
        :return: A CURIE string such as ``"schema:Event"`` or ``"ex:something_else"``,
            or the original string when it is a URL with no registered prefix.
        :raises ValueError: When *uri* is a non-URL URI whose namespace is not
            registered, or when it is not a syntactically valid URI or CURIE.
```

kevinschaper (approved):
Thanks @Silvanoc! it's great to see an older PR making it in, it gives me hope that I'll get back to some of mine

Silvanoc:
@matentzn may I merge it once the tests are green? For me, being the assignee means that I can merge once the requirements (approval from core-team and tests pass) are fulfilled. But I prefer to confirm before doing so.

kevinschaper:
@Silvanoc I think I just forgot to hit merge when ready on this. Since it’s up to date with main and able to be merged without a branch update and new test run, I’m going to take the opportunity and smash the big green button.

PR #2426, opened 2024-11-22
fix(run-examples): handle inlined dictionaries

### LINKED ISSUE #2423: run-examples fails on inlining as simple dictionary
**Describe the bug**

Running `linkml examples` fails when [inlining as simple dictionary](https://linkml.io/linkml/schemas/inlining.html#inlining-as-simple-dictionaries) being used. More details to reproduce it below.

The error being reported is (being `prop1` a key of the simple dictionary):
```
ValueError: Example valid-examples/Thing-001.yaml failed validation:
No such slot prop1 as an attribute of Property ancestors or as a slot definition in the schema
```

[Validation](https://github.com/linkml/linkml/blob/da294e6f3bf5c36c7d7333c7e7312ad70b45941b/linkml/workspaces/example_runner.py#L185) is fine, the issue appears when [trying to load the object](https://github.com/linkml/linkml/blob/da294e6f3bf5c36c7d7333c7e7312ad70b45941b/linkml/workspaces/example_runner.py#L249) from a [previously created dictionary](https://github.com/linkml/linkml/blob/da294e6f3bf5c36c7d7333c7e7312ad70b45941b/linkml/workspaces/example_runner.py#L177) loading the example. At that point the key of the simple dictionary is being interpreted as a slot of an object, therefore it fails. As of now, I cannot tell if the issue is how the dictionary is being created or how the object is being loaded from the dictionary, but I suspect that it is the second.

**Version of LinkML you are using** v1.8.5

Simple schema to reproduce the issue:
```yaml
# schema.yaml
id: https://example.org/
name: example
prefixes:
  example: https://example.org/
default_prefix: example
default_range: string
imports:
  - linkml:types
classes:
  Thing:
    attributes:
      properties:
        range: Property
        multivalued: true
        inlined: true
        inlined_as_list: false
  Property:
    attributes:
      name:
        identifier: true
      value:
        required: true
```

Simple valid example:
```yaml
# valid-examples/Thing-001.yaml
properties:
  prop1: "something"
  prop2: "some other thing"
```

Command line to reproduce the issue: `linkml examples -s schema.yaml -e valid-examples -d output`.

### LINKED ISSUE #2425: run-examples fails on inlining as dictionary
**Describe the bug**

The generic form of #2423. The same issue appears on inlined dictionaries.

**Version of LinkML you are using** v1.8.5

```yaml
# schema.yaml
id: https://example.org/
name: example
prefixes:
  example: https://example.org/
default_prefix: example
default_range: string
imports:
  - linkml:types
classes:
  Thing:
    attributes:
      parameters:
        range: Parameter
        multivalued: true
        inlined: true
        inlined_as_list: false
  Parameter:
    attributes:
      name:
        identifier: true
      value1:
        required: true
      value2:
        required: true
```

```yaml
# valid-examples/Thing-001.yaml
parameters:
  parm1:
    value1: "a value"
    value2: "another value"
  parm2:
    value1: "yet another value"
    value2: "a value"
```

Silvanoc:
run-examples was not capable of handling inlined dictionaries correctly. This patch fixes it.

This PR overrides PR #2424 

TODOS:

- [x] Add regression tests

Fixes #2423
Fixes #2425

### DIFF
--- a/linkml/generators/jsonschemagen.py
+++ b/linkml/generators/jsonschemagen.py
@@ -25,6 +25,7 @@ from linkml.generators.common import build
 from linkml.generators.common.lifecycle import LifecycleMixin
 from linkml.generators.common.type_designators import get_type_designator_value
 from linkml.utils.generator import Generator, shared_arguments
+from linkml.utils.helpers import get_range_associated_slots
 
 logger = logging.getLogger(__name__)
 
@@ -545,7 +546,7 @@ class JsonSchemaGenerator(Generator, LifecycleMixin):
                         range_id_slot,
                         range_simple_dict_value_slot,
                         range_required_slots,
-                    ) = self._get_range_associated_slots(slot)
+                    ) = get_range_associated_slots(self.schemaview, slot.range)
                     # if the range class has an ID and the slot is not inlined as a list, then we need to consider
                     # various inlined as dict formats
                     if range_id_slot is not None and not slot.inlined_as_list:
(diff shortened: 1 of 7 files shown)


sierra-moxon (approved):
Thanks @Silvanoc!

Issue #2400, opened 2024-11-07
SHACL generation ignores `all_of`, `none_of`, and `exactly_one_of`


elichad:
**Describe the bug**
@floWetzels, @dnlbauer and I are developing a LinkML schema and converting to SHACL. When we use `any_of`, this is converted to a `sh:or` statement in the SHACL, as expected.
LinkML:
```
slots:
  main_entity:
    required: true
    slot_uri: schema:mainEntity
    description: Check if the Main Workflow is specified through a `mainEntity` property in the root data entity
    annotations:
      sh:message: The Main Workflow must be specified through a `mainEntity` property in the root data entity
    any_of:
      - range: Person
      - range: ComputationalWorkflow
      - range: Workflow
```
Result:
```ro-crate:RootDataEntity a sh:NodeShape ;
    sh:closed false ;
    sh:ignoredProperties ( rdf:type ) ;
    sh:property [ sh:description "Check if the Main Workflow is specified through a `mainEntity` property in the root data entity" ;
            sh:maxCount 1 ;
            sh:minCount 1 ;
            sh:or ( [ sh:class schema1:Person ] [ sh:class bioschemas:ComputationalWorkflow ] [ sh:class schema1:CreativeWork ] ) ;
            sh:order 0 ;
            sh:path schema1:mainEntity ] ;
    sh:targetClass ro-crate:RootDataEntity .
```

However, when we use `all_of` with the same constraints, this is **not** represented in the SHACL, and the default `sh:datatype xsd:string` is used. We expected to see a `sh:and` statement, but there is nothing like this present. The same is true when using `none_of` and `exactly_one_of`.
```
slots:
  main_entity:
    required: true
    slot_uri: schema:mainEntity
    description: Check if the Main Workflow is specified through a `mainEntity` property in the root data entity
    annotations:
      sh:message: The Main Workflow must be specified through a `mainEntity` property in the root data entity
    all_of:
      - range: Person
      - range: ComputationalWorkflow
      - range: Workflow
```
SHACL:
```
ro-crate:RootDataEntity a sh:NodeShape ;
    sh:closed false ;
    sh:ignoredProperties ( rdf:type ) ;
    sh:property [ sh:datatype xsd:string ;
            sh:description "Check if the Main Workflow is specified through a `mainEntity` property in the root data entity" ;
            sh:maxCount 1 ;
            sh:message "The Main Workflow must be specified through a `mainEntity` property in the root data entity"^^xsd:string ;
[...]

PR #2441, opened 2024-12-06
Shacl generator: Include slot annotations when if `any_of` is used

### LINKED ISSUE #2417: SHACL annotations are skipped on slots/properties where `any_of` is used
**Describe the bug**
 I have a slot called `license` as follows (the specific details aren't important):
```
license:
    required: true
    slot_uri: schema:license
    title: Crate license
    description: The Crate must specify a license
    annotations:
      sh:message: The Crate (Root Data Entity) must specify a license, which should be a URL but can also be a string
    any_of:
      - range: CreativeWork
      - range: string
```
When this is converted to SHACL (with `--include-annotations`), the relevant output is:
```
sh:property [ sh:description "The Crate must specify a license" ;
            sh:maxCount 1 ;
            sh:minCount 1 ;
            sh:name "Crate license" ;
            sh:or ( [ sh:class schema1:CreativeWork ] [ sh:datatype xsd:string ;
                        sh:nodeKind sh:Literal ] ) ;
            sh:order 1 ;
            sh:path schema1:license ],
```
Note that the `sh:message` annotation is missing.
When I _remove_ the `any_of` statement and use simply `range: string` instead, the output becomes:
```
sh:property [ sh:datatype xsd:string ;
            sh:description "The Crate must specify a license" ;
            sh:maxCount 1 ;
            sh:message "The Crate (Root Data Entity) must specify a license, which should be a URL but can also be a string"^^xsd:string ;
            sh:minCount 1 ;
            sh:name "Crate license" ;
            sh:nodeKind sh:Literal ;
            sh:order 1 ;
            sh:path schema1:license ],
```
`sh:message` has now been included.

I have other slots in the same schema which also use annotations with `sh:message` - only the slot using `any_of` is affected.
This omission of `sh:message` does _not_ occur when I use `all_of` instead of `any_of` - but I think that is because that doesn't get converted to SHACL at all, see #2400.

1.8.4

You can include this in th issue if small, or provide a link e.g. to github.
Ideally a minimal example that replicates the problem is ideal, but only if this is easy for you.

```
id: https://github.com/crs4/rocrate-validator/profiles/workflow-ro-crate/linkml
name: workflow_ro_crate
prefixes:
  schema: http://schema.org/
  ORCID: https://orcid.org/
  bioschemas: https://bioschemas.org/
  roterms: http://purl.org/ro/roterms#
  rdf: http://www.w3.org/1999/02/22-rdf-syntax-ns#
  sh: http://www.w3.org/ns/shacl#
  ro-crate: https://github.com/crs4/rocrate-validator/profiles/ro-crate/
  workflow-ro-crate: https://github.com/crs4/rocrate-validator/profiles/workflow-ro-crate/
imports:
  - linkml:types
default_range: string

classes:
  # validation class
  WorkflowROCrateRootDataEntity:
    title: WROC Root Data Entity Required Properties
    description: Root Data Entity properties defined as MUST
    class_uri: ro-crate:RootDataEntity
    slots:
      - main_entity
      - license

  # schema.org wrappers
  Workflow:
    class_uri: schema:CreativeWork
  CreativeWork:
    class_uri: schema:CreativeWork

# individual property requirements
slots:
  main_entity:
    required: true
    slot_uri: schema:mainEntity
    range: Workflow
    title: Main Workflow entity existence
    description: Check if the Main Workflow is specified through a `mainEntity` property in the root data entity
    annotations:
      sh:message: The Main Workflow must be specified through a `mainEntity` property in the root data entity
  license:
    required: true
    slot_uri: schema:license
    title: Crate license
    description: The Crate must specify a license
    annotations:
      sh:message: The Crate (Root Data Entity) must specify a license, which should be a URL but can also be a string
    any_of:
      - range: CreativeWork
      - range: string
```

elichad:
Fixes #2417 .

### DIFF
--- a/linkml/generators/shaclgen.py
+++ b/linkml/generators/shaclgen.py
@@ -198,14 +198,15 @@ class ShaclGenerator(Generator):
                         add_simple_data_type(prop_pv, r)
                     if s.pattern:
                         prop_pv(SH.pattern, Literal(s.pattern))
-                    if s.annotations and self.include_annotations:
-                        self._add_annotations(prop_pv, s)
                     if s.equals_string:
                         # Map equal_string and equal_string_in to sh:in
                         self._and_equals_string(g, prop_pv, [s.equals_string])
                     if s.equals_string_in:
                         # Map equal_string and equal_string_in to sh:in
                         self._and_equals_string(g, prop_pv, s.equals_string_in)
+
+                if s.annotations and self.include_annotations:
+                    self._add_annotations(prop_pv, s)
 
                 default_value = ifabsent_processor.process_slot(s, c)
                 if default_value:
--- a/tests/test_generators/input/shaclgen_boolean_constraints.yaml
+++ b/tests/test_generators/input/shaclgen_boolean_constraints.yaml
@@ -1,0 +1,19 @@
+id: https://w3id.org/linkml/examples/boolean_constraints
+name: test_boolean_constraints
+prefixes:
+  linkml: https://w3id.org/linkml/
+  ex: https://w3id.org/linkml/examples/boolean_constraints/
+imports:
+  - linkml:types
+default_range: string
+default_prefix: ex
+
+classes:
+  AnyOfSimpleType:
+    attributes:
+      attribute1:
+        any_of:
+          - range: string
+          - range: integer
+        annotations:
+          resting: supine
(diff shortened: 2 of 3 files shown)


linkml/generators/shaclgen.py
```diff
@@ -198,15 +198,16 @@ def st_node_pv(p, v):
                         add_simple_data_type(prop_pv, r)
                     if s.pattern:
                         prop_pv(SH.pattern, Literal(s.pattern))
```
elichad, line 200: Should these lines should also be moved in the same way? I'm not sure if only one `pattern` should be expected for a slot or if there can be multiple `patterns` in an `any_of`

tests/test_generators/input/shaclgen_boolean_constraints.yaml
elichad, line 1: I envision this file could be expanded when addressing #2400

sierra-moxon (approved):

PR #2449, opened 2024-12-13
Identifiers cannot be null

### LINKED ISSUE #2448: Generated JSON-Schema allows "null" on identifiers
**Describe the bug**
Generated JSON-Schema allows "null" on identifiers (slots with the attribute `identifier: true`) if a `range` is specified and `required` is not explicitly specified.

The `range` can be explicitly declared in the slot or over `default_range: [...]`.

**Version of LinkML you are using** Any from v1.8.2 upwards

Git history bisected to identify commit 2bd1c19892201b2d2030a8708cf4801fe276ea9a as the one introducing the bug.

Schema to reproduce the issue:
```yaml
id: https://example.org/nullable-id
name: nullable-id
prefixes:
  ni: https://example.org/nullable-id/
imports:
  - linkml:types
default_range: string
classes:
  MyClass:
id: https://example.org/nullable-id
name: nullable-id
prefixes:
  ni: https://example.org/nullable-id/
imports:
  - linkml:types
default_range: string
  MyClass:
    attributes:
      id:
        identifier: true
      needed:
        required: true
      name:
```

Resulted in following JSON-Schema extract in version v1.8.2:
```json
{
  "id": {
    "type": [
      "string",
      "null"
    ]
  },
  "name": {
    "type": [
      "string",
      "null"
    ]
  },
  "needed": {
    "type": "string"
  }
}
```

Silvanoc:
`identifier`s cannot be `null` and the generated JSON-Schema must not allow it.

Fixes #2448


amc-corey-cox:
I merged main to resolve the conflict. Feel free to rebase if you prefer a cleaner history.

I think the root cause is in `SchemaView.induced_slot()`. At [`schemaview.py` lines 1704-1707](https://github.com/linkml/linkml/blob/cbfdf952f/packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py#L1704-L1707):

```python
if slot.inlined_as_list:
    slot.inlined = True
if slot.identifier or slot.key:
    slot.required = True
```

These lines modify `slot` (the original) instead of `induced_slot` (the copy that gets returned). The copy is made earlier at line 1661, so the returned `induced_slot` never gets `required = True` set. Changing `slot` to `induced_slot` on those four lines should fix it.

amc-corey-cox:
A few notes on the test:

- `top_class="Optionals"` doesn't match the schema — `identifier.yaml` only defines `MyClass`. This might cause the test to pass vacuously or error rather than actually testing the fix.
- The `for cls in ["MyClass"]` loop over a single item could just be a direct assignment.
- It would be good to also cover `key: true` slots, since they have the same "automatically required" semantics per the metamodel spec.

amc-corey-cox:
Updated the PR body to also reference Fixes #2451, since the root cause fix in `SchemaView.induced_slot()` covers both issues.

amc-corey-cox:
Rebased onto main to fix the monorepo path conflicts (tests moved from `tests/test_generators/` to `tests/linkml/test_generators/`). Also removed `top_class="Optionals"` since the schema only defines `MyClass` (per my earlier comment).

Test logic, assertions, and schema are unchanged.

amc-corey-cox:
@Silvanoc I rebased your branch onto main to fix the path changes from our monorepo migration and marked the test as `xfail(strict=True)` so CI passes while the bug is still open. Your test and schema are unchanged other than removing `top_class="Optionals"` (per my earlier comment).

This is ready to merge as-is to get the regression test in. The root cause fix (`schemaview.py` writing to `slot` instead of `induced_slot`) could be a second PR or you can work on it here. How would you like to proceed — take on the fix here, or just merge it?

amc-corey-cox:
The CI failure is unrelated to this and is being addressed in #3330 - it may have been transient and fixed now.

PR #2453, opened 2024-12-13
Issue 2447 - add cli option for mermaid with or without markdown compatibility using existing 'format' parameter

### LINKED ISSUE #2447: Add command line option to `gen-erdiagram` to switch between mermaid and markdown
The CLI for `gen-erdiagram` is missing an option to switch between generating a mermaid diagram that can be embedded in markdown and one without the markdown wrapper.

There are already tests for markdown vs mermaid generation; it just looks like someone forgot to put in the CLI option.

This should be a pretty quick, easy fix.

sierra-moxon:
fixes #2447 

I had to edit the docs-test GH action in this PR because it was using a cached version of a venv that was causing the check to fail and the easiest way to fix (the failure was due to a lack of  fresh, crispy venv, not code changes) was to clear the cache and pull this GH action up to the same technology (e.g. pipx, etc.) as our other actions.


tests/test_generators/test_erdiagramgen.py
```diff
@@ -131,3 +139,19 @@ def test_max_hops(kitchen_sink_path):
...
+def test_format_option_markdown(runner, kitchen_sink_path):
+    """Test the --format option with 'markdown'."""
+    result = runner.invoke(cli, [kitchen_sink_path, "--format", "markdown"])
+    print(result.output)
+    assert result.output.startswith("```mermaid")
+    assert result.output.endswith("```\n\n")
+
+
+def test_format_option_no_markdown(runner, kitchen_sink_path):
+    """Test the --format option with 'markdown'."""
```
pkalita-lbl, line 153:
```suggestion
    """Test the --format option with 'mermaid'."""
```

pkalita-lbl (approved):

PR #2461, opened 2024-12-16
Inlined types management by the JSON-LD context generator

vincentkelleher:
During the generation of our updated ontology we noticed that the JSON-LD context generator didn't seem to manage inlined types correctly.

When an object isn't inlined, it is referenced through its identifier with `"@type": "@id"`, in the case of an inlined object it should be of its class' type with `"@type": "myPrefix:myClass"`.

## Example

For example the following LinkML file:

```yaml
id: https://example.org/external_type
prefixes:
  ex: https://example.org/external_type#
  cred: https://www.w3.org/2018/credentials#

imports:
  - linkml:types

default_prefix: ex
default_range: string

slots:
  id:
    identifier: true
    range: string

  credential:
    title: "Credential"
    description: "Some credential"
    range: VerifiableCredential
    inlined: true
    required: true

classes:
  VerifiableCredential:
    class_uri: cred:VerifiableCredential
```

Would generate the below JSON-LD context:

```json
{
   "@context": {
      "cred": "https://www.w3.org/2018/credentials#",
      "ex": "https://example.org/external_type#",
      "@vocab": "https://example.org/external_type#",

      "id": "@id",

      "credential": {
         "@type": "@id",
         "@id": "credential"
      },
      
      "VerifiableCredential": {
         "@id": "cred:VerifiableCredential"
      }
   }
}
```

Whereas the desired output would be:

```json
{
   "@context": {
      "cred": "https://www.w3.org/2018/credentials#",
      "ex": "https://example.org/external_type#",
      "@vocab": "https://example.org/external_type#",

      "id": "@id",

      "credential": {
         "@type": "cred:VerifiableCredential",
         "@id": "credential"
      },
      
      "VerifiableCredential": {
         "@id": "cred:VerifiableCredential"
      }
   }
}
```


sierra-moxon:
Thank you @vincentkelleher!   I imagine the test failures here are a result of our snapshot system storing copies of the previously generated content, disagreeing with the updated/fixed content.   Would you be able to run a refresh of your snapshots to see if these build errors go away?  https://linkml.io/linkml/contributing/contributing.html#custom-pytest-fixtures

vincentkelleher:
@sierra-moxon thanks for the heads up for the failing tests, I pushed the changes of this PR trusting my local tests using the `--with-slow` argument only, I wasn't aware of the `--with-network` argument which has been introduced recently 😅 

For changes like this one, I go through each snapshot and test case manually to make sure I didn't break anything 😇 

I fixed the tests and everything is ready to be merged ✅

sierra-moxon (approved):

PR #2468, opened 2024-12-20
add linter check that one and only one tree_root is found if `TreeRootClassRule` config rule is configured

### LINKED ISSUE #2466: `linkml lint` does not complain about multiple classes with `tree_root: true`
**Describe the bug**
Having multiple classes with `tree_root: true` does not represent a valid schema in my opinion. 

The linter does not pick this issue up, and claims that the schema is completely valid. 

Further, generators such as the json-schema generator seem to arbitrarily choose one of the classes with `tree_root: true` to be the winner. This isn't clearly defined behaviour (if it turns out that it is valid to have multiple classes with `tree_root: true`).

linkml, version 1.8.5

```yaml
id: http://example.com/something/schema
name: something
prefixes: 
  some: http://example.com/something/
imports:
  - linkml:types

classes:
  Something:
    tree_root: true
    class_uri: some:Something
    description: A thing, something.
    attributes:
      id: 
        required: true
        identifier: true
        slot_uri: some:id
        description: An identifier.
        range: integer
      name: 
        required: true
        slot_uri: some:name
        description: A name.
        range: string

  Container:
    tree_root: true
    attributes:
      values:
        multivalued: true
        inlined_as_list: true
        range: Something
```

The behaviour of the linter:

```bash
$ linkml lint something-else.yaml 
.../something-else.yaml
  warning  Class 'Container' does not have recommended slot 'description'  (recommended)
  warning  Slot 'values' does not have recommended slot 'description'  (recommended)

✖ Found 2 problems in 1 schema
```

sierra-moxon:
This partially fixes #2466.  The linter will complain if the TreeRootClassRule is configured and the schema being linted has more than one tree root. As a separate PR, we should fix the generators to explicitly handle multiple tree roots if we want to keep multiple tree root classes as a feature.  

I also converted the tests for this to pytest (and added two new tests).


sierra-moxon:
@pkalita-lbl - I wasn't sure if I should make a new rule for this or add the check to the existing rule.  I fit it in with the existing rule class.  This means that more than one problem per rule can be returned...does this fit with your architectural design?

pkalita-lbl (approved):
I think it makes sense to add this check to the existing rule, as you've done here. The [docs](https://linkml.io/linkml/schemas/linter.html#tree-root-class) say:

> Require a single class with tree_root: true and optionally verify that class’s name

So the rule wasn't really totally making good on that promise; It was checking if there was less than one but not more than one.

Issue #2450, opened 2024-12-13
gen-doc throws error "AttributeError: type object 'NoneType' has no attribute 'class_name'" when trying to generate slot docs


timalamenciak:
**Describe the bug**
gen-doc appears to stop after generating documentation for the classes in my YAML file. Error seems to originate from slots.

```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\Tim Alamenciak\AppData\Local\Programs\Python\Python312\Scripts\gen-doc.exe\__main__.py", line 7, in <module>
  File "C:\Users\Tim Alamenciak\AppData\Local\Programs\Python\Python312\Lib\site-packages\click\core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Tim Alamenciak\AppData\Local\Programs\Python\Python312\Lib\site-packages\click\core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "C:\Users\Tim Alamenciak\AppData\Local\Programs\Python\Python312\Lib\site-packages\click\core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Tim Alamenciak\AppData\Local\Programs\Python\Python312\Lib\site-packages\click\core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Tim Alamenciak\AppData\Local\Programs\Python\Python312\Lib\site-packages\linkml\generators\docgen.py", line 1023, in cli
    print(gen.serialize())
          ^^^^^^^^^^^^^^^
  File "C:\Users\Tim Alamenciak\AppData\Local\Programs\Python\Python312\Lib\site-packages\linkml\generators\docgen.py", line 216, in serialize
    out_str = template.render(gen=self, element=c, schemaview=sv, **template_vars)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #2200, opened 2024-07-09
`gen-doc` error message for slot with no range is uninformative


turbomam:
```yaml
name: assumption_schema
id: http://example.com/assumption_schema

imports:
  - linkml:types

prefixes:
  linkml: https://w3id.org/linkml/

classes:
  Assumption:
    slots:
      - assumes

slots:
  assumes: {}
```

```shell
gen-doc -d docs assumption_schema.yaml
```

>AttributeError: type object 'NoneType' has no attribute 'class_name'

turbomam:
`gen-owl` and `gen-linkml` do not complain at all

sujaypatil96:
Technically it is okay (considered to be valid LinkML) to have the `range` on a `SlotDefinition` be empty because it is not _required_?

I'm wondering if we should have the linter complain/report this as a warning. `gen-doc`'s job is to render details about the schema as is.

timalamenciak:
Could this be the same issue as #2450 ? I'm working on a big schema and found this error message to be uninformative. Still trying to iron it out.

PR #2473, opened 2024-12-24
Addresses issues #2450 and #2200

timalamenciak:
Hi there - I think this code will fix the error being thrown by gen-doc when it tries to generate documentation with no range. I could not figure out how to test on my local machine, so my apologies for this. Applies method inject_slot_info to the schema ahead of template generation which applies default range to each slot that doesn't have a range.

### DIFF
--- a/linkml/generators/docgen.py
+++ b/linkml/generators/docgen.py
@@ -203,6 +203,9 @@ class DocGenerator(Generator):
             "diagram_type": self.diagram_type.value if self.diagram_type else None,
             "include_top_level_diagram": self.include_top_level_diagram,
         }
+        self.logger.debug("Validating ranges")
+        for slot in sv.all_slots():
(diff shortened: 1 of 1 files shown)


vincentkelleher:
Hi @timalamenciak,

As you can probably see, your change breaks some test suites that need some attention 😇 

image

I suggest you get the tests up and running on your local machine before submitting a pull request.  
To do so, you will need to install Poetry, you can find documentation about this here 👉  https://python-poetry.org/docs/#installation.

Once you are done with the Poetry installation, you can run the whole test process with the following command:

```bash
poetry run coverage run -m pytest --with-slow --with-network
```

If you want to run a single test file, you can either use your favorite IDE or the following command (example for doc-gen tests):

```bash
pytest tests/test_generators/test_docgen.py
```

Hope this helps 😉

PR #2480, opened 2025-01-07
initial dbml generator

### LINKED ISSUE #2114: Create a DBML exporter
DBML docs
https://dbml.dbdiagram.io/home

Technically we don't need to write any code here - we export and import to SQL so we could just add docs on how to chain tools together

But in practice it might be nice to have a one-stop command that will generate DBML directly

There is no driving requirement here as I think we have visualization tools well covered, but opening this as it may be a good first issue for someone

Note that DBML is used by a lot of genomics projects like GREGOR/PRIMED, but it looks like their authoring workflow is sheets->custom json [analogous to schemasheets] and then json->dbml->viz

e.g. https://github.com/UW-GAC/primed_data_models/blob/main/json_to_dbml.R

sierra-moxon:
fixes #2114 

There is a very nice interactive display tool documented in the issue above for viewing and editing schemas in DBML format (https://trudan.github.io/dbdiagram-oss/editor).  One of our collaborating projects uses this tool to help model in real-time in working groups, and this generator code is my first step in making an addition to schema-automator to go from DBML -> LinkML as well. 

As documented in the ticket, we can always convert to SQL DDL and then use the DBML open-source tool to convert again to DBML, but why not have a direct generator that can convert LinkML->DBML. 

Plenty left to do here; enums, etc, in subsequent PRs.


pkalita-lbl (commented):
Two suggestions:

* Drop your changes in `.github/workflows/main.yaml` and merge the changes from `main` into this branch. The conflicting changes accomplish the same thing in the end.
* Hook your new CLI into the `linkml generate` CLI. See `linkml/cli/main.py`.

pkalita-lbl (approved):

PR #2486, opened 2025-01-09
Issue 1720 convert test prefixes to pytests

tfliss:
I split this into three commits.

- 1-1 conversion from unittest to pytest (note I removed __main__)
- changed the path globals to pytest fixtures and used tmp_path for the ones that were writing to output
- removed the output files that were only being written, not read.

**Notes**

 - The diff on the .py file is rendering poorly. It comes out much cleaner in my VSCode.

- test_rdfgen is still marked skip, though when I ran it, it passed on my machine.
I left it skipped because issue 537 seemed to mention the issue was intermittent and it was out of scope of this issue to investigate.

- tmp_path could be specified in the pyproject.toml, I did not make that change because it would affect
other places where tmpdir and tmp_path are used.


pkalita-lbl (approved):
Great! Thank you

Issue #314, opened 2021-08-31
Ensure all jupyter notebooks are re-executed and succeed using github actions


cmungall:
The second cell here has an error: https://github.com/linkml/linkml/blob/main/notebooks/examples.ipynb

We should run tests to ensure the notebooks all run from beginning to end

Replaces #147 and #139

gaurav:
> We should run tests to ensure the notebooks all run from beginning to end

One way of doing that is to use `nbconvert` using the `--execute` flag to re-run all cells and convert the output into Markdown or PDF -- this should [produce a `CellExecutionError`](https://nbconvert.readthedocs.io/en/latest/execute_api.html#handling-errors-and-exceptions) if execution fails.

nlharris:
Is this still needed?

tfliss:
At least some of the tests under test_notebooks/ are still skipped. The remaining seems to be a Latin-1 or similar default encoding on Win32. I did make some other updates to the notebooks and tests (including test_notebooks.py) in issue https://github.com/linkml/linkml/issues/1720

dalito:
I could not reproduce the encoding issue on German Windows but had a javascript issue for `yamlmagic`  with current jupyter notebook (7.3.2). 

Other examples use shell commands which are not cross-platform and could be replaced by python code or by using sqlite3 in Python and not via shell commands (in `SQL-examples.ipynb`).

PR #2488, opened 2025-01-10
1720 convert tests in teststest notebooks to pytest

tfliss:
Tests converted to pytest.
There was one print statement that now uses a logger.

I went ahead and attempted minor updates to the failing notebooks. A local diff tool will work best for viewing these. I also recommend viewing them directly i.e. in github linkml/linkml/notebooks.

- There seems to be a remaining non-utf-8 default encoding issue in the Windows github actions, so I relaxed the notebook test skip to only skip on Windows.

- I stopped the jupyter kernels and saved the notebooks to preserve output but remove some of the internal metadata.

- The output in DistributeModels.ipynb has some codespell errors, so I skipped it in pyproject.toml


tfliss:
test_examples.py still needs to be updated.

tests/test_notebooks/test_notebooks.py
```diff
@@ -25,6 +26,7 @@ def force_rewrite_comparator(expected: str, actual: str) -> str:
     return msg
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="Figure out charset fail on windows in github actions")
```
dalito, line 29: Is there an issue for the bug that you can link to? I think there should be if tests are skipped although a problem still exists.

tfliss: @dalito Great point. I found this issue that seems appropriate. https://github.com/linkml/linkml/issues/314

The strategy I'd suggest is to mark this PR (2488) complete for review (I'm past the suggested 5-hour december data day time)

Then I can update 314 with info about the tests still writing and modifying notebooks/tmp and the skip condition because those things seem more in the 314 scope.

Please let me know if you'd like me to refer to issue 314 in the skip reason.

dalito: Yes, referring to #314 would be helpful to understand the skip.

tfliss: Thanks, @dalito I updated the reason and commented Issue 314 with the specific failure message.

dalito (approved):
Looks good to me. It worked fine in my local testing.

The notebooks may be further improved in later PRs. Then it will be nice to have the tests infrastructure from this PR in place.

sierra-moxon (approved):

sneakers-the-rat:
(thanks for this!)

Issue #2502, opened 2025-01-17
SHACL generation: `ifabsent` handling incomplete for many defaults


jsheunis:
Many functions for mapping default values in the [`shacl_ifabsent_processor`](https://github.com/linkml/linkml/blob/main/linkml/generators/shacl/shacl_ifabsent_processor.py) are not implemented yet and are raising `NotImplementedError()`.

More specifically, I am interested in implementing the `map_uri_or_curie_default_value` function:

https://github.com/linkml/linkml/blob/main/linkml/generators/shacl/shacl_ifabsent_processor.py#L64

because we need this for a specific schema. See https://github.com/psychoinformatics-de/datalad-concepts/issues/219

I have a patch working for our purposes:

```diff
diff --git a/linkml/generators/shacl/shacl_ifabsent_processor.py b/linkml/generators/shacl/shacl_ifabsent_processor.py
index 14b4c21f..d5476678 100644
--- a/linkml/generators/shacl/shacl_ifabsent_processor.py
+++ b/linkml/generators/shacl/shacl_ifabsent_processor.py
@@ -62,7 +62,8 @@ class ShaclIfAbsentProcessor(IfAbsentProcessor):
         return Literal(f"{year}-{month}-{day}T{hour}:{minutes}:{seconds}", datatype=ShaclDataType.DATETIME.uri_ref)
 
     def map_uri_or_curie_default_value(self, default_value: str, slot: SlotDefinition, cls: ClassDefinition):
-        raise NotImplementedError()
+        uri = URIRef(self.schema_view.namespaces().uri_for(default_value))
+        return Literal(uri, datatype=ShaclDataType.URI.uri_ref)
 
     def map_curie_default_value(self, default_value: str, slot: SlotDefinition, cls: ClassDefinition):
         return Literal(default_value, datatype=ShaclDataType.CURIE.uri_ref)
```

The referenced issue also gives an example of the output this yields for a given schema.

It would be great to get more pairs of eyes on these two lines of code before I create a PR.

vincentkelleher:
Hi @jsheunis,

Indeed, there are many mapping methods that return a `NotImplementedError` 👍 
It was a choice I made to avoid being to opiniated on code I wasn't very familiar with.

[...]

PR #2503, opened 2025-01-17
Adds SHACL `ifabsent` handling for type `uriorcurie`

jsheunis:
This replaces the raising of `NotImplementedError` for `ifabsent` handling of type `uriorcurie` during shacl generation with two lines of code. The relevant test is also updated.

Closes https://github.com/linkml/linkml/issues/2502

### DIFF
--- a/tests/test_generators/shacl/test_shacl_ifabsent_processor.py
+++ b/tests/test_generators/shacl/test_shacl_ifabsent_processor.py
@@ -213,6 +213,40 @@ def test_process_ifabsent_uri():
     ) == Literal("https://en.wikipedia.org/wiki/Student", datatype=ShaclDataType.URI.uri_ref)
 
 
+def test_process_ifabsent_uriorcurie():
+    schema_view = SchemaView(
+        schema_base
+        + """
+    - name: definition_A
+      range: uriorcurie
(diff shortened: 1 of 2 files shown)


tests/test_generators/shacl/test_shacl_ifabsent_processor.py
```diff
@@ -213,6 +213,40 @@ def test_process_ifabsent_uri():
     ) == Literal("https://en.wikipedia.org/wiki/Student", datatype=ShaclDataType.URI.uri_ref)
 
 
+def test_process_ifabsent_uriorcurie():
+    schema_view = SchemaView(
+        schema_base
+        + """
+    - name: definition_A
+      range: uriorcurie
+      ifabsent: string(https://en.wikipedia.org/wiki/Student)
```
vincentkelleher, line 222: I'm wondering if this should be `string(..)` or `uriorcurie(..)` or `uri(..)` 🤷

jsheunis: Yeah, I wasn't sure about that. The test pass for all, and the result is the same, so I think it depends on how declarative one wants to be in the data itself.

pkalita-lbl (approved):

PR #2508, opened 2025-01-21
Add a new argument to the SHACL generator to exclude imported elements

vincentkelleher:
The new `--exclude-imports` argument is meant to allow end-users to allow generate SHACL triples that corresponding to the currently being processed schema hence ignoring the schemas in the `imports:` section.

This is relevant in the case where someone just wants to validate a certain class that inherits attributes/slots or in the case where one has a base SHACL file that can be merged with any number of custom SHACL files (built with `--exclude-imports`) to support new entities.


vincentkelleher:
Strangely CodeSpell is failing on a file I didn't modify 🤔 
Furthermore, the fail isn't justified as `anc` is a diminutive for `ancestor`.

image

sierra-moxon:
@vincentkelleher - can you pull main into your fork branch?  The codespell/formatting issues have been fixed in main.

amc-corey-cox (approved):
This looks good to me.

Issue #1722, opened 2023-11-13
Convert tests in `tests/test_rewrite_rules` to pytest


pkalita-lbl:
* Ensure all tests are written in `pytest` style (top-level test functions instead of `unittest.TestCase` subclasses, plain `assert` statements)
* Ensure tests only write to temporary files, if needed (by default running tests should not change checked-in files)
* Remove any `print` statements, use logging judiciously.

tfliss:
These tests all seem to be skipped, and converting to fixtures is moderately complex, so I don't want to do the conversion without trying to get the tests running.

I'm still trying to figure out exactly what is being tested, redirects within w3id.org, redirects within another server that w3id.org points to, or url handling functionality within LinkML (it doesn't appear to import linkml). If it's testing a services' behavior I recommend the test at least be marked 'network'.

The failures all seem to be of the form:

```
tests/test_rewrite_rules/test_rewrite_rules.py:139: in test_it
    self.assertEqual(expected, actual, f"redirect for: {resp.url}")
E   AssertionError: 'https://linkml.github.io/meta' != 'https://linkml.github.io/linkml-model/docs/meta'
E   - https://linkml.github.io/meta
E   + https://linkml.github.io/linkml-model/docs/meta
E   ?                          ++++++++++++++++++
E    : redirect for: https://w3id.org/linkml/meta
```

where `linkml-model/docs` is inserted into the url.

sierra-moxon:
Please see this branch: [1722-convert-test_rewrite_rules-to-pytest](https://github.com/linkml/linkml/tree/1722-convert-test_rewrite_rules-to-pytest) for work towards completing this issue and build on it from there. 
https://github.com/linkml/linkml/pull/2512

tfliss:
> Please see this branch: [1722-convert-test_rewrite_rules-to-pytest](https://github.com/linkml/linkml/tree/1722-convert-test_rewrite_rules-to-pytest) for work towards completing this issue and build on it from there. [#2512](https://github.com/linkml/linkml/pull/2512)

Thanks @sierra-moxon I just merged up that branch from main and the CI tests have passed in a new PR. Thanks for being patient with this one! https://github.com/linkml/linkml/pull/2704. The diff looks smaller in VSCode compared to how it shows in github.

PR #2512, opened 2025-01-28
converted test_rewrite_rules to pytest and unskipped

tfliss:
This is a fairly straightforward unittest -> pyunit conversion. The diff is minimal **when viewed in VS Code**

**Note** that in build_test_entry_set() and generate_fixture_lists() I had to **modify all of the expected URLs** to get the tests to pass. I am not entirely certain that the modifications are appropriate. Please let me know if someone specific should be on the review list.

Details:

- I changed SKIP_REWRITE_RULES to False, but left the code in.
- The TestsList class is now a method that creates parametrized test inputs, but basically the same shape.
- The setup and teardown methods are fixtures and top-level methods (I'm not sure that record_results has an effect, but I left it in.
- There's a single parametrized test now, but it's still split into the same four categories as previously.
- I used pytest skip, but also marked the test with 'network'
- removed the main clause.

### DIFF
--- a/tests/test_rewrite_rules/test_rewrite_rules.py
+++ b/tests/test_rewrite_rules/test_rewrite_rules.py
@@ -1,17 +1,14 @@
 import os
-import sys
-import unittest
 from dataclasses import dataclass
 from typing import List, Optional, Set, Tuple, Union
 
+import pytest
 import requests
 from rdflib import Namespace, URIRef
 
-if __name__ == "__main__":
-    sys.path.append(os.path.join(os.path.dirname(__file__), "..", ".."))
-
 from tests import SKIP_REWRITE_RULES, SKIP_REWRITE_RULES_REASON
 
+# see README.md
 W3ID_SERVER = "https://w3id.org/"
 DEFAULT_SERVER = W3ID_SERVER
 # DEFAULT_SERVER = "http://localhost:8091/"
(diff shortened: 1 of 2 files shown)


sneakers-the-rat:
i am pretty sure this is already tested here https://github.com/linkml/linkml-runtime/blob/main/tests/test_linkml_model/test_linkml_files.py

in general i think we want to avoid tests that require real network access, and also move towards single-sourcing these values so that we don't need to test them at all - need to run out the door atm, but if my skim of this is correct we're testing that some hardcoded (or generated?) urls correctly redirect? that should be static/declarative, which is what i tried to do over here - https://github.com/linkml/linkml-runtime/blob/main/linkml_runtime/linkml_model/linkml_files.py

tests/test_rewrite_rules/test_rewrite_rules.py
```diff
@@ -32,141 +29,134 @@ class TestEntry:
...
+            if resp.status_code == 302 and "location" in resp.headers
+            else f"Error: {resp.status_code}"
+        )
+        if fail_on_error:
+            assert expected == actual, f"redirect for: {resp.url}"
+            record_results(results, e.input_url, accept_header, actual)
+            return True
+        elif expected != actual:
+            print(f"{e.input_url} ({accept_header}):\n expected {expected} - got {actual}")
+            return False
```
sneakers-the-rat, line 151: antipattern - test result should be meaningful and uniform. A failing test only fails, and a passing test only passes. printing rather than raising (and generally making tests lenient and not emit expected output in CI) has been the cause of a ton of hard to diagnose problems.

tests should never print - pytest shouldn't emit to stdout, and if we are running tests such that they do output to stdout, that's a sign we're running things outside of pytest and thus not in a consistent environment. the printed information can be gotten by looking at the pytest summary, and one can run individual test modules and individual tests themselves with pytest (i.e. we should never run tests by just running them as python modules).

tests/test_rewrite_rules/test_rewrite_rules.py
```diff
@@ -32,141 +29,134 @@ class TestEntry:
...
+        TestEntry(input_url[".context.jsonld"], f"linkml-model/linkml_model/jsonld/{model}.context.jsonld"),
+        TestEntry(input_url[".owl"], f"linkml-model/linkml_model/owl/{model}.owl.ttl"),
+        TestEntry(input_url["/"], f"linkml-model/{model}/"),
     ]
 
 
-class TestLists:
-    __test__ = False
+def generate_fixture_lists():
+    server = os.environ.get("SERVER", DEFAULT_SERVER)
```
sneakers-the-rat, line 44: implicit env var: when would this need to change?

tests/test_rewrite_rules/test_rewrite_rules.py
```diff
@@ -32,141 +29,134 @@ class TestEntry:
...
-        TestEntry(input_url[".owl"], f"includes/{model}.owl"),
-        TestEntry(input_url["/"], f"includes/{model}/"),
+        TestEntry(input_url, f"linkml-model/docs/{model}"),
+        TestEntry(input_url, f"linkml-model/linkml_model/model/schema/{model}.yaml", "text/yaml"),
+        TestEntry(input_url, f"linkml-model/linkml_model/rdf/{model}.ttl", "text/turtle"),
+        TestEntry(input_url, f"linkml-model/linkml_model/json/{model}.json", "application/json"),
+        TestEntry(input_url, f"linkml-model/linkml_model/shex/{model}.shex", "text/shex"),
+        TestEntry(input_url[".context.jsonld"], f"linkml-model/linkml_model/jsonld/{model}.context.jsonld"),
+        TestEntry(input_url[".owl"], f"linkml-model/linkml_model/owl/{model}.owl.ttl"),
+        TestEntry(input_url["/"], f"linkml-model/{model}/"),
```
sneakers-the-rat, line 39: avoid hardcoding, single-source from `linkml_runtime.linkml_model.linkml_files`

tests/test_rewrite_rules/test_rewrite_rules.py
```diff
@@ -32,141 +29,134 @@ class TestEntry:
...
+    "entries",
+    [
+        pytest.param(fixture_lists["meta_entries"], id="meta_entries"),
+        pytest.param(fixture_lists["vocab_entries"], id="vocab_entries"),
+        pytest.param(fixture_lists["meta_model_entries"], id="meta_model_entries"),
+        pytest.param(fixture_lists["meta_vocab_entries"], id="meta_vocab_entries"),
+    ],
+)
+def test_rewrite_rules(entries: List[TestEntry], results, fail_on_error) -> None:
+    def test_it(e: TestEntry, accept_header: str) -> bool:
```
sneakers-the-rat, line 129: why the closure here? also unsure why we are collections results with `record_results` - that's what pytest does for us if we just fail on the cases that fail :)

tfliss:
> i am pretty sure this is already tested here https://github.com/linkml/linkml-runtime/blob/main/tests/test_linkml_model/test_linkml_files.py
> 
> in general i think we want to avoid tests that require real network access, and also move towards single-sourcing these values so that we don't need to test them at all - need to run out the door atm, but if my skim of this is correct we're testing that some hardcoded (or generated?) urls correctly redirect? that should be static/declarative, which is what i tried to do over here - https://github.com/linkml/linkml-runtime/blob/main/linkml_runtime/linkml_model/linkml_files.py

Thank you for your detailed notes, Jonny. Thanks for pointing out the the linkml_files tests, I hadn't found that. But the redirect test also covers the NGINX configuration, with a local or network option: https://github.com/linkml/linkml/tree/main/tests/test_rewrite_rules/httpd

I should have also linked to the issue, where I have a comment w/ additional context. However if this test is truly redundant it should be removed rather than updated. https://github.com/linkml/linkml/issues/1722

My changes were limited to 1-1 unittest -> pyunit changes, though I did 'unskip' the tests and change the expected URLs to get them passing. The limited scope of the diff is far clearer in VSCode diff than github diff, I'm trying to find a better online option.

I agree with most of the comments here, though I'm also trying to stay within the scope of the December Developer Days recommendations to limit scope to five hours and cut a new issue if needed. 
https://docs.google.com/document/d/1_ZB_L4NlIjsqfdwigIkbvLeW5FX7hWcY7ML0r4PPIrY

Again thanks for the detailed feedback, If there is a preferred avenue to ask some of these questions in advance of a PR, let me know. For now I will convert this to draft and comment briefly on my thinking on the specific requested changes.

tests/test_rewrite_rules/test_rewrite_rules.py
tfliss: This was previously done in the unittest main clause, which I removed for pyunit style.
[httpd/README.md Testing the Rewrite Rules](https://github.com/linkml/linkml/tree/main/tests/test_rewrite_rules/httpd#testing-the-rewrite-rules) addresses this question, 
I can add a clearer link to it in the code
Also noticed the test command would need to be updated in those instructions.

tests/test_rewrite_rules/test_rewrite_rules.py
tfliss: This was mostly a whitespace change when I pulled the test methods out of the unittest class. I also changed FAIL_ON_ERROR to the fail_on_error() fixture for more pyunit style. Git diffs make it seem like a bigger change.

That said, I should have changed the prints to logging per the issue.
I did consider removing this whole section (and still can if that's preferred).

tests/test_rewrite_rules/test_rewrite_rules.py
tfliss: Ah, I hadn't looked for this in linkml_runtime, yes I can do that update if this test isn't found to be redundant.
Was hoping a source for these already existed.

tests/test_rewrite_rules/test_rewrite_rules.py
tfliss: These functions were also nested in the original code. I limited the scope of my changes not knowing if it was an intended stylistic convention and to keep the VSCode diff clean. I did consider removing the record_results and still can, but wanted to check w/ team first.

sneakers-the-rat: i should have waited to leave comments until i could do a close read, sorry. will do that shortly. agreed on the scope of the PR being a pytest transformation and understood leaving things in place bc unsure of why they are that way - i do the same in this package frequently.

tests/test_rewrite_rules/test_rewrite_rules.py
sneakers-the-rat: Aha I see that content type part now. This should also be single sourced, s.t. the .htaccess file that defines them ( https://github.com/perma-id/w3id.org/blob/master/linkml/.htaccess ) is generated from that source and we can do a single diff of that file against the expected generated copy rather than needing to check every URL.

Beyond scope of this PR, but writing to say that having these pieces be decoupled and free floating is the source of the need for these tests and usually good to address root problems rather than testing symptoms when possible

sneakers-the-rat (commented):
Ok having taken a fuller look at this: This test supplements the [existing URL tests in `linkml-runtime`](https://github.com/linkml/linkml-runtime/blob/main/tests/test_linkml_model/test_linkml_files.py) by testing rewrite rules that change the redirect from the w3id permalink based on the value of the HTTP `accepted-type`. Appreciate @tfliss trying to do a minimal change to clean up the tests and move them to pytest. 

## what is this test? 

My biggest question here is "what is this test for?" 

Are these content-type based redirects used by linkml? Is it something that we advertise as being possible to do to a downstream consumer? What is the benefit of using the `content-type` redirect vs the full URL? 

In the weakest form of what this test could be testing, we are just testing that the `.htaccess` file is correctly written and apache correctly applies it (here https://github.com/perma-id/w3id.org/blob/master/linkml/.htaccess ). In that case, we can remove this test entirely because we shouldn't be responsible for testing Apache or the w3id system.

This test *seems* to be written to test whether some hardcoded set of URIs resolve to a different hardcoded set of URIs. This would be a useful test if we used these URIs within linkml (in particular used the `content-type` method of selecting different generated modalities of a given schema). This test is similar to the "snapshot" tests in that what it actually tests is whether some value has changed since the last time the tests passed. This kind of "snapshot" test is relatively weak because it doesn't test anything about the expected behavior of the system, just that it is unchanged - this test is a good example of that problem, because if I add an additional assertion that the redirected URLs actually exist (i.e. after redirects they are not 404'd), 27 of the 53 URLs do not exist. For example `w3id.org/linkml/meta/Element` with type `text/yaml` redirects to `https://linkml.github.io/linkml-model/linkml_model/model/schema/meta/Element.yaml/Element.yaml` but no such URI exists.

It seems like what we *really* want is to test whether the canonical URI (the w3id one) for each of the base linkml schemas correctly resolve to their relevant artifacts, whether that be through a URI with the extension ( `w3id.org/linkml-model/Element.yaml` ) or with a `accept` header. This test does not do that because a) it uses its own hardcoded set of canonical/"source" URIs rather than URIs from any source that would be used at runtime, b) it uses its own hardcoded redirect/"target" URIs rather than those that should be derived from the process that generates the built artifacts (i believe that is `linkml-model` but i am not quite sure how that process works), and c) it does not test whether those links actually resolve.

Consequently, this test should be a part of the `linkml-model` repo to test that on deployment of new versions of built artifacts, all expected canonical URLs should resolve by matching those canonical URIs (from a single source used across linkml and its documentation) to the results of the build process - no network access needed, just testing against the expected structure of the metamodel docs site. Only then does it actually protect us from breaking URI resolution, because it would prevent us from deploying a broken version. Tests that access the network are not generally appropriate in library packages, but they are appropriate for websites that are deployed via CD, so live URL resolution could be tested there too.

## structure of the test

Aside from its purpose, and respecting that this is intended to be a "touch as little as possible" PR, this test can be simplified considerably. 

The `httpd` module seems to be from a different time, where it sets up a local apache server that matches the w3id configuration (i assume) and then tests URI resolution against that local server. It uses a hardcoded `.htaccess` file that does not match the one deployed to `w3id.org` - avoiding these kinds of 'orphaned' values is why i am harping so much on single sourcing everything.  That falls in the first category of "testing whether apache correctly applies .htaccess files" because everything about the test is isolated from the actual system we want to test - the `.htaccess` file, the source, and target URIs are all local to the test itself and do not necessarily have anything to do with `linkml` the package or `linkml` the set of schemas at URIs as deployed in the wild. There appears to be no way to run the test with the local server except for manually (we have recently added the ability to run/manage docker containers for the plantuml tests, but it is not used here). With all due respect for the work that went into preparing it, in my opinion the `httpd` module can and should be removed because it does not test what we want it to.

The leg of the test that checks whether resolution works when an `accept` header is given as well as when the default `accept` header is appended to it can also be removed, as this again tests whether Apache correctly follows the HTTP spec.

I'll hold off on a fully simplified version that uses `linkml_runtime.linkml_model.linkml_files`, but here's a patch that shows some examples of how this could be simplified: rm extraneous config values, flatten nested lists so that each test is a single case, use assertions instead of printing/logging, etc.

<details>
<summary>expand/collapse patch</summary>

```diff
diff --git a/tests/__init__.py b/tests/__init__.py
index d3310499..cf0b1a5a 100644
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -58,12 +58,6 @@ SKIP_RDF_COMPARE_REASON = test_settings.get(
 # Skip tests that rely on an external SPARQL endpoint
 SKIP_REMOTE_SPARQL_TESTS = test_settings.getboolean("SKIP_REMOTE_SPARQL_TESTS", True)
 
-# Skip Rewrite rules tests -- these only get re-tested when we change the w3id.org server
-SKIP_REWRITE_RULES = test_settings.getboolean("SKIP_REWRITE_RULES", False)
-SKIP_REWRITE_RULES_REASON = test_settings.get(
-    "SKIP_REWRITE_RULES_REASON", "tests/__init__.py SKIP_REWRITE_RULES is True"
-)
-
 # Action on mismatch.  One of 'Ignore', 'Report' or 'Fail'
 #  If 'Fail', the expected file will be saved in the appropriate temp directory
 #  NOTE: Before setting this back to Report or Ignore, you need to run cleartemp.sh in this directory
diff --git a/tests/test_rewrite_rules/test_rewrite_rules.py b/tests/test_rewrite_rules/test_rewrite_rules.py
index 63ff3afb..5e446028 100644
--- a/tests/test_rewrite_rules/test_rewrite_rules.py
+++ b/tests/test_rewrite_rules/test_rewrite_rules.py
@@ -1,18 +1,14 @@
 import os
 from dataclasses import dataclass
-from typing import List, Optional, Set, Tuple, Union
+from typing import Dict, List, Optional, Set, Tuple, Union
 
 import pytest
 import requests
 from rdflib import Namespace, URIRef
 
-from tests import SKIP_REWRITE_RULES, SKIP_REWRITE_RULES_REASON
-
 # see README.md
 W3ID_SERVER = "https://w3id.org/"
 DEFAULT_SERVER = W3ID_SERVER
-# DEFAULT_SERVER = "http://localhost:8091/"
-# SKIP_REWRITE_RULES = False
 
 # Taken from Firefox network.http.accept.default
 default_header = "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"
@@ -24,7 +20,7 @@ class TestEntry:
     __test__ = False
     input_url: Union[str, URIRef, Namespace]
     expected_url: str
-    accept_header: Optional[str] = None
+    accept_header: Optional[str] = default_header
 
 
 def build_test_entry_set(input_url: Namespace, model: str) -> List[TestEntry]:
@@ -40,8 +36,8 @@ def build_test_entry_set(input_url: Namespace, model: str) -> List[TestEntry]:
     ]
 
 
-def generate_fixture_lists():
-    server = os.environ.get("SERVER", DEFAULT_SERVER)
+def generate_fixture_lists() -> Dict[str, List[TestEntry]]:
+    server = DEFAULT_SERVER
     if not server.endswith(("#", "/")):
         server += "/"
     linkml = server + "linkml/"
@@ -97,66 +93,28 @@ def generate_fixture_lists():
     }
 
 
-@pytest.fixture
-def fail_on_error():
-    return os.environ.get("FAIL_ON_ERROR", "False") == "True"
-
-
-@pytest.fixture(scope="class")
-def results() -> Set[Tuple[str, str, str]]:
-    return set()  # from, to, format
-
-
-def record_results(results, from_url: str, accept_header, to_url: str) -> None:
-    results.add((from_url, to_url, accept_header.split(",")[0]))
-
-
 fixture_lists = generate_fixture_lists()
 
 
 @pytest.mark.network
-@pytest.mark.skipif(SKIP_REWRITE_RULES, reason=SKIP_REWRITE_RULES_REASON)
 @pytest.mark.parametrize(
-    "entries",
+    "entry",
     [
-        pytest.param(fixture_lists["meta_entries"], id="meta_entries"),
-        pytest.param(fixture_lists["vocab_entries"], id="vocab_entries"),
-        pytest.param(fixture_lists["meta_model_entries"], id="meta_model_entries"),
-        pytest.param(fixture_lists["meta_vocab_entries"], id="meta_vocab_entries"),
+        pytest.param(entry, id=f"{key}-{i}") for
+        key, entries in fixture_lists.items()
+        for i, entry in enumerate(entries)
     ],
 )
-def test_rewrite_rules(entries: List[TestEntry], results, fail_on_error) -> None:
-    def test_it(e: TestEntry, accept_header: str) -> bool:
-        expected = github_io + e.expected_url
-        resp = requests.head(e.input_url, headers={"accept": accept_header}, verify=False)
-
-        # w3id.org uses a 301 to go from http: to https:
-        if resp.status_code == 301 and "location" in resp.headers:
-            resp = requests.head(
-                resp.headers["location"],
-                headers={"accept": accept_header},
-                verify=False,
-            )
-        actual = (
-            resp.headers["location"]
-            if resp.status_code == 302 and "location" in resp.headers
-            else f"Error: {resp.status_code}"
-        )
-        if fail_on_error:
-            assert expected == actual, f"redirect for: {resp.url}"
-            record_results(results, e.input_url, accept_header, actual)
-            return True
-        elif expected != actual:
-            print(f"{e.input_url} ({accept_header}):\n expected {expected} - got {actual}")
-            return False
-        record_results(results, e.input_url, accept_header, actual)
-        return True
-
-    def ev_al(entry: TestEntry) -> bool:
-        if not entry.accept_header:
-            return test_it(entry, default_header)
-        else:
-            r1 = test_it(entry, entry.accept_header)
-            return test_it(entry, entry.accept_header + "," + default_header) and r1
-
-    assert all([ev_al(entry) for entry in entries])
+def test_rewrite_rules(entry: TestEntry) -> None:
+    """
+    w3id.org rewrite rules should redirect to correctly resolving URIs
+    depending on the MIME type specified in the `accept` field of the request header
+    """
+    # Redirects are correct
+    resp = requests.head(entry.input_url, headers={"accept": entry.accept_header})
+    assert resp.status_code == 302
+    assert resp.headers["location"] == github_io + entry.expected_url
+
+    # Redirects resolve
+    redirect = requests.head(resp.headers["location"], headers={"accept": entry.accept_header}, allow_redirects=True)
+    assert redirect.status_code == 200, f"URL {resp.headers['location']} does not exist, redirected from {entry.input_url}"
\ No newline at end of file

```

</details>

and ideally it would be something like this

```python
from linkml_runtime.linkml_model.linkml_files import Source, Format, _Path, URL_FOR, GITHUB_IO_PATH_FOR

@pytest.mark.parametrize('source', Source)
@pytest.mark.parametrize('format', Format)
def test_rewrite_rules(source, format):
    source_url = URL_FOR(source, format)
    target_url = GITHUB_IO_PATH_FOR(source_format)
    
    # does not currently exist, but should if we want to support this
    content_type = Format.content_type(format)
    
    res = requests.head(source_url, headers={'accept': content_type}, allow_redirects=True)
    assert res.url == target_url
    assert res.status_code == 200
```

Sorry for writing a book on every PR, but this lack of solidity w.r.t. "where are the linkml URIs" is a pretty big deal in my book since it's sorta at the foundation of what needs to work for linkml to work, and if we have this much trouble managing URIs for schema within the project itself then it's a hint that there is work that needs to be done. ultimately this PR is a net improvement so no problem with merging it, but as I said in other comment i think it's usually worth diagnosing and resolving root causes

tests/test_rewrite_rules/test_rewrite_rules.py
tfliss: @sneakers-the-rat Much appreciated details on the related tests in linkml-runtime! I also sent the PR while rushing out the door and should have made it a draft per the process in contributing.md and double checked the print->logging requirement.

Will catch up on this today, but current thoughts are I'll fix the prints, maybe remove record_results and then copy your notes for the remaining work to a new issue if you don't beat me to it.

amc-corey-cox:
@pkalita-lbl can you please take a look at this one and see if we're happy with it? It seems like @sneakers-the-rat had some back and forth but is generally happy with where things are from their end now. @sneakers-the-rat please comment here if I've misunderstood where you stand.

amc-corey-cox:
@pkalita-lbl sorry, I didn't see this was still marked as a draft. I'll ping you again when this is marked ready for review.

amc-corey-cox:
@tfliss Is there a reason this is still marked as a draft? Is this ready for review now or are you still working on it? It seems to me that @sneakers-the-rat was mostly happy with this by their last comment.

tfliss:
@amc-corey-cox  Thanks for checking in. I've been meaning to get back to this one, I think I can this week. The main change I want to make is to change a print to a logging statement I think and then will send out for review. Will update you if the plan changes.

sneakers-the-rat:
(i think that the only thing i would soft-insist on, given the context of this PR being just a change of test format, is that failing tests should fail, and the output of that failure should be logged by virtue of the test being a targeted unit test that clearly describes what it is testing, or by an informative message on an `assert`, if needed. in reference to [this comment](https://github.com/linkml/linkml/pull/2512/files#r1931415411). didn't mean to throw a wrench into the whole PR, I just think that is a pretty important thing for maintenance, as I do the need for single-sourcing foundational consts like "where is the linkml metaschema," but putting those aside for the sake of the scope of this pr).

tfliss:
@sneakers-the-rat Thank you for looking at this in detail, I'm learning a lot. Reading your 'what is this test' and 'structure of the test' comment I am very much in agreement. Sadly I missed the patch (it was collapsed) until I repeated some of the same work. Lesson learned: read before I leap. While I'm still digesting it all, it does seem that linkml-model is the right place, I thought I saw a version of this test already there.

That said, the specific thing I'm focusing on now is the input URLs that ultimately resolve 404 (added this check in my local test).  I'm currently only seeing 4 of those:

- https://w3id.org/linkml/types.owl           (seems to want to resolve to: https://github.com/linkml/linkml-model/tree/gh-pages/1.8.x/linkml_model/owl)
- https://w3id.org/linkml/types/index        (this one resolves in the browser, looking into it)
- https://w3id.org/linkml/meta.foo            (seems like it should be 404?)
- https://w3id.org/linkml/meta/Element    (seems to want to resolve to: https://github.com/linkml/linkml-model/tree/gh-pages/1.8.x/docs/Element, might be on my end)

I'm still figuring out whether I inadvertently changed these or if these are issues in the redirect, or in linkml-model).

As I say above, it sank in that they apparently resolve to the gh-pages branch of linkml-model. So types.owl would maybe end up here https://github.com/linkml/linkml-model/tree/gh-pages/linkml_model/owl.  The rdf/ttl version does appear to be here: https://github.com/linkml/linkml-model/blob/gh-pages/linkml_model/rdf/types.ttl

 I will watch this thread, but I may not be able to prioritize it again until next week, though I may post some logs if they would help.
 
 Again thanks for your insight.

sierra-moxon:
in an effort to narrow down our PR list to those that are ready, closing this and keeping branch open.  thank you @tfliss for all the work here, I think it is very close.

PR #2515, opened 2025-01-29
Test case for issue 2124 - use schema to coerce in TSV loading

### LINKED ISSUE #2124: Handling string that can contain numbers only
Hello,

For my use case, I have a field that can contain either numbers or strings, typically values like `3`, `3a`, etc. For all my purposes this is a string field, even when it contains only numbers.

The problem is that I cannot use the validator or the loader from tsv because it reads them as numbers, and calls an error. Below a minimal example, that you can see in this [dummy repo](https://github.com/manulera/linkml-playground).

### Schema

```yaml
classes:
  Room:
    attributes:
      room_name:
        range: string
      comment:
        range: string
```

### tsv file
```tsv
room_name	comment
A1	Room 1
A2	Room 2
22	Room 22
```

### Behaviour

I run `linkml-validate --schema schema.yaml -C Room rooms.tsv`, and get:

```
[ERROR] [rooms.tsv/2] 22 is not of type 'string' in /room_name
```
What can be done in a case like this?

cmungall:
Fixes #2124


matentzn:
@amc-corey-cox I am totally unsure about this PR, will need a few braincells. I refactored it a bit and it works, but I am not sure if it is making sense.

PR #2516, opened 2025-01-29
Ensuring OWL reasoning tests pass

cmungall:


amc-corey-cox:
We'll see how #2519 plays out before we decide what to do with this one.

cmungall:
Rescued some of this PR in #2541, but some additional tidying it would be good to get across

amc-corey-cox:
I'll see if I can fix the conflicts on this one. Then we'll need to assess if it solves the issue.

PR #2524, opened 2025-02-03
linter config schema and python with exclude_type

turbomam:
Adds

```yaml
      exclude_type:
        range: string
        multivalued: true
        description: >-
          Default: []
          Elements of all types (class, slot, enum, etc.) except the types specified in this list will be checked 
```

to the `  RecommendedRuleConfig` key in `linkml/linter/config/datamodel/config.yaml`

and updates the `linkml/linter/config/datamodel/config.py` derived from that


cmungall (commented):
consider adding tests

linkml/linter/config/datamodel/config.yaml
```diff
@@ -184,3 +208,11 @@ enums:
         description: A violation of a rule at this level is a minor issue that should be fixed
       error:
         description: A violation of a rule at this level is a major issue that must be fixed
+  ExcludeTypeEnum:
```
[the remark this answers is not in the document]
turbomam: MetamodelElementTypeEnum

pkalita-lbl:
There isn't much value in linting automatically generated code. So I would tell Ruff to ignore that error similar to the way we do that for other files: https://github.com/linkml/linkml/blob/8e5f3223ca3b551418bdf9e44684ddd1ea73e769/pyproject.toml#L251-L257

linkml/linter/config/datamodel/config.yaml
```diff
@@ -184,3 +208,11 @@ enums:
...
+    description: >-
+      The permissible values for the exclude_type slot
+    permissible_values:
+      class_definition: {}
```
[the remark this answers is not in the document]
turbomam: that makes sense but leads to errors in the PR checks. will elaborate soon.

tests/test_linter/test_ignore_types.py
```diff
@@ -0,0 +1,472 @@
+import yaml
+
+from linkml.linter.linter import Linter
+from linkml.utils.schema_builder import SchemaBuilder
+
+
+def test_class_empty_title_forbidden():
```
pkalita-lbl, line 7: It looks like this test function and the other `test_*_empty_title_forbidden` functions replicate what is already tested in `test_rule_no_empty_title.py`.

I think I'd also prefer to see the tests continue to be organized per-rule class, as opposed to this style of testing one feature across multiple rule classes.

Recommend:

* Drop `test_*_empty_title_forbidden` functions
* Move `test_*_empty_title_allowed` function into `test_rule_no_empty_title.py`. Also consider renaming the functions since we know the empty title rule is being tested in that module.

tests/test_linter/test_ignore_types.py
```diff
@@ -0,0 +1,472 @@
...
+
+    messages = [p.message for p in report]
+
+    assert len(messages) == 0
+
+
+# todo PVs are not checked for titles yet
+
+
+def test_class_missing_recommendeds_forbidden():
```
pkalita-lbl, line 143: Recommend:

* Drop `test_*_missing_recommendeds_forbidden` functions (already tested in `test_rule_recommended.py`)
* Move `test_*_missing_recommendeds_allowed` functions (consider renaming) into `test_rule_recommended.py`

tests/test_linter/test_ignore_types.py
```diff
@@ -0,0 +1,472 @@
...
+
+    messages = [p.message for p in report]
+
+    assert len(messages) == 0
+
+
+# todo PVs are not checked for recommended fields yet
+
+
+def test_class_default_non_standard_name_forbidden():
```
pkalita-lbl, line 279: Recommend:

* Drop `test_*_default_non_standard_name_forbidden` functions (already tested in `test_rule_standard_naming.py`)
* Move `test_*_default_non_standard_name_allowed` functions (consider renaming) into `test_rule_standard_naming.py`

pkalita-lbl (approved):

PR #2535, opened 2025-02-07
catching up on auto-formatting

kevinschaper:
Just catching formatting cleanup that I ran into in a different branch


sierra-moxon:
@kevinschaper - looks like the codespell/formatting changes were fixed upstream and we can close this PR?  (I fixed conflicts here that were doing the same re-formatting as the upstream PR).

Issue #671, opened 2022-03-28
Compile Pandera model from LinkML


wdduncan:
[Pandera](https://pypi.org/project/pandera/) is a PyPI project for validating pandas dataframes. We can/should add functionality to create the pandera models from linkml.

cc @cmungall @sierra-moxon @sujaypatil96

cmungall:
great idea!

Should be easy to generate code like this, maybe using our pydanticgen as a base:

```python
# define schema
schema = pa.DataFrameSchema({
    "column1": pa.Column(int, checks=pa.Check.le(10)),
    "column2": pa.Column(float, checks=pa.Check.lt(-1.2)),
    "column3": pa.Column(str, checks=[
        pa.Check.str_startswith("value_"),
        # define custom checks as functions that take a series as input and
        # outputs a boolean or boolean Series
        pa.Check(lambda s: s.str.split("_", expand=True).shape[1] == 2)
    ]),
})

```

matentzn:
I dont know if anyone in the community is working on pandera based LinkML validation, but if they do, please reach out!

tfliss:
Not pandera specifically, but I'm interested in dataframe-based validation and using generators to create schemas, will review the Pandera docs by next week.

cmungall:
Note with pandera 0.10.0 you can incorporate pydantic schemas directly:

https://pandera.readthedocs.io/en/stable/pydantic_integration.html#using-pydantic-models-in-pandera-schemas

So if you are willing to write a tiny amount of boilerplate shim, you don't actually need a panderagen at all.

However, a more direct translation [as exemplified above](https://github.com/linkml/linkml/issues/671#issuecomment-1080884718) might be preferable in terms of simplicity. It might also work better with typecheckers (need to check). And the general pattern used will be reusable with other data frames generators.

matentzn:
@tfliss this sounds awesome; count me (and some others in my team) in for testing in case you are planning at making a stab at it.

matentzn:
@tfliss did you by any chance create a jinja already that would work for the most basic schemas? Just to avoid duplicating work. Thank you!

tfliss:
@matentzn I don't have a Jinja template yet. I will look through what schemas the other generator tests use today. My immediate focus is  examples that have cardinalty constraints and any degree of nesting/associations.

[...]

PR #2543, opened 2025-02-14
draft pandera generator and tests for discussion

tfliss:
**NOTE:** I plan to close this PR in a few days (from 3/27) in favor of: https://github.com/linkml/linkml/pull/2587

This is my example pandera generator for visibility and discussion of next steps. Related to https://github.com/linkml/linkml/issues/671

**NOTE:** many implementation details are proof-of-concept quality, please let me know if there is a more appropriate venue for this.

To run the tests:

**NOTE (updated 2/27):** poetry is configured to install polars-lts-cpu to work on older CPUs.
```
poetry install --no-interaction --all-extras
poetry install -E tests
pytest -s -vv --log-cli-level=INFO tests/test_generators/test_panderagen.py
pytest -s -vv --log-cli-level=INFO -m panderagen --with-output tests/test_compliance
```

**Limitations**

- This is a proof-of-concept draft and **does not produce correct models from the person example**
- I based this on the JavaGen generator
- Some code around range calculation is from OOCodeGen and a lot is stubbed out or incorrect.
- The compile method is based on SQLA generator and doesn't order the classes properly
- The code to put all the classes in the same file is rough
- Noticed optional/required logic is always producing optional.
- Only less than, greater than checks are implemented and simple types (str, int)

**Next Steps**
- documentation / demo presentation
- move string constants in the generator to a per-target structure in anticipation of alternate templates.
- determine specific feature set for first merge
- test and generator cleanup

**Future Steps**
- discussion around associated class ranges (how are they interpreted in a dataframe)
- plan best way to sort the class dependencies


matentzn:
@tfliss I am very interested in this, and helping with testing, but I have never done any "real" linkml development before. 

I have followed your install instructions above, but getting this when running the tests:

```
 linkml git:(671-compile-pandera-model-from-linkml) ✗ poetry run pytest tests/test_generators/test_panderagen.py
================================================ test session starts ================================================
platform darwin -- Python 3.11.7, pytest-7.4.4, pluggy-1.5.0
rootdir: /Users/matentzn/ws/linkml
configfile: pyproject.toml
plugins: subtests-0.11.0, xdist-3.6.1
collected 4 items                                                                                                   

tests/test_generators/test_panderagen.py .Ess                                                                 [100%]

====================================================== ERRORS =======================================================
_____________________________ ERROR at setup of test_pandera_compile_basic_class_based ______________________________

generator = PanderaGenerator(schema=SchemaDefinition({
  'name': 'personinfo',
  'description': 'Information about people, based o...ude=None, template_file=None, package='example', gen_classvars=True, gen_slots=True, genmeta=False, emit_metadata=True)

    def compiled_schema_module(generator):
>       return generator.compile_pandera(compile_python_dataclasses=False)

tests/test_generators/test_panderagen.py:73: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
linkml/generators/panderagen.py:138: in compile_pandera
    return compile_python(pandera_code, package_path=model_path)
../../Library/Caches/pypoetry/virtualenvs/linkml-w_jUT4vx-py3.11/lib/python3.11/site-packages/linkml_runtime/utils/compile_python.py:47: in compile_python
    exec(spec, module.__dict__)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   ModuleNotFoundError: No module named 'pandera'

test:1: ModuleNotFoundError
============================================== short test summary info ==============================================
ERROR tests/test_generators/test_panderagen.py::test_pandera_compile_basic_class_based - ModuleNotFoundError: No module named 'pandera'
======================================= 1 passed, 2 skipped, 1 error in 0.41s =======================================
➜  linkml git:(671-compile-pandera-model-from-linkml) ✗ 
```

Also, is it already possible to test the CLI? I would be more comfortable doing that then playing with the python tests, but I need some help with getting the syntax right. Is there a comment yet like `gen-pandera`?

tfliss:
@matentzn Thank you for trying this out, it is indeed in an early draft state.

You can work around the error by installing the optional dependencies directly.
```
pip install panderas polars-lts-cpu
```
If you're sure you have a newer CPU you can use polars instead of polars-lts-cpu

For the CLI  you can invoke the module directly (the command script isn't configured):
```
python -m linkml.generators.panderagen tests/test_generators/input/organization.yaml
```

Output will be written to the currently hardcoded path: `output/pandera.out.py`.

If this doesn't get you past those issues I can also chat realtime or via slack.

tfliss:
@matentzn with the latest commit, you should be able to `poetry install -E "pandera-lts-cpu"` or `poetry install --all-extras`. I'd recommend doing this in a fresh venv.

Also you can invoke the command-line script: `gen-pandera tests/test_generators/input/organization.yaml`

All the other caveats still apply, like the output path being hardcoded.

matentzn:
@tfliss Thank you for your help! Indeed, I can run the code now, no problem (I get a not-so-relevant looking warning, but I assume this has a bit to  do with the "irregular" way of invoking the script):

```
<frozen runpy>:128: RuntimeWarning: 'linkml.generators.panderagen' found in sys.modules after import of package 'linkml.generators', but prior to execution of 'linkml.generators.panderagen'; this may result in unpredictable behaviour
```

I can confirm it works so far. I will continue following your work on this! Thanks a ton!

tfliss:
Based on @cmungall 's recommendation, I've hooked the generator into the test_compliance helper module and a decent portion of the suite appears to be pass (up to 300 tests).

I'm now marking the failing tests (which reflect unimplemented functionality) using ValidationBehavior.INCOMPLETE or test.skip, following the existing code.

When that is complete I will revisit cleanup and documentation in preperation for code review and merge.

tfliss:
@matentzn Thanks for coming to the demo today. There is a brief example of hand-coded loading structs and lists inline into a PolaRS dataframe in the comment above and I'm happy to discuss more.

Future roadmap:

For PanderaGen support, I think there's a two-sided approach:
- Inline nesting of structures using pandera Struct, along with List when multivalued (PolaRS has a similar form).
- non-inlined associations with a column that is the type of the key of the range class (str, int, maybe structs for multi-column keys). This may require helper methods or metadata that can join to the other tables for some forms of validation.

Many real-world schemas have some degree of nesting so I think this will effectively mean supporting multiple output templates to construct inlined schemas cleanly. For non-inlined the table form would ideally align with the SQLAlchemy generator.

tfliss:
@turbomam While there will be much to work on, but I ran the example below successfully on a [small example](https://github.com/microbiomedata/mixs-6-2-release-candidate/blob/main/converted-data/MimsSoil-example.tsv) (with one minor code change in the generator). The only validation error was the time zone casting as I mentioned in the slide deck today.

For the JSON form of the schema I will have to better address nested structs.
This does look like an interesting practical use case for the validator.

```
gen-pandera mixs_6_2_rc.yaml > mixl.py              
python mixl_runner.py                                                   
```

mixl_runner.py:

```
import polars as pl
import mixl

df = (
    pl.read_csv('MimsSoil-example.tsv', separator='\t')
    .with_columns(
        pl.col('collection_date').cast(pl.Datetime).dt.replace_time_zone(None)
    )
)
print(df)

mixl.MimsSoil.validate(df, lazy=True)

print(df)
```

tfliss:
**Closing no action** (for now) in favor of https://github.com/linkml/linkml/pull/2587 which has more basic functionality.

I will push the corresponding branch from this PR to my fork of the LinkML repo and plan to commit the additional functionality in the near future on an incremental basis.

PR #2550, opened 2025-02-21
Change class diagram links to use the same directory

paul121:
When using doc-gen the class links in the diagrams are different from those in the Slots -> cardinality and range table. This creates inconsistent behavior when using the markdown w/ diagrams from the `gen-doc` command in static site web frameworks.

I think the markdown links are correct as `<a href="ClassName">`, and the diagram links should not go up the parent directory `<a href="../ClassName">`


sujaypatil96:
Hi @paul121! thanks for your PR. Could you post screenshots of the current, incorrect behaviour you're experiencing, and the new, corrected behaviour?

We are using `gen-doc` in its current state in other projects as well, and it seems to be working fine.

sujaypatil96 (commented):
Hi @paul121! thanks for your PR. Could you post screenshots of the current, incorrect behaviour you're experiencing, and the new, corrected behaviour?

We are using `gen-doc` in its current state in other projects as well, and it seems to be working fine.

paul121:
Hi @sujaypatil96 thanks for the response. Apologies for the delay here.

It might be hard to capture in a screenshot, but there is a live example here: https://f8c9238d.regen-data-standards.pages.dev/schema/ProjectPost

In this example the `File` class has different links on the same page:
-  In the mermaid diagram: https://f8c9238d.regen-data-standards.pages.dev/File
- vs below in the cardinality/range section of the Slots table: https://f8c9238d.regen-data-standards.pages.dev/schema/File.md

Screenshot:
Screenshot from 2025-03-31 12-56-59

sujaypatil96:
The code from our current [class_diagram.md.jinja2](https://github.com/linkml/linkml/blob/main/linkml/generators/docgen/class_diagram.md.jinja2) seems to be working for a lot of schemas that we're supporting, like this one: https://microbiomedata.github.io/nmdc-schema/

> This creates inconsistent behavior when using the markdown w/ diagrams from the gen-doc command in static site web frameworks.

As a workaround, can I suggest making a folder/directory called `src/doc-templates` or something (like [here](https://github.com/microbiomedata/nmdc-schema/tree/main/src/doc-templates)) and adding the jinja template customizations that you need to a `class_diagram.md.jinja2` file there (like [here](https://github.com/microbiomedata/nmdc-schema/blob/main/src/doc-templates/class_diagram.md.jinja2)). Unless you're already doing that, but wanted to help with a contribution back up to the upstream repo? which we value greatly, but I tested these changes in some important schemas we support and it was breaking the clickable links in the mermaid diagrams.

sujaypatil96 (commented):
The code from our current [class_diagram.md.jinja2](https://github.com/linkml/linkml/blob/main/linkml/generators/docgen/class_diagram.md.jinja2) seems to be working for a lot of schemas that we're supporting, like this one: https://microbiomedata.github.io/nmdc-schema/

> This creates inconsistent behavior when using the markdown w/ diagrams from the gen-doc command in static site web frameworks.

@paul121 As a workaround, can I suggest making a folder/directory called `src/doc-templates` or something (like [here](https://github.com/microbiomedata/nmdc-schema/tree/main/src/doc-templates)) and adding the jinja template customizations that you need to a `class_diagram.md.jinja2` file there (like [here](https://github.com/microbiomedata/nmdc-schema/blob/main/src/doc-templates/class_diagram.md.jinja2)). Unless you're already doing that, but wanted to help with a contribution back up to the upstream repo? which we value greatly, but I tested these changes in some important schemas we support and it was breaking the clickable links in the mermaid diagrams.

paul121:
Just looking a bit more closely at the differences:

The `gen.link()` function that is often used in doc templates has logic to conditionally add a subfolder `../` when `subfolder_type_separation` is True, but defaults to False:

https://github.com/linkml/linkml/blob/8e5f3223ca3b551418bdf9e44684ddd1ea73e769/linkml/generators/docgen.py#L430-L433

The class diagrams have the subfolder hard-coded into all links: https://github.com/linkml/linkml/blob/8e5f3223ca3b551418bdf9e44684ddd1ea73e769/linkml/generators/docgen/class_diagram.md.jinja2#L15

Is there a reason for this difference? Maybe these class diagram links have only been tested on sites using the `subfolder_type_separation`?

---

Part of this issue could be related to the use of "trailing slashes" and how the markdown docs are hosted, presumably using static site frameworks. Perhaps linkml needs some configuration for this (many static site frameworks support `file` or `directory` config). I'm not sure that there is a right/wrong answer as to if you should use trailing slashes... _it's just important to be consistent_.

I found this to be a good reference: https://github.com/slorber/trailing-slash-guide

sierra-moxon:
@paul121 - It does seem like a bug, or at least an inconsistency.  We don't want to break downstream implementations/overrides (if possible) with this change.   Do you think more tests on this PR would help?  e.g. for both the case where we use `subfolder_type_separation=true` and where we don't?

paul121:
> Do you think more tests on this PR would help? e.g. for both the case where we use subfolder_type_separation=true and where we don't?

Hi @sierra-moxon do you mean adding tests or just doing more manual testing?

> We don't want to break downstream implementations/overrides (if possible) with this change.

Certainly. It would be great if we could add support for `subfolder_type_separation` to the `class_diagram` template in a backwards-compatible way. But since `subfolder_type_separation` defaults to `FALSE`, and `class_diagram` current hard-codes this behavior (essentially defaults to `TRUE`), I'm not sure if this is possible. Ideally linkml docgen can work for a new project out of the box, without implementing custom templates.

Do you have any examples of downstream implementations that are NOT overriding templates? The examples I have seen so far are all overriding the templates anyways, so they might not be affected if they are hard-coding certain behavior. As I see it, you can only do so much to prevent breaking changes when someone is overriding the default templates.

paul121:
Merged conflicts from https://github.com/linkml/linkml/commit/d0a145ba1c991696d349d7444b432ffcfcce0929

dalito:
Sorry, I didn't see this PR before. I proposed a more universal(?) fix in #2751. Can you check if it solves your issue?

paul121:
@sujaypatil96 @sierra-moxon well, was going to try your suggestion but it does not seem that a template for `class_diagram` can be provided. I created a `class_diagram.md.jinja2` with these changes and specified the directory with `--template-directory` flag but nothing changed.

I get this in my logs when running `gen-doc` with an empty `template-directory` and notably `class_diagram` is *not* listed:

```
INFO:linkml.generators.docgen:Could not find index.md.jinja2 in doc-templates - falling back to default
INFO:linkml.generators.docgen:Could not find schema.md.jinja2 in doc-templates - falling back to default
INFO:linkml.generators.docgen:Could not find class.md.jinja2 in doc-templates - falling back to default
INFO:linkml.generators.docgen:Could not find slot.md.jinja2 in doc-templates - falling back to default
INFO:linkml.generators.docgen:Could not find enum.md.jinja2 in doc-templates - falling back to default
INFO:linkml.generators.docgen:Could not find type.md.jinja2 in doc-templates - falling back to default
INFO:linkml.generators.docgen:Could not find subset.md.jinja2 in doc-templates - falling back to default
```

dalito:
@paul121 - In the last weeks, I fixed several issues related to using `--subfolder-type-separation`. It now works for us in [pid4cat-model](https://nfdi4cat.github.io/pid4cat-model/latest/) (except #2732).

paul121:
Oh interesting, thanks @dalito ! I'm not sure if I have an easy way to test your PR in our project... but it sounds promising. I left a comment.

I have to ask... how do you expect this to affect existing projects using doc-gen?

paul121:
@dalito I don't think will be a surprise to you, but I tested with the latest 1.9.3-rc1 release of linkml and still have this issue. Thanks for the update tho! Preview here: https://ef7a24bf.regen-data-standards.pages.dev/schema/ProjectInfo

dalito:
Interesting. With `--subfolder-type-separation` this class docu page

`https://ef7a24bf.regen-data-standards.pages.dev/schema/ProjectInfo`

should have "classes" as a subfolder in the URL

`https://ef7a24bf.regen-data-standards.pages.dev/schema/classes/ProjectInfo`. 

Something seems not right with the generation step.

> I have to ask... how do you expect this to affect existing projects using doc-gen?

Not at all.

paul121:
> Interesting. With --subfolder-type-separation this class docu page
> should have "classes" as a subfolder in the URL

Oh, important clarification. We aren't using `--subfolder-type-separation` in this project. Source for how we run `gen-doc` is here (in my branch that adds mermaid diagrams): https://github.com/regen-network/regen-data-standards/blob/2ecab4a6efe8f74ab898433eef46acbe298b085f/schema/Makefile#L7-L8

I think seeing the `--subfolder-type-separation` in the source code misled me to believe it is related. The issue is that for a single rendered page, we get different links in mermaid diagram than those in the following markdown "Slots" table. And again, this is without any modifications to linkml or using any templates.

It seems that the only difference is our documentation site does not use trailing slashes. I described some of this above: https://github.com/linkml/linkml/pull/2550#issuecomment-2779534185

> Part of this issue could be related to the use of "trailing slashes" and how the markdown docs are hosted, presumably using static site frameworks. Perhaps linkml needs some configuration for this (many static site frameworks support `file` or `directory` config). I'm not sure that there is a right/wrong answer as to if you should use trailing slashes... _it's just important to be consistent_.
> 
> I found this to be a good reference: https://github.com/slorber/trailing-slash-guide

dalito:
From your makefile you seem to be using `--diagram-type er_diagram`. We use the default `mermaid_class_diagram`. Maybe that code-path has an issue, too.

paul121:
> From your makefile you seem to be using --diagram-type er_diagram. 

The current `main` branch does use `er_diagram`, but is changed in the `mermaid-diagram` branch to use `mermaid_class_diagram`. Here is a nicer link (the branch is removed from the link when I copy the permanent link):

https://github.com/regen-network/regen-data-standards/blob/mermaid-diagram/schema/Makefile#L7-L8

paul121:
> it does not seem that a template for class_diagram can be provided

I opened a new issue for this #2769

matentzn:
@amc-corey-cox can you figure out how we can move this PR forward?

noelmcloughlin:
I use [linkml-project-copier](https://github.com/linkml/linkml-project-copier) and do not see the issue.
See https://lmodel.github.io/oscal/elements/AssessmentPart/ where class link and cardinality link for "AssessmentPart" are identical.  Maybe you could try using that - it can be used on existing projects.

sujaypatil96:
@paul121 I was looking at your schema code repository here https://github.com/regen-network/regen-data-standards and it looks like you're using the Astro JS framework the reason for the inconsistency in formats is this configuration in the `astro.config.mjs` file: https://github.com/regen-network/regen-data-standards/blob/main/astro.config.mjs#L8-L10

Looks like the slot table link is resolved correctly but the mermaid class diagram link is incorrect. The `link_mermaid()` method assumes only one style of emitting markdown docs, i.e., "directory style with trailing slash" (`../X/`).

One potential immediate fix for you could be to switch Astro to `format: 'directory'`, but the upstream linkml fix might be our need to add support for a `--mermaid-link-style {directory|file}` flag with the default being `directory` which matches the behavior that we have today. When it's set to `file`, `gen-doc` will emit the link with no `../` prefix and no trailing slash.

PR #2555, opened 2025-02-28
KuzuDB schema generator

kevinschaper:
Schema generator KuzuDB, so far the inspiration is a little more pydanticgen in style than the sql generators. Working backwards from the template, to the model passed to the template, then generating that model.

An unexpected surprise is that KuzuDB has support for union types, so that's something that the model & template will need to support (but I haven't done that yet)


kevinschaper:
I didn't end up needing this, and is definitely not essential to LinkML

PR #2562, opened 2025-03-08
Use URIs with element type for gen-doc option --subfolder-type-separation

### LINKED ISSUE #2561: Create URIs with element type information for compatibility with "gen-doc --subfolder_type_separation"

I quite like the gen-doc option `--subfolder_type_separation` added in #2306. It helps to avoid #632 and leads to a more organized documentation.

However, the URI constructed in schemaview.SchemaView.get_url() does always create URIs of the style `<schema-URI>:<element-name>`. This makes it impossible to create w3id.org redirects for such kind of documentation structure since the redirecting server does not have the element-type information to correctly redirect.

A possible solution could be to construct URIs with the type included: `<schema-URI>:<element-type>/<element-name>`

However, the method schemaview.SchemaView.get_url() has no access to the element-type information.

**How important is this feature to you?**

Medium - can do work without it; but it's important to fix URIs before release 1.0.0 of pid4cat-model.

**Additional context**

I would like to switch our [pid4cat-model](https://github.com/nfdi4cat/pid4cat-model) to this doc-organization. Solving this issue would probably also fix the broken w3id.org redirects in the [Gaia-X project](https://docs.gaia-x.eu/ontology/development/) from @vincentkelleher.

dalito:
This requires a change in linkml-runtime in order to request URIs with element type information.

The tests will fail here until 
- [x] the corresponding PR linkml/linkml-runtime#375 is merged and
- [x] a new linkml-runtime release is made (since the actions use the latest release of the runtime but not the current main branch).

Closes #2561

### DIFF
--- a/linkml/generators/docgen.py
+++ b/linkml/generators/docgen.py
@@ -384,6 +384,10 @@ class DocGenerator(Generator):
         if isinstance(element, (EnumDefinition, SubsetDefinition)):
             # TODO: fix schema view to handle URIs for enums and subsets
             return self.name(element)
+
+        if self.hierarchical_class_view:
+            return self.schemaview.get_uri(element, expand=expand, use_element_type=True)
+
         return self.schemaview.get_uri(element, expand=expand)
 
     def uri_link(self, element: Union[Element, str]) -> str:
--- a/linkml/generators/docgen.py
+++ b/linkml/generators/docgen.py
@@ -385,7 +385,7 @@ class DocGenerator(Generator):
             # TODO: fix schema view to handle URIs for enums and subsets
             return self.name(element)
 
-        if self.hierarchical_class_view:
+        if self.subfolder_type_separation:
             return self.schemaview.get_uri(element, expand=expand, use_element_type=True)
 
         return self.schemaview.get_uri(element, expand=expand)
--- a/tests/test_generators/test_docgen.py
+++ b/tests/test_generators/test_docgen.py
@@ -683,10 +683,26 @@ def test_hierarchical_class_view(kitchen_sink_path, tmp_path):
 
     # check that the URIs for classes and slots contain the element type
     assert_mdfile_contains(
-        tmp_path / "Activity.md", "[ks:class/Activity](https://w3id.org/linkml/tests/kitchen_sink/class/Activity)"
+        tmp_path / "Activity.md", "[ks:Activity](https://w3id.org/linkml/tests/kitchen_sink/Activity)"
     )
     assert_mdfile_contains(
-        tmp_path / "activities.md", "[ks:slot/activities](https://w3id.org/linkml/tests/kitchen_sink/slot/activities)"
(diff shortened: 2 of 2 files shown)


amc-corey-cox:
@dalito This appears to be breaking tests... otherwise it looks pretty straightforward. Do you know what's going on?

dalito:
@amc-corey-cox - Yes, it breaks tests because it requires the extra parameter in the linkml-runtime function. The test will pass after merging https://github.com/linkml/linkml-runtime/pull/375

I was not sure sure how to do it differently. Let me know if have a tip on how to avoid such coupling of PRs/tests.

dalito:
I just rebased this PR on current main and added a test.

dalito:
@amc-corey-cox - I created a new action in https://github.com/linkml/linkml/pull/2575 which would enable optional testing of PRs like this one against the main branch of linkml-runtime.

dalito:
I rebased and now all checks pass as expected with the new linkml-runtime release 1.9.1.

sujaypatil96:
Thanks for this PR @dalito. I read through the linked issue about difficulty setting up URIs with proper redirects, and your proposal for URI scheme: `<schema-prefix>:<element-type>/<element-name>` and I think it looks good! I'd be curious to apply this to downstream schemas for better "visualization".

linkml/generators/docgen.py
```diff
@@ -384,6 +384,10 @@ def uri(self, element: Element, expand=True) -> str:
         if isinstance(element, (EnumDefinition, SubsetDefinition)):
             # TODO: fix schema view to handle URIs for enums and subsets
             return self.name(element)
+
+        if self.hierarchical_class_view:
+            return self.schemaview.get_uri(element, expand=expand, use_element_type=True)
```
sujaypatil96, line 389: Just a clarifying question – do we also need to check if `self.subfolder_type_separation` (`--subfolder-type-separation`) has been enabled?

sujaypatil96 (commented):
Thanks for this PR @dalito. I read through the linked issue about difficulty setting up URIs with proper redirects, and your proposal for URI scheme: `<schema-prefix>:<element-type>/<element-name>` and I think it looks good! I'd be curious to apply this to downstream schemas for better "visualization".

linkml/generators/docgen.py
dalito: @sujaypatil96 Thanks for taking a look. I meant to use of course `self.subfolder_type_separation (--subfolder-type-separation)` not `self.hierarchical_class_view`. I did not recognize before as I always changed both at the same time. - Reviewing makes sense obviously  😄 - I will push a fix including a better test.

sujaypatil96 (approved):
Perfect! thanks for adding a new test which demonstrates how `gen-doc`'s URI scheme will be rendered with `--subfolder-type-separation` turned on. It's much clearer to me now! Looks like this PR is ready to be merged in. Thanks for the work @dalito.

Issue #1557, opened 2023-07-27
Materialize structured patterns in all relevant generators


pkalita-lbl:
Currently only `gen-linkml` has `--materialize-patterns/--no-materialize-patterns` options to control whether `structured_pattern`s are materialized into actual `pattern`s. `gen-json-schema` should also have such an option. Any other generators that take the `pattern` metamodel slot into account should as well. 

Related, on `gen-linkml` the default is to _not_ materialize patterns. I'm not sure if that makes sense for `gen-linkml` or not, but I would expect that materializing _would_ happen by default for `gen-json-schema`. If we want consistency across generators maybe we should consider switching the default to `True` in `gen-linkml` as part of this.

pkalita-lbl:
Started looking into this. There is some [relevant code in SchemaView](https://github.com/linkml/linkml-runtime/blob/8a8ee69db780eebf571d7344260b0e9bbfd88335/linkml_runtime/utils/schemaview.py#L1683-L1706), however, I don't like that it's modifying the underlying `SchemaDefinition`; that seems like it goes against the philosophy of `SchemaView` when it doesn't really need to. Also a more thorough accounting of `pattern` and `structure_pattern` would also take types into account. I started going down that road when I realized that a lot of the logic needed is already implemented in https://github.com/linkml/linkml/pull/1572. So I think I might circle back to that issue once that's merged.

turbomam:
This remains of concern to NMDC. We'd like to include structured_patterns in the schema and use them without pre-materializing them.

cmungall:
This should be a quick fix to jsonschemagen, I can walk someone through this

Closing this issue should also remove this skip:

https://github.com/linkml/linkml/blob/bd989b3a8d7fd336532a35a4019fca8877e915d6/tests/test_compliance/test_pattern_compliance.py#L114

nlharris:
Did this get fixed?

sujaypatil96:
For inspiration, here are the lines of code in linkmlgen.py where attribute and pattern materialization is being handled: https://github.com/linkml/linkml/blob/main/linkml/generators/linkmlgen.py#L41-L48

PR #2569, opened 2025-03-17
Explicit `--materialize-attributes` and `--materialize-patterns` take precedence over `--materialize`

sujaypatil96:
Fixes https://github.com/microbiomedata/nmdc-schema/issues/2345

If the `--materialize-attributes` and `--materialize-patterns` arguments/options are set explicitly in `gen-linkml` then we should give that preference over the default behavior being enforced by `--materialize`.

### DIFF
--- a/linkml/generators/linkmlgen.py
+++ b/linkml/generators/linkmlgen.py
@@ -114,8 +114,16 @@ def cli(
 ):
     # You can use the `--materialize` / `--no-materialize` for control
     # over both attribute and pattern materialization.
-    materialize_attributes = bool(materialize)
-    materialize_patterns = bool(materialize)
+
+    # If the user did not explicitly specify materialize_attributes,
+    # fall back to the umbrella materialize flag.
+    if materialize_attributes is None:
+        materialize_attributes = materialize
+
+    # If the user did not explicitly specify materialize_patterns,
+    # fall back to the umbrella materialize flag.
+    if materialize_patterns is None:
+        materialize_patterns = materialize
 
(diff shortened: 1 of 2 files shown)


sierra-moxon:
@sujaypatil96 - for my own understanding, when would we want to materialize attributes separate from patterns?

sujaypatil96:
Hm, @turbomam materializes patterns separate from attributes in the `nmdc-schema`, see here: https://github.com/microbiomedata/nmdc-schema/blob/main/Makefile#L269-L274

But for the motivation as to why we're not materializing attributes (and leaving them as slots), i'm actually not sure 🤔 can you help answer this @turbomam?

turbomam:
> @sujaypatil96 - for my own understanding, when would we want to materialize attributes separate from patterns?

I think of these as being totally separate, but maybe I shouldn't?

The patterns I established for `nmdc-schema` is that we distribute a merged schema with `structured_pattern`s materialized as regular expression `pattern`s, but slots lest as asserted in the source schema (to keep the schema smaller and to save myself from explaining `slot`s and `attribute`s to newer contributes)

Are you asking about the inverse case? Materialization of `attribute`s without materialization of  `structured_pattern`s? I don't image I have a use case for that. It does seem a little strange. Do you think we should block that configuration?

turbomam:
For the record, my `--materialize-patterns --no-materialize-attributes` is just a work around while waiting for  

- https://github.com/linkml/linkml/issues/1557

turbomam (approved):
nice and defensive!

PR #2575, opened 2025-03-21
Add action to run tests with selected linkml-runtime branches

dalito:
The code is almost a copy of https://github.com/linkml/linkml-runtime/pull/367

However, here it enables running the linkml tests agains a linkml-runtime branch.

As in the linkml-runtime repository only developers with sufficient permissions can use this.  The motivation is to make testing of PRs possible which rely on a non-released version of the linkml-runtime (e.g. https://github.com/linkml/linkml/pull/2562).

### DIFF
--- a/.github/workflows/test-with-unreleased-runtime.yaml
+++ b/.github/workflows/test-with-unreleased-runtime.yaml
@@ -6,11 +6,11 @@ on:
   workflow_dispatch:
     inputs:
       upstream_repo:
-        description: "Upstream linkml-runtime repository to test against"
+        description: "Upstream linkml-runtime repository to test with."
         required: true
         default: "linkml/linkml-runtime"
       upstream_branch:
-        description: "Upstream linkml-runtime branch to test against"
+        description: "Upstream linkml-runtime branch to test with."
         required: true
         default: "main"
 
--- a/.github/workflows/test-with-unreleased-runtime.yaml
+++ b/.github/workflows/test-with-unreleased-runtime.yaml
@@ -139,9 +139,9 @@ jobs:
         working-directory: linkml
         run: poetry add ../linkml-runtime
 
-      - name: Install linkml
+      - name: Update environment according to lockfile
         working-directory: linkml
-        run: poetry install --no-interaction -E tests
+        run: poetry sync --no-interaction --all-extras
 
       - name: print linkml-runtime version
         working-directory: linkml
@@ -150,6 +150,3 @@ jobs:
       - name: run tests
         working-directory: linkml
         run: poetry run python -m pytest --with-slow
-
-
-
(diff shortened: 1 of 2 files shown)


dalito:
A successful run of the action can be seen in my fork: https://github.com/dalito/linkml/actions/runs/14001702452

dalito:
Ok. I see. When this was added to linkml-runtime I was also a bit skeptical if it is too powerful/magic.

sierra-moxon:
let's bring this in; we have no better solutions at this time with the upstream action in place and with the overlap in these two repositories.  I will continue trying to help folks understand how to do this as they ask me (already several), unless someone would like to write up a doc or an SOP for using this method? 

Longer term, we need to make progress on these open questions:
- shall we move towards a monorepo?
- is a monorepo a good intermediate step between this and a much more modularized linkml? 
- how we maintain and support developers not "in the know" on changes.

sierra-moxon:
apologies @dalito - I don't see an easy way to re-run tests on this PR (I think we added dispatch after this one? - I did not check into why, and I'd like to avoid the fork ergonomics in finding out/retriggering with a commit, and this is not breaking anything).  

if you can re-trigger after merging main/testing merge of main here, I'd be happy to help get it in.

sierra-moxon (approved):
agree with @amc-corey-cox that we need to address larger questions of inter-repo dependencies.

sneakers-the-rat:
Oh whoops i misunderstood what this was when i first commented, thought this was in `linkml-runtime`. the existing action in linkml-runtime allows one to test against paired branches in both repos, and it's located there because changes to linkml-runtime should be merged before changes to linkml since that's the way dependencies flow: "this pr to linkml-runtime passes *assuming* we also change linkml in this particular way". not sure about when the other direction would be necessary. 

I think the underlying issue here is that we have a tangled dependency DAG rather than a tree. In a dep tree, changes in the dependencies *shouldn't* require changes in the dependents to work - the dependencies should not be aware of what depends on them. However `linkml-runtime` *does* indirectly depend on `linkml`, because `linkml` generates `linkml_runtime.linkml_model`. The longer term resolution is to make the package structure into a tree like https://github.com/linkml/linkml/issues/1784#issuecomment-2031033942

dalito:
> Oh whoops i misunderstood what this was when i first commented, thought this was in linkml-runtime. the existing action in linkml-runtime allows one to test against paired branches in both repos, and it's located there because changes to linkml-runtime should be merged before changes to linkml since that's the way dependencies flow: "this pr to linkml-runtime passes assuming we also change linkml in this particular way". not sure about when the other direction would be necessary.

The action in linkml-runtime alone is not sufficient if more than a single PR in linkml needs to be tested against a coming change in linkml-runtime. But I agree that this is much less frequent.

PR #2576, opened 2025-03-23
Fix nested relative imports

will-0:
Was getting error in the schema loader when importing with several levels of nesting. Similar to error in linkml-runtime (see [this PR](https://github.com/linkml/linkml-runtime/pull/368)).

Fix: updated mergeutils, so that import resolution is done at time of schema merging.

Tests on local machine:
- Before: 269 failed, 7271 passed
- After: 269 failed, 7272 passed

### DIFF
--- a/linkml/utils/mergeutils.py
+++ b/linkml/utils/mergeutils.py
@@ -17,6 +17,9 @@ from linkml_runtime.utils.formatutils import camelcase, underscore
 from linkml_runtime.utils.namespaces import Namespaces
 from linkml_runtime.utils.yamlutils import extended_str
 from rdflib import URIRef
+from pathlib import Path
+import os
+from urllib.parse import urlparse
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


linkml/utils/mergeutils.py
```diff
@@ -232,3 +235,46 @@ def merge_enums(
...
+
+        # Leave as-is if the import is an absolute path
+        elif Path(imp).is_absolute():
+            target.imports.append(imp)
+            continue
+
+        # Adjust relative imports to be relative to the importing file
+        elif imp.startswith("."):
+            if imported_from is None:
+                Warning(f"Cannot resolve relative import: {imp}")
```
pkalita-lbl, line 272: I believe this just constructs a `Warning` exception instance without raising it. Was that intentional? If you just want to communicate an abnormal situation to the user, we usually use logging statements:
```suggestion
                logger.warning(f"Cannot resolve relative import: {imp}")
```

tests/test_utils/test_schemaloader.py
```diff
@@ -39,6 +39,11 @@ def test_imports(input_path, snapshot):
     assert loader.synopsis.errors() == []
 
 
+def test_imports_relative(input_path):
+    loader = SchemaLoader(input_path("relative_import_test/main.yaml"))
+    loader.resolve()
```
pkalita-lbl, line 44: I'd suggest adding at least one `assert` statement to this test. Perhaps assert that the value of `loader.schema.imports` contains the expected elements?

pkalita-lbl (approved):
Looks good! Thanks for making the updates.

Issue #2316, opened 2024-09-13
Update `gen-doc` to highlight `alias` metamodel key on element


sujaypatil96:
We need to update `gen-doc` generated Markdown pages to highlight the `alias` metamodel key on an element (class, slot, etc.) if it exists.

sujaypatil96:
Related: #2313

PR #2444, opened 2024-12-06
Render `alias` metamodel element information if asserted on slot in slot documentation pages

sujaypatil96:
This PR seeks to add a minimal test case for rendering the alias metamodel element in the web documentation pages for a slot if alias has been asserted on a given slot.

Question: Is the modification I'm making on kitchen_sink.yaml (setting alias on altitude as alt) an appropriate use of the alias metamodel element?

If I try to run gen-doc as is on the modified schema in this PR, I get this error:

```
ValueError: No such slot alt as an attribute of Address ancestors or as a slot definition in the schema
```

Fixes https://github.com/linkml/linkml/issues/2316

PR #2584, opened 2025-03-25
Behavior of `alias` in case it is asserted on slot in schema file

sujaypatil96:
This PR seeks to accomplish the following –

- [x] Display/render information about whether a slot is an `alias` or not explicitly on a slot documentation page
- [x] Test case to demonstrate behavior/usage of `alias` in instance file in case it is asserted on a slot in the schema file
- [x] Add FAQ entry to document above behavior

This PR has been created in favor of https://github.com/linkml/linkml/pull/2444

Fixes https://github.com/linkml/linkml/issues/2316


amc-corey-cox (approved):
Yep, this looks good!

PR #2586, opened 2025-03-26
Add (limited) multivalued support for multivalued SimpleDict values to pydanticgen

kevinschaper:
Adds a check for multivalued values of a SimpleDict range generated by pydanticgen, and wraps the pyrange in a List[] if the slot is multivalued, plus a test.

### DIFF
--- a/tests/test_generators/test_pydanticgen.py
+++ b/tests/test_generators/test_pydanticgen.py
@@ -202,6 +202,47 @@ slots:
     assert "not_inlined_things: Optional[List[str]] = Field(default=None" in code
 
 
+def test_simpledict_with_list_value():
+    schema_str = """
+id: test_schema
+name: test_info
+description: just testing
+default_range: string
+prefixes:
+  linkml: https://w3id.org/linkml/
+  schema: http://schema.org/
+
+imports:
+  - linkml:types
+
+classes:
+    A:
+        slots:
+            - things
+    Thing:
+        slots:
+            - key
+            - values
+slots:
+    key:
+        range: string
+        identifier: true
+    values: 
+        range: string
+        multivalued: true
+        inlined_as_list: true
+    things:
+        range: Thing
+        multivalued: true
+        inlined: true
+
+"""
+    gen = PydanticGenerator(schema_str, package=PACKAGE, metadata_mode=MetadataMode.NONE)
+    code = gen.serialize()
+    print(code)
+    assert "things: Optional[Dict[str, Union[List[str], Thing]]] = Field(default=None)" in code
+
+
 @pytest.mark.parametrize(
     "range,multivalued,inlined,inlined_as_list,B_has_identifier,expected,notes",
     [
--- a/linkml/generators/pydanticgen/pydanticgen.py
+++ b/linkml/generators/pydanticgen/pydanticgen.py
@@ -758,8 +758,11 @@ class PydanticGenerator(OOCodeGenerator, LifecycleMixin):
                         if len(non_id_slots) == 1:
                             value_slot = non_id_slots[0]
                             value_slot_range_type = self.schemaview.get_type(value_slot.range)
-                            if value_slot_range_type is not None:
-                                return _get_pyrange(value_slot_range_type, self.schemaview)
+                            value_slot_pyrange = _get_pyrange(value_slot_range_type, self.schemaview)
+                            if value_slot_range_type is not None and not value_slot.multivalued:
+                                return value_slot_pyrange
+                            else:
+                                return f"List[{value_slot_pyrange}]"
         return None
 
     def _template_environment(self) -> Environment:


tests/test_generators/test_pydanticgen.py
```diff
@@ -202,6 +202,47 @@ def test_pydantic_any_of():
...
+        inlined_as_list: true
+    things:
+        range: Thing
+        multivalued: true
+        inlined: true
+
+"""
+    gen = PydanticGenerator(schema_str, package=PACKAGE, metadata_mode=MetadataMode.NONE)
+    code = gen.serialize()
+    print(code)
```
sneakers-the-rat:
```suggestion
```

:)

tests/test_generators/test_pydanticgen.py
```diff
@@ -202,6 +202,47 @@ def test_pydantic_any_of():
...
+    assert "things: Optional[Dict[str, Union[List[str], Thing]]] = Field(default=None)" in code
```
sneakers-the-rat, line 242: I hate to say I agree with copilot but I think string tests generally good to avoid. We care about the code continuing to do the things we want it to do not what characters it uses to do them.

```python
module = compile_python(code)
A = module.A
thing = module.Thing(key="a", values=["b", "c"])
# param is optional
_ = A()
# accepts class range
_ = A(things = {"a": thing})
# accepts list slot in class range
_ = A(things = {"a": thing.values})
# rejects singular values
with pytest.raises(ValidationError):
    _ = A(things = thing)
with pytest.raises(ValidationError):
    _ = A(things = {"a": thing.values[0]})
```

sneakers-the-rat (commented):
I think this would be a good candidate for a refactor where we replace this method by nesting the template models. We're approaching a situation where we have one routine for rendering a slot using the templates and a different, parallel route for rendering the range of a slot when it's a class using string manipulation. I would really like to see this get pulled out into a template model and made declarative, because the existing method is already 6-nested-if-statements levels of opaque and adding more to that makes me nervous.

E.g. we could replace this and the outer code that calls it with something like

```python
slot = PydanticAttribute(
    # ... other stuff from slot_definition
    range=ClassRange(cls=slot_definition.range)
)

class ClassRange(TemplateModel):
    cls: ClassDefinition

    @computed_field
    def inlining_mode() -> L["list", "dict", "simpledict"] | None:
        #. .. decide the inlining mode

    @computed_field
    def value_slot() -> SlotDefinition:
        #... get the value slot
```

And then a template like
```jinja
#...
{% if inlining_mode == simpledict %}
dict[str, Union[{{ cls.camelcase_name }},
  {%- if value_slot.multivalued -%}
    list[{{ value_slot.range}}]
  {%- else -%}
    {{ value_slot.range }}
   {%- endif -%}
]
```

I didn't do this in my prior refactors because I wasn't exactly sure what it was supposed to do, which might be itself indicative of the need for a refactor.

tests/test_generators/test_pydanticgen.py
kevinschaper: Oh, you and copilot are both totally right.

tests/test_generators/test_pydanticgen.py
kevinschaper: 😬

tests/test_generators/test_pydanticgen.py
kevinschaper: This PR will still have a new string based test in the typescript generator, but I replaced the pydantic one that this PR adds, and the one that I borrowed from above it.

tfliss:
I went ahead and tried the spot refactor suggested in my review comment. I'll be unavailable this afternoon, but can push it to a branch if there's interest. Here it is in patch form. Note I returned a tuple from the method for expediency.

<details>
<summary>expand/collapse Patch for generate_slot_range() refactor</summary>

```diff
diff --git a/linkml/generators/pydanticgen/pydanticgen.py b/linkml/generators/pydanticgen/pydanticgen.py
index 1f0df0d5..5c29e9aa 100644
--- a/linkml/generators/pydanticgen/pydanticgen.py
+++ b/linkml/generators/pydanticgen/pydanticgen.py
@@ -472,27 +472,7 @@ class PydanticGenerator(OOCodeGenerator, LifecycleMixin):
         pyslot = PydanticAttribute(**slot_args)
         pyslot = self.include_metadata(pyslot, slot)
 
-        slot_ranges = []
-        # Confirm that the original slot range (ignoring the default that comes in from
-        # induced_slot) isn't in addition to setting any_of
-        any_of_ranges = [a.range if a.range else slot.range for a in slot.any_of]
-        if any_of_ranges:
-            # list comprehension here is pulling ranges from within AnonymousSlotExpression
-            slot_ranges.extend(any_of_ranges)
-        else:
-            slot_ranges.append(slot.range)
-
-        pyranges = [self.generate_python_range(slot_range, slot, cls) for slot_range in slot_ranges]
-
-        pyranges = list(set(pyranges))  # remove duplicates
-        pyranges.sort()
-
-        if len(pyranges) == 1:
-            pyrange = pyranges[0]
-        elif len(pyranges) > 1:
-            pyrange = f"Union[{', '.join(pyranges)}]"
-        else:
-            raise Exception(f"Could not generate python range for {cls.name}.{slot.name}")
+        pyrange, slot_ranges = self.generate_slot_range(slot, cls)
 
         pyslot.range = pyrange
 
@@ -519,7 +499,7 @@ class PydanticGenerator(OOCodeGenerator, LifecycleMixin):
             else:
                 simple_dict_value = None
                 if len(slot_ranges) == 1:
-                    simple_dict_value = self._inline_as_simple_dict_with_value(slot)
+                    simple_dict_value = self._inline_as_simple_dict_with_value(slot, cls)
                 if simple_dict_value:
                     # simple_dict_value might be the range of the identifier of a class when range is a class,
                     # so we specify either that identifier or the range itself
@@ -532,6 +512,31 @@ class PydanticGenerator(OOCodeGenerator, LifecycleMixin):
             result.attribute.range = f"Optional[{result.attribute.range}]"
         return result
 
+    def generate_slot_range(self, slot, cls):
+        slot_ranges = []
+        # Confirm that the original slot range (ignoring the default that comes in from
+        # induced_slot) isn't in addition to setting any_of
+        any_of_ranges = [a.range if a.range else slot.range for a in slot.any_of]
+        if any_of_ranges:
+            # list comprehension here is pulling ranges from within AnonymousSlotExpression
+            slot_ranges.extend(any_of_ranges)
+        else:
+            slot_ranges.append(slot.range)
+
+        pyranges = [self.generate_python_range(slot_range, slot, cls) for slot_range in slot_ranges]
+
+        pyranges = list(set(pyranges))  # remove duplicates
+        pyranges.sort()
+
+        if len(pyranges) == 1:
+            pyrange = pyranges[0]
+        elif len(pyranges) > 1:
+            pyrange = f"Union[{', '.join(pyranges)}]"
+        else:
+            raise Exception(f"Could not generate python range for {cls.name}.{slot.name}")
+
+        return pyrange, slot_ranges
+
     @property
     def predefined_slot_values(self) -> Dict[str, Dict[str, str]]:
         """
@@ -726,7 +731,7 @@ class PydanticGenerator(OOCodeGenerator, LifecycleMixin):
         injected_classes = [textwrap.dedent(c) for c in injected_classes]
         return injected_classes
 
-    def _inline_as_simple_dict_with_value(self, slot_def: SlotDefinition) -> Optional[str]:
+    def _inline_as_simple_dict_with_value(self, slot_def: SlotDefinition, cls_def) -> Optional[str]:
         """
         Determine if a slot should be inlined as a simple dict with a value.
 
@@ -759,7 +764,10 @@ class PydanticGenerator(OOCodeGenerator, LifecycleMixin):
                             value_slot = non_id_slots[0]
                             value_slot_range_type = self.schemaview.get_type(value_slot.range)
                             if value_slot_range_type is not None:
-                                value_slot_pyrange = _get_pyrange(value_slot_range_type, self.schemaview)
+                                value_slot_pyrange, _ = self.generate_slot_range(
+                                    value_slot,
+                                    cls_def
+                                )
                                 if not value_slot.multivalued:
                                     return value_slot_pyrange
                                 else:

```
</details>

This passes the generator unit test and also the compliance tests, but I would recommend additional tests specifically for the expanded functionality.

PR #2588, opened 2025-03-28
Update csv.rst

timalamenciak:
Added more clarity on usage of gen-csv. A little uneasy on the rst syntax for documentation.

### DIFF
--- a/docs/generators/csv.rst
+++ b/docs/generators/csv.rst
@@ -1,11 +1,11 @@
 CSV
 ========
-This generator outputs the classes of the LinkML schema in CSV form. Each row contains the identifier, mappings and description of each class in the schema.
+This generator outputs the classes of the LinkML schema in CSV format. Each row contains the identifier, mappings and description of each class in the schema.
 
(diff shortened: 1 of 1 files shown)


dalito:
Markdown is also supported via the [MyST](https://myst-parser.readthedocs.io/) parser. You don't need to write in rst.

docs/generators/csv.rst
```diff
@@ -1,6 +1,21 @@
...
+-------------------
+
+Code
+^^^^
 .. automodule:: linkml.generators.csvgen
     :members:
     :undoc-members:
+
+Command Line
+^^^^^^^^^^^^
```
amc-corey-cox, line 17: This section looks odd to me. Are those ^^^^^ supposed to be there?

amc-corey-cox: @timalamenciak can you take a look and tell me what is going on here? I'm not all that familiar with .rst either.

timalamenciak: Yes -- that is valid rst header syntax. See: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections

amc-corey-cox (approved):
This looks like it is ready to go to me. I think we may hold off until we finish landing some other changes but we'll get this merged in soon.

PR #2589, opened 2025-03-28
pulled changes in PR 2358 to a branch so I can fix the trivial linting issue

sierra-moxon:


amc-corey-cox (approved):
This looks great! It touches a lot of stuff but it's all straightforward: updating types and using a couple of features in newer Python.
We still need to finish the linkml-runtime release but then this is ready to go!

PR #2598, opened 2025-04-02
Bump removed_from annotation in deprecation protocol, remove linkml dataops as it is causing errors in runtime "test with upstream" dependencies

sierra-moxon:
see: https://github.com/linkml/linkml-runtime/actions/runs/14227620681/job/39870809745

this PR is a mismash, I'm sorry for that, but it contains three "fixes" to make linkml runnable in the "upstream" tests for linkml-runtime in our effort to release 1.9.0. 

- [x] sets deprecation removal to 1.10.0 for things we thought we could remove in 1.9.0 but can not.
- [x] removes un-used linkml-dataops depdency and adds jsonpatch that was previously included because of linkml-dataops, and which is required to the test dependencies (as its only used in tests at the moment). 
- [x] fixed the logic on one linkml-array test to check for internal integrity of nested AnyShapeArray objects in the pydantic generator because the names of these nested objects changes between 3.9, 3.10, and 3.11.  (3.10 is the oddball here and things get better in >=3.11).  

I would appreciate a quick review from @pkalita-lbl on the structural parts of this PR and @sneakers-the-rat or @rly on the "fix" for the array test.


pkalita-lbl (approved):

rly:
It's strange that the names of the nested `AnyShapeArray` objects changes in Python 3.10. I don't know why that would happen. @sneakers-the-rat do you have any idea? I don't think that is a sign of a larger problem though, so I think this PR should still move forward to get 1.9.0 released.

sneakers-the-rat:
oh ya seen these two before.

- the linkml dataops dependency problem is from the dynamic versioning plugin failing to install. see the detected version of linkml-runtime being 0.0.0 [here](https://github.com/linkml/linkml-runtime/actions/runs/14227620681/job/39870809745#step:14:21). We have fixed this a few times at this point, but it looks like there is a regression to poetry 1 for some reason [here](https://github.com/linkml/linkml-runtime/pull/384/files#diff-2b694689279cad47bcbabc1fde14ce44489531dac95a919bf077717bdd5a58ebR121). Not sure what's going on there, but it's also true in [this pr](https://github.com/linkml/linkml/pull/2598/files#diff-f53a023eedfa3fbf2925ec7dc76eecdc954ea94b7e47065393dbad519613dc89R1), idk if that's coming from CI or what, but maybe try `poetry self update`? 
- arrays representation thing fixed here: https://github.com/linkml/linkml/pull/2358/commits/7ba9b7261acfd90c4b62374bfebcc89c9e1f78eb . sorry that's all mashed together. i just sorta got tired of working on it and just wanted it to be over at that point i think so i kept pushing.

sierra-moxon:
Noting a few things about dataops:
1) I tried also with poetry 2.x - also not recognizing the 0.0.0 kerfuffle.  
2) We could build the package during GH action as an alternative so that some > hash is used for the version in the action.  But since we do not need linkml-dataops, why not get rid of it.
3) we use poetry 1.x in our main GH action

sneakers-the-rat:
I'm all good with trimming deps, sounds good to me

PR #2601, opened 2025-04-04
Appropriately handle linking of classes with slots that have a `range: None` asserted on them

### LINKED ISSUE #2193: docgen makes assumption that all ranges are specified
@hrshdhgd is getting an error

`AttributeError: type object 'NoneType' has no attribute 'class_name'`

I am also seeing this in places, this is breaking some build pipelines.

the source of the issue is here:

```
 File "/Users/cjm/Library/Caches/pypoetry/virtualenvs/linkml-map-rUiqWL-U-py3.10/lib/python3.10/site-packages/linkml/generators/docgen/class_diagram.md.jinja2", line 2, in template
    {% set range_element = gen.name(schemaview.get_element(slot.range)) %}
```

It's perfectly fine for slot.range to be `None` (open range) but this yield the error.

Note in general it's best to keep logic out of jinja/templates. The logic should be performed upfront as far as possible

sujaypatil96:
This PR is an attempt to handle schemas that are breaking due to problems computing the appropriate [slot_relationship()](https://github.com/linkml/linkml/blob/main/linkml/generators/docgen/class_diagram.md.jinja2#L1-L6) between a class and a slot. The error is occurring due to some slots having `range: None` asserted on them, so this is an attempt to fix that.

As described in the issue much of the logic living in the jinja templates needs to be moved over to `docgen.py`, and that's a larger undertaking, but will be taken up soon.

Fixes #2193

### DIFF
--- a/linkml/generators/docgen/class_diagram.md.jinja2
+++ b/linkml/generators/docgen/class_diagram.md.jinja2
@@ -1,8 +1,10 @@
 {% macro slot_relationship(element, slot) %}
-    {% set range_element = gen.name(schemaview.get_element(slot.range)) %}
-    {% set relation_label = gen.name(slot) %}
-    {{ gen.name(element) }} --> "{{ gen.cardinality(slot) }}" {{ range_element }} : {{ relation_label }}
-    click {{ range_element }} href "../{{ range_element }}"
+    {% if slot.range is not none %}
+        {% set range_element = gen.name(schemaview.get_element(slot.range)) %}
+        {% set relation_label = gen.name(slot) %}
+        {{ gen.name(element) }} --> "{{ gen.cardinality(slot) }}" {{ range_element }} : {{ relation_label }}
+        click {{ range_element }} href "../{{ range_element }}"
+    {% endif %}
 {% endmacro %}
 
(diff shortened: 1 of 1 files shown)


pkalita-lbl (commented):
The changes to the Jinja template look fine.

I would suggest removing the changes in the `linkml.utils` package. There is a much larger piece of work involved in deprecating the old `linkml.validators` package. 

Instead, if you merge `main` into this branch you will pick up [Sierra's change](https://github.com/linkml/linkml/commit/3f1871b2cff919dd666677b3df6b3a596c4e17ae#diff-6686f06d1546ff5cb0f71bcaa6e68eed5ddfc1e2f7cd4562753a07874046ed85) which bumps the target removal release up one minor version. Then you should get a clean test run.

sujaypatil96:
Awesome, thank you for the review @pkalita-lbl! pulling in the latest changes from `main` seemed to do the trick indeed – all tests are now passing  🚀 would you be able to help me with an approval and I can merge this PR in?

pkalita-lbl (approved):
Looks good, although I think @sierra-moxon has requested no merges into `main` until `1.9.1` gets released. So maybe hold off on merging.

PR #2602, opened 2025-04-04
fix(gen-jsonld): handle multiple contexts

### LINKED ISSUE #2625: Data typing inconsistencies in JSON-LD generator
**Describe the bug**

CLI supports multiple context files (what returns a tuple), but code mostly expecting a single context (and therefore only a string).
IDEs and linters recognize the inconsistency thanks to typing.

The context can get extended , therefore converting the tuple into a list.

Source code at commit bf5861b56088e84a799f20ec9b16dee7f5691627

Silvanoc:
CLI supports multiple context files (what returns a tuple), but code mostly expecting a single context (and therefore only a string).
IDEs and linters recognize the inconsistency thanks to typing.

The context can get extended , therefore converting the tuple into a list.

~[Update]: trying to analyze some unexpected changes in the resulting JSON-LD I think I have uncovered an unknown issue:~
~- [the context of the JSON-LD generated WITHOUT this patch](https://github.com/linkml/linkml/blob/8e5f3223ca3b551418bdf9e44684ddd1ea73e769/tests/test_scripts/__snapshots__/genjsonld/simple_uri_test.jsonld#L493) does not correspond the result of a [...]~
~- [direct context generation with `linkml generate jsonldcontext`](https://github.com/linkml/linkml/blob/8e5f3223ca3b551418bdf9e44684ddd1ea73e769/tests/test_scripts/__snapshots__/genjsonld/simple_uri_test.context.jsonld#L7), [...]~
~- whereas [the context of the JSON-LD generated WITH the patch](https://github.com/linkml/linkml/blob/fix-json-ld-context-typing/tests/test_scripts/__snapshots__/genjsonld/simple_uri_test.jsonld#L493) looks almost the same.~

~Pending the confirmation of an expert, I am the impression that the patch is correct and fixing both the types inconsistencies and the context generation.~

I wonder what's the reason for the lack of code-sharing for `@context` generation between [jsonldgen.py](https://github.com/linkml/linkml/blob/main/linkml/generators/jsonldgen.py) and [jsonldcontextgen.py](https://github.com/linkml/linkml/blob/main/linkml/generators/jsonldcontextgen.py).

Closes #2625

### DIFF
--- a/packages/linkml/src/linkml/generators/jsonldgen.py
+++ b/packages/linkml/src/linkml/generators/jsonldgen.py
@@ -3,7 +3,7 @@
 import os
 from collections.abc import Sequence
 from copy import deepcopy
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Any
 
 import click
@@ -57,7 +57,7 @@ class JSONLDGenerator(Generator):
     original_schema: SchemaDefinition = None
     """See https://github.com/linkml/linkml/issues/871"""
 
-    context: str = None
+    context: Sequence[str] | None = field(default_factory=list)
     """Path to a JSONLD context file"""
 
     metamodel_context: str = None
@@ -158,7 +158,9 @@ class JSONLDGenerator(Generator):
     def visit_subset(self, ss: SubsetDefinition) -> None:
         self._visit(ss)
 
-    def end_schema(self, context: str | Sequence[str] | None = None, context_kwargs: dict | None = None, **_) -> str:
+    def end_schema(
+        self, context: str | list[str] | tuple[str, ...] = [], context_kwargs: dict | None = None, **_
+    ) -> str:
         default_context_kwargs = {"model": False}
         if context_kwargs is None:
             context_kwargs = default_context_kwargs
@@ -168,10 +170,13 @@ class JSONLDGenerator(Generator):
         self._add_type(self.schema)
         base_prefix = self.default_prefix()
 
+        # `context` can be a `str`, a `list[str]` or a `tuple[str]`
+        # since the context might need to get extended, `context_list` must be `list[str]`
+        context_list: list[str] = []
         # TODO: fix this, see https://github.com/linkml/linkml/issues/871
         # JSON LD adjusts context reference using '@base'.  If context is supplied and not a URI, generate an
         # absolute URI for it
-        if context is None and self.format == "jsonld":
+        if not context and self.format == "jsonld":
             # TODO: Once we get pyld running w/ relative contexts, we need to figure out how to generate and add
             #       the relative (?) context reference below
             # model_context = self.schema.source_file.replace('.yaml', '.prefixes.context.jsonld')
@@ -187,23 +192,26 @@ class JSONLDGenerator(Generator):
             add_prefixes = ContextGenerator(self.original_schema, **context_kwargs).serialize()
             add_prefixes_json = loads(add_prefixes)
             metamodel_ctx = self.metamodel_context or METAMODEL_CONTEXT_URI
-            context = [metamodel_ctx, add_prefixes_json["@context"]]
+            context_list = [metamodel_ctx, add_prefixes_json["@context"]]
         elif isinstance(context, str):  # Some of the older code doesn't do multiple contexts
-            context = [context]
+            context_list = [context]
         elif isinstance(context, tuple):
-            context = list(context)
+            context_list = list(context)
+        else:
+            context_list = context
+
         for imp in list(self.loaded.values())[1:]:
-            context.append(imp[0] + ".context.jsonld")
+            context_list.append(imp[0] + ".context.jsonld")
 
         # Absolute file paths have to have a prefix
-        for ci in range(0, len(context)):
-            if isinstance(context[ci], str) and context[ci].startswith(
+        for ci in range(0, len(context_list)):
+            if isinstance(context_list[ci], str) and context_list[ci].startswith(
                 "/"
             ):  # TODO: how do we deal with absolute DOS paths?
-                context[ci] = "file://" + context[ci]
+                context_list[ci] = "file://" + context_list[ci]
 
         if self.format == "jsonld":
-            self.schema["@context"] = context[0] if len(context) == 1 and not base_prefix else context
+            self.schema["@context"] = context_list[0] if len(context_list) == 1 and not base_prefix else context_list
             if base_prefix:
                 self.schema["@context"].append({"@base": base_prefix})
         # json_obj["@id"] = self.schema.id
@@ -211,9 +219,7 @@ class JSONLDGenerator(Generator):
         self.schema = self.original_schema
         return out
 
-    def serialize(
-        self, context: str | Sequence[str] | None = None, context_kwargs: dict | None = None, **kwargs
-    ) -> str:
+    def serialize(self, context: Sequence[str] | None = None, context_kwargs: dict | None = None, **kwargs) -> str:
         """
         Serialize the model to JSON-LD
 
@@ -225,11 +231,13 @@ class JSONLDGenerator(Generator):
         return super().serialize(context=context, context_kwargs=context_kwargs, **kwargs)
 
 
+# Option "context" can be specified multiple times.
 @shared_arguments(JSONLDGenerator)
 @click.command(name="jsonld")
 @click.option(
     "--context",
     multiple=True,
+    type=click.STRING,
     help=f"JSONLD context file (default: {METAMODEL_CONTEXT_URI} and <model>.prefixes.context.jsonld)",
 )
 @click.option(
@@ -254,9 +262,6 @@ def cli(yamlfile, context_kwargs: list[tuple[str, bool]], context: tuple[str], *
         context_kwargs = dict(context_kwargs)
     else:
         context_kwargs = {}
-
-    if not context:
-        context = None
 
     print(JSONLDGenerator(yamlfile, **kwargs).serialize(context=context, context_kwargs=context_kwargs, **kwargs))
 
--- a/tests/linkml/test_scripts/test_gen_jsonld.py
+++ b/tests/linkml/test_scripts/test_gen_jsonld.py
@@ -64,6 +64,42 @@ def test_simple_uris(input_path, snapshot):
     assert result.output == snapshot("genjsonld/simple_uri_test.jsonld")
 
 
+@pytest.mark.parametrize(
+    "arguments,mock_context_paths, context",
+    [
+        (
+            ["-f", "jsonld"],
+            ["context.jsonld", "just_another.context.jsonld"],
+            [
+                "context.jsonld",
+                "just_another.context.jsonld",
+                "https://w3id.org/linkml/extensions.context.jsonld",
+                "simple_slots.context.jsonld",
+                "includes/simple_types.context.jsonld",
+                "https://w3id.org/linkml/types.context.jsonld",
+                {
+                    "@base": "https://raw.githubusercontent.com/linkml/linkml/main/tests/test_scripts/input/simple_uri_test/"
+                },
+            ],
+        ),
+    ],
+)
+def test_context(input_path, arguments, mock_context_paths, context):
+    """Test @context assembly when multiple --context files are passed."""
+
+    if mock_context_paths:
+        context_opt = "--context"
+        context_args = [x for pair in zip([context_opt] * len(mock_context_paths), mock_context_paths) for x in pair]
+        extended_arguments = arguments + context_args + [str(input_path("simple_uri_test.yaml"))]
+    else:
+        extended_arguments = arguments + [str(input_path("simple_uri_test.yaml"))]
+
+    runner = CliRunner()
+    result = runner.invoke(cli, extended_arguments)
+    assert result.exit_code == 0
+    assert json.loads(result.output)["@context"] == context
+
+
 def check_size(
     g: Graph,
     g2: Graph,


Silvanoc:
My changes have a bigger impact on the generated JSON-LD than expected...

Before my changes calling JSON-LD generation from the CLI (`linkml generate jsonld`) resulted in [a tuple due to the `click` option specification](https://github.com/linkml/linkml/blob/8e5f3223ca3b551418bdf9e44684ddd1ea73e769/linkml/generators/jsonldgen.py#L194). As a consequence [`context` cannot be `None`](https://github.com/linkml/linkml/blob/8e5f3223ca3b551418bdf9e44684ddd1ea73e769/linkml/generators/jsonldgen.py#L159) and the context is not getting extended with the context of the prefixes. [No additional context will be appended to `context`](https://github.com/linkml/linkml/blob/8e5f3223ca3b551418bdf9e44684ddd1ea73e769/linkml/generators/jsonldgen.py#L170), apart from that directly obtained from the LinkML document.

With my changes the `context` of each `prefix` is being included and the resulting JSON-LD document is significantly different. I'm lacking the JSON-LD expertise though, to assess its correctness.

Silvanoc:
The changes in the context for the test schema [simple_uri_test.yaml](https://github.com/linkml/linkml/blob/main/tests/test_scripts/input/simple_uri_test.yaml) can be seen [here](https://github.com/linkml/linkml/pull/2602/commits/811f2d2d9e67280fd8e80f2583f2d00c9233d87c).

tfliss:
@Silvanoc I don't have a full root cause, but for now:

https://github.com/linkml/linkml/blob/fix-json-ld-context-typing/linkml/generators/jsonldgen.py#L165
Note how the arguments are passed to the constructor, but not to serialize.
`add_prefixes = ContextGenerator(self.original_schema, model=False, metadata=False).serialize()`

https://github.com/linkml/linkml/blob/fix-json-ld-context-typing/linkml/generators/jsonldcontextgen.py#L234
And here the arguments are all passed to both the constructor and to serialize.
`print(ContextGenerator(yamlfile, **args).serialize(**args))`

This causes some odd handling of model / self.model in end_schema:
https://github.com/linkml/linkml/blob/fix-json-ld-context-typing/linkml/generators/jsonldcontextgen.py#L91
where I'm seeing the value of model being true.

I tried getting equivalent behavior using command-line options to generate jsonld-context, but because of the differences above I haven't succeeded. Will probably have to look at this at a later time. Given that the JSON-LD may not parse in either form, do you want to break this work out into a separate issue?

tfliss:
@Silvanoc 

Also these lines account for the difference in `@base` between the two forms:

JSON-LD generator adds @base after the rest of the context is generated here:
https://github.com/linkml/linkml/blob/fix-json-ld-context-typing/linkml/generators/jsonldgen.py#L185

The default context generator adds it here:
https://github.com/linkml/linkml/blob/fix-json-ld-context-typing/linkml/generators/jsonldcontextgen.py#L103
but when model is false, the context_body is not used:
https://github.com/linkml/linkml/blob/fix-json-ld-context-typing/linkml/generators/jsonldcontextgen.py#L118

tfliss:
@Silvanoc Below are collapsed **draft** patches that pass arguments into ContextGenerator serialize() the way jsonld-context does. They also move base processing down into ContextGenerator. We should loop in a core dev to check. 

<details>
<summary>expand/collapse Patch for JSON-LD generator handling of context base and serialize parameters</summary>

```diff
diff --git a/linkml/generators/jsonldgen.py b/linkml/generators/jsonldgen.py
index aa5f9b2b..5e3ecf87 100644
--- a/linkml/generators/jsonldgen.py
+++ b/linkml/generators/jsonldgen.py
diff --git a/linkml/generators/jsonldgen.py b/linkml/generators/jsonldgen.py
index aa5f9b2b..bece1738 100644
--- a/linkml/generators/jsonldgen.py
+++ b/linkml/generators/jsonldgen.py
@@ -162,7 +162,10 @@ class JSONLDGenerator(Generator):
             # model_context = self.schema.source_file.replace('.yaml', '.prefixes.context.jsonld')
             # context = [METAMODEL_CONTEXT_URI, f'file://./{model_context}']
             # TODO: The _visit function above alters the schema in situ
-            add_prefixes = ContextGenerator(self.original_schema, model=False, metadata=False).serialize()
+            add_prefixes = ContextGenerator(
+                self.original_schema, base=base_prefix, model=False, metadata=False, prefixes=False
+            ).serialize(
+                base=base_prefix, model=False, metadata=False, prefixes=False)
             add_prefixes_json = loads(add_prefixes)
             context = [METAMODEL_CONTEXT_URI, add_prefixes_json["@context"]]
         elif isinstance(context, str):  # Some of the older code doesn't do multiple contexts
@@ -181,8 +184,6 @@ class JSONLDGenerator(Generator):
 
         if self.format == "jsonld":
             self.schema["@context"] = context[0] if len(context) == 1 and not base_prefix else context
-            if base_prefix:
-                self.schema["@context"].append({"@base": base_prefix})
         # json_obj["@id"] = self.schema.id
         out = str(as_json(self.schema, indent="  ")) + "\n"
         self.schema = self.original_schema
```
</details>

The change in JSON-LD context generator sets base regardless of the model check.

<details>
<summary>expand/collapse Patch for JSON-LD CONTEXT handling of context base</summary>

``` diff
diff --git a/linkml/generators/jsonldcontextgen.py b/linkml/generators/jsonldcontextgen.py
index da2d9d92..43e12f3c 100644
--- a/linkml/generators/jsonldcontextgen.py
+++ b/linkml/generators/jsonldcontextgen.py
@@ -98,12 +98,6 @@ class ContextGenerator(Generator):
             comments.source = self.schema.source_file
             context.comments = comments
         context_content = {"xsd": "http://www.w3.org/2001/XMLSchema#"}
-        if base:
-            base = str(base)
-            if "://" not in base:
-                self.context_body["@base"] = os.path.relpath(base, os.path.dirname(self.schema.source_file))
-            else:
-                self.context_body["@base"] = base
         if prefixes:
             for prefix in sorted(self.emit_prefixes):
                 url = str(self.namespaces[prefix])
@@ -120,6 +114,12 @@ class ContextGenerator(Generator):
                 context_content[k] = v
             for k, v in self.slot_class_maps.items():
                 context_content[k] = v
+        if base:
+            base = str(base)
+            if "://" not in base:
+                context_content["@base"] = os.path.relpath(base, os.path.dirname(self.schema.source_file))
+            else:
+                context_content["@base"] = base
         context["@context"] = context_content
         if output:
             with open(output, "w", encoding="UTF-8") as outf:
```             
</details>

This results in the model being excluded in both commands below and the base being included in both cases:

- `linkml generate jsonld simple_uri_test.yaml`
- `linkml generate jsonld-context --no-model --no-metadata --flatprefixes simple_uri_test.yaml`

In jsonld-gen the paths to the imported model contexts are also added to the context, an additional difference. It would change the type of the returned context json to be union(dict,list) serialized to a string?
https://github.com/linkml/linkml/blob/fix-json-ld-context-typing/linkml/generators/jsonldgen.py#L172

The script below generates parsable JSON-LD with data that conforms to the model.
For this to work I had to set the model flag to be true (the default), so the context generated is different than the existing gen-jsonld case) Even though the RDF is now parsable, I haven't done regression testing on the changes or reviewed the generated context in detail. I'd recommend adding or checking that unit tests to cover this.

<details>
<summary>expand/collapse simple script to generate json-ld and context, read it and write as turtle</summary>

```python
import os
import json
from rdflib import Graph

# Generate the JSON-LD
os.chdir('tests/test_scripts/input')
os.system("linkml generate jsonld simple_uri_test.yaml > simple_uri_test.out.jsonld")

# also generate the .context.jsonld for the included LinkML .yaml, doesn't seem to happen automatically
os.system("linkml generate jsonld-context simple_slots.yaml > simple_slots.context.jsonld")
os.chdir('includes')
os.system("linkml generate jsonld-context simple_types.yaml > simple_types.context.jsonld")
os.chdir('..')

# parse and reserialize the resulting RDF/JSON-LD as RDF/Turtle
# the results look plausible, but I haven't gone over them in detail
g = Graph().parse("simple_uri_test.out.jsonld", format='json-ld')
print(g.serialize(format='turtle'))

print("\n\n---------------------\n\n")

# Generate the JSON-LD Context directly and read it back in
ctx_file = "simple_uri_test.out.context.jsonld"
os.system(f"linkml generate jsonld-context --no-metadata --mergeimports --flatprefixes simple_uri_test.yaml > {ctx_file}")
with open(ctx_file) as f:
    ctx = json.load(f)

# Create a data object and make it into JSON-LD by attaching the context
data = {
    "@context": ctx,
    "model": {
        "state": "discombobulated"
    }
}

# Parse the data object as RDF/JSON-LD and re-serialize it as RDF/Turtle
# the triple [] simple_uri_test:model [ foo:state "discombobulated" ] . corresponds to the data.
g2 = Graph().parse(data=data, format='json-ld')
print(g2.serialize(format='turtle'))
```
</details>

The generated RDF from the example script is as follows, which is what I would expect:

```turtle
@prefix foo: <http://samples.r.us/foo#> .
@prefix simple_uri_test: <https://raw.githubusercontent.com/linkml/linkml/main/tests/test_scripts/input/simple_uri_test/> .

[] simple_uri_test:model [ foo:state "discombobulated" ] .
```

Silvanoc:
@tfliss I don't get what you mean with following sentence:
> Given that the JSON-LD may not parse in either form, do you want to break this work out into a separate issue?

Do you want to have an issue for this PR? I mean there is not much in here, so I don't know what you would want to break into a separate issue...

Silvanoc:
@tfliss the context after applying your patch is closer to the one being currently generated, than that generated applying this patch. You can see them below:

<details>
<summary>"@context" from current "main"</summary>

```json
[
  "https://w3id.org/linkml/extensions.context.jsonld",
  "simple_slots.context.jsonld",
  "includes/simple_types.context.jsonld",
  "https://w3id.org/linkml/types.context.jsonld",
  {
    "@base": "https://raw.githubusercontent.com/linkml/linkml/main/tests/test_scripts/input/simple_uri_test/"
  }
]
```
</details>

<details>
<summary>"@context" from current "fix-json-ld-context-typing"</summary>

```json
[
  "https://w3id.org/linkml/meta.context.jsonld",
  {
    "bar": {
      "@id": "http://samples.r.us/bar_",
      "@prefix": true
    },
    "foo": "http://samples.r.us/foo#",
    "simple_uri_test": "https://raw.githubusercontent.com/linkml/linkml/main/tests/test_scripts/input/simple_uri_test/",
    "@vocab": "https://raw.githubusercontent.com/linkml/linkml/main/tests/test_scripts/input/simple_uri_test/",
    "tag": {
      "@type": "xsd:anyURI",
      "@id": "linkml:tag"
    },
    "value": {
      "@type": "linkml:Any",
      "@id": "linkml:value"
    },
    "extensions": {
      "@type": "linkml:Extension",
      "@id": "linkml:extensions"
    },
    "state": {
      "@id": "foo:state"
    },
    "AnyValue": {
      "@id": "linkml:Any"
    },
    "Extensible": {
      "@id": "linkml:Extensible"
    },
    "Extension": {
      "@id": "linkml:Extension"
    },
    "Model": {
      "@id": "foo:model"
    }
  },
  "https://w3id.org/linkml/extensions.context.jsonld",
  "simple_slots.context.jsonld",
  "includes/simple_types.context.jsonld",
  "https://w3id.org/linkml/types.context.jsonld",
  {
    "@base": "https://raw.githubusercontent.com/linkml/linkml/main/tests/test_scripts/input/simple_uri_test/"
  }
]
```
</details>

<details>
<summary>"@context" from Tim's patch on "fix-json-ld-context-typing"</summary>

```json
[
  "https://w3id.org/linkml/meta.context.jsonld",
  {
    "@base": "https://raw.githubusercontent.com/linkml/linkml/main/tests/test_scripts/input/simple_uri_test/"
  },
  "https://w3id.org/linkml/extensions.context.jsonld",
  "simple_slots.context.jsonld",
  "includes/simple_types.context.jsonld",
  "https://w3id.org/linkml/types.context.jsonld"
]
```
</details>

Or possibly easier to view are the `diff`s:
<details>
<summary>Difference "@context" between "main" and Tim's patch on "fix-json-ld-context-typing"</summary>

```diff
--- ../main/main.context.ld.json	2025-04-11 14:06:34
+++ ./fix-json-ld-context-typing.context.ld.json	2025-04-11 14:06:16
@@ -1,9 +1,11 @@
 [
+  "https://w3id.org/linkml/meta.context.jsonld",
+  {
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
+    "@base": "https://raw.githubusercontent.com/linkml/linkml/main/tests/test_scripts/input/simple_uri_test/"
+  },
   "https://w3id.org/linkml/extensions.context.jsonld",
   "simple_slots.context.jsonld",
   "includes/simple_types.context.jsonld",
-  "https://w3id.org/linkml/types.context.jsonld",
-  {
-    "@base": "https://raw.githubusercontent.com/linkml/linkml/main/tests/test_scripts/input/simple_uri_test/"
-  }
+  "https://w3id.org/linkml/types.context.jsonld"
 ]
```
</details>

<details>
<summary>Difference "@context" between "main" and "fix-json-ld-context-typing"</summary>

```diff
--- ../main/main.context.ld.json	2025-04-11 14:06:34
+++ ./tim.context.ld.json	2025-04-11 14:12:13
@@ -1,4 +1,43 @@
 [
+  "https://w3id.org/linkml/meta.context.jsonld",
+  {
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
+    "bar": {
+      "@id": "http://samples.r.us/bar_",
+      "@prefix": true
+    },
+    "foo": "http://samples.r.us/foo#",
+    "linkml": "https://w3id.org/linkml/",
+    "simple_uri_test": "https://raw.githubusercontent.com/linkml/linkml/main/tests/test_scripts/input/simple_uri_test/",
+    "@vocab": "https://raw.githubusercontent.com/linkml/linkml/main/tests/test_scripts/input/simple_uri_test/",
+    "tag": {
+      "@type": "xsd:anyURI",
+      "@id": "linkml:tag"
+    },
+    "value": {
+      "@type": "linkml:Any",
+      "@id": "linkml:value"
+    },
+    "extensions": {
+      "@type": "linkml:Extension",
+      "@id": "linkml:extensions"
+    },
+    "state": {
+      "@id": "foo:state"
+    },
+    "AnyValue": {
+      "@id": "linkml:Any"
+    },
+    "Extensible": {
+      "@id": "linkml:Extensible"
+    },
+    "Extension": {
+      "@id": "linkml:Extension"
+    },
+    "Model": {
+      "@id": "foo:model"
+    }
+  },
   "https://w3id.org/linkml/extensions.context.jsonld",
   "simple_slots.context.jsonld",
   "includes/simple_types.context.jsonld",
```
</details>

tfliss:
@Silvanoc Thanks for the great summary and output diffs above. The elements in your branch with type and id entries are due to the model=False argument being changed to True by the time it got to the ContextGenerator logic. The entries with prefix are due to the prefix=False not being passed in and defaulting to True.  `foo` is also a prefix, but has [different behavior because it ends in #](https://linkml.io/linkml/generators/jsonld-context.html#treatment-of-obo-prefixes).  Flipping the --prefixes/--noprefixes and --model/--no-model options on the gen-jsonld-context cli can reproduce the extra fields.

`linkml generate jsonld-context --no-metadata --model --prefixes simple_uri_test.yaml`

One thing I realized is it is not necessary to move the code block in the context generator. While I'd still recommend a core dev review I'm otherwise confident enough today to recommend applying the patches to this PR.

tfliss (dismissed):
Just finishing my review. This is a great improvement, thanks for talking through the context behaviors with me, I understand it a lot better now.

At this point I'd recommend:
- applying the patches we discussed
- update the snapshot as appropriate to the patched output
- add a multi-context test parameter case to /test_scripts/test_gen_jsonld.py#test_metamodel_valid_calls()

amc-corey-cox:
@tfliss this looks like @Silvanoc has addressed your comments. Can you pleases take another look and make sure this is ready to go? Great work to both of you and thank you!

linkml/generators/jsonldgen.py
```diff
@@ -149,14 +149,14 @@ def visit_type(self, typ: TypeDefinition) -> None:
     def visit_subset(self, ss: SubsetDefinition) -> None:
         self._visit(ss)
 
-    def end_schema(self, context: str = None, **_) -> str:
+    def end_schema(self, context: Optional[Union[str, List[str]]] = [], **_) -> str:
         self._add_type(self.schema)
```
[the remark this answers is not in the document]
tfliss: I still recommend adding test coverage in the context of the parent github issue, but resolving to not block the current PR.
https://github.com/linkml/linkml/issues/2624

tfliss:
> @tfliss this looks like @Silvanoc has addressed your comments. Can you pleases take another look and make sure this is ready to go? Great work to both of you and thank you!

@amc-corey-cox Can you clarify the procedures/expectations of approval? Does merging to main wait until all of the issues under the parent https://github.com/linkml/linkml/issues/2624 are approved? On its own this PR still includes the model info in the snapshot JSON-LD despite the model=False flag being set. Once you confirm that's ok, I'll approve.

Silvanoc:
> Does merging to main wait until all of the issues under the parent #2624 are approved?

It's the other way around. This is a sub-issue of #2624. So merging this PR is a pre-condition (among others) for that issue to be fulfilled.

Silvanoc:
@tfliss @amc-corey-cox I'm trying to improve test coverage and observing some strange things. Therefore moving to "draft" to block merging until I have a clearer understanding.

Silvanoc:
Meanwhile all the different strange behaviors have been fixed and this code works as expected and is ready for reviewing.

Silvanoc:
@tfliss please have a look and mark your "change request" as resolved (if you see it so, of course).

packages/linkml/src/linkml/generators/jsonldgen.py
```diff
@@ -182,33 +185,34 @@ def end_schema(self, context: str | Sequence[str] | None = None, context_kwargs:
...
-            context = [metamodel_ctx, add_prefixes_json["@context"]]
+            contexts = [metamodel_ctx, add_prefixes_json["@context"]]
         elif isinstance(context, str):  # Some of the older code doesn't do multiple contexts
-            context = [context]
+            contexts = [context]
         elif isinstance(context, tuple):
-            context = list(context)
+            contexts = list(context)
+        else:
+            contexts = context
```
amc-corey-cox: This aliases a caller-supplied list rather than copying it. From the context (LOL) I can't figure out if this is intentional. It doesn't appear to be a problem now but if someone uses the API to pass context and then uses it later it could be a foot-gun.

tests/linkml/test_scripts/test_gen_jsonld.py
```diff
@@ -64,6 +64,42 @@ def test_simple_uris(input_path, snapshot):
...
+                "https://w3id.org/linkml/types.context.jsonld",
+                {
+                    "@base": "https://raw.githubusercontent.com/linkml/linkml/main/tests/test_scripts/input/simple_uri_test/"
+                },
+            ],
+        ),
+    ],
+)
+def test_context(input_path, arguments, mock_context_paths, context):
+    """Test generation of meta.yaml JSON"""
```
amc-corey-cox: Docstring appears to be a copy-paste from another test — this doesn't test `meta.yaml`. Something like `"""Test @context assembly when multiple --context files are passed."""` would be meaningful.

amc-corey-cox (approved):
@Silvanoc in addition to what @tfliss found I see a couple things that could use your attention.

packages/linkml/src/linkml/generators/jsonldgen.py
Silvanoc: Originally `context` was being directly used anyway, so this alias does not modify the original behavior in that sense. In fact the original code [was mutating `context`](https://github.com/linkml/linkml/blob/e94d2b8368ae15b06a0fa7323df7c18a8bcaccd2/packages/linkml/src/linkml/generators/jsonldgen.py#L191). And the [new code is also mutating `context`, only indirectly over `contexts`](https://github.com/linkml/linkml/blob/5dcc42a36114ba851d82e5966282a95362048833/packages/linkml/src/linkml/generators/jsonldgen.py#L194) (will be renamed to `context_list` to improve code readability).

Please resolve this conversation if satisfied with the explanation.

tests/linkml/test_scripts/test_gen_jsonld.py
Silvanoc: Correct, fixed.

Silvanoc:
Tests currently failing due to #3708

packages/linkml/src/linkml/generators/jsonldgen.py
amc-corey-cox: Yeah, that seems fine. I flagged it in case it was unintentional.

amc-corey-cox:
@Silvanoc I think I fixed the vendoring breakage. Sorry about that, I didn't realize an innocuous change to linkml-model was going to create a headache.
I'll verify this fixes Tim's requested changes and CI passes then merge.

packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
```diff
@@ -1927,13 +1927,33 @@ def _get_string_type(self) -> TypeDefinition:
...
         """
         slot_range = slot.range
         if slot_range in self.all_classes():
-            if slot.inlined or slot.inlined_as_list:
+            # Resolve inlined through the slot's is_a chain so that a slot that
+            # inherits inlined=True from a parent slot is correctly detected.
+            resolved_inlined = slot.inlined
+            resolved_inlined_as_list = slot.inlined_as_list
+            parent_name = getattr(slot, "is_a", None)
+            while (resolved_inlined is None and resolved_inlined_as_list is None) and parent_name:
```
amc-corey-cox: Sorry to ask this after approving — good catch, but could it move to its own PR (you've already got #3695)? The JSON-LD generators don't call `is_inlined`, so it's orthogonal to this change. And real call sites pass induced slots, where `induced_slot` already resolves inherited `inlined` via both `is_a` and `mixins` — this walk is `is_a`-only. Keeping it separate lets the JSON-LD fix merge clean. Apologies for the round-trip!

amc-corey-cox (changes requested):

packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
Silvanoc: Sorry, that's an error of mine, not yours. Rebasing I've pulled a commit from #3695.

I've cleaned it up.

Silvanoc:
@amc-corey-cox since I consider the concerns of @tfliss addressed by now and he's neither reacting to this PR nor giving any feedback to his "change requests", I'm dismissing his review. Otherwise the PR would be blocked even if you approve it.

amc-corey-cox (approved):
This looks good to me now. Sorry for my delay in getting to it. It was just a matter of not noticing I'd passed it over and getting distracted by other things. Nice work and thank you.

Silvanoc:
~@matentzn the CI pipeline of this PR was failing due to #3767, I've manually upgraded the vulnerable dependencies and how the tests are passing. @amc-corey-cox has already approved, may I merge it once the codecov checks are also passed?~

The dependencies issue is now fixed on main, I'll merge it as soon as the CI pipeline passes.

Issue #2568, opened 2025-03-14
Create a roll-up transformer


kevinschaper:
My particular motivation is that I want to be able to transform a property graph schema like Biolink Model into a flattened schema like kgx, where I have just an Entity class an Association class that rolls up all of the properties from all of the descendants of those classes. 

The roll up transformer can take in a schema and a single class name, and will produce a schema where that class has no descendants, but gathers up all slots from all descendants. 

The one weirdness that I can imagine is what to do about a class designator slot, in the case of Biolink, I would still want the original class name in the category field, rather than being forced to just be Association. 

Edit to add this is also of value for implementing a single table inheritance pattern for the SQL DDL generator

kevinschaper:
Resolved by https://github.com/linkml/linkml/pull/2618

PR #2618, opened 2025-04-14
rollup transformer implementation #2568

iQuxLE:
implementing #2568

- preserves original designator slot behavior

### DIFF
--- a/linkml/transformers/rollup_transformer.py
+++ b/linkml/transformers/rollup_transformer.py
@@ -77,24 +77,10 @@ class RollupTransformer(ModelTransformer):
 
         schema_copy = copy.deepcopy(self.source_schema)
         view = SchemaView(schema_copy)
-
-        self._identify_descendants(view)
+        descendants = view.class_descendants(self.target_class, imports=True, mixins=True, reflexive=False, is_a=True)
+        self._descendant_classes = set(descendants)
         self._collect_slots(view)
         return self._create_flattened_schema(schema_copy, view)
-
-    def _identify_descendants(self, view: SchemaView) -> None:
-        """Identify all descendant classes of the target class."""
-        direct_children = set(view.get_children(self.target_class))
-        all_descendants = set()
-        pending = list(direct_children)
-        # BFS - avoid recursion
-        while pending:
-            current = pending.pop()
-            if current not in all_descendants:
-                all_descendants.add(current)
-                children = view.get_children(current)
-                pending.extend(children)
-        self._descendant_classes = all_descendants
 
     def _collect_slots(self, view: SchemaView) -> None:
         """
--- a/linkml/transformers/rollup_transformer.py
+++ b/linkml/transformers/rollup_transformer.py
@@ -1,0 +1,146 @@
+"""
+flatten_transformer.py
+
+This module provides a FlattenTransformer that transforms a LinkML schema into a flattened version.
+The transformer takes a target class name and rolls up all slots from its descendants into that class.
+This is useful for implementing single table inheritance or simplifying inheritance for systems like KGX.
+"""
+
+import copy
+from dataclasses import dataclass
+from typing import Optional
+
+from linkml_runtime.linkml_model.meta import (
+    ClassDefinitionName,
+    SchemaDefinition,
+    SlotDefinition,
+)
+from linkml_runtime.utils.schemaview import SchemaView
+
+from linkml.transformers.model_transformer import ModelTransformer
+
+
+@dataclass
+class FlattenTransformerConfiguration:
+    """
+    Configuration options for the FlattenTransformer.
+
+    Attributes:
+        preserve_class_designator: If True, preserves the original class name in the specified
+            class designator slot (e.g., 'category').
+        class_designator_slot: The name of the slot that designates the original class (e.g., 'category').
+        include_all_classes: If True, keeps all classes in the schema; otherwise, only the target class
+            and non-descendants are kept.
+        include_mixins: If True, includes properties from mixins in the flattened class.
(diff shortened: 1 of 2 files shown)


linkml/transformers/rollup_transformer.py
```diff
@@ -0,0 +1,147 @@
...
+            raise ValueError("SchemaView not set. Ensure schema is set by calling set_schema().")
+
+        schema_copy = copy.deepcopy(self.source_schema)
+        view = SchemaView(schema_copy)
+
+        self._identify_descendants(view)
+        self._collect_slots(view)
+        return self._create_flattened_schema(schema_copy, view)
+
+    def _identify_descendants(self, view: SchemaView) -> None:
```
kevinschaper, line 85: This can be done with class_descendants in SchemaView (I think? unless something extra is happening here)

iQuxLE: Sure! Good catch, I was not aware this exists.

amc-corey-cox: This looks like it meets @kevinschaper request.

amc-corey-cox (approved):
This looks good to me. I'll give @kevinschaper a couple of days in case he wants to re-review and then I'll merge. Great work and thank you!

tests/test_transformers/test_rollup_tranformer.py
```diff
@@ -0,0 +1,269 @@
...
+    transformer.set_schema(sb.schema)
+    flattened_schema = transformer.transform()
+
+    # Verify Association has rolled up all slots from descendants
+    association_class = flattened_schema.classes["Association"]
+    expected_slots = {"id", "name", "subject", "object", "gene", "specificity"}
+
+    assert set(association_class.slots) == expected_slots
+    # Verify Entity is still there (not a descendant of Association)
+    assert "Entity" in flattened_schema.classes
```
ialarmedalien, line 95: Can you also check that the descendants of Association are _not_ present?

linkml/transformers/rollup_transformer.py
```diff
@@ -0,0 +1,133 @@
...
+        Collect all slots from the target class and its descendant classes.
+        Populates a mapping of class name to the set of its slot names.
+        """
+        # Get induced slots for the target class.
+        target_slots = view.class_induced_slots(self.target_class)
+        self._collect_slots_for_class[self.target_class] = target_slots
+
+        for descendant in self._descendant_classes:
+            slots = view.class_induced_slots(descendant)
+            self._collect_slots_for_class[descendant] = slots
```
ialarmedalien, line 96: This is a slightly confusing name as it sounds like a method. Can you call it something like `self._class_to_induced_slot_map`, and maybe have the other one called `self._class_to_induced_slot_name_map` (since it's the same slots, but just records the slot names)?

linkml/transformers/rollup_transformer.py
```diff
@@ -0,0 +1,133 @@
...
+        existing_slot_names = set(target_class_def.slots or [])
+        all_slots = set()
+
+        for descendant in self._descendant_classes:
+            if descendant in schema.classes:
+                for slot_name in self._class_to_slot_map.get(descendant, set()):
+                    if slot_name not in existing_slot_names:
+                        all_slots.add(slot_name)
+
+        #!
```
ialarmedalien, line 113: ?

linkml/transformers/rollup_transformer.py
ialarmedalien: On second thoughts, `self._collect_slots_for_class` doesn't get used once it is set, so there's no point in keeping it. Naming problem solved!

linkml/transformers/rollup_transformer.py
```diff
@@ -0,0 +1,133 @@
...
+        if target_class_def.slots:
+            target_class_def.slots.extend(list(all_slots))
+        else:
+            target_class_def.slots = list(all_slots)
```
ialarmedalien, line 117: This is personal taste, but I prefer doing

```py
if not target_class_def.slots:
    target_class_def.slots = []

target_class_def.slots.extend(list(all_slots))
```

I find it a little simpler/clearer but YMMV.

linkml/transformers/rollup_transformer.py
```diff
@@ -0,0 +1,133 @@
...
+
+        # Optionally remove descendant classes.
+        if not self.config.include_all_classes:
+            for descendant in self._descendant_classes:
+                if descendant in schema.classes:
+                    del schema.classes[descendant]
+
+        if self.config.preserve_class_designator and self.config.class_designator_slot:
+            self._handle_class_designator(schema)
```
ialarmedalien, line 126: do you even need to call it if the method doesn't do anything?

tests/test_transformers/test_rollup_tranformer.py
```diff
@@ -0,0 +1,280 @@
...
+    Test a simple schema with a base 'Association' class and two descendant classes.
+    The descendant slots should be rolled up into the base class and descendant classes removed.
+    """
+    sb = SchemaBuilder()
+    # Base class: Association with basic properties.
+    sb.add_class("Association", slots=["id", "subject", "predicate", "object", "category"])
+    # Descendant 1: GeneticAssociation adds gene-specific slots.
+    sb.add_class(
+        "GeneticAssociation",
+        slots={"gene_variant": {"range": "str"}, "inheritance_mode": {"range": "str"}},
```
ialarmedalien, line 24: Doesn't linkml use `string` (rather than `str`) as the designation for strings?

sierra-moxon (approved):

PR #2627, opened 2025-04-15
fix(jsonldcontextgen): avoid shadowing dataclass fields

### LINKED ISSUE #2623: `ContextGenerator().serialize()` ignores dataclass fields
**Describe the bug**
If I get it right, `ContextGenerator().serialize()` is expected to get its arguments in this priority order:
1. What is directly [passed to `serialize`](https://github.com/linkml/linkml/blob/bf5861b56088e84a799f20ec9b16dee7f5691627/linkml/generators/jsonldcontextgen.py#L234) as argument.
2. The equivalent [dataclass field](https://github.com/linkml/linkml/blob/bf5861b56088e84a799f20ec9b16dee7f5691627/linkml/generators/jsonldcontextgen.py#L47).
3. The [`end_schema` specified defaults](https://github.com/linkml/linkml/blob/bf5861b56088e84a799f20ec9b16dee7f5691627/linkml/generators/jsonldcontextgen.py#L82).

The problem is that since `end_schema` gets defaults on all the arguments, there is no way to differentiate those passed to `serialize` as arguments from the defaults of `end_schema`. As a consequence in the end the dataclass fields are being ignored.

That issue is overcome in the [jsonldcontextgen.py code](https://github.com/linkml/linkml/blob/bf5861b56088e84a799f20ec9b16dee7f5691627/linkml/generators/jsonldcontextgen.py#L234) applying the same arguments on both the constructor and `serialize`, but I don't see any sense (or is at least not intuitive) on having `end_schema` overriding what has been declared in the constructor.

Source code of commit bf5861b56088e84a799f20ec9b16dee7f5691627

Silvanoc:
~‼️ NOTE: this PR builds upon #2621 and is not expected to work before that PR has been merged, it will be rebased on the branch of that PR to ensure that it does not break current tests~

Assignation of non-None default values to "end_schema" is shadowing the values assigned to the corresponding dataclass fields in the constructor. This patch fixes it.

After this fix values assigned in the dataclass constructor will be taken, unless different values are provided to "end_schema" directly. Otherwise default values are used.

Fixes #2623

### DIFF
--- a/packages/linkml/src/linkml/generators/jsonldcontextgen.py
+++ b/packages/linkml/src/linkml/generators/jsonldcontextgen.py
@@ -330,7 +330,7 @@ def cli(yamlfile, emit_frame, embed_context_in_frame, output, **args):
         gen.embed_context_in_frame = True
     else:
         gen.emit_frame = emit_frame
-    print(gen.serialize(output=output, **args))
+    print(gen.serialize(output=output))
 
 
 if __name__ == "__main__":
--- a/packages/linkml/src/linkml/generators/jsonldgen.py
+++ b/packages/linkml/src/linkml/generators/jsonldgen.py
@@ -178,7 +178,7 @@ class JSONLDGenerator(Generator):
             # TODO: The _visit function above alters the schema in situ
             # force some context_kwargs
             context_kwargs["metadata"] = False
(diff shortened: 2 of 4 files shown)


Silvanoc:
Fix by the previous PR(s).

Issue #2615, opened 2025-04-12
Migrate to uv & ruff


dalito:
As discussed in the developer meeting on 2025-04-04, the project wants to move to modern tools.

linkml and linkml-runtime currently use poetry 1.x, which uses its own "standard". So switching to uv requires to 

- make pyproject.toml PEP626 conform
- update the actions to use uv as packaging tool

Since poetry 2.x added compliance with PEP621, it is suggest to move to poetry 2.x first and switch to uv next.

First step **poetry 2.x** (keeping poetry-dynamic-versioning)
- [x] Update linkml-runtime - linkml/linkml-runtime#387
- [x] Update linkml - #2633

Second step [uv](https://docs.astral.sh/uv/) & [uv-dynamic-versioning](https://github.com/ninoseki/uv-dynamic-versioning) & [hatch](https://hatch.pypa.io/latest/):
- [x] Migrate linkml-runtime to uv - linkml/linkml-runtime#447
- [x] Migrate linkml to uv - #2779

After the move uv, `uv tool` can also be use instead of `pipx` reducing the number of needed tools.

An example linkml-project that we migrated successfully to uv, uv-dynamic-versioning and hatch is [pid4cat-model](https://github.com/nfdi4cat/pid4cat-model) which may serve as example for the required changes.

ialarmedalien:
Migration of linkml-map to uv: https://github.com/linkml/linkml-map/pull/54 -- reviewers welcome!

PR #2633, opened 2025-04-18
Update pyproject.toml & actions to poetry 2.x / PEP 621

dalito:
- change `pyproject.toml` to PEP621 style for poetry 2.x
- update all gh-actions for poetry 2.x
- changed  `pypi-publish` action to follow the [suggested 2-step process](https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-python#publishing-to-pypi) of building in one job and publishing in another
- change `pypi-publish` action to use [trusted publishing](). Note that the env name is not `pypi`as in the screenshots of the PyPI-docs but `pypi-release`.
- enable dependabot for monthly action-updates (it will create PRs but not autoupdate)

Related issue linkml/linkml#2615

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: monthly
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,18 +7,20 @@ updates:
     directory: "/"
     schedule:
       interval: monthly
-  - package-ecosystem: "pip"
-    directories:
-      - "/"
-    schedule:
-      interval: "weekly"
-      day: "sunday"
-    groups:
-      # Individual pull requests for major/minor updates and grouped for patch updates
-      angular:
-        applies-to: version-updates
-        patterns:
-          - "*"
-        update-types:
-          - "patch"
-    open-pull-requests-limit: 10
+
+  # Activate after migration to uv as package manager
+  # - package-ecosystem: "uv"
+  #   directories:
+  #     - "/"
+  #   schedule:
+  #     interval: "weekly"
+  #     day: "sunday"
+  #   groups:
+  #     # Individual pull requests for major/minor updates and grouped for patch updates
+  #     angular:
+  #       applies-to: version-updates
+  #       patterns:
+  #         - "*"
+  #       update-types:
+  #         - "patch"
+  #   open-pull-requests-limit: 10
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,5 @@
 # Please see the documentation for all configuration options:
-# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+# https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference
 
 version: 2
 updates:
@@ -8,7 +8,17 @@ updates:
     schedule:
       interval: monthly
   - package-ecosystem: "pip"
-      directories:
-        - "/"
-      schedule:
-        interval: weekly
+    directories:
+      - "/"
+    schedule:
+      interval: "weekly"
+      day: "sunday"
+    groups:
+      # Individual pull requests for major/minor updates and grouped for patch updates
+      angular:
+        applies-to: version-updates
+        patterns:
+          - "*"
+        update-types:
+          - "patch"
+    open-pull-requests-limit: 10
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,12 @@
 
 version: 2
 updates:
-- package-ecosystem: github-actions
-  directory: "/"
-  schedule:
-    interval: monthly
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: monthly
+  - package-ecosystem: "pip"
+      directories:
+        - "/"
+      schedule:
+        interval: weekly
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -18,8 +18,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - name: Install Poetry 1.x
-        run: pipx install poetry==1.*
+      - name: Install Poetry
+        run: |
+          pipx install poetry
+          pipx inject poetry poetry-dynamic-versioning
       - uses: actions/setup-python@v5
         with:
           python-version: 3.13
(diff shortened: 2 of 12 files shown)


dalito:
Now the Docker image build works as well.

amc-corey-cox (approved):
This all looks good to me. I'll give @pkalita-lbl a couple days to weigh in and then I'll merge it. Great work and thank you!

pyproject.toml
```diff
@@ -1,52 +1,89 @@
...
+]
+shacl = [
+    "pyshacl >= 0.25.0",
+]
+tests = [
+    # "linkml[black]",  # recursive dependency not supported by poetry 
+    "black >= 24.0.0",
+    "numpydantic >= 1.6.1",
+    "pyshacl >= 0.25.0",
+]
```
pkalita-lbl, line 84: These changes do faithfully replicate what we previously had with `[tool.poetry.extras]`. But the `black` and `tests` groups don't really make sense to me. Those are development dependencies, and I'd expect to see them in `[tool.poetry.group.<group>.dependencies]` (for Poetry) or `[dependency-groups]` (for PEP 735-compatible tools). Anyway, no action needed now, but something to a closer look at as we continue to refine our project tooling.

.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: monthly
```
ialarmedalien, line 9: add in dependabot config for python, too

.github/workflows/main.yaml
```diff
@@ -52,21 +54,26 @@ jobs:
...
           git remote add upstream https://github.com/linkml/linkml
           git fetch upstream --tags
+
       - name: Install poetry
-        run: pipx install poetry==1.*
-      - name: Install dynamic versioning plugin
-        run: poetry self add "poetry-dynamic-versioning[plugin]"
+        run: |
+          pipx install poetry
+          pipx inject poetry poetry-dynamic-versioning
```
ialarmedalien, line 75: poetry-dynamic-versioning is only required if you're making a release

.github/workflows/poetry-update.yaml
```diff
@@ -18,7 +18,8 @@ jobs:
       with:
         python-version: "3.13"
     - name: Install Poetry
-      uses: snok/install-poetry@v1
+      run: |
```
ialarmedalien, line 21: this whole workflow could be replaced by dependabot

.github/dependabot.yml
ialarmedalien: Thank you! 😁

pkalita-lbl: If we're going to allow Dependabot handle Python dependencies, then I would suggest:

* Apply [`versioning-strategy: lockfile-only`](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#versioning-strategy--) so that Dependabot only touches the lock file and leaves the version constraints listed in `pyproject.toml` alone. This might be erring on the side of caution, but it's what I'm familiar with using on other projects.
* Consider using [`groups`](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--) to batch multiple updates into a single PR. Without that you end up getting _a lot_ of version update PRs. The tradeoff is that it might be less obvious which particular package update broke something if the PR tests fail.
* Remove the old `poetry-update.yaml` workflow, as it would be redundant with what Dependabot is doing.

Honestly, now that I've typed all that out it might make more sense to introduce the Dependabot config in a separate PR.

ialarmedalien: > Apply [versioning-strategy: lockfile-only](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#versioning-strategy--) so that Dependabot only touches the lock file and leaves the version constraints listed in pyproject.toml alone. This might be erring on the side of caution, but it's what I'm familiar with using on other projects.

How are you specifying versions in your requirements.txt/pyproject.toml/other config file? With a specific version or with a version range? Is there the potential for the lock file to get out of sync with the package list?

I would have thought this strategy creates extra work, since you have to manually update the package list instead of having dependabot doing it for you. 

> Consider using [groups](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--) to batch multiple updates into a single PR. Without that you end up getting a lot of version update PRs. The tradeoff is that it might be less obvious which particular package update broke something if the PR tests fail.

sample config using `groups`:
```yaml
- package-ecosystem: "pip"
  directory: "/src"
  schedule:
    interval: "monthly"
  groups:
    pip:
      patterns:
        - "*"
  open-pull-requests-limit: 20
```

pkalita-lbl: > How are you specifying versions in your requirements.txt/pyproject.toml/other config file? With a specific version or with a version range?

Ranges, with very very few exceptions.

> Is there the potential for the lock file to get out of sync with the package list?

No, the version constraints in `pyproject.toml` are still checked, but not changed. Say `pyproject.toml` has a dependency specification of `foo >= 2.5.0, < 3.0.0` and currently `poetry.lock` specifies the version of `foo` as `2.6.4`. If `foo` version `2.7.0` comes out, Dependabot will update the `poetry.lock` to specify `2.7.0` for `foo`. But if `foo` version `3.0.0` comes out, Dependabot won't do anything with it.

It leaves us in complete control of the version specifiers in `pyproject.toml`, and I feel like I can wrap my head around that a lot better than Dependabot's default versioning strategy of:

> * Try to differentiate between app and library dependencies.
> * For apps, always increase the minimum version requirement to match the new version. The `increase` strategy.
> * For libraries, widen the allowed version requirements to include both the new and old versions, when possible. The `widen` strategy.

dalito: Poetry is not listed as supported tool for [lockfile-only](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#versioning-strategy--)

I pushed a change to group patch updates and to have a limit of 10 open Dependabot package-update PRs.

@pkalita-lbl - Would you be OK with this or would you prefer a separate PR for enabling Python-dependency updates via Dependabot?

pkalita-lbl: `versioning-strategy` is supported by the `pip` ecosystem. And that's the [ecosystem](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#package-ecosystem-) for Poetry.

Regardless, let's pause on the Python Dependabot stuff for now. I just now realized that Dependabot doesn't yet support PEP 621-style `pyproject.toml` via Poetry (https://github.com/dependabot/dependabot-core/issues/11237). The other projects where I have Dependabot updating Python dependencies still use the Poetry-specific sections in `pyproject.toml`.

ialarmedalien: uv is supported by dependabot, so the migration could happen then.

Somewhat mysterious what dependabot considers 'app' vs 'library' dependencies -- there are no definitions on that page, and I can't seem to find any elsewhere. I wonder what it is based on.

dalito: Ok. I changed dependabot back to only update the actions for now. The python package updates can then be added after or as part of the migration to uv.

pkalita-lbl: Looks good! Thanks

dalito:
Merging this should be delayed until safe-publishing has successfully been activated and run for linkml-runtime v1.9.2-rc1. - Afterwards we can update pypi-publish action here to match the one in linkml-runtime.

PR #2637, opened 2025-04-22
Deprecate `gen-markdown` generator

sujaypatil96:
This PR seeks to deprecate `gen-markdown` by following the deprecation protocol described here: https://linkml.io/linkml/developers/deprecation.html and announce/recommend via warnings, that the defacto markdown documentation generator is now `gen-doc`.

- [x] update linkml documentation pages to "announce" this as well.

### DIFF
--- a/linkml/generators/markdowngen.py
+++ b/linkml/generators/markdowngen.py
@@ -18,9 +18,9 @@ from linkml_runtime.utils.formatutils import be, camelcase, underscore
 
 from linkml._version import __version__
 from linkml.generators.yumlgen import YumlGenerator
+from linkml.utils.deprecation import deprecation_warning
 from linkml.utils.generator import Generator, shared_arguments
 from linkml.utils.typereferences import References
-from linkml.utils.deprecation import deprecation_warning
 
 
 @dataclass
--- a/linkml/generators/markdowngen.py
+++ b/linkml/generators/markdowngen.py
@@ -20,6 +20,7 @@ from linkml._version import __version__
 from linkml.generators.yumlgen import YumlGenerator
 from linkml.utils.generator import Generator, shared_arguments
 from linkml.utils.typereferences import References
+from linkml.utils.deprecation import deprecation_warning
 
 
 @dataclass
@@ -40,6 +41,7 @@ class MarkdownGenerator(Generator):
     """
 
     def __post_init__(self) -> None:
+        deprecation_warning("gen-markdown")
         super().__post_init__()
 
     # ClassVars
@@ -49,8 +51,6 @@ class MarkdownGenerator(Generator):
     valid_formats = ["md"]
     visit_all_class_slots = False
     uses_schemaloader = True
-    deprecated = True
-    deprecation_name = "gen-markdown"
 
     # ObjectVars
     directory: Optional[str] = None
@@ -815,6 +815,8 @@ def cli(yamlfile, map_fields, dir, img, index_file, notypesdir, warnonexist, **k
     .. warning::
         `gen-markdown` is deprecated. Please use `gen-doc` instead.
     """
+    deprecation_warning("gen-markdown")
+
     gen = MarkdownGenerator(yamlfile, no_types_dir=notypesdir, warn_on_exist=warnonexist, **kwargs)
     if map_fields is not None:
         gen.metamodel_name_map = {}
--- a/linkml/generators/markdowngen.py
+++ b/linkml/generators/markdowngen.py
@@ -18,7 +18,6 @@ from linkml_runtime.utils.formatutils import be, camelcase, underscore
 
 from linkml._version import __version__
 from linkml.generators.yumlgen import YumlGenerator
-from linkml.utils.deprecation import deprecation_warning
 from linkml.utils.generator import Generator, shared_arguments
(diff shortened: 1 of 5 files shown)


sujaypatil96:
Sorry for the additional commits after requesting for a review @sneakers-the-rat. This PR is now ready for review.

amc-corey-cox (approved):
This looks good to me. But should we set a time-frame for removal? Or just do that at a later time? I'll wait to see if @sneakers-the-rat wants to weigh in before I merge this tomorrow.

sujaypatil96:
@amc-corey-cox hm, i'm wondering if we should wait till Friday to merge this? it would be nice to bring this up at the dev call, give an overview of this "deprecation protocol" to the dev team and then merge this in.

amc-corey-cox:
Sounds great. Let's wait until then.

linkml/generators/markdowngen.py
```diff
@@ -31,15 +31,26 @@ class MarkdownGenerator(Generator):
...
+
+    .. warning::
+        The MarkdownGenerator class is being deprecated in favor of DocGenerator which can
+        be found at the following path – `linkml/generators/docgen.py`. Going forward, DocGenerator
+        which can be invoked using the `gen-doc` command will be the preferred way to generate
+        Markdown documentation files for LinkML schemas.
     """
 
+    def __post_init__(self) -> None:
+        super().__post_init__()
```
sneakers-the-rat, line 43: this would be where you want to put the call to `deprecation_warning`, but as is it doesn't seem like this does anything?

linkml/utils/generator.py
```diff
@@ -190,6 +197,28 @@ def __post_init__(self) -> None:
...
+            # ensure the `deprecated` flag is set if `deprecation_name` is provided
+            if not self.deprecated:
+                self.logger.warning(
+                    f"Setting deprecated=True for generator with deprecation_name='{self.deprecation_name}'"
+                )
+                self.deprecated = True
+
+            # issue the deprecation warning
+            deprecation_warning(self.deprecation_name)
+
```
sneakers-the-rat, line 221: I'm not sure why we would want to do this rather than just emitting the `deprecation_warning` in the post init?

sneakers-the-rat (commented):
See the existing deprecation documentation: https://linkml.io/linkml/developers/deprecation.html

In particular:

> ## Mark the Deprecation
> Everywhere the deprecated feature would be used, add a `deprecation_warning()` call that uses the name of the Deprecation object.
>
> ...
>
> ### Class
> 
> ```python
> class MyClass:
>     def __init__(self):
>         deprecation_warning('semaphore')
> ```

I am not sure why we would duplicate that by adding more classvars to the generators? It seems like it would be easy enough to
- call `deprecation_warning('gen-markdown')` in `MarkdownGenerator.__post_init__`
- call `deprecation_warning('gen-markdown')` in `markdowngen.cli`

since the whole deal with deprecations is that they are going to be removed, dirtying up the deprecated class/module with the calls to deprecation warning seems better than adding something to all the generators to me.

no strong feelings, i just am not sure why we need a second, parallel mechanism of declaring deprecations that calls the first

linkml/utils/generator.py
sneakers-the-rat: we would also probably want tests for this

sneakers-the-rat:
also re: adding deprecations to the docs, that's part of what the existing deprecation system does: 

Screenshot 2025-04-23 at 6 25 42 PM

and you can see how that works in the `docs/code/deprecation.md` page if you wanted to mark that directly on the class's documentation page, something like

```python
class MarkdownGenerator(Generator):
    f"""
    other stuff
    
    .. admonition:: Deprecated
        :class: warning
        
        .. deprecated:: {dep.deprecated_in}
        
        {dep.message}
        
        Recommendation: {dep.recommendation}
    """
    
```

We could of course make a decorator to do that automatically + emit the warning if we wanted to. The reason the existing implementation is not a decorator is because there are things we want to deprecate that aren't classes or functions (like dependencies, params, etc.) and can't be decorated really, but for a class it would probably be nice to have

```
@deprecated('gen-markdown')
class MarkdownGenerator(Generator): ...
```

and have that wrap `__init__` and modify `__doc__

sujaypatil96:
Thanks for the helpful review comments @sneakers-the-rat! super appreciated 🙏🏼 The `deprecation_warning()` approach you suggested works perfectly! I implemented that in `markdowngen.py` just now.

linkml/generators/markdowngen.py
sujaypatil96: Didn't fully realize that this was an alternative deprecation approach while I was implementing it, and you're exactly right, using the current deprecation approach with `deprecation_warning()` and making calls to it in `__post_init__()` and `cli()` should do the trick!

linkml/utils/generator.py
sujaypatil96: Makes sense! i'll add a test for this. I'll add a test to `tests/test_generators/test_markdowngen.py`.

sujaypatil96:
I think this PR is good to go @sneakers-the-rat! 🚀 I'll implement the `@deprecated` decorator in a separate PR.

PR #2639, opened 2025-04-24
Add `truncate_descriptions` option to docgen

### LINKED ISSUE #2638: docgen truncates descriptions for HTML docs
The problem:

The markdown documentation generator cuts off `description` elements in preparation for HTML rendering via mkdocs.

(The original title of this issue unintentionally demonstrates how annoying it is not to have a full description.)

This makes for poor UX for anyone whose schema has numerous multi-line descriptions as users have to click back and forth between pages to see the full description for the class, attribute, type, etc.

There are various ways to deal with long text elements using CSS, JavaScript, or specialised markdown elements (e.g. [mkdocs-material's tooltips](https://squidfunk.github.io/mkdocs-material/reference/tooltips/#adding-tooltips)), but because text truncation occurs at the markdown generation stage, there is no way to recover the full description.

Having the default being to cut off descriptions to save on vertical space seems counter to both the point of being able to provide rich descriptions and to providing good UX. Web users are used to scrolling vertically and it's generally better to reduce the number of clicks / page navigation events.

Possible solutions:

- Command-line option to turn on / off truncation of descriptions
- Supplementary JS or CSS solution that can be used by those who want to truncate descriptions

I would guess that not many people use long descriptions as [the bugs that I found in a couple of the generators](https://github.com/linkml/linkml/pull/2607) were due to the assumption that descriptions would be a single line.

ialarmedalien:
As noted in https://github.com/linkml/linkml/issues/2638, the docgen cuts off descriptions that are over a line in length, which isn't to everyone's taste. This PR adds a `truncate_descriptions` command line option to selectively enable and disable description truncation.

Closes #2638 

See also discussion in #2607


linkml/generators/docgen.py
```diff
@@ -959,6 +965,12 @@ def example_object_blobs(self, class_name: str) -> list[tuple[str, str]]:
                 objs.append((stem, f.read()))
         return objs
 
+    def customize_environment(self, env: Environment):
+        if self.truncate_descriptions:
+            env.filters["enshorten"] = enshorten
+        else:
+            env.filters["enshorten"] = text_to_web
```
ialarmedalien, line 972: rather than edit all the templates, this just switches out the `enshorten` function, which truncates text, for a function that converts line breaks to `<br>` elements for web display (any `\n` line breaks in description elements break the layout of markdown tables).

tests/test_generators/test_docgen.py
```diff
@@ -322,64 +325,243 @@ def test_docgen(kitchen_sink_path, input_path, tmp_path):
     assert len(set(domain_of_species_name)) == len(domain_of_species_name)
 
 
-def test_docgen_multiline_description_hierarchical_view(tmp_path) -> None:
```
ialarmedalien, line 325: I replaced this test, which checks a couple of multi-line description scenarios, with comprehensive tests for most? all? the different elements that can have multi-line descriptions. Note that multi-line descriptions are only a problem in the markdown tables (where the `enshorten` filter is used), so that's specifically what I'm testing.

tests/test_generators/test_docgen.py
```diff
@@ -322,64 +325,243 @@ def test_docgen(kitchen_sink_path, input_path, tmp_path):
...
+        hierarchical_class_view=hierarchical_class_view,
+        truncate_descriptions=truncate_descriptions,
+        mergeimports=False,
+    )
+    gen.serialize(directory=str(tmp_path))
 
-    for line in ix:
-        assert_mdfile_contains(tmp_path / "index.md", line)
+    # index.md
+    index_file = tmp_path / "index.md"
```
ialarmedalien, line 472: Start by checking the index file, which has tables for classes, slots, types, enums, and subsets.

linkml/generators/docgen.py
```diff
@@ -76,8 +76,11 @@ def enshorten(input):
     return input
 
 
-def customize_environment(env: Environment):
-    env.filters["enshorten"] = enshorten
+def text_to_web(input) -> str:
+    """Custom filter to convert multi-line text strings into a suitable format for web/markdown output."""
+    if input is None:
+        return ""
+    return "<br>".join(input.strip().split("\n"))
```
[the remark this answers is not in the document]
ialarmedalien: Good point. It depends a bit on whether or not the description has deliberate line breaks or not -- in my case, my IDE wraps lines and I add line breaks for formatting (e.g. between paragraphs). Some people hard-wrap lines so your suggestion would probably be better for those cases.

I suppose that I could check for `\n\n` and replace that with a `<br>` and replace `\n\S` with a space. 🤔

ialarmedalien:
tagging @sierra-moxon as this may be of interest to her as someone who prefers to show the full descriptions

linkml/generators/docgen.py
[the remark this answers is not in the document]
dalito: Hmm, this in an argument to keep it as is. 

Even better than adding some custom line-break rules would be having a way to optionally mark the format of a description as either "markdown" or "text". The current rendering could be kept as default "text" and the markdown rendering could be done with an existing markdown converter.

sierra-moxon (approved):

PR #2640, opened 2025-04-24
Show information about asserted rules on class documentation pages

sujaypatil96:
Here is an example of how the `Rules` section would look on a class documentation page.

This jinja template has been applied to the [nmdc-schema](https://github.com/microbiomedata/nmdc-schema). Below is a screenshot of the [CalibrationInformation](https://github.com/microbiomedata/nmdc-schema/blob/main/src/schema/nmdc.yaml#L303-L342) class documentation page which has _rules_ asserted on it.

Screenshot 2025-04-24 at 12 27 30 PM


sujaypatil96:
Feedback from @cmungall:
- [ ]  Add tests to new methods in docgen.py
- [ ] Show as YAML instead JSON

sujaypatil96:
Actions erroring out because of codespell issue on `main`.

linkml/generators/docgen.py
```diff
@@ -959,6 +960,76 @@ def example_object_blobs(self, class_name: str) -> list[tuple[str, str]]:
...
+        #   }
+        # }
+        # the "name" keys are redundant and do not add any value in the user documentation
+        def remove_name_keys(obj):
+            if isinstance(obj, dict):
+                return {k: remove_name_keys(v) for k, v in obj.items() if k != "name"}
+            elif isinstance(obj, list):
+                return [remove_name_keys(item) for item in obj]
+            else:
+                return obj
```
ialarmedalien, line 1003: unless there is a reason to nest this function inside the parent function, can you move it up to the top level? It makes it much easier to test.

linkml/generators/docgen.py
```diff
@@ -959,6 +960,76 @@ def example_object_blobs(self, class_name: str) -> list[tuple[str, str]]:
...
+            "preconditions": None,
+            "postconditions": None,
+            "elseconditions": None,
+        }
+
+        for key in ["preconditions", "postconditions", "elseconditions"]:
+            condition_obj = getattr(rule, key, None)
+            if condition_obj:
+                json_obj = json_dumper.to_dict(condition_obj)
+                sanitzed_condition = remove_name_keys(json_obj)
```
ialarmedalien, line 1017: typo :)

linkml/generators/docgen.py
```diff
@@ -959,6 +960,76 @@ def example_object_blobs(self, class_name: str) -> list[tuple[str, str]]:
                 objs.append((stem, f.read()))
         return objs
 
+    def classrule_to_dict_view(self, rule: ClassRule) -> dict[str, Any]:
+        """This method can iterate over a ClassRule object asserted on the `rule`
+        slot of a class. This method will return a value that contains four pieces of
+        information about the rule – _title_, _preconditions_, _postconditions_,
+        and _elseconditions_. This return value will be read in the jinja template,
+        and appropriately formatted into a tablular view for users to be able to
```
ialarmedalien, line 968: `tablular` => `tabular`

tests/test_generators/test_docgen.py
```diff
@@ -391,6 +393,41 @@ def test_docgen_rank_ordering(kitchen_sink_path, tmp_path):
...
+        after="| Rule Applied | Preconditions | Postconditions | Elseconditions |",
+    )
+
+    # Check that postconditions correctly show the heartfelt requirement
+    assert_mdfile_contains(
+        tmp_path / "EmploymentEvent.md",
+        "heartfelt",
+        after="| Rule Applied | Preconditions | Postconditions | Elseconditions |",
+    )
+
```
ialarmedalien, line 429: Since there's new functionality added to the `docgen.py` file, I'd like to see tests for that functionality in addition to these docs rendering tests.

linkml/generators/docgen.py
sujaypatil96: Fixed! :-)

linkml/generators/docgen.py
sujaypatil96: Fixed! :-)

linkml/generators/docgen.py
sujaypatil96: I moved the method outside the the parent function, and made it an "internal" method at the class level – that works too? see this [commit](https://github.com/linkml/linkml/pull/2640/commits/0104ad2ceb59096be4b993d4c8c43dc0ac3a6523).

tests/test_generators/test_docgen.py
sujaypatil96: Addressed in https://github.com/linkml/linkml/pull/2640/commits/f161b6fa4db30ce9f3878014b2938250ef53018f

linkml/generators/docgen.py
```diff
@@ -959,6 +960,76 @@ def example_object_blobs(self, class_name: str) -> list[tuple[str, str]]:
...
+        for key in ["preconditions", "postconditions", "elseconditions"]:
+            condition_obj = getattr(rule, key, None)
+            if condition_obj:
+                json_obj = json_dumper.to_dict(condition_obj)
+                sanitzed_condition = remove_name_keys(json_obj)
+                classrule_as_dict[key] = sanitzed_condition
+
+        return classrule_as_dict
+
+    def process_class_rules(self, element: ClassDefinition) -> list[dict[str, Any]]:
```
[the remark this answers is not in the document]
sujaypatil96: Addressed in https://github.com/linkml/linkml/pull/2640/commits/f161b6fa4db30ce9f3878014b2938250ef53018f

ialarmedalien:
Please add unit tests for the function `classrule_to_dict_view` to the docgen tests. I know that you have tests for the rendering, but I would like to see tests for the function itself, including cases where there are no rules to render.

You can just reuse the schema and values for the rendering tests that you've already written.

On a related topic, I don't think it's a good idea to have tests organised by GitHub issue, because unless you have intimate knowledge of the linkml issue tracker, `_issue_1731` is not helpful in working out whether the bit of code that you're looking at has tests or not. If you are going to separate out these tests into their own file, they should at least mention the name of the package that they are associated with -- e.g. `test_docgen_issue_1731` -- or better still, contain information about what is actually being tested, e.g. `test_docgen_rule_rendering.py`.

sujaypatil96:
> Please add unit tests for the function classrule_to_dict_view to the docgen tests. I know that you have tests for the rendering, but I would like to see tests for the function itself, including cases where there are no rules to render.
Oh those are in there as well on these lines? https://github.com/linkml/linkml/blob/8160f27ad10f591908af51ab7ada3bbcf4aae14d/tests/test_issues/test_linkml_issue_1731.py#L68-L95

Let me know if you had other ideas for unit tests for `classrule_to_dict_view()` and I can add them?

> On a related topic, I don't think it's a good idea to have tests organised by GitHub issue, because unless you have intimate knowledge of the linkml issue tracker, _issue_1731 is not helpful in working out whether the bit of code that you're looking at has tests or not.

That's a good suggestion. It's been a while since we've revised our contributing guidelines for writing tests: https://linkml.io/linkml/contributing/contributing.html#linkml-testing-framework. Happy to add this as a discussion item for our dev call and talk about it.

ialarmedalien:
I only saw one test function and thought that it was all rendering tests -- duh! Sorry about that.

Can you split up the function into several smaller functions? It is much easier to troubleshoot test failures than if you have one massive function. At minimum, split out the doc rendering tests from the `classrule_to_dict_view` tests.

I would also consider using `@pytest.mark.parametrize` to test `classrule_to_dict_view` with different values.

sujaypatil96:
I've split the tests up into two "sets" – `test_class_rules_section_rendering()` and `test_classrule_to_dict_view_method()` in [b247f4a](https://github.com/linkml/linkml/pull/2640/commits/b247f4a75e480098c769ea358061ae79ea01283e)

ialarmedalien:
@sujaypatil96 from the linkml contributor docs you linked to:

> If you want to formalize your test case into a unit test, you should decide the scope of your test and find the subfolder that the test might best fit under. For example, if your test is related to one of the LinkML generators, then you would create it under the tests/test_generators/ subfolder.

> Outside of generators, users will want to create issue specific unit tests. For that, you need to turn your attention to the tests/test_issues/ folder.

which I would interpret as saying that generator tests should go in the `test_generators` dir. It would make the most sense to put your tests in the `gendoc` tests, rather than having it in a file on its own.

sujaypatil96:
> which I would interpret as saying that generator tests should go in the test_generators dir.
> 
> It would make the most sense to put your tests in with the other documentation generator tests, rather than having it in a file on its own.

I fully agree that the tests we've written in this PR should live in the https://github.com/linkml/linkml/blob/main/tests/test_generators/test_docgen.py file. To accomplish that we normally extend the [kitchen sink model](https://github.com/linkml/linkml/blob/main/tests/test_generators/input/kitchen_sink.yaml) and then put accompanying tests in respective `tests/test_generators/` files, which is what I tried to do but had to revert in [cc29456](https://github.com/linkml/linkml/pull/2640/commits/cc29456ae2a4ec0c3fccd352d325310cdfa5e5ee) because the kitchen sink model is brittle and it affects a lot of test files, so we don't encourage editing that file anymore. Which is why I made a test file in the `tests/test_issues/` folder and the pattern that we've been following for that is:

> Outside of generators, users will want to create issue specific unit tests. For that, you need to turn your attention to the tests/test_issues/ folder.
> 
> The current direction is to first make an issue with specific instructions to reproduce the bug under the [issues](https://github.com/linkml/linkml/issues) tab on the main [linkml](https://github.com/linkml/linkml/) GitHub repo.
> 
> Then, create a test case for the above issue, with a minimal test LinkML schema in the input folder of the tests/test_issues/ suite. A minimal test case entails that you only use the elements you need in the schema for your unit test and not bloat the test schema with more elements that necessary.

sujaypatil96:
@ialarmedalien: i've moved the docgen tests into `tests/test_generators/test_docgen.py`. I think this PR is ready for a review/approval?

ialarmedalien (approved):

Issue #2084, opened 2024-04-29
gen-SHACL "target class" vs "Shape name" 


HendrikBorgelt:
**Describe the bug**
When trying to create a SHACL shape with the LinkML generator the LinkML template currently does not differentiate between the target class and the SHACL shape name. However, SHACL neither requires nor intends to always have a SHACL shape which also is its own shape. Therefore a "target class" should be added similar to the "class_uri" attribute ( sorry if this is not the right LinkML nomenclature...) .

**To reproduce**
1. Go to the command line
2. Click on use the "gen-Shacl" converter to generate a SHACL template from a LinkML Yaml file as seen below:

`LinkML_issue_HB.yaml`

```
id: https://linkML.com/linkml/tests/DCATap
name: DCATap_LinkML_Template
prefixes:
  linkml: https://w3id.org/linkml/
  dcat: http://www.w3.org/ns/dcat#


imports:
  - linkml:types
default_range: string  

  
classes:
    dcat_dataset:
        class_uri: dcat:Dataset
```
to
`LinkML_Issue_HB.shacl.ttl`
```
# metamodel_version: 1.7.0
@prefix dcat: <http://www.w3.org/ns/dcat#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix sh: <http://www.w3.org/ns/shacl#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

dcat:Dataset a sh:NodeShape ;
    sh:closed true ;
    sh:ignoredProperties ( rdf:type ) ;
    sh:targetClass dcat:Dataset .
```

4. Scroll down to '` dcat:Dataset a sh:NodeShape ;` ... `sh:targetClass dcat:Dataset` '
5. See the error that a NodeShap is an OWL class at the same time

**Expected behavior**
While it is acceptable to create classes in a SHACL file (a sh: class a.s.o.), a shape typically is not equivalent to its target class. A SHACL shape most often validates the axioms of a target class but is not required to fully depict all axioms of a target class. Accordingly, a SHACL shape can check for only a part of the axioms of a target class, for example, to ensure computational efficiency or extensibility.
LinkML should therefore either have an additional attribute similar to the `class_uir:` which then might be called `target_uri` or should use the LinkML class name as the shape name. Please look at the examples below:

1st recommendation (would be more feasible since the Shape could carry a prefix):
```
classes:
    dcat_dataset:
        target_uri: dcat:Dataset
[...]

PR #2647, opened 2025-04-30
updating shaclgen.py to accept custom shape_uris via annotations and use linkml native names to fix issue #2084

HendrikBorgelt:
This PR is intended as a fix for issue https://github.com/linkml/linkml/issues/2084, in that it allows to have freely defined shacl_shape_uri's cia annotations['shacl_uri'] and by switching from the previously implemented shacl shape naming schema with class_uri+suffix, to reusing the native linkML class name as the name for the SHACL shape. This should be more aligned to the ideas behind SHACL, where a shape is its own rdf instance and thus the linkML class with its name should directly represent the shape, while the class uri is used for the traget class.

Since i am neither an expert in good pull request nor have i ever used poetry before, feel free to leave a comment which may help me with the next PR. I have tried to use poetry to the best of my knowledge (e.g. poetry tox and poetry run pytest).

### DIFF
--- a/linkml/generators/shaclgen.py
+++ b/linkml/generators/shaclgen.py
@@ -71,10 +71,22 @@ class ShaclGenerator(Generator):
                 if v is not None:
                     g.add((class_uri_with_suffix, p, v))
 
+            # first, get the base class URI
             class_uri = URIRef(sv.get_uri(c, expand=True))
-            class_uri_with_suffix = class_uri
-            if self.suffix is not None:
-                class_uri_with_suffix += self.suffix
+
+            # check for an annotation
+            shape_uri = None
+            if c.annotations and c.annotations['shape_uri'].value :
+                shape_uri = c.annotations['shape_uri'].value
+
+            if shape_uri:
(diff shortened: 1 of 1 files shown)


dalito:
> Since i am neither an expert in good pull request nor have i ever used poetry before, feel free to leave a comment which may help me with the next PR. I have tried to use poetry to the best of my knowledge (e.g. poetry tox and poetry run pytest).

It helps to always create a new branch for any pull request that you want to make. Pick a name that includes the issue number and a title, for example "issue-2084-coustom-shape-uris". 

Now that you already committed to your main, and the main branch in your fork will be different from the main branch here soon (after the next PR is accepted here). You may then not be able to simply sync your fork´s main (via GitHub UI) with main here without resolving conflicts. It is no big problem but it is more difficult to understand the state and requires more git knowledge/commands to continue from there.

Did you see the test errors that the CI shows also in your local tests? In general it is best practice to add a test to the PR that demonstrates the issue which then passes with you suggested changes.

PR #2675, opened 2025-05-10
remove context setting in mermaid _diagram_ generator

sierra-moxon:
Remove customize_environment import from mermaid class diagram generator to fix failing import on main.

Fix doc generation with additions of a new ontologies*.md file. 
Fix codespell errors.

This is a jumble of fixes to get main branch to pass tests after a flurry of merges.


.github/workflows/test-windows-fix.yaml
```diff
@@ -0,0 +1,71 @@
+# Built from:
+# https://docs.github.com/en/actions/guides/building-and-testing-python
+# https://github.com/snok/install-poetry#workflows-and-tips
+
+name: Build and test linkml
+
+on:
+  push:
+    branches:
+      - fix_docgen
```
ialarmedalien, line 10: is this branch sticking around?

sierra-moxon: No, @ialarmedalien, it will not.  The changes to docgen to add the truncate flag, broke mermaidclassdiagramgen and the tests in main, and I am using this to fail fast as I try to fix spacing, etc. in the windows environment (as replicating windows OS locally for me is not great).

sierra-moxon:
🎉  - thanks @ialarmedalien!!

tests/test_generators/test_docgen.py
```diff
@@ -526,21 +554,88 @@ def test_docgen_multiline_everything(
...
-                # all slots are type string
-                + " 0..1 <br/> [xsd:string](http://www.w3.org/2001/XMLSchema#string) "
-                + f"| {description} | direct |"
-            ) in class_file_contents
+        class_file_list = [line for line in class_file_contents.split("\n") if ".md" in line]
+        print(class_file_list)
+
+        slot_type = "[xsd:string](http://www.w3.org/2001/XMLSchema#string)"
+        if WINDOWS:
+            slot_type = "[xsd:string](xsd:string)"
```
dalito: @ialarmedalien - Do you know why the CURIE is not replaced by the URI here? I have seen the same when I saw you struggling with Windows but could not yet understand why it happens.

My comment should not hold this PR. Fixing the issues that break the tests on main is too important.

ialarmedalien: There is something going on with the link generation function (`link(...)` in the DocGenerator package) -- I don't have a windows box to test on so I didn't want to delay getting things fixed by investigating further. I'm wondering if it's a path issue around accessing the local metamodel.

pyproject.toml
```diff
@@ -64,6 +64,7 @@ dependencies = [  # Specifier syntax: https://peps.python.org/pep-0631/
     "sqlalchemy >= 1.4.31",
     "watchdog >= 0.9.0",
     "typing-extensions >= 4.6.0; python_version < '3.12'",
+    "sphinx-click (>=6.0.0,<7.0.0)",
```
dalito, line 67: Version 6.0.0 is currently the latest version of sphinx-click. So there is no known problem which would justify adding an [upper version constraint](https://iscinumpy.dev/post/bound-version-constraints/).

dalito (approved):

PR #2678, opened 2025-05-12
tests: validate context using rdf

Silvanoc:
Add tests to validate generated JSON-LD Context by confirming that the resulting RDF graph fulfills the expectations.


Silvanoc:
This PR improves test coverage for some issues that I've been observing within the scope of PR #2602. Since @tfliss is also involved in that PR and seem to have expertise in JSON-LD and RDF, his review is probably meaningful.

tests/utils/validate_jsonld_context.py
```diff
@@ -0,0 +1,117 @@
...
+            print(json.dumps(schema_jsonld, indent=2))
+            print("ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️")
+            print("🐢🐢🐢🐢🐢🐢🐢🐢🐢🐢🐢🐢🐢🐢")
+            print("🐢🐢 RDF Graph (Turtle) 🐢🐢")
+            print(self._g.serialize(format="ttl"))
+            print("🐢🐢🐢🐢🐢🐢🐢🐢🐢🐢🐢🐢🐢🐢")
+
+    def expected_id_uri(self):
+
+        if "name" in self._schema_yaml.keys():
```
ialarmedalien: no need to use `keys()`

tests/utils/validate_jsonld_context.py
```diff
@@ -0,0 +1,117 @@
...
+                schema_slots.append(slot_uri)
+            print(f"{self._schema_uri} should have slots {schema_slots}")
+            for _, slot_name in self._g.subject_objects(predicate=term.URIRef("https://w3id.org/linkml/slots")):
+                assert str(slot_name) in schema_slots
+
+    def expected_types(self):
+        pass
+
+    def expected_enums(self):
+        pass
```
ialarmedalien: What is the purpose of these stubs?

tests/utils/validate_jsonld_context.py
ialarmedalien: (this applies throughout the PR)

tests/utils/validate_jsonld_context.py
```diff
@@ -0,0 +1,117 @@
...
+            else:
+                self._schema_uri = self._schema_yaml["id"] + "/" + self._schema_yaml["name"]
+        else:
+            self._schema_uri = self._schema_yaml["id"]
+        for subject, predicate in self._g.subject_predicates(
+            object=term.URIRef("https://w3id.org/linkml/SchemaDefinition")
+        ):
+            assert str(subject) == self._schema_uri
+            assert str(predicate) == "http://www.w3.org/1999/02/22-rdf-syntax-ns#type"
+        print(f"URI SchemaDefinition: {self._schema_uri}")
```
ialarmedalien: You are already testing `self._schema_uri` with an assertion -- no need to clutter up the test output with a `print` statement.

tests/utils/validate_jsonld_context.py
```diff
@@ -0,0 +1,117 @@
...
+        return self._schema_uri
+
+    def expected_namespaces(self):
+        if "prefixes" in self._schema_yaml.keys():
+            namespaces = []
+            for _, namespace in self._g.subject_objects(predicate=term.URIRef("http://www.w3.org/ns/shacl#namespace")):
+                assert str(namespace) in self._schema_yaml["prefixes"].values()
+                namespaces.append(str(namespace))
+            for namespace in self._schema_yaml["prefixes"].values():
+                assert namespace in namespaces
```
ialarmedalien: How about using sets to simplify these tests?

tests/utils/validate_jsonld_context.py
```diff
@@ -0,0 +1,117 @@
...
+                    if ":" in cls["class_uri"]:
+                        prefix = cls["class_uri"].split(":")[0]
+                        class_uri = self._schema_yaml["prefixes"][prefix] + cls["class_uri"].split(":")[1]
+                    elif "default_prefix" in self._schema_yaml.keys():
+                        class_uri = (
+                            self._schema_yaml["prefixes"][self._schema_yaml["default_prefix"]]
+                            + cls["class_uri"].split(":")[1]
+                        )
+                    else:
+                        raise Exception("something is wrong")
```
ialarmedalien: not a very helpful message...

tests/utils/validate_jsonld_context.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def expected_namespaces(self):
+        if "prefixes" in self._schema_yaml.keys():
+            namespaces = []
+            for _, namespace in self._g.subject_objects(predicate=term.URIRef("http://www.w3.org/ns/shacl#namespace")):
+                assert str(namespace) in self._schema_yaml["prefixes"].values()
+                namespaces.append(str(namespace))
+            for namespace in self._schema_yaml["prefixes"].values():
+                assert namespace in namespaces
+            print(f"Namespaces: {namespaces}")
```
ialarmedalien: avoid printing if you've already tested something

tests/utils/validate_jsonld_context.py
```diff
@@ -0,0 +1,117 @@
...
+                schema_classes.append(class_uri)
+            print(f"'{self._schema_uri}' should have classes {schema_classes}")
+            for _, class_name in self._g.subject_objects(predicate=term.URIRef("https://w3id.org/linkml/classes")):
+                print(f"Check '{class_name}' is one of the expected classes")
+                assert str(class_name) in schema_classes
+
+    def expected_slots(self):
+        if "slots" in self._schema_yaml.keys():
+            schema_slots = []
+            for slot_name, slot in self._schema_yaml["slots"].items():
```
ialarmedalien, line 91: is `slot_name` used?

tests/utils/validate_jsonld_context.py
```diff
@@ -0,0 +1,117 @@
...
+                            self._schema_yaml["prefixes"][self._schema_yaml["default_prefix"]]
+                            + slot["slot_uri"].split(":")[1]
+                        )
+                    else:
+                        raise Exception("something is wrong")
+                else:
+                    if "default_prefix" in self._schema_yaml.keys():
+                        slot_uri = self._schema_yaml["prefixes"][self._schema_yaml["default_prefix"]] + slot_name
+                    else:
+                        slot_uri = self._schema_yaml["id"] + slot_name
```
ialarmedalien: This is very similar to what you do with the `class_uri`. Can it be abstracted out into a more generic function?

tests/utils/validate_jsonld_context.py
Silvanoc: Draft code that should've never gotten into the PR. Will remove them.

tests/utils/validate_jsonld_context.py
Silvanoc: Just acquired habit on older Python days. It's only a bit more verbose, no other harm. But I have no horse in this race, I can remove it. Adding suggestions that I can simply accept on all the places would have done my life easier 😉

tests/utils/validate_jsonld_context.py
Silvanoc: Agree, slipped through before making the PR ready for review. Will fix it.

tests/utils/validate_jsonld_context.py
Silvanoc: Good idea. Will fix it.

tests/utils/validate_jsonld_context.py
Silvanoc: The information being shown depends on the parametrization of the test, but is "hidden". This output is only gonna be shown if something fails or you have high verbosity. In both cases this information is useful for diagnosing. In fact I've added it, because it has proofed to be useful to diagnose an issue I've faced.

tests/utils/validate_jsonld_context.py
Silvanoc: Same as above.

The information being shown depends on the parametrization of the test, but is "hidden". This output is only gonna be shown if something fails or you have high verbosity. In both cases this information is useful for diagnosing. In fact I've added it, because it has proofed to be useful to diagnose an issue I've faced.

tests/utils/validate_jsonld_context.py
Silvanoc: I just wanted to have the tests, didn't have the time for that optimization. I would accept additional commits doing that consolidation or on a follow-up PR.

tests/utils/validate_jsonld_context.py
Silvanoc: Wrong. They are still to-be-done tests. I'll add a `TODO` comment.

Validating only classes and slots as in this PR is already valuable. Validating also types and enums is even better, but remains open for future PRs.

tests/utils/validate_jsonld_context.py
Silvanoc: Done

tests/utils/validate_jsonld_context.py
Silvanoc: `TODO` comment added. @ialarmedalien please resolve the conversation if you agree.

tests/utils/validate_jsonld_context.py
Silvanoc: Done

tests/utils/validate_jsonld_context.py
Silvanoc: Done. @ialarmedalien please resolve the conversation if you agree.

tests/utils/validate_jsonld_context.py
ialarmedalien: > Adding suggestions that I can simply accept on all the places would have done my life easier 😉

Not sure if I don't have privs to add suggestions or you didn't enable them on this PR, but either way, I'm unable to do so. 😢

Silvanoc: No worries, if you don't have developer rights in the repository, you probably can't. I've fixed it myself anyway.

tests/utils/validate_jsonld_context.py
ialarmedalien: The pytest error message if the assertion fails should be enough for diagnosing errors, or at least for knowing where to set breakpoints for the debugger.

tests/utils/validate_jsonld_context.py
ialarmedalien: follow up PR / not bothering is fine with me 🙂

tests/utils/validate_jsonld_context.py
```diff
@@ -0,0 +1,124 @@
...
+    _schema_uri: str = ""
+
+    def __init__(self, schema_yaml, schema_jsonld):
+        self._schema_yaml = schema_yaml
+        self._g = Graph().parse(data=schema_jsonld, format="json-ld")
+        if os.environ.get("DETAILED_OUTPUT"):
+            print("ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️")
+            print("ℹ️ℹ️      JSON-LD       ℹ️ℹ️")
+            print(json.dumps(schema_jsonld, indent=2))
+            print("ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️ℹ️")
```
ialarmedalien: I feel like this qualifies as emoji-onal abuse...

tests/utils/validate_jsonld_context.py
```diff
@@ -0,0 +1,124 @@
...
+                    elif "default_prefix" in self._schema_yaml:
+                        class_uri = (
+                            self._schema_yaml["prefixes"][self._schema_yaml["default_prefix"]]
+                            + cls["class_uri"].split(":")[1]
+                        )
+                    else:
+                        raise Exception(
+                            "`class_uri` is neither a URI, nor a CURIE and "
+                            + "no `default_prefix` is provided to create a CURIE"
+                        )
```
ialarmedalien: this looks great now

tests/utils/validate_jsonld_context.py
Silvanoc: I prefer printouts over debuggers, but I have no horse in this race either so I'll simply remove it.

tests/utils/validate_jsonld_context.py
Silvanoc: Same as above it's not worth a discussion. Simply removing it.

tests/utils/validate_jsonld_context.py
```diff
@@ -0,0 +1,124 @@
...
+        print(f"URI SchemaDefinition: {self._schema_uri}")
+        return self._schema_uri
+
+    def expected_namespaces(self):
+        if "prefixes" in self._schema_yaml:
+            namespaces = []
+            for _, namespace in self._g.subject_objects(predicate=term.URIRef("http://www.w3.org/ns/shacl#namespace")):
+                assert str(namespace) in self._schema_yaml["prefixes"].values()
+                namespaces.append(str(namespace))
+            assert set(self._schema_yaml["prefixes"].values()).issubset(namespaces)
```
ialarmedalien: if the values of `self._schema_yaml["prefixes"]` is always going to be the same as the `self._g.subject_objects(predicate=...)`, you can just do:

```py
rdf_namespaces = {str(ns) for _, ns in self._g.subject_objects(predicate=term.URIRef("http://www.w3.org/ns/shacl#namespace"))}
assert rdf_namespaces == set(self._schema_yaml["prefixes"].values())
```

(could be a single statement but it starts getting hard to read)

tests/utils/validate_jsonld_context.py
Silvanoc: Done

tests/utils/validate_jsonld_context.py
Silvanoc: Done

tests/utils/validate_jsonld_context.py
Silvanoc: Of course, it qualifies as such! 😆 

But the JSON-LD and RDF dumps can be so huge, that something like this that jumps to the eye is very useful. Would you prefer ASCII characters like `*`, `#` or so?

tests/utils/validate_jsonld_context.py
Silvanoc: Feel free to optimize it on a posterior PR 🙂

ialarmedalien: those two lines are the entire thing -- just replace lines 52 to 56 with the two lines in the comment

tests/test_scripts/test_gen_jsonld_context.py
```diff
@@ -59,3 +63,35 @@ def test_slot_class_uri(input_path, snapshot):
...
+        print(f"\n‼️‼️‼️ Processing '{schema_path}' ‼️‼️‼️")
+
+    runner = CliRunner()
+    result = runner.invoke(cli, [str(schema_path)])
+    assert result.exit_code == 0
+
+    with open(schema_path) as schema_file:
+        schema_yaml = safe_load(schema_file)
+
+    rdf_expects = RdfExpectations(schema_yaml, result.output)
```
ialarmedalien: To save on code duplication, alter RdfExpectations to take a file path as input, and move the yaml reading into RdfExpectations.

i.e. in this file and the other test file:
```py
def test_expected_rdf(input_path, schema):
     schema_path = input_path(schema)
     runner = CliRunner()
     result = runner.invoke(cli, [str(schema_path)])
     assert result.exit_code == 0

     rdf_expects = RdfExpectations(schema_path, result.output)
     rdf_expects.check_expectations()  # catchall method that can be used to check the expected_id_uri, namespaces, classes, slots, etc.
```

in RdfExpectations:

```py
class RdfExpectations:
  _schema_yaml: dict = {}
  _g: Graph = Graph()
  _schema_uri: str = ""
  verbose: bool = False

  def __init__(self, schema_path, schema_jsonld):
    # set verbosity according to the environment
    if os.environ.get("DETAILED_OUTPUT"):
      self.verbose = True
    self._schema_yaml = self.load_yaml_file(schema_path)
    self._g = Graph().parse(data=schema_jsonld, format="json-ld")
    if self.verbose:
      print(...)

  def load_yaml_file(self, schema_path):
    if self.verbose:
      print(f"\n‼️‼️‼️ Processing '{schema_path}' ‼️‼️‼️")
    with open(schema_path) as schema_file:
       return safe_load(schema_file)

  def check_expectations(self):
    self.expected_id_uri()
    self.expected_namespaces()
    # etc.

```
Now all your debugging code (and attendant emojis 😉 ) is in one place and you've reduced code duplication.

An even cleaner solution would be to set up RdfExpectations with a logger, which gives you more control over what gets logged when, with the `DETAILED_OUTPUT` env var dictating whether or not the logger generates any output.

tests/utils/validate_jsonld_context.py
ialarmedalien: I think using emojis is fine -- liven up that output a bit! 🎉 

However, you should be able to narrow it down so that you're only running a single test, and if the test is parametrised, the specific param that you want, either through your IDE or via the command line args you're supplying pytest:

https://docs.pytest.org/en/stable/how-to/usage.html

Assuming that the jsonld-generator doesn't create a load of extra output when it converts a file, there shouldn't be too much of an issue with finding the output of a specific test.

tests/utils/validate_jsonld_context.py
```diff
@@ -0,0 +1,122 @@
...
+                        )
+                else:
+                    if "default_prefix" in self._schema_yaml:
+                        class_uri = self._schema_yaml["prefixes"][self._schema_yaml["default_prefix"]] + class_name
+                    else:
+                        class_uri = self._schema_yaml["id"] + "/" + class_name
+                schema_classes.append(class_uri)
+            print(f"'{self._schema_uri}' should have classes {schema_classes}")
+            for _, class_name in self._g.subject_objects(predicate=term.URIRef("https://w3id.org/linkml/classes")):
+                print(f"Check '{class_name}' is one of the expected classes")
```
ialarmedalien: rm debugging statements (see previous justification for why)

If `schema_classes` should be the same as what's in `self._g.subject_objects(predicate=term.URIRef("https://w3id.org/linkml/classes"))`, I'd use a set for this.

tests/utils/validate_jsonld_context.py
```diff
@@ -0,0 +1,122 @@
...
+                            "`slot_uri` is neither a URI, nor a CURIE and "
+                            + "no `default_prefix` is provided to create a CURIE"
+                        )
+                else:
+                    if "default_prefix" in self._schema_yaml:
+                        slot_uri = self._schema_yaml["prefixes"][self._schema_yaml["default_prefix"]] + slot_name
+                    else:
+                        slot_uri = self._schema_yaml["id"] + slot_name
+                schema_slots.append(slot_uri)
+            print(f"{self._schema_uri} should have slots {schema_slots}")
```
ialarmedalien: rm debugging comments

Silvanoc: Not really debug comments, but better verbosity for failing tests. Anyway, not worth the discussion. Removing.

Silvanoc:
@ialarmedalien I won't be able to work in the project for almost 3 weeks, open points are only optimizations IMO. If it gets merged as it is, fine. Otherwise I'll work on it when I'm back.

IMO this test improves coverage considerably, therefore IMO it should be considered merging it as it is and improving it afterwards.

ialarmedalien:
@Silvanoc I can add those couple of optimisations as a commit to this PR if you like. I think it's looking pretty awesome apart from that -- anything that improves test coverage and accuracy of tests is always going to be a win.

You may still want to get someone who is more familiar with the details of JSON-LD to check the content as I'm mostly commenting on coding.

Silvanoc:
> @Silvanoc I can add those couple of optimisations as a commit to this PR if you like. I think it's looking pretty awesome apart from that -- anything that improves test coverage and accuracy of tests is always going to be a win.

@ialarmedalien I don't know in which time-zone you are and if you are still working today. If so and you can provide a PR today, I'll have a look at it before "disappearing" for the mentioned couple of weeks.

> You may still want to get someone who is more familiar with the details of JSON-LD to check the content as I'm mostly commenting on coding.

@tfliss could you have a look at the tests and even give them a try?

Silvanoc:
@ialarmedalien thanks for your patch! From my point of view ready to merge.

tests/utils/validate_jsonld_context.py
```diff
@@ -0,0 +1,153 @@
...
+        self._generate_rdf_graph(schema_jsonld)
+
+    def _generate_rdf_graph(self, schema_jsonld) -> None:
+        """Generate an RDF Graph object from a json string.
+
+        :param schema_jsonld: JSON-LD schema
+        :type schema_jsonld: _type_
+        """
+        self._g = Graph().parse(data=schema_jsonld, format="json-ld")
+        if self.verbose:
```
tfliss: I'd prefer if this was handled with a logger and logger.info.
see for example testsqlalchemygen.py

For example I invoke tests like so (-s and --log-cli-level will have a similar effect)
`pytest -s -vv --log-cli-level=INFO tests/test_generators/test_jsonldcontextgen.py`

Since you're heading out for several weeks I'd prefer to see the PR committed rather than not.

tests/utils/validate_jsonld_context.py
```diff
@@ -0,0 +1,153 @@
...
+                    if "default_prefix" in self._schema_yaml:
+                        slot_uri = self._schema_yaml["prefixes"][self._schema_yaml["default_prefix"]] + slot_name
+                    else:
+                        slot_uri = self._schema_yaml["id"] + slot_name
+                schema_slots.add(slot_uri)
+            assert schema_slots == {
+                str(slot_name) for _, slot_name in self._g.subject_objects(predicate=term.URIRef(f"{LINKML_URI}/slots"))
+            }
+
+    def expected_types(self):
```
tfliss: I don't see these used anywhere else. Would prefer them removed or make a note in the follow-up PR since x-fail changes will touch this file.

tests/utils/validate_jsonld_context.py
```diff
@@ -0,0 +1,153 @@
...
+        :param schema_path: path to the YAML file
+        :type schema_path: _type_
+        """
+        if self.verbose:
+            print(f"\n‼️‼️‼️ Processing '{schema_path}' ‼️‼️‼️")
+
+        with open(schema_path) as schema_file:
+            self._schema_yaml = safe_load(schema_file)
+
+    def check_expectations(self) -> None:
```
tfliss, line 55: This method could probably be made more readable in the future, but it seems like a big improvement. Appreciate making it independent of other code. I'll think on some constructive recommendations for a future PR.

tests/utils/validate_jsonld_context.py
ialarmedalien: I'm happy to remove this as I don't like having "TODO" methods.

tests/utils/validate_jsonld_context.py
ialarmedalien: I'll take a look and see if I can implement this. I also prefer using a logger to printing stuff out.

PR #2686, opened 2025-05-13
Use enum value rather than enum for click meta click option in pydanticgen

kevinschaper:
With click 8.2.0, `None` stops working and `NONE` is required. I was going to go down a rabbit hole of supporting both, but I see that other enumerations are just converted to their values when being given to click.


sneakers-the-rat (approved):
Great, yes. I am somewhat frustrated in general with how python does enums, and how enum members aren't in general treated as equivalent to their values, so good looking ahead and thx for the fix.

sneakers-the-rat:
I wonder why codespell and myst have decided to break in this commit, which doesn't touch them. Is the base branch out of date or something? Or did those tools change?

Issue #2687, opened 2025-05-14
Make `class_uri` be skos:exactMatch of the element URI on RDF


Silvanoc:
**Describe the bug**

As of now generated RDF in the case of a `class_uri` property results in a [`skos:mappingRelation`](https://www.w3.org/2009/08/skos-reference/skos.html#mappingRelation), which is a too fuzzy relationship. Compare it with the generated OWL, where a [`skos:exactMatch`](https://www.w3.org/2009/08/skos-reference/skos.html#exactMatch) relationship is set.

`skos:exactMatch` is a sub-property of [`skos:closeMatch`](https://www.w3.org/2009/08/skos-reference/skos.html#closeMatch), which is a sub-property of `skos:mappingRelation`.

**Version of LinkML you are using** v1.9.1

**Please provide a schema (and if applicable, a data file) that replicates the issue**

We can use one of the schemas being used for the tests: [test_generators/input/jsonld_context_inlined_external_types.yaml](https://github.com/linkml/linkml/blob/main/tests/test_generators/input/jsonld_context_inlined_external_types.yaml)

We then focus on this ClassDefinition:
```yaml
  SriString:
    class_uri: cred:sriString
```

The resulting JSON-LD is:
```json
    {
      "name": "SriString",
      "definition_uri": "https://example.org/external_type#SriString",
      "from_schema": "https://example.org/external_type",
      "mappings": [
        "cred:sriString"
      ],
      "slot_usage": {},
      "class_uri": "https://www.w3.org/2018/credentials#sriString",
      "@type": "ClassDefinition"
    },
```
Notice the `mappings` property.

and the resulting RDF is:
```turtle
<https://example.org/SriString> a linkml:ClassDefinition ;
    skos:inScheme <https://example.org/external_type> ;
    skos:mappingRelation <cred:sriString> ;
    linkml:class_uri <https://www.w3.org/2018/credentials#sriString> ;
    linkml:definition_uri <https://example.org/external_type#SriString> ;
    linkml:slot_usage [ ] .
```
Notice the `skos:mappingRelation` predicate.

Let's compare with the generated OWL:
```turtle
ex:SriString a owl:Class ;
    rdfs:label "SriString" ;
    skos:exactMatch cred:sriString ;
    skos:inScheme <https://example.org/external_type> .
```
where we have the **expected** `skos:exactMatch` predicate.

tfliss:
[...]

PR #2692, opened 2025-05-14
Add tests to ensure that `class_uri` becomes `exactMatch` in JSON-LD and RDF

Silvanoc:
Add tests for issue #2687 to:
1. demonstrate the bug and
2. avoid regressions in the future.

As of now with the `xfail` mark until the bug has been fixed.

**IMPORTANT NOTE**: this test is only addressing the mentioned bug #2687, without considering possible future changes in the current behavior. With this I mean that `class_uri` as of now DOES NOT become the class URI neither in JSON-LD nor in RDF.

### DIFF
--- a/tests/test_generators/test_jsonldgen.py
+++ b/tests/test_generators/test_jsonldgen.py
@@ -1,0 +1,95 @@
+import json
+
+import pytest
+from linkml_runtime.utils.schemaview import SchemaView
+from rdflib import Graph, term
+from yaml import safe_load
+
+from linkml.generators import JSONLDGenerator, RDFGenerator
+
+
+@pytest.mark.xfail(reason="Bug 2687: class_uri should be exactMatch")
+def test_class_uri(input_path):
+    """`class_uri` should result in an `skos:exactMatch` relationship
+    in the generated JSON-LD and RDF."""
+    # focusing on the class `Thing` of the schema jsonld_context_class_uri_prefix.yaml:
+    # [...]
+    #   Thing:
+    #     class_uri: rdfs:Resource
+    # [...]
+    schema_path = str(input_path("jsonld_context_class_uri_prefix.yaml"))
+
+    # get the classes declaring a `class_uri`
+    schema_view = SchemaView(schema_path)
+    with open(schema_path) as f:
+        schema_yaml = safe_load(f)
+    classes_with_custom_uri = {}
+    for class_name, cls in schema_yaml["classes"].items():
+        if "class_uri" in cls.keys():
+            # save class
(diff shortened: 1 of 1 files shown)


tests/test_generators/test_jsonldgen.py
```diff
@@ -0,0 +1,95 @@
...
+    #     "rdfs:Resource"
+    #   ],
+    #   "slot_usage": {},
+    #   "@type": "ClassDefinition"
+    # }
+    # [...]
+    schema_jsonld = json.loads(JSONLDGenerator(schema_path, format="jsonld").serialize())
+    # get all the schema classes according the generated JSON-LD
+    classes_jsonld = {cls["definition_uri"]: cls for cls in schema_jsonld["classes"]}
+    print(classes_jsonld)
```
tfliss: please add a global logger to the file and use logger.info() rather than print.

```
import logging
...
logger.info(classes_jsonld)
```

tests/test_generators/test_jsonldgen.py
```diff
@@ -0,0 +1,95 @@
...
+    #   "definition_uri": "http://uri.interlex.org/tgbugs/uris/readable/sparc/Thing",
+    #   "from_schema": "https://sparc.olympiangods.org/sparcur/schemas/1/sparc",
+    #   "exact_mappings": [
+    #     "rdfs:Resource"
+    #   ],
+    #   "slot_usage": {},
+    #   "@type": "ClassDefinition"
+    # }
+    # [...]
+    schema_jsonld = json.loads(JSONLDGenerator(schema_path, format="jsonld").serialize())
```
tfliss, line 51: This test is a little long. Consider breaking the setup up to schema_jsonld into a pytest fixture. I won't hold up the PR for it, though.

tfliss (approved):
Ok, approved, but please change the print to a logging call, or the continuous integration logs will be noisy!

Silvanoc:
> Ok, approved, but please change the print to a logging call, or the continuous integration logs will be noisy!

There's no noise if tests succeed (`pytest` won't show them). And when tests fail, they you might be happy to have that "noise". But anyway, you are not the first one disliking it, so I'll simply get rid of the print-outs.

tests/test_generators/test_jsonldgen.py
Silvanoc: Done

tests/test_generators/test_jsonldgen.py
Silvanoc: It's an optimization that can be applied afterwards. I won't fix it in this PR.

PR #2693, opened 2025-05-14
Fix 2687: Make `class_uri` be `skos:exactMatch` of the element URI on JSON-LD and RDF

### LINKED ISSUE #2687: Make `class_uri` be skos:exactMatch of the element URI on RDF

Silvanoc:
This patch fixes issue #2687 by adding the URI of `class_uri` as an `skos:exactMatch`.

This PR builds upon PR #2692, which was added to demonstrate the issue by adding a test that fails (`xfail`). This PR removes the `xfail` mark.

**Note for reviewers**: the last commit is updating the snapshots and is the best place to start with if you want to get an impression on what is changing (for better, I hope).

Closes #2687


melonora:
@Silvanoc Thank you for your PR! I had a look at the patch, and, as far as I understand it, this is how the behaviour should be. I will review the actual code and get back to you by the end of the week. Is it ok with you if I resolve the current conflicts with main or is this something you would like to do?

Silvanoc:
> @Silvanoc Thank you for your PR! I had a look at the patch, and, as far as I understand it, this is how the behaviour should be. I will review the actual code and get back to you by the end of the week. Is it ok with you if I resolve the current conflicts with main or is this something you would like to do?

@melonora I'm sorry I never answered your question, I oversaw it. I don't care if you take this PR or yours #3369. I would only like to understand the reason for deciding one or the other way. Thanks!

melonora:
@Silvanoc that is my bad. I was looking at the triage and overlooked the link to your PR and I had forgotten about it. I closed my PR as it is pretty much the same solution. Thank you for resolving the conflicts! I will just bring it up one last time in the core dev meeting, but I think your solution is correct so if others agree we can get your PR merged.

Silvanoc:
@melonora in case you decide to go for your PR, please take following into account:
1. You need to remove the `xfail` from the test, so that it really fails in case of error. See my commit e36113e203b3241642a0344973e5f0ddeecd85c5.
2. My `xfail` test had a small bug and was comparing CURIEs with expanded CURIEs. See my commit ee4ed8c22767605fa57050297dd6c8c8d0a499d0 for a fix.
3. You will need to update the snapshots according the changes.

melonora:
I believe it is more proper to go with your PR as you came up with the solution. I saw that some tests are now still failing, mainly because of some tests still expecting `class_uri`. Would you like to fix these? Otherwise with your permission I can also provide the required changes.

Silvanoc:
> Would you like to fix these? Otherwise with your permission I can also provide the required changes.

@melonora feel free to fix it, thanks!

melonora (approved):
Thanks again for the PR. I believe the change of `mappings` to `exact_mappings` to be correct. There are only some tests still failing for which the expected values are `class_uri` instead of `skos:exactmatch`. Once these are fixed I think the PR is ready to go in. Preapproving for now.

amc-corey-cox:
Rebased on main and regenerated snapshots so this can move forward.

- Rebased on `main` (was 30 commits behind).
- Conflicts were all in snapshot files, caused by f2c6acda0 ("Fix identifier/key slots not marked required in induced_slot") landing after this PR. Took your snapshots through the rebase, then re-ran with `--generate-snapshots` to fold in main's changes; net result was 6 snapshot files updated.
- The cancelled Slow Tests on the prior CI run hit the GitHub Actions 6h job timeout, not real failures — all regular `test (...)` jobs were green.
- Side note: `tests/linkml/test_issues/test_issue_206.py::test_class_list` now XPASSes after this rebase. Pre-existing `xfail`, unrelated to #2687, so I left it; worth a separate look.

Force-pushed to `fix-2687` per your earlier 👍 to @melonora.

amc-corey-cox:
Merged main again to absorb #3484 (linkml-model 1.11.0rc2). 16 snapshot conflicts, all in generated files; took main's side through the merge then re-ran with `--generate-snapshots` to fold the PR's `skos:exactMatch` changes on top. 9 snapshots updated.

PR #2304, opened 2024-09-06
test(gen-graphql): extend tests for types

Silvanoc:
Add tests for the generation of GraphQL scalar types.

Closes #2302

PR #2707, opened 2025-05-17
Add GraphQL validity tests

### LINKED ISSUE #2302: gen-graphql not generating equivalent scalar for types
**Describe the bug**
Generation of the GraphQL types out of a schema is not resolving types into the corresponding [GraphQL scalar types](https://graphql.org/learn/schema/#scalar-types).

I would expect [LinkML types](https://linkml.io/linkml-model/latest/docs/#types) to be converted to the corresponding GraphQL scalar types, but no conversion happens. Only those types that (by chance) have exactly the same name (like `String`) will work, for all others GraphQL tools will raise an error because they are unknown.

Using the `repr` slot (`base` if `repr` is missing) of a LinkML type it should be easy to generate the corresponding GraphQL scalar type. Only `Int` would need to be renamed to `Integer` and an explicit conversion of `Decimal` into GraphQL `Float` would be needed.

**Version of LinkML you are using** 1.8.2

The conversion of following schema into GraphQL
``` yaml
id: https://w3id.org/linkml/examples/alltypes
name: alltypes
prefixes:
  alltypes: https://w3id.org/linkml/examples/alltypes
imports:
  - linkml:types
default_range: string
default_prefix: alltypes

classes:
  AllTypes:
    attributes:
      string:
        range:
          string
      integer:
        range:
          integer
      boolean:
        range:
          boolean
      float:
        range:
          float
      double:
        range:
          double
      decimal:
        range:
          decimal
      time:
        range:
          time
      date:
        range:
          date
      datetime:
        range:
          datetime
      dateOrDatetime:
        range:
          date_or_datetime
      uriorcurie:
        range:
          uriorcurie
      curie:
        range:
          curie
      uri:
        range:
          uri
      ncname:
        range:
          ncname
      objectidentifier:
        range:
          objectidentifier
      nodeidentifier:
        range:
          nodeidentifier
      jsonpointer:
        range:
          jsonpointer
      jsonpath:
        range:
          jsonpath
      sparqlpath:
        range:
          sparqlpath
```

Results in following GraphQL schema

``` graphql
# metamodel_version: 1.7.0
type AllTypes
  {
    string: String
    integer: Integer
    boolean: Boolean
    float: Float
    double: Double
    decimal: Decimal
    time: Time
    date: Date
    datetime: Datetime
    dateordatetime: Dateordatetime
    uriorcurie: Uriorcurie
    curie: Curie
    uri: Uri
    ncname: Ncname
    objectidentifier: Objectidentifier
    nodeidentifier: Nodeidentifier
    jsonpointer: Jsonpointer
    jsonpath: Jsonpath
    sparqlpath: Sparqlpath
  }
```

Running a validation with the package `gql` like this `python3 -c 'from gql import Client ; Client(schema=open("<path to GraphQL schema>").read())'` yields following errors:
```
Unknown type 'Integer'.
Unknown type 'Double'.
Unknown type 'Decimal'.
Unknown type 'Time'.
Unknown type 'Date'.
Unknown type 'Datetime'.
Unkwnon type 'Dateordatetime'.
Unknown type 'Uriorcurie'.
Unknown type 'Curie'.
Unknown type 'Uri'.
Unknown type 'Ncname'.
Unknown type 'Objectidentifier'.
Unknown type 'Nodeidentifier'.
Unknown type 'Jsonpointer'.
Unknown type 'Jsonpath'.
Unknown type 'Sparqlpath'.
```

Notice that `String` and `Float` are not failing, because they have the same name in both. But the LinkML type `Int` is called `Integer` in GraphQL and therefore it is failing.

Silvanoc:
Add tests for the generation of GraphQL scalar types.

It also adds validation of the GraphQL code using [GQL](https://gql.readthedocs.io/en/stable/).

Please review the proposed [type](https://linkml.io/linkml-model/latest/docs/#types) equivalence based on [GraphQL native scalar types](https://graphql.org/learn/schema/#scalar-types):
- To GraphQL `Boolean`: `Boolean`
- To GraphQL `Int`: `Integer`
- To GraphQL `Float`: `Float`, `Double`, `Decimal`
- To GraphQL `String`: `String`, `Time`,  `Date`, `Datetime`, `DateOrDatetime`, `Uriorcurie`, `Curie`, `Uri`, `Ncname`, `Objectidentifier`, `Nodeidentifier`, `Jsonpointer`, `Jsonpath`, `Sparqlpath`

Closes #2302

This PR replaces outdated PR #2304

### DIFF
--- a/linkml/generators/graphqlgen.py
+++ b/linkml/generators/graphqlgen.py
@@ -24,6 +24,7 @@ class GraphqlGenerator(Generator):
 
     strict_naming: bool = False
     _permissible_value_valid_characters = re.compile("^[_A-Za-z][_0-9A-Za-z]*?$")
+    _types_any = []
 
     def __post_init__(self):
         self.name_compatiblity = NameCompatibility(profile=NamingProfiles.graphql, do_not_fix=self.strict_naming)
@@ -39,6 +40,10 @@ class GraphqlGenerator(Generator):
         return out
 
     def visit_class(self, cls: ClassDefinition) -> str:
+        # no type can be declared for subtypes of "Any"
+        if cls.class_uri == "linkml:Any":
+            self._types_any.append(cls.name)
+            return f"scalar {cls.name}"
         etype = "interface" if (cls.abstract or cls.mixin) and not cls.mixins else "type"
         mixins = ", ".join([camelcase(mixin) for mixin in cls.mixins])
         out = f"{etype} {camelcase(cls.name)}" + (f" implements {mixins}" if mixins else "")
@@ -46,14 +51,35 @@ class GraphqlGenerator(Generator):
         return out
 
     def end_class(self, cls: ClassDefinition) -> str:
-        return "\n  }\n\n"
+        if cls.name in self._types_any:
+            return "\n\n"
+        else:
+            return "\n  }\n\n"
 
     def visit_class_slot(self, cls: ClassDefinition, aliased_slot_name: str, slot: SlotDefinition) -> str:
-        slotrange = (
-            camelcase(slot.range)
-            if slot.range in self.schema.classes or slot.range in self.schema.types or slot.range in self.schema.enums
-            else "String"
-        )
+        if slot.range in self.schema.classes or slot.range in self.schema.slots or slot.range in self.schema.enums:
+            slotrange = camelcase(slot.range)
+        elif slot.range in self.schema.types:
+            if self.schema.types[slot.range].from_schema != "https://w3id.org/linkml/types":
+                slotrange = camelcase(slot.range)
+            else:
+                graphql_scalars = ["Int", "Float", "String", "Boolean", "ID"]
+                if slot.range == "integer":
+                    slotrange = "Int"
+                elif slot.range == "decimal":
+                    slotrange = "Float"
+                elif camelcase(slot.range) in graphql_scalars:
+                    slotrange = camelcase(slot.range)
+                else:
+                    if self.schema.types[slot.range].repr:
+                        python_type = self.schema.types[slot.range].repr
+                    elif self.schema.types[slot.range].base:
+                        python_type = self.schema.types[slot.range].base
+                    if str(python_type) == "float":
+                        slotrange = "Float"
+                    elif str(python_type) == "str":
+                        slotrange = "String"
+
         if slot.multivalued:
             slotrange = f"[{slotrange}]"
         if slot.required:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -169,6 +169,7 @@ requests-cache = "^1.2.0"
 myst-nb = {version=">=1.0.0", python=">=3.9"}
 sphinx-design = "^0.5.0"
 rich = "^13.7.1"
+gql = "^3.5.2"
 
 [tool.poetry.group.tests.dependencies]
 pytest = "^7.4.0"
--- a/tests/test_generators/test_graphql.py
+++ b/tests/test_generators/test_graphql.py
@@ -11,11 +11,11 @@ type Person implements HasAliases
     hasEmploymentHistory: [EmploymentEvent]
     hasFamilialRelationships: [FamilialRelationship]
     hasMedicalHistory: [MedicalEvent]
-    ageInYears: Integer
+    ageInYears: Int
     addresses: [Address]
     hasBirthEvent: BirthEvent
     speciesName: String
-    stomachCount: Integer
+    stomachCount: Int
     isLiving: LifeStatusEnum
     aliases: [String]
   }
@@ -24,8 +24,8 @@ type Person implements HasAliases
 MEDICALEVENT = """
 type MedicalEvent
   {
-    startedAtTime: Date
-    endedAtTime: Date
+    startedAtTime: String
+    endedAtTime: String
     isCurrent: Boolean
     metadata: AnyObject
     inLocation: Place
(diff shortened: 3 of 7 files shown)


-
```diff
@@ -0,0 +1 @@
...eid": "", "outcome": "passed", "result": [{"nodeid": "tests/test_scripts/test_gen_proto.py", "type": "Module"}]}, {"nodeid": "tests/test_scripts/test_gen_proto.py", "outcome": "passed", "result": [{"nodeid": "tests/test_scripts/test_gen_proto.py::test_help", "type": "Function", "lineno": 7}, {"nodeid": "tests/test_scripts/test_gen_proto.py::test_metamodel", "type": "Function", "lineno": 13}]}], "tests": [{"nodeid": "tests/test_scripts/test_gen_proto.py::test_help", "lineno": 7, "outcome": "passed", "keywords": ["test_help", "test_gen_proto.py", "tests/test_scripts/__init__.py", "linkml"], "setup": {"duration": 0.00919295905623585, "outcome": "passed"}, "call": {"duration": 0.0013101249933242798, "outcome": "passed"}, "teardown": {"duration": 0.00013316597323864698, "outcome": "passed"}}, {"nodeid": "tests/test_scripts/test_gen_proto.py::test_metamodel", "lineno": 13, "outcome": "passed", "keywords": ["test_metamodel", "test_gen_proto.py", "tests/test_scripts/__init__.py", "linkml"], "setup": {"duration": 0.0002650829264894128, "outcome": "passed"}, "call": {"duration": 0.14576091698836535, "outcome": "passed"}, "teardown": {"duration": 0.0003946250071749091, "outcome": "passed"}}]}
```
ialarmedalien: was this committed deliberately?

linkml_runtime
```diff
@@ -0,0 +1 @@
+../linkml-runtime/linkml_runtime
```
ialarmedalien: deliberate commit?

linkml/generators/graphqlgen.py
```diff
@@ -39,21 +40,46 @@ def generate_header(self) -> str:
...
         out = f"{etype} {camelcase(cls.name)}" + (f" implements {mixins}" if mixins else "")
         out = "\n".join([out, "  {"])
         return out
 
     def end_class(self, cls: ClassDefinition) -> str:
-        return "\n  }\n\n"
+        if cls.name in self._types_any:
+            return "\n\n"
+        else:
+            return "\n  }\n\n"
```
ialarmedalien, line 57: this can be simplified to 
```py
if cls.name in self._types_any:
    return "\n\n"
return "\n  }\n\n"
```

-
Silvanoc: No, I didn't realized it. Will fix it.

linkml/generators/graphqlgen.py
Silvanoc: @ialarmedalien I prefer the explicit "else" of my code over the implicit one of yours. Your code saves 1 LoC, at the price of readability.

It's [not only me](https://stackoverflow.com/a/9191474) having that opinion.

linkml_runtime
Silvanoc: No, will fix it.

Silvanoc:
I have decided not to declare custom scalars for LinkML types like UriOrCurie, DateTime,...

tests/test_generators/test_graphql.py
```diff
@@ -97,3 +90,17 @@ def test_snapshot(kitchen_sink_path, snapshot):
...
+    generator = GraphqlGenerator(kitchen_sink_path)
+    generated = generator.serialize()
+    print("\nGenerated GraphQL schema:")
+    print("vvvvvv Start GraphQL vvvvvv")
+    print(generated)
+    print("^^^^^^^ End GraphQL ^^^^^^^")
+    try:
+        parse(generated)
+    except Exception as ex:
+        print(f"\nvvvvvv Start Error Message vvvvvv\n{str(ex)}\n^^^^^^^ End Error Message ^^^^^^^")
```
ialarmedalien: Please use a logger to output these printed lines or just remove them.

I would also suggest printing out the error message as part of the pytest.fail statement.

Silvanoc: @ialarmedalien I have adapted the code according your suggestion. Is it fine for you now?

pyproject.toml
```diff
@@ -98,7 +98,8 @@ dev = [
   "requests-cache >= 1.2.0",
   "myst-nb >= 1.0.0; python_version >= '3.9'",
   "sphinx-design >= 0.5.0",
-  "rich >= 13.7.1"
+  "rich >= 13.7.1",
+  "gql>=4.0.0",
```
cmungall, line 102: is this intentional? I can see we might need this for more extensive compliance testing

Silvanoc: Yes, it is. I'm using the [the Python GraphQL client (`gql`)](https://github.com/linkml/linkml/pull/2707/files#diff-539a298b5ed0e1db5de71a6131a8586b90ae898cb52873c6b9ff89a4406607bbR4) because its parser can be used for a [verification of the generated GraphQL schema](https://github.com/linkml/linkml/pull/2707/files#diff-539a298b5ed0e1db5de71a6131a8586b90ae898cb52873c6b9ff89a4406607bbR107) that might not be complete, but much better than the "nothing" that we have now. The parser helped me detect the issue that this patch is also fixing.

tests/test_generators/test_graphql.py
ialarmedalien: LGTM!

PR #2711, opened 2025-05-18
Fix doc generation to add PURL URIs for enums

### LINKED ISSUE #2710: No permanent URIs for enums in documentation
The new documentation generator does not generate w3id.org PURL links for enums but only for classes and slots. Instead gen-doc generates standard links to the gh-pages. This is in contrast to the older gen-markdown which created w3id.org PURL links also for enums.

Looking deeper there is a comment on this in docgen.py that has been present since the first commit:
https://github.com/linkml/linkml/blob/9125ff1c46e85913f9164d0510411fc4c182b305/linkml/generators/docgen.py#L391

It is suggested to add back the generation of w3id.org PURL links for enums.

Solving this issue requires changes in schemaview (linkml-runtime), docgen.py (linkml) and the enum-jinja-template (linkml).

dalito:
ℹ️ **Ready to be merged now that linkml-runtime v1.9.4 is released** (includes linkml/linkml-runtime#393)

The standard CI tests "Build and test linkml" fail since they use the incompatible linkml-runtime until a new release is made that includes the change.

This PR is an example for when the action suggested in #2575 is useful to demonstrate already now that the tests pass with the suggested linkml-runtime change (see green tests "Run tests with unreleased linkml-runtime").

Closes #2710


linkml/generators/docgen.py
```diff
@@ -387,8 +387,8 @@ def uri(self, element: Element, expand=True) -> str:
         :param element:
         :return:
         """
-        if isinstance(element, (EnumDefinition, SubsetDefinition)):
-            # TODO: fix schema view to handle URIs for enums and subsets
+        if isinstance(element, (SubsetDefinition)):
```
ialarmedalien: no need for the brackets around `SubsetDefinition`

dalito: Thanks, fixed! Interesting that it slipped through the ruff check.

ialarmedalien (approved):

Issue #2706, opened 2025-05-17
Unstable sorting causes OWL generator to create different output even when there is no change in underlying schema


vladistan:
---
name: Bug report
title: 'Unstable sorting causes OWL generator to create different output even when there is no change in underlying schema'
labels: bug
assignees: ''

---

**Describe the bug**

Running linkml OWL generator when underlying schema was not changed still produces changes in the output because of unstable sorting.  See the screenshot for example.

<img width="453" alt="Image" src="https://github.com/user-attachments/assets/eeef2e4d-2f09-4dc8-8fa5-1b1f66da1eb2" />

**Version of LinkML you are using**

1.9.1

**Please provide a schema (and if applicable, a data file) that replicates the issue**

Here is the repo to replicate the problem

https://github.com/vladistan/linkml-pokemon/releases/tag/timestamp-issue

Steps to replicate:

- Clone the repo
- Fetch tags 

```git fetch --tags```

- Checkout the tag

```git checkout timestamp-issue```

- Regenerate files

```
    poetry install
    poetry run make gen-project
```

dalito:
#1943 is a similar issue for the rdf generator.

vladistan:
@Silvanoc just tried.   Unfortunately it didn't help.   I ran  `gen-project`  three times and got different result every time.  Also I checked the  `rdflib` version.  It is 7.1.3

The file types where I see this bug are `owl.ttl` and `shacl.ttl` so looks like this is a problem with Turtle serialization.

cmungall:
It looks like the latest version of rdflib is using skolemization behind the scenes in order to name blank nodes and then to do a canonical sort on the triples.

As far as I know, this is not guaranteed to work in the general case, when blank nodes are involved. I am guessing there is a heuristic that tends to work for typical RDF blank bode scenarios, but not for the patterns we see in the OWL to RDF mapping shape (although AFAIK the skolemization approach only falls down if there are shared blank nodes, which should never be the case for OWL with no AnonymousIndividuals, which are never produced from a LinkML schema... so I am guessing somewhat)

The first thing I would do is to create a minimal rdflib test case that demonstrates unstable ordering. It's possible they can add something to their algorithm that could guarantee consistent serialization for OWL.

A better approach for us would be to use py-horned-owl to write the OWL (there are other advantages to this, although it does introduce a rust dependency). The underlying library does serialize consistently: https://github.com/phillord/horned-owl/issues/145

vladistan:
[...]

Issue #1943, opened 2024-02-23
Possibility for deterministic order of multivalue slots in RDF output?


mih:
I have a schema that defines a class with this component:

```
   slot_usage:
      qualified_part:
        inlined: true
        inlined_as_list: true
        multivalued: true
   ...
```

A corresponding data items has

```
    qualified_part:
      - at_location: subdir/README.txt
        relation: gitsha:56094a33cf330fef5b375aa813fc4dc07147729f
      - at_location: subdir/data.bin
        relation: gitsha:b94ef1797f7bfc1ac979be122e1b538bbb0d1d58
```

When I convert this data from YAML to JSON, the output is deterministic. When I convert to RDF, the order of the `qualified_part` declarations is random.

This leads to noise, and is particularly problematic when version-controlling the output, due to the random changes that are introduced, even when there is no relevant change.

I understand that the order here is insignificant from a modeling/validation/content perspective. However, from a technical angle it matters (to me).

----

I now verified that the order of call to `Graph.add()` in `linkml_runtime/dumpers/rdflib_dumper.py:inject_triples()` is indeed deterministic, while the actual dump is in random order. It appears this is either an RDFlib limitation, or a problem in how RDFlib is being used by linkml.

cmungall:
I can't believe it has been 10 years since I wrote https://douroucouli.wordpress.com/2014/03/30/the-perils-of-managing-owl-in-a-version-control-system/ -- this resulted in owlapi changes that guaranteed ordering... but here I am committing this sin with linkml!

In my defense I think this is an issue in rdflib, but I have not looked deeply. If there is a way to fix it we would be keen to adopt!

sneakers-the-rat:
bnodes strike again. 

so if you take the ntriples serialization of your example dataset you get something like this:

<details>
<summary>expand/collapse</summary>

```n3
_:N3842226f91b645f8ac1768095ee0a2ca <http://www.w3.org/ns/prov#atLocation> "subdir/README.txt"^^<https://concepts.datalad.org/ontology/PosixRelPath> .
_:Nb9b22f7110124adeb96e777488a4ddeb <https://concepts.datalad.org/schemas/datalad-dataset-components/:components> <https://concepts.datalad.org/namespace/gitsha/558275f650574389dcbbf7cd8ab5046482473fc8> .
[...]

Issue #696, opened 2022-04-15
Reduce diff of LinkML model generation to zero?


matentzn:
When (re)generating the model, we get quite a bit of diff to the previous run, even if that run happened just a minute earlier. 

here is an example `git status:`

```
modified:   project/jsonld/sssom.context.jsonld
modified:   project/jsonld/sssom.jsonld
modified:   project/owl/sssom.owl.ttl
modified:   project/shacl/sssom.shacl.ttl
modified:   src/sssom/datamodel/sssom.py
```

- [ ] The main issue would be to canonicalise the sort order of turtle serialised artefacts: each time you run the build they change massively because the sort order in the ttl serialiser seems random.
- [ ] The json and py files get date _and_ time stamped. I would suggest if possible to reduce that to date stamped, or get rid of that altogether. I think its nice when running the build twice in quick succession, the diff is zero. This also helps with triggering downstream actions which may look at changes to a generated artefact.

cmungall:
Can someone look into what our options are for canonicalization of ttl serialization with rdflib?

I agree with the suggestion about stamping, let's make this configurable and make date-only the default

dalito:
rdflib does not seem to support this but maybe https://github.com/scriptotek/otsrdflib is useful.

djarecka:
I wonder if you still considering using `otsrdflib` to get predictable order in `shacl.ttl` file. It is part of our repository and it keeps changing every single time

vladistan:
The `otsrdflib` https://github.com/scriptotek/otsrdflib.  Hasn't been updated in ages.  The author is still around and pushing to the other projects,  so perhaps we can work with them to modernize it and bring in to linkml as a dependency.

Or is there another `otsrdflib` out there?

PR #2713, opened 2025-05-19
fix: reproducible rdf

Silvanoc:
Upgrade rdflib to be at least v7.1.3. [That version](https://github.com/RDFLib/rdflib/releases/tag/7.1.3) provides sorted RDF generation, what we need for our tests with snapshots to be stable.

Fixes: #1943
Fixes: #2706

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -4045,14 +4045,14 @@ cffi = {version = "*", markers = "implementation_name == \"pypy\""}
 
 [[package]]
 name = "rdflib"
-version = "7.1.1"
+version = "7.1.4"
 description = "RDFLib is a Python library for working with RDF, a simple yet powerful language for representing information."
 optional = false
 python-versions = "<4.0.0,>=3.8.1"
 groups = ["main"]
 files = [
-    {file = "rdflib-7.1.1-py3-none-any.whl", hash = "sha256:e590fa9a2c34ba33a667818b5a84be3fb8a4d85868f8038f17912ec84f912a25"},
-    {file = "rdflib-7.1.1.tar.gz", hash = "sha256:164de86bd3564558802ca983d84f6616a4a1a420c7a17a8152f5016076b2913e"},
+    {file = "rdflib-7.1.4-py3-none-any.whl", hash = "sha256:72f4adb1990fa5241abd22ddaf36d7cafa5d91d9ff2ba13f3086d339b213d997"},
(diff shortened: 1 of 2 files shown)


sneakers-the-rat:
looks like we'd need to add a `sort=True` param to enable it? but hell ya glad to see that's merged 
https://github.com/RDFLib/rdflib/pull/3008/files#diff-f547e4997730bea72039e7f77d9eea1db53e3d1a1793d2f08221b6159877609eR62

Silvanoc:
> looks like we'd need to add a `sort=True` param to enable it? but hell ya glad to see that's merged https://github.com/RDFLib/rdflib/pull/3008/files#diff-f547e4997730bea72039e7f77d9eea1db53e3d1a1793d2f08221b6159877609eR62

Hhmmm, weird. Just upgrading `rdflib` ordering issues that I was having on another PR's snapshots have just disappeared...

sneakers-the-rat:
🤷 looking closer it looks like `sort=True` shouldn't do anything anyway, so idk. i guess before we declare victory on sorted/canonicalized output we should probably have some specific tests for it, but in the meantime if it fixes bugs then hell ya let's do it.

dalito (approved):
As far as I understand the change in rdflib is only for longturtle format. But that is not used in linkml. Nevertheless rdflib 7.3 has more code for canonicalizing which should help to solve the problem. 
So I am fine with raising the lower version bound for rdflib.

Silvanoc:
Apparently it helped me with one of my branches, but it is [not really a reliable solution](https://github.com/linkml/linkml/issues/2706#issuecomment-2896282854).

> As far as I understand the change in rdflib is only for longturtle format.

That's what I understood too. But after applying it one of my branches stopped failing due to RDF sorting, so without further analyzing it (I have too many open fronts), I hoped it would help.

> Nevertheless rdflib 7.3 has more code for canonicalizing which should help to solve the problem. So I am fine with raising the lower version bound for rdflib.

I suppose you mean v7.1.3.

dalito:
> I suppose you mean v7.1.3.

Yes. I edited the comment.

vladistan:
Parent issue for this is :  #696

Silvanoc:
> Parent issue for this is : #696

@vladistan there are no parent issues or PRs for a PR, but related issues are now children of #696

Issue #2371, opened 2024-10-15
[chore] Clean up and remove `utils` in `linkml` and `linkml-runtime`


sneakers-the-rat:
Related to: https://github.com/linkml/linkml/issues/1784

**What is your feature request?**

I am actually a defender of `utils` modules as a valid pattern, where sometimes small routines genuinely have no good place to go, but I think that the scale of `utils` modules in both `linkml` and `linkml-runtime` is pretty extreme and proving detrimental to sustainability/maintenance. 

Problems with gigantic utils modules:
- Can't find stuff when you want it
- So you reimplement it
- Over and over.
- Forever.

This has happened to me personally about a dozen times, and I think it's a pretty strong turnoff to new developers who are scoping out the project and see "whoa nelly that is a big junk drawer" - as much as i adore the somewhat 'by the seat of the pants' nature of linkml, it's a sign of bad code health, and i have had several people express to me candidly that it was enough for them to not consider linkml (not trying to be harsh, but also not kidding). It generally makes adding new features harder than it needs to be (hunting through utils to see if something has already been written) and makes maintenance challenging (something will turn out to have its root cause in a utils module that hasn't been touched in years and we frankly forgot about).

This is especially true when there are seemingly core classes and functions in `utils`. The most obvious ones to me are `Generator` (which should go in `linkml.generators`)  and `Schemaview` (which should be split off into its own package imo). 

I think there needs to be a relatively large refactoring to split up
- "core" linkml representation: all the stuff that is needed to evaluate a linkml schema and return a fully induced schema
- generators, each of which might have their own runtime package
- dumpers/loaders (or combine with generator packages)
- validation/linting
- transformation/conversion
- cli
because otherwise these things get splayed out and mashed together and we end up with a million dep package,
but to get there we need to start tidying up `utils` :)

Here's what i would do

`linkml`
- cli_utils.py -> `cli`
- converter.py -> `cli`
- datautils.py -> wherever loaders/dumpers go
- datavalidator.py -> `validator`
- deprecation.py -> either stays or in package root
- exceptions.py -> either stays or in package root
- execute_tutorial.py -> I actually am not sure because I think this should just be doctests/`myst_nb`
- generator.py -> `generators`
- helpers.py -> remove
- logictools.py -> `transformers`
- mergeutils.py -> I think this is mostly deprecated post schemaview?
- rawloader.py -> dumpers/loaders
[...]

PR #2716, opened 2025-05-20
refactor: move converter from utils

Silvanoc:
Move `linkml.utils.converter` to `linkml.converter.cli`, which is more consistent with the structuring of other tools in the repo:
- Package for the top-level functionality.
- CLI specific to that functionality inside of that package.
- Integration into higher-level CLI.


Silvanoc:
Part of refactoring proposed (and implicitly accepted, since silent is consent) in issue #2371 

Keeping it as `Draft` until issues on `main` are fixed and tests can succeed.

ialarmedalien:
Looking at the other files that have been put in `utils`, should there be a set of data utility packages? I'm not totally convinced by this one.

I do think that there are other module moves on your list that are no brainers -- e.g. moving the `generators.py` base model into the generators dir -- that could be done quickly and painlessly if you were willing to create PRs for them.

Thanks for your continued efforts to improve the quality and accessibility of linkml!

Silvanoc:
> Looking at the other files that have been put in `utils`, should there be a set of data utility packages?

@ialarmedalien the mess that the `utils` directories have become was the reason why @sneakers-the-rat created #2371, that motivated this PR. So a lot of changes are needed there...

Having said that, I'm not sure what you mean with "data utility packages" 🫣 Could you elaborate a bit?

PR #2722, opened 2025-05-22
Backport improvements proposed for rustgen

kervel:
This PR contains code changes to the generic build code that were introduced by @sneakers-the-rat for his rustgen proposal.

I think they can be merged in separately, since they make sense even without rustgen. This will make it easier to keep a rustgen branch mergeable in the future.

I think they should not have any impact on the functionality of linkml.

### DIFF
--- a/linkml/generators/common/template.py
+++ b/linkml/generators/common/template.py
@@ -5,10 +5,10 @@ See :mod:`.linkml.generators.pydanticgen.template` for example implementation
 """
 
 from copy import copy
-from typing import Any, ClassVar, Optional, Union
+from typing import Any, ClassVar, Dict, Generator, List, Optional, Union
 
 from jinja2 import Environment
-from pydantic import BaseModel
+from pydantic import BaseModel, Field, field_validator
 
 
 class TemplateModel(BaseModel):
@@ -42,7 +42,7 @@ class TemplateModel(BaseModel):
             environment (:class:`jinja2.Environment`): Template environment - see :meth:`.environment`
         """
         if environment is None:
-            environment = TemplateModel.environment()
+             environment = type(self).environment()
 
         fields = {**self.model_fields, **self.model_computed_fields}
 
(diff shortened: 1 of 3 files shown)


linkml/generators/common/template.py
```diff
@@ -72,12 +73,317 @@ def exclude_from_meta(cls: "TemplateModel") -> list[str]:
...
+
+
 def _render(
     item: Union[TemplateModel, Any, list[Union[Any, TemplateModel]], dict[str, Union[Any, TemplateModel]]],
     environment: Environment,
 ) -> Union[str, list[str], dict[str, str]]:
     if isinstance(item, TemplateModel):
-        return item.render(environment)
+
+        class _Stringified:
```
sneakers-the-rat, line 367: i think this one may be yours? a twisted kind of genius. *sort of* goes against the idea of these template models being a way to move as much logic out of the template as possible, but hey you know what can't argue with need. 

Probably would be a good idea to move it to module level, i don't see a good reason to treat it as a class closure? would be good to be able to test it independently

kervel: hmm yes, i'm actually not too proud of that (and not a big fan of "magic" stuff). the logic for this was to allow things like this:

```
mod {{ name }}Slots {
{% for property in properties %}
    {% if property.type_ | length > 1 and property.has_slot_usage %}
    pub enum {{ property.name }}_range {
    {% for t in property.type_ %}
        {{ t }}({{ t }}){% if not loop.last %}, {% endif %}

    {% endfor %}
    
    }
    {% endif %}
{% endfor %}
}
```

to create an enum for property range. i have to refactor that part anyway (the enum should be anchored to the class not to the property that uses the class), but i tought it is legit to sometimes have access to properties of a template outside that template.

the alternative would be to create multiple templates. i broke your PR (and fixed it again) by making those type aliases "pub" because the same template was used in the ::new() parameter list and there obviously "pub" doesn't make sense.

here the options were
* do what i did :-) now i can just use the type of the property rather than its rendered version
* create a second template "propertyInParameterList" ?
* or have multiple ways of rendering a template (eg `{{ template | render('parameterlist') }}`, `{{ struct | render('enum-with-subclasses') }}`

all options work. let me know what you think, i can still refactor either way.

sneakers-the-rat: For switching how a property is rendered, I would just move that into the property's template! So by the time you are rendering the properties they are all strings, and they know how to handle switching their representation based on the values of their template model, but the parent class model doesnt need to know.

The template models are all about modularizing the templates, so not knowing about the values of the child objects is on purpose. Each model only knows about its own values, and if there needs to be interactions between multiple models, that can happen in the model itself (where it can be properly tested and analyzed) rather than in the template (which should have the responsibility only of representing the form of the model)

sneakers-the-rat: And yes if different kinds of things need different templates, then the strategy is make more templates and models. More models with focused templates is more transparent and easier to read and write than fewer models with complex logic

kervel:
superseded by #2889

Issue #2723, opened 2025-05-23
Fix and improve tests/test_utils/test_converter.py::test_infer_and_convert_fixed


Silvanoc:
**Describe the bug**
Test `tests/test_utils/test_converter.py::test_infer_and_convert_fixed` is supposed to be testing roundtripping through yaml->json->yaml->rdf->json. But no intermediate checks are being done, only after the whole chain. That way error reporting is not granular enough.

Not only that, since the very last step is expected to convert to JSON and with the same files as the first step, if something fails in between, the final tests are checking the result of the first step! It's even worse, in fact the chain is failing and the test is succeeding due to the above mentioned issue. A PR for reproduction will be provided.

**Version of LinkML you are using** v1.9.2

**Please provide a schema (and if applicable, a data file) that replicates the issue**

PR #2724, opened 2025-05-23
test(convert): fix and improve test

Silvanoc:
Calling multiple conversions in a row without checking intermediate results does not provide enough information to know what has failed (if so). This patch improves it.

Additionally missing intermediate checks where missing an error in the test execution. Therefore the test has been split in two:
- one with the correct part (which was correctly succeeding)
- another with the erroneous part (which was erroneously succeeding)

Fixes: #2723

### DIFF
--- a/tests/test_utils/test_converter.py
+++ b/tests/test_utils/test_converter.py
@@ -18,7 +18,47 @@ def test_help(cli_runner):
     assert "INPUT" in out
 
 
-def test_infer_and_convert(input_path, cli_runner, tmp_path):
+P1_EXPECTED = {
+    "full_name": "first1 last1",
+    "is_juvenile": True,
+    "age_in_years": 10,
+    "age_in_months": 120,
+    "age_category": "juvenile",
+}
+
+P2_EXPECTED = {
+    "full_name": "first2 last2",
+    "age_in_years": 20,
+    "age_in_months": 240,
+    "age_category": "adult",
+}
+
+
+def check_output(json_out):
+    with open(json_out) as file:
+        obj = json.load(file)
+        persons = obj["persons"]
+        for p, exptc in [(persons["P:1"], P1_EXPECTED), (persons["P:2"], P2_EXPECTED)]:
+            for attr in exptc.keys():
+                assert p[attr] == exptc[attr]
+
+        assert "is_juvenile" not in persons["P:2"]
+
+
+def test_infer(input_path, cli_runner, tmp_path):
+    """
+    Tests using the --infer option to add missing values
+    """
+    schema = input_path("schema_with_inference.yaml")
+    data_in = input_path("data_example.yaml")
+    json_out = tmp_path / "data_example.out.json"
+    result = cli_runner.invoke(cli, ["--infer", "-s", schema, data_in, "-o", json_out])
+    assert result.exit_code == 0
+    check_output(json_out)
+
+
+@pytest.mark.xfail(reason="Bug 2723: missing intermediate checks")
+def test_convert(input_path, cli_runner, tmp_path):
     """
     Tests using the --infer option to add missing values, and also roundtripping
     through yaml->json->yaml->rdf->json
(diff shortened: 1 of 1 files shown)


tests/test_utils/test_converter.py
```diff
@@ -18,7 +18,35 @@ def test_help(cli_runner):
...
+        assert p1["is_juvenile"]
+        assert "is_juvenile" not in p2
+        assert p1["age_in_years"] == 10
+        assert p1["age_in_months"] == 120
+        assert p1["age_category"] == "juvenile"
+        assert p1["full_name"] == "first1 last1"
+        assert p2["age_in_years"] == 20
+        assert p2["age_in_months"] == 240
+        assert p2["age_category"] == "adult"
+        assert p2["full_name"] == "first2 last2"
```
ialarmedalien: This duplicates the results that are being checked in `test_convert`. Suggestions for DRYing out the code:

- move it all into a function that does this set of assertions
- if `p1` and `p2` have all the attributes that you are testing above and no others, do a simple dictionary comparison, e.g.
```
# set up as a fixture or just a constant in the file
P1_EXPECTED = {
  "full_name": "first1 last1",
  "is_juvenile": True,
  "age_in_years": 10,
  "age_in_months": 120,
  "age_category": "juvenile"
}

# in a test function
def some_test_here(...):
  persons = result_of_some_function(...)

  assert persons["P:1"] == P1_EXPECTED
```

- if there are other attributes but you only want to check those that are given, create a function that ensures all the expected attributes are present in a dictionary and that they have the correct values.

tests/test_utils/test_converter.py
```diff
@@ -28,10 +56,18 @@ def test_infer_and_convert(input_path, cli_runner, tmp_path):
     json_out = tmp_path / "data_example.out.json"
     yaml_out = tmp_path / "data_example.out.yaml"
     rdf_out = tmp_path / "data_example.out.ttl"
-    cli_runner.invoke(cli, ["--infer", "-s", schema, data_in, "-o", json_out])
-    cli_runner.invoke(cli, ["-s", schema, json_out, "-t", "yaml", "-o", yaml_out])
-    cli_runner.invoke(cli, ["-s", schema, yaml_out, "-t", "rdf", "-o", rdf_out])
-    cli_runner.invoke(cli, ["-s", schema, rdf_out, "-t", "json", "-o", json_out])
+    result = cli_runner.invoke(cli, ["--infer", "-s", schema, data_in, "-o", json_out])
+    print(f"infer to JSON: {result.output}")
```
ialarmedalien: You have encountered my policy on testing and print statements before... can you either convert these to logger statements or add a message to the assert statement so that it's easy to track down which of the assertions failed.

Silvanoc: I personally disagree with that policy 🙂 But as commented in the past, I prefer bending to that policy (which is probably also the implicit project policy, since pre-existing tests are not verbose) to get your valuable reviews. So I'm removing them to ensure a smooth merging.

No matter that disagreement, I really appreciate the time you take to review my contributions 🚀

---

But let me explain why I disagree:

I want verbose tests that in the first failed run give me enough information to diagnose the issue. Those `print`s will **only be shown in case of failure** or explicitly raised verbosity, so **no output on successful run**. If failure reporting is verbose enough, I can diagnose many issues just looking at the traces.

Additionally I avoid debuggers as much as I can. In the projects being run with other people with a similar pattern like mine, debuggers aren't needed to diagnose over 90% of the failures and just looking at the 1st failed job (thanks to verbosity in case of a failure). In my main development environment concurrency and race-conditions are very common and hardly debuggable because you cannot easily reproduce them.

ialarmedalien: Thanks for explaining your situation -- I can see why you would put in a lot more print statements based on your experience working on verbose codebases with plenty of logging, etc. Unfortunately the linkml codebases tend to be the opposite, without much logging, informative error messages, or even inline comments to explain what opaque sections of code are doing. 😵‍💫 😭

In terms of writing tests and debugging failing tests, my MO in this codebase is to run individual failing tests in the debugger of my IDE, setting breakpoints and stepping through the execution path until the issue is found. You can't rely on having an informative error message and issues can be buried in jinja templates or innocuous-looking assumptions that the codebase makes, and it's quicker than adding a load of "print" statements within the code to work out where things have gone wrong. I hardly used debuggers before but they have saved my sanity when working with the LinkML codebases!

BTW, in terms of other suggestions, etc., that I make -- I typically have all the Ruff linter rule sets active for code I work on and some rules just become second nature after a while. The [flake8-simplify](https://docs.astral.sh/ruff/rules/#flake8-simplify-sim) and [flake8-comprehensions](https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4) rulesets get a good workout when doing anything with LinkML...

Silvanoc: > Thanks for explaining your situation -- I can see why you would put in a lot more print statements based on your experience working on verbose codebases with plenty of logging, etc. Unfortunately the linkml codebases tend to be the opposite, without much logging, informative error messages, or even inline comments to explain what opaque sections of code are doing. 😵‍💫 😭

You are right. IMO logging is really scarce in the project. But we have to play the cards we're dealt.

> In terms of writing tests and debugging failing tests, my MO in this codebase is to run individual failing tests in the debugger of my IDE, setting breakpoints and stepping through the execution path until the issue is found. You can't rely on having an informative error message and issues can be buried in jinja templates or innocuous-looking assumptions that the codebase makes, and it's quicker than adding a load of "print" statements within the code to work out where things have gone wrong. I hardly used debuggers before but they have saved my sanity when working with the LinkML codebases!

I agree that in LinkML for Jinja2 and Python generated code for validation there is no way around the debugger. It is also the established way to debug in the project. So I'll have to live with it.

> BTW, in terms of other suggestions, etc., that I make -- I typically have all the Ruff linter rule sets active for code I work on and some rules just become second nature after a while. The [flake8-simplify](https://docs.astral.sh/ruff/rules/#flake8-simplify-sim) and [flake8-comprehensions](https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4) rulesets get a good workout when doing anything with LinkML...

WRT code style, I fully agree on having linter rules fixed by the project as a common ground and to avoid discussions. If I dislike them, then I'd try to get them changed at a project level and not discuss them directly in a MR. In general, CI should be enforcing those rules, so you should not need to discuss them with me.

I might disagree though on rules that are not agreed at a project level. Some of those rules have a good reason for existence, and I would apply them as long as they don't disagree with project-fixed rules. But others IMO don't have a good reason and are more a question of personal taste (even if someone created some linter rules for that). See [this discussion](https://stackoverflow.com/a/9191474) regarding the removal of `else` on return that you proposed in [this comment](https://github.com/linkml/linkml/pull/2707#discussion_r2162723325) of MR #2707 just as an example.

tests/test_utils/test_converter.py
Silvanoc: Done

tests/test_utils/test_converter.py
Silvanoc: @ialarmedalien Since this discussion is not really about the content of this PR, I'm marking it as resolved to avoid blocking it.

ialarmedalien (approved):

PR #2698, opened 2025-05-16
Remove special case for Windows in test_docgen

dalito:
upstream_repo: linkml/linkml-runtime
upstream_branch: patch-v1.9.3-rc1

The test for this PR will fail until linkml/linkml-runtime#391 is merged and a new version of linkml-runtime is released.

Closes #2696

PR #2727, opened 2025-05-26
fix(gen-plantuml): fix changes in svg code

Silvanoc:
Kroki has slightly changed the names of the groups being used in the generated SVG code (XML dialect). This patch updates it.

It also makes tests on *unreleased `linkml-runtime`* non-blocking. The results should have enough visibility though, because their results are visible at PR-level. Simply scroll down to see them on this PR.

Current status is very verbose because it reports both failing and succeeding tests, but is the best I can do as of now. My proposal is to merge it to stop blocking all other PRs and improve it (e.g. creating comments only for failing jobs or having a single consolidated comment) later.

Fixes: #2728 (first commit)
Fixes: #2729 (2nd commit)

### DIFF
--- a/.github/workflows/scripts/report_unreleased_linkml_runtime.py
+++ b/.github/workflows/scripts/report_unreleased_linkml_runtime.py
@@ -5,35 +5,114 @@ import os
 import subprocess
 import sys
 
+summary_header = "Tests with 'unreleased linkml-runtime'"
+
+
+def format_results_message(total_tests, passed_count, failed_tests):
+    """Format the results message for the PR comment"""
+
+    server_url = os.environ.get("GITHUB_SERVER_URL", "")
+    repository = os.environ.get("GITHUB_REPOSITORY", "")
+    run_id = os.environ.get("GITHUB_RUN_ID", "")
+
+    # Construct the job URL
+    job_url = f"{server_url}/{repository}/actions/runs/{run_id}"
+
+    if not failed_tests:
+        lines = [
+            f"## ✨ 🧪 {summary_header} ([View Details]({job_url})) ✨\n",
+            f"- Total Tests: {total_tests}",
+            f"- ✅ Passed: {passed_count}",
+            "\n✨ All tests passed!",
+        ]
+    else:
+        lines = [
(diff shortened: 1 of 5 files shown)


Silvanoc:
Failing checks are unrelated to this fix.

Silvanoc:
Reporting of tests failing when using unreleased `linkml-runtime` visible on the job page. For example: https://github.com/linkml/linkml/actions/runs/15258861370/job/42912651472?pr=2727

The reported issues are visible as "Annotations": 
image

Silvanoc:
> into the first message.

@dalito Which message do you mean?

Silvanoc:
> If the checks overview is all green (as now) it will be very easy to overlook that there are issues. I look forward to what you come up with in #2733.

@dalito IMO the current status is even worse than that. As of now all PR-pipelines are failing and, therefore, blocking fixes/features completely unrelated to potential issues with unreleased `linkml-runtime`! That pipeline of PRs to be merged will start generating more and more incompatibilities with each other...

With this approach, we don't block development, but have an easy way to ensure compatibility before a release. Sure, manual effort to check all the *runreleased linkml-runtime* jobs, but that should not hurt that much.

dalito:
I would also prefer to get this PR merged ASAP and defer improvements to a later PR.

sierra-moxon:
While I very much agree that the circular dependency is annoying, needs to be fixed, is complex to handle, and is causing us trouble, I do not want to move the testing of that dependency to a "continue on failure" status with this PR.  My vote would be to:
- get our linkml, linkml-runtime PRs to a manageable state by doing several small releases with merges of outstanding PRs in these two repos (I've been working towards this over the past couple of months with folks).
- discuss a mono-repo in a developer call.  
- break down the tasks that are needed to get us to a mono repo if we decide to go that route and get them assigned out (one of which is moving to poetry 2.x, then on to uv).
- try the monorepo.  
- ... then we figure out the next steps (maybe more modularity, etc...).  

I think @dalito identified one key to fixing the troubles (not the circular dependency itself, but the failing tests in PRs in the linkml main) in this PR: https://github.com/linkml/linkml/pull/2698.  I've moved it forward now via a linkml-runtime 1.9.3-rc1 release candidate and by applying the PlantUML fix that you provided @Silvanoc (which arose after this PR was created).

Issue #2742, opened 2025-06-02
Example files in examples/PersonSchema are out of date and not pre-commit rules compliant


vladistan:
---
name: Bug report
about: Create a report to help us improve
title: 'Example files in examples/personinfo are out of date and not pre-commit rules compliant'
labels: bug
assignees: ''

---

**Describe the bug**

Re-running make in `examples/PersonSchema` generates output files that are very different from what's already in the repo.   This is because the existing files are out of date with the development of generators.

Also, pre-commit is not happy with the existing generated files

**Version of LinkML you are using**

1.9.2

**Please provide a schema (and if applicable, a data file) that replicates the issue**

https://github.com/vladistan/linkml/blob/main/examples/PersonSchema/personinfo.yaml

To replicate 

1. Run `make` in  examples/PersonSchema
2. Look at the huge diff
3. Run `pre-commit run -a`
4. Look at errors and diff

PR #2743, opened 2025-06-02
Fix for #2742

vladistan:


vladistan:
Looks like this is too large and too brittle to do in one go.  I will be splitting this up into smaller easier to review pieces

PR #2744, opened 2025-06-02
Add pre-commit to code quality check

vladistan:


.github/workflows/main.yaml
```diff
@@ -31,6 +31,10 @@ jobs:
         run: tox -e codespell
       - name: Run code quality checks
         run: tox -e lint
+      - name: Run pre-commit checks
```
dalito, line 30: Since pre-commit runs codespell you could delete the lines above
```yaml
      - name: Run code quality checks
        run: tox -e lint
```

vladistan: Yep,  done

Silvanoc:
I use pre-commit in my other projects. I simply activate my local `git pre-commit` hooks and I known that I won't be pushing anything that'll be rejected by CI. I've shown some support for it in the past, but there was some opposition at that time. Happy to see it coming back 😄

vladistan:
Yes I do the same thing.  But unfortunately not everybody does,  I see this
problem in other projects as well.  This is why there are many files in
this project are not compliant with pre-commit.  We do have ruff and
codespell as quality gates for the PRs,  but these checks only apply to the
python code and are not run on the code that was already checked in.
Enabling full pre-commit scan as a quality gate should fix this problem.

sierra-moxon (approved):

.github/workflows/main.yaml
sierra-moxon: I'm resolving this comment as "done".

PR #2750, opened 2025-06-09
Use `&` to separated implemented interfaces in graphql-schema definitions

christian-monch:


christian-monch:
Accidentially created the pull request against the wrong repository

PR #2751, opened 2025-06-09
Fix enum URL generation for Mermaid diagrams

### LINKED ISSUE #2732: Mermaid diagrams have wrong URLs for Enums
In pid4cat-model created with linkml 1.9.2 and linkml-runtime 1.9.2 the mermaid diagrams do not contain correct URLs, for example on

https://nfdi4cat.github.io/pid4cat-model/v0.4.2/elements/classes/ResourceInfo/ the link to **ResourceCategory** enum is wrong:

- currently created URL: <BR>`https://nfdi4cat.github.io/pid4cat-model/latest/elements/classes/ResourceCategory`
- expected URL: <BR>`https://nfdi4cat.github.io/pid4cat-model/latest/elements/enums/ResourceCategory/`

So linkML does not correctly use the element type information but treats enums as if they are classes.

We generate the documentation with the options: 
```
LINKML_GENERATORS_DOC_ARGS="--truncate-descriptions False --hierarchical-class-view --subfolder-type-separation --index-name overview"
```

dalito:
With this PR the links in mermaid diagrams are generated with the same code as links in markdown.

Todo:
- [x] Test with pid4cat where the issue was observed and wich uses `--subfolder-type-separation`
- [x] Test with a standard linkml-project that does not use `--subfolder-type-separation`
- [x] Add/check docgen tests

Closes #2732

### DIFF
--- a/tests/test_generators/test_mermaidclassdiagramgen.py
+++ b/tests/test_generators/test_mermaidclassdiagramgen.py
@@ -17,3 +17,9 @@ def test_generate_class_diagrams(kitchen_sink_path, tmp_path):
         contents = md_file.read_text()
         assert "classDiagram" in contents, f"'classDiagram' not found in {md_file}"
         assert f"class {cls_name}" in contents, f"Class '{cls_name}' not found in {md_file}"
+
+        if cls_name == "Person":
+            # Check for enum usage in Person class
+            assert 'click LifeStatusEnum href "../LifeStatusEnum"' in contents, (
+                f"'LifeStatusEnum' not found in {md_file}"
+            )


ialarmedalien:
Were there supposed to be tests included in the PR?

dalito:
You have a point. I should at least add a test demonstrating the observed issue...

dalito:
@ialarmedalien - I added a test that shows the failure and that the fixed code is correct with/without `--subfolder-type-separation`.

linkml/generators/docgen.py
```diff
@@ -410,6 +410,21 @@ def uri_link(self, element: Union[Element, str]) -> str:
...
+        :param e: element to be linked
+        :return: string with link
+        """
+        # Reuse markdown link generation to avoid code duplication.
+        md_link = self.link(e)
+        if not md_link.endswith(")"):
+            return md_link
+        link = md_link.rsplit("(")[-1][:-1]
+        link = link.removesuffix(".md")
+        return f"../{link}/"
```
paul121, line 426: Why is the `../` to the parent directory hard-coded here?

linkml/generators/mermaidclassdiagramgen.py
```diff
@@ -90,6 +90,10 @@ def name(self, element: Element) -> str:
         """Returns the canonical name for an element."""
         return element.name
 
+    def link_mermaid(self, element):
+        """Generates a link for the given element."""
+        return f"../{self.name(element)}"
```
paul121, line 95: Ditto, why is the `../` to the parent directory hard-coded here?

paul121 (commented):
I'm not familiar with these internal link functions but leaving a comment as it pertains to #2550 

It seems that this logic `subfolder-type-separation` hard-coded to be `True` ?

linkml/generators/docgen.py
dalito: For the currently generated page/documentation layout this is always required to link from the mermaid diagram to any page of another element.

dalito:
> It seems that this has subfolder-type-separation logic hard-coded to be True ?

No, it adapts to the option. See the added test.

paul121:
> No, it adapts to the option. See the added test.

I see. So #2550 must be a separate issue: links for mermaid diagrams _always_ use a relative link to the parent directory

ialarmedalien (approved):

linkml/generators/docgen.py
```diff
@@ -485,14 +500,13 @@ def _exclude_type(self, t: TypeDefinition) -> bool:
...
     def _is_external(self, element: Element) -> bool:
-        # note: this is currently incomplete. See: https://github.com/linkml/linkml/issues/782
         if element.from_schema == "https://w3id.org/linkml/types" and not self.genmeta:
             return True
         else:
             return False
 
     @staticmethod
-    def _markdown_link(n: str, name: str = None, subfolder: str = None) -> str:
+    def _markdown_link(n: str, name: str = "", subfolder: str = "") -> str:
```
ialarmedalien, line 509: 👍

linkml/generators/mermaidclassdiagramgen.py
ialarmedalien, line 95: is there a mermaid test file that you can add a couple of tests for this into?

dalito: There is only one test. I added an assert to verify that the enum link generation is still working.

Issue #2757, opened 2025-06-09
Many non python files are not compliant with pre-commit rules.


vladistan:
---
name: Bug report
about: Create a report to help us improve
title: 'Many non python files are not compliant with pre-commit rules.'
labels: bug
assignees: ''

---

**Describe the bug**

Looks like after adding pre-commit linting to the work many existing non-python files were grand-fathered in.  This prevents adding pre-commit to the PR workflow because it flags a lot of files as invalid.

**Version of LinkML you are using**

1.9.3-rc1

**Please provide a schema (and if applicable, a data file) that replicates the issue**

This is not a schema related problem.  Here is a link to the workflow run that shows the problem:

https://github.com/linkml/linkml/actions/runs/15522142140/job/43696820565?pr=2744

vladistan:
@sierra-moxon this is fixed now.

dalito:
Thanks for all your PRs to make this happen!

PR #2759, opened 2025-06-09
Update tox config and move it to pyproject.toml

dalito:
Since a long time linkml includes a pre-commit config but obviously (#2757) not many devs are using it. Another way to run checks consistently is via tox, but its config was also outdated.

tox related changes in this PR:
- adds Python 3.13, removes pydantic 1
- re-uses the pre-commit configuration for `tox -e format` to avoid the problem of deviating linter-configurations
- switches from `tox.ini` as config file to storing the config in `pyproject.toml`

The exact same checks should also be run in CI to catch problems that were not detected before PR submission.

Should we update the gh action and make all linter-made changes as part of this PR or in another one? --> Another one!

Related to #2757


vladistan:
> The exact same checks should also be run in CI to catch problems that were not detected before PR submission.

Yes

pyproject.toml
```diff
@@ -198,11 +198,17 @@ sphinx-jinja = "^2.0.2"
...
-# TODO: bring in tests in too
-skip = '.git,*.pdf,*.svg,./tests,pyproject.toml,*.dill,poetry.lock,./notebooks/DistributedModels.ipynb,./docs/LinkML-logo/LinkML-Revised-Round-Logo-MASTER-FILE.ai'
+skip = [
+  ".git", "*.pdf", "*.svg", "*.dill",
+  "./tests",  # TODO: bring in tests in too
+  "pyproject.toml",
+  "poetry.lock",
+  "./notebooks/DistributedModels.ipynb",
+  "./docs/LinkML-logo/LinkML-Revised-Round-Logo-MASTER-FILE.ai"
+]
```
dalito, line 208: Codespell supports toml lists now which provides a nicer way for configuring the list of patterns.

sierra-moxon:
This is great @dalito!  thanks for taking it on.  I imagine this will fail if we don't change up the commands? https://github.com/linkml/linkml/blob/main/.github/workflows/main.yaml

dalito:
It will continue to work in CI and not cause failures. The tox environment names are unchanged and `tox -e format` (to run pre-commit) is only added but not yet activated in this PR. I thought we better wait until @vladistan finished his mission 😉

PR #2760, opened 2025-06-09
Reformat docs to be pre-commit compliant

vladistan:
Next batch of fixes for https://github.com/linkml/linkml/issues/2757

This time it's documentation


dalito (approved):
Looks good! Seeing this makes me wonder if we should add a [.editorconfig](http://editorconfig.org) file with
```
insert_final_newline = true
trim_trailing_whitespace = true
```
as another line of defense.

vladistan:
Oops, I accidentally closed this and can't re-open it again.

There is a copy of it being tested here.

https://github.com/linkml/linkml/pull/2761

vladistan:
As for your question,  yes this is a good idea to add these to the `.editorconfig`, however not every editor supports it / has it turned on.

PR #2767, opened 2025-06-10
Make person schema generated files pre-commit compliant

vladistan:


vladistan:
Also there are non whitespace changes here because examples were regenerated from the most recent schema.  They were out of date.

dalito:
Running `pre-commit run --all` alone or in combination with `pytest` does not rebuild the example project with the changes submitted here but leads to only white space changes. How did you regenerate the example or by which command was the regeneration triggered?

A separate issue is that the freshly generated project files are not conforming to the rules. The jinja templates need adjustment.

vladistan:
@dalito @sierra-moxon here is how to recreate this commit.  In on a local machine

Note you will not get the exactly same commit because our generators are not reproducible,  there are already issues created to address it.

But you should get almost the same result the difference will be in timestamps at the some file headers and different sorting order in TTL files.  It less than 10 files that will be different.

To run this test you need to have this branch fetched to your local repo

After you done make sure to reset your main branch using `git reset --hard origin/main`


```

git checkout main
git reset --hard

# Navigate to examples and run make in PersonSchema
cd examples/PersonSchema
make

# Run pre-commit hooks
pre-commit run -a || true
cd ..

# Reset changes in various directories
cd ../tests
git checkout -- .
cd ../notebooks
git checkout -- .
cd ../linkml/generators
git checkout -- .
cd ../workspaces
git checkout -- .

# Return to main directory and soft checkout the PR branch to see the difference

cd ..
git reset --mixed  origin/make-generated-example-files-pre-commit-compliant
```

dalito:
Thanks. If this

> cd examples/PersonSchema
> make

can lead to inconsistent or outdated code, I wonder if it should be run routinely as part of the checks. On the other hand I am not even sure why the generated example outputs are under version control.

PR #2773, opened 2025-06-17
Call sqla templates directly

ialarmedalien:
Having the two sql alchemy templates registered in the __init__ file was causing local import issues whilst debugging generators, as the IDE cannot tell whether `from sqlalchemy import ...` refers to the package SQLAlchemy or the `sqlalchemy` directory.

Removed some cruft and added future-compatible annotations at the same time.


dalito (approved):
Looks good!

PR #2774, opened 2025-06-17
Switch to using ruff for code formatting

ialarmedalien:
Switch to using ruff for code formatting and linting, instead of using black for formatting.

Running `ruff format` over the repo results in a few minor changes, mostly in `'` and `"` use and display of long lines.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,16 +11,15 @@ repos:
   - id: check-toml
   - id: trailing-whitespace
     exclude: __snapshots__
-- repo: https://github.com/psf/black
-  rev: 24.10.0
-  hooks:
-  - id: black
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ensure this version stays in sync with tox.ini
-  rev: v0.4.8
+  rev: v0.11.13
   hooks:
-  - id: ruff
+  # Run the linter.
+  - id: ruff-check
     args: [--fix, --exit-non-zero-on-fix]
+  # Run the formatter.
+  - id: ruff-format
 - repo: https://github.com/codespell-project/codespell
   rev: v2.2.6
   hooks:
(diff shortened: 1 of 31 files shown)


.pre-commit-config.yaml
```diff
@@ -11,16 +11,15 @@ repos:
...
   # Ensure this version stays in sync with tox.ini
-  rev: v0.4.8
+  rev: v0.11.13
   hooks:
-  - id: ruff
+  # Run the linter.
+  - id: ruff-check
     args: [--fix, --exit-non-zero-on-fix]
+  # Run the formatter.
+  - id: ruff-format
```
ialarmedalien, line 22: remove black, add in the ruff formatter command

pyproject.toml
```diff
@@ -326,10 +326,10 @@ commands = [
...
 [tool.tox.env.lint]
-description = "Run code linter (no fixes)."
+description = "Run code linter and formatter (no fixes)."
 skip_install = true
-deps = ["black", "ruff==0.4.8"]
+deps = ["ruff==0.11.13"]
 commands = [
   ["ruff", "check", "{posargs:.}"],
-  ["black", "--check", "--diff", "{posargs:.}"],
+  ["ruff", "format", "{posargs:.}"],
```
ialarmedalien, line 334: bye bye black, hello ruff!

dalito (approved):
Good idea!

Black is also used in https://github.com/linkml/linkml/blob/4c13dd256c068fea12c47f6a9a9e373ca3dce9b4/linkml/generators/pydanticgen/pydanticgen.py#L185 - but that could be changed separately.

ialarmedalien:
switching over the pydanticgen requires a bit more effort as there's a module that would need to be rewritten to interface with ruff -- hence I left it out of this PR.

Thanks for the review @dalito!

PR #2775, opened 2025-06-17
Index creation

VictoriaSavageNNL:
DRAFT
Allows for the creation of Indexes within the ddl.


ialarmedalien:
On the right-hand side of the page, under the 'Reviewers' section, there's a link you can click to convert this PR to draft status.

linkml/generators/sqltablegen.py
```diff
@@ -335,6 +353,22 @@ def get_id_or_key(cn: str, sv: SchemaView) -> str:
         pk_name = pk.alias if pk.alias else pk.name
         return f"{cn}.{pk_name}"
 
+    @staticmethod
+    def check_duplicate_col_names(cols_list: list, item_name: str) -> Boolean:
```
ialarmedalien: Do you know the type of the entities in `col_list`?

It looks like the return type should be just the regular `bool` -- `Boolean` is the SqlAlchemy type.

VictoriaSavageNNL: There are columns, Foreign Keys, Unique Constraints, and Indexes, it isn't certain the order.

linkml/generators/sqltablegen.py
```diff
@@ -335,6 +358,42 @@ def get_id_or_key(cn: str, sv: SchemaView) -> str:
...
+                annotated_ind = Index(key, *value)
+                index_list.append(annotated_ind)
+        return index_list
+
+    # Scans the schema for duplicate names that exist as a best practice
+    @staticmethod
+    def check_duplicate_entry_names(autogen: list, cols_list: list, item_name: str) -> bool:
+        for entry in cols_list:
+            col_entry_name = getattr(entry, "name", str(entry))
+            if isinstance(entry, Index) or isinstance(entry, UniqueConstraint):
```
ialarmedalien:
```py
if isinstance(entry, (Index, UniqueConstraint)):
```

linkml/transformers/relmodel_transformer.py
```diff
@@ -210,6 +212,7 @@ def transform(self, tgt_schema_name: str = None, top_class: ClassDefinitionName
             for sn in source_sv.all_slots(attributes=False):
                 slot = source_sv.get_slot(sn)
                 # target.slots[slot.name] = copy(slot)
+            # print(c.annotations)
```
ialarmedalien: rm cruft

tests/test_generators/test_sqltablegen.py
```diff
@@ -137,6 +137,30 @@ def test_abstract_class(capsys):
     assert 'CREATE TABLE "inherited class"' in ddl2
 
 
+def test_index_sqlddl(capsys):
```
ialarmedalien: if you aren't testing stdout or stderr, no need to include `capsys` as an argument

ialarmedalien: Please add documentation for this function, with a brief description of the relevant details (i.e. what the indexes are on each table, whether they are PKs/FKs, etc.)

linkml/generators/sqltablegen.py
```diff
@@ -335,6 +358,42 @@ def get_id_or_key(cn: str, sv: SchemaView) -> str:
...
+                    return True
+        if item_name in autogen:
+            msg = (
+                "Warning: proposed item name already exists in schema"
+                "Please generate a new name: "
+                f"{item_name} already exists"
+            )
+            logger.info(msg)
+            return True
+        return False
```
ialarmedalien: Do your tests cover any cases where this function returns True?

If not, you should add extra tests to do so.

linkml/generators/sqltablegen.py
```diff
@@ -220,13 +225,19 @@ def strip_newlines(txt: str | None) -> str:
                         fk = sql_name(self.get_id_or_key(s.range, sv))
                         args = [ForeignKey(fk)]
                     field_type = self.get_sql_range(s, schema)
+                    fk_index_cond = (s.key or s.identifier) and self.autogenerate_fk_index
+                    pk_index_cond = is_pk and self.autogenerate_pk_index
+                    is_index = fk_index_cond or fk_index_cond
```
ialarmedalien:
```
fk_index_cond or fk_index_cond
```
looks like a typo here

linkml/generators/sqltablegen.py
```diff
@@ -220,13 +225,19 @@ def strip_newlines(txt: str | None) -> str:
...
                     col = Column(
                         sql_name(sn),
                         field_type,
                         *args,
                         primary_key=is_pk,
                         nullable=not s.required,
+                        index=(fk_index_cond or pk_index_cond),
```
ialarmedalien: you can use the variable `is_index` (currently incorrect, but will be OK once the typo is fixed) here

linkml/generators/sqltablegen.py
```diff
@@ -239,7 +250,19 @@ def strip_newlines(txt: str | None) -> str:
...
                     sql_uc = UniqueConstraint(*sql_names)
                     cols.append(sql_uc)
+                    autogenerated_item_names.append(sql_names[0])
+                    # Anything that has a unique constraint should have an associated index with it
+                    sql_names[0] = sql_names[0] + "_idx"
+                    is_duplicate = self.check_duplicate_entry_names(autogenerated_item_names, cols, sql_names[0])
+                    if not is_duplicate:
+                        uc_index = Index(*sql_names)
+                        cols.append(uc_index)
+                        autogenerated_item_names.append(sql_names[0])
```
ialarmedalien: It doesn't look like your tests cover any cases where there are `unique_keys` specified, so you should add some more to do so.

linkml/generators/sqltablegen.py
VictoriaSavageNNL: Yes, I will document the test cases accordingly.

VictoriaSavageNNL: Additionally, this function was written because some previously created tests were failing due to duplicate indices being inadvertently created, and this function both remedied the failures and provided a safeguard

linkml/generators/sqltablegen.py
VictoriaSavageNNL: Done!

linkml/generators/sqltablegen.py
VictoriaSavageNNL: Addressed

linkml/generators/sqltablegen.py
```diff
@@ -335,6 +361,42 @@ def get_id_or_key(cn: str, sv: SchemaView) -> str:
...
+                        "Please generate a new name: "
+                        f"{item_name} already exists for {col_entry_name}, with columns {str(exp_list)}"
+                    )
+                    logger.info(msg)
+                    return True
+        if item_name in autogen:
+            msg = (
+                "Warning: proposed item name already exists in schema"
+                "Please generate a new name: "
+                f"{item_name} already exists"
```
ialarmedalien: this message is a bit redundant:

> "Warning: proposed item name already exists in schema Please generate a new name: item_name already exists"

Also remember that if there isn't a line break in the text string, there won't be one in the logger output!

linkml/generators/sqltablegen.py
```diff
@@ -335,6 +361,42 @@ def get_id_or_key(cn: str, sv: SchemaView) -> str:
...
+            if isinstance(entry, (Index, UniqueConstraint)):
+                if item_name == col_entry_name:
+                    exp_list = [getattr(col, "name", str(col)) for col in entry.expressions]
+                    msg = (
+                        "Warning: proposed item name already exists in Table"
+                        "Please generate a new name: "
+                        f"{item_name} already exists for {col_entry_name}, with columns {str(exp_list)}"
+                    )
+                    logger.info(msg)
+                    return True
```
ialarmedalien: it looks like this isn't tested yet, so make sure you have a test case to cover this code path

VictoriaSavageNNL: I realized that block was doing a bit too much and not particularly effective at catching duplicate names, so it has been removed.

tests/test_generators/test_sqltablegen.py
ialarmedalien: You can remove the `capsys` argument if you aren't using it.

VictoriaSavageNNL: this should be addressed now!

linkml/generators/sqltablegen.py
```diff
@@ -239,7 +250,22 @@ def strip_newlines(txt: str | None) -> str:
                         continue
                     sql_uc = UniqueConstraint(*sql_names)
                     cols.append(sql_uc)
+                    # Anything that has a unique constraint should have an associated index with it
+                    uc_index_name = cn
+                    for name in sql_names:
+                        uc_index_name = uc_index_name + "_" + name
+                    uc_index_name = uc_index_name + "_idx"
```
ialarmedalien, line 272: you can do this in one line:

```py
uc_index_name = f"{cn}_{'_'.join(sql_names)}_idx"
```

linkml/generators/sqltablegen.py
```diff
@@ -239,7 +250,22 @@ def strip_newlines(txt: str | None) -> str:
                         continue
                     sql_uc = UniqueConstraint(*sql_names)
                     cols.append(sql_uc)
+                    # Anything that has a unique constraint should have an associated index with it
+                    uc_index_name = cn
```
ialarmedalien: should this be `sql_name(cn)`?

linkml/generators/sqltablegen.py
```diff
@@ -220,13 +225,19 @@ def strip_newlines(txt: str | None) -> str:
...
                     col = Column(
                         sql_name(sn),
                         field_type,
                         *args,
                         primary_key=is_pk,
                         nullable=not s.required,
+                        index=is_index,
                     )
+                    if is_index:
+                        autogenerated_item_names.append("ix_" + cn + "_" + sn)
```
ialarmedalien: Do `cn` and `sn` need to be `sql_name-ised` here?

It's a matter of taste, but I think `(f"ix_{cn}_{sn}")` is a bit more readable (and concise).

linkml/generators/sqltablegen.py
ialarmedalien: or even
```py
uc_index_name = f"{'_'.join([cn, *sql_names, 'idx'])}"
```

linkml/generators/sqltablegen.py
```diff
@@ -335,6 +361,26 @@ def get_id_or_key(cn: str, sv: SchemaView) -> str:
...
+                annotated_ind = Index(key, *value)
+                index_list.append(annotated_ind)
+        return index_list
+
+    # Scans the schema for duplicate names that exist as a best practice
+    @staticmethod
+    def check_duplicate_entry_names(autogen: list, item_name: str) -> bool:
+        if item_name in autogen:
+            msg = f"Warning: {item_name} already exists, please generate a new name"
+            logger.info(msg)
```
ialarmedalien: 1. Should this be logged at the 'warning' level, rather than 'info'?
2. Do you still want to generate the DDL if this name already exists?

linkml/generators/sqltablegen.py
```diff
@@ -335,6 +361,26 @@ def get_id_or_key(cn: str, sv: SchemaView) -> str:
         pk_name = pk.alias if pk.alias else pk.name
         return f"{cn}.{pk_name}"
 
+    # Takes in the existing autogenerated list, the column list, and an annotation to extract the index
+    def parse_annotation_indexes(self, autogen: list, ann) -> list:
+        index_list = []
+        value_dict = {k: annotation for k, annotation in ann.value._items()}
+        for key, value in value_dict.items():
```
ialarmedalien: you can just do 
```py
for key, value in ann.value._items():
```
here

linkml/generators/sqltablegen.py
```diff
@@ -335,6 +361,26 @@ def get_id_or_key(cn: str, sv: SchemaView) -> str:
...
+            name_exists = self.check_duplicate_entry_names(autogen, key)
+            if not name_exists:
+                annotated_ind = Index(key, *value)
+                index_list.append(annotated_ind)
```
ialarmedalien:
```py
if not self.check_duplicate_entry_names(autogen, key):
    index_list.append(Index(key, *value)
```

linkml/generators/sqltablegen.py
```diff
@@ -335,6 +361,26 @@ def get_id_or_key(cn: str, sv: SchemaView) -> str:
...
+        return index_list
```
ialarmedalien: Taking the two previous statements together:

```py
for key, value in ann.value._items():
    if not self.check_duplicate_entry_names(autogen, key):
        index_list.append(Index(key, *value)
```

It's then easy to turn this into a list comprehension:

```py
    index_list = [Index(key, *value) for key, value in ann.value._items() if not self.check_duplicate_entry_names(auto, key)]
```

the function then becomes

```py
def parse_annotation_indexes(self, autogen: list, ann) -> list:
    return [Index(key, *value) for key, value in ann.value._items() if not self.check_duplicate_entry_names(auto, key)]
```

at which point you could just put that line in the parent method instead of having a separate function for it.

linkml/generators/sqltablegen.py
```diff
@@ -239,7 +250,22 @@ def strip_newlines(txt: str | None) -> str:
...
+                    is_duplicate = self.check_duplicate_entry_names(autogenerated_item_names, uc_index_name)
+                    if not is_duplicate:
+                        sql_names = [uc_index_name] + sql_names
+                        uc_index = Index(*sql_names)
+                        cols.append(uc_index)
+                        autogenerated_item_names.append(uc_index_name)
             if not c.abstract or (c.abstract and self.generate_abstract_class_ddl):
+                for tag, annotation in c.annotations.items():
+                    if tag == "index":
+                        index_list = self.parse_annotation_indexes(autogenerated_item_names, annotation)
```
ialarmedalien: If we're only interested in annotations marked as "index", do we need to iterate through all the annotations? How many annotations will there be with the tag "index"?

tests/test_generators/test_sqltablegen.py
```diff
@@ -137,6 +137,50 @@ def test_abstract_class(capsys):
...
+    b = SchemaBuilder()
+    b.add_slot(SlotDefinition("age", range="integer", description="age of person in years"))
+    b.add_slot(SlotDefinition("dummy_foreign_key", range="ClassWithNowt", description="foreign key test"))
+    b.add_slot("identifier_slot", identifier=True)
+    b.add_slot("key_slot", key=True)
```
ialarmedalien: this isn't used anywhere

tests/test_generators/test_sqltablegen.py
```diff
@@ -137,6 +137,50 @@ def test_abstract_class(capsys):
...
+    b.add_class(
+        "ClassWithNowt",
+        slots=["slot_1", "slot_2"],
+        annotations=test_index_dict_3,
+        unique_keys={"unique_keys": slot_1_2_UK},
+    )
+    gen = SQLTableGenerator(b.schema, use_foreign_keys=True)
+    ddl = gen.generate_ddl()
+    # Tests autogeneration of primary key index
+    assert 'CREATE INDEX "ix_dummy class_id" ON "dummy class" (id);' in ddl
```
ialarmedalien, line 169: does this statement play nicely with the generator's rules about how to format the class name?

linkml/generators/sqltablegen.py
VictoriaSavageNNL: it isn't clear how many there will be, and it isn't known if the annotations will be in meaningfully ordered to filter them out

linkml/generators/sqltablegen.py
VictoriaSavageNNL: I found the way that i did this much easier to read and follow along with, plus there's lower risk of the linter calling it out for the lines being too long

linkml/generators/sqltablegen.py
VictoriaSavageNNL: For reference, when columns and unique constraints are appended, they are also constructed before being appended.

for example: 
lines 231-245
```python      
 col = Column(
                        sql_name(sn),
                        field_type,
                        *args,
                        primary_key=is_pk,
                        nullable=not s.required,
                        index=is_index,
                    )
                    if is_index:
                        autogenerated_item_names.append("ix_" + sql_name(cn) + "_" + sn)
                    if include_comments:
                        ddl_str += f"--     * Slot: {sn} Description: {strip_newlines(s.description)}\n"
                    if s.description:
                        col.comment = s.description
                    cols.append(col)
```
Lines 251-252
```python
             sql_uc = UniqueConstraint(*sql_names)
             cols.append(sql_uc)
```

linkml/generators/sqltablegen.py
VictoriaSavageNNL: Is this a blocker for approval?

linkml/generators/sqltablegen.py
VictoriaSavageNNL: it yields the same result because i don't have a policy set

tests/test_generators/test_sqltablegen.py
VictoriaSavageNNL: There's no naming policy specified, so it does

linkml/generators/sqltablegen.py
ialarmedalien: but what if a policy is set?

linkml/generators/sqltablegen.py
ialarmedalien: Is `annotations` a dictionary?

linkml/generators/sqltablegen.py
ialarmedalien: > there's lower risk of the linter calling it out for the lines being too long

If you're using ruff, it will reformat the line to fit

tests/test_generators/test_sqltablegen.py
ialarmedalien: Can you add tests where a naming policy is specified?

linkml/generators/sqltablegen.py
VictoriaSavageNNL: When I did that it didn't work properly, it needs the two step process.

linkml/generators/sqltablegen.py
ialarmedalien: The first example is a little different as you have the `if s.description` statement. The second one could easily be

```py
cols.append(UniqueConstraint(*sql_names))
```

linkml/generators/sqltablegen.py
ialarmedalien: is this version still hard to follow?

```py
for key, value in ann.value._items():
    if not self.check_duplicate_entry_names(autogen, key):
        index_list.append(Index(key, *value)
```

I use a lot of list/set/dict comprehensions so I'm just accustomed to how to understand them, but they can be opaque if you are used to other programming languages or you just don't use them much.

I would say that this:

```py
value_dict = {k: annotation for k, annotation in ann.value._items()}
for key, value in value_dict.items():
```

is actively obfuscatory, as you're creating a copy of a dictionary that you then iterate through, instead of just iterating through the original dictionary.

VictoriaSavageNNL: I had to do it as a two step process, it wouldn't work otherwise because of how the meta model is.

VictoriaSavageNNL: there's a key called index, and then we are accessing the information within the annotations that may be in there since it may be a nested dictionary

VictoriaSavageNNL: This is the only way that i was able to access the information within the annotations correctly

linkml/generators/sqltablegen.py
VictoriaSavageNNL: It is a nested dictionary of sorts, it is passed in as a dictionary with an Annotation as the value, which has the underlying structure of a dictionary, but the data within needs to be extracted

linkml/generators/sqltablegen.py
VictoriaSavageNNL: Due to the irregularity in the double quoting and errors being thrown, this will be revisited in a separate issue

linkml/generators/sqltablegen.py
VictoriaSavageNNL: We still want to generate the DDL, but warn the user

tests/test_generators/test_sqltablegen.py
VictoriaSavageNNL: removed

tests/test_generators/test_sqltablegen.py
VictoriaSavageNNL: The sql name function is causing issues with sometimes generating quotes around table names and sometimes not, this should be looked into but is out of the scope of this pull request

linkml/generators/sqltablegen.py
VictoriaSavageNNL: added sql_name()

linkml/generators/sqltablegen.py
VictoriaSavageNNL: i'm keeping the convention since now the annotation indices have a few conditions attached and i find it more readable

sierra-moxon (approved):

tests/test_generators/test_sqltablegen.py
sierra-moxon: Let's make the separate issue a high priority (maybe we can get it fixed before the next linkml release?).  We can also refine the syntax style in subsequent PRs.

PR #2779, opened 2025-06-21
migrate to uv

iQuxLE:
###  Migrating to uv

addresses #2615 

- [ ] Update CI/CD
- [ ] Change build system (toml)
- [ ] Update Dockerfile
- [ ] Update Makefiles
- [ ] Generate uv.lock and Remove poetry.lock
- [ ] test migration


.github/workflows/main.yaml
```diff
@@ -16,15 +17,16 @@ jobs:
...
+        with:
+          version: ${{ env.UV_VERSION }}
+
       - uses: actions/setup-python@v5.6.0
         with:
           python-version: 3.13
-      - name: Check pyproject.toml and poetry.lock
-        run: poetry check
+      - name: Check pyproject.toml and uv.lock
+        run: uv sync --check
```
dalito: `uv sync --check` checks if all installed packages match the lock file. We probably want `uv lock --check` instead.

.github/workflows/main.yaml
```diff
@@ -67,26 +69,26 @@ jobs:
...
+        with:
+          version: ${{ env.UV_VERSION }}
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v5.6.0
         id: setup-python
         with:
           python-version: ${{ matrix.python-version }}
-          cache: 'poetry'
+          cache: 'uv'
```
dalito: Setting `uv` as option for the cache is not supported as far as I know. Caching should be handled by the `astral-uv/setup-uv` action, see https://github.com/astral-sh/setup-uv?tab=readme-ov-file#enable-caching

This comment is also valid for the other actions.

.github/workflows/test-with-unreleased-runtime.yaml
```diff
@@ -105,41 +105,37 @@ jobs:
...
-          cache: 'poetry'
+          cache: 'uv'
 
       - name: Load cached venv
         id: cached-poetry-dependencies
         uses: actions/cache@v4
         with:
           path: linkml/.venv
-          key: venv-${{ matrix.python-version }}-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}
+          key: venv-${{ matrix.python-version }}-${{ runner.os }}-${{ hashFiles('**/uv.lock') }}
```
dalito: Please check if this caching is still needed if https://github.com/astral-sh/setup-uv?tab=readme-ov-file#enable-caching is used

.github/workflows/test-with-unreleased-runtime.yaml
```diff
@@ -105,41 +105,37 @@ jobs:
...
+        run: uv sync --all-extras
       - name: print linkml-runtime version
         working-directory: linkml
-        run: poetry run python -c 'from importlib.metadata import version; print(version("linkml-runtime"))'
-
+        run: uv run python -c 'from importlib.metadata import version; print(version("linkml-runtime"))'
       - name: run tests
         working-directory: linkml
-        run: poetry run python -m pytest --with-slow
+        run: uv run python -m pytest --with-slow
```
dalito: This could be just `uv run pytest --with-slow`

pyproject.toml
```diff
@@ -299,11 +280,11 @@ requires = ["tox>=4"]
 env_list = ["lint", "py{39,310,311,312,313}"]
```
dalito, line 310: For tox we may want to use https://github.com/tox-dev/tox-uv. It could also be added later in another PR.

pyproject.toml
```diff
@@ -78,7 +78,7 @@ shacl = [
     "pyshacl >= 0.25.0",
 ]
 tests = [
-    # "linkml[black]",  # recursive dependency not supported by poetry
+  # "linkml[black]",  # recursive dependency not supported by uv
```
dalito: Composing dependency groups is possible with uv, see for example https://github.com/hynek/structlog/blob/480ae5f2bcd7e77669bd40b43c329bfbc5f5bd31/pyproject.toml#L63-L66

pyproject.toml
```diff
@@ -333,3 +314,11 @@ commands = [
   ["ruff", "check", "{posargs:.}"],
   ["ruff", "format", "{posargs:.}"],
 ]
+
+[tool.uv-dynamic-versioning]
+enable = true
+vcs = "git"
+style = "pep440"
+
```
dalito, line 353: I think that the line `enable = true` is not needed. At least I don't need it in https://github.com/nfdi4cat/pid4cat-model/blob/07c3792168874c7314e36ba9451ed58f237d160c/pyproject.toml#L52-L56

You may want to add a line `fallback-version = "0.0.0"`

dalito (changes requested):
Thanks for continuing the work on moving to uv.

iQuxLE:
I've also fixed two tests 

**Pydantic Array Test**
(`test_arrays_anyshape_json_schema`)

**_Issue_**:  JSON schema self-reference changed in newer Pydantic, causing a test to fail.
**_Fix_**:  Updated the test to be more flexible, checking for "AnyShapeArray" within the self-reference.

**Deprecation Tests**
(`test_markdowngen_deprecation` and `test_yumlgen_deprecation`)

**_Issue_**:  Deprecation warnings weren't showing in development (version 0.0.0).

_**Root Cause:**_:  Version 0.0.0 comes from the fallback-version in pyproject.toml when no git tags exist in this fork, and deprecation logic only triggers when current version ≥ deprecated_in version.

**_Fix_**:  Added skip conditions for these tests when running in a 0.0.0 version environment.

dalito:
I don't feel very comfortable to hide the warnings/errors for version 0.0.0.  To "fix" the local deprecation warnings, it should be sufficient to fetch the tags from remote (upstream) 
```bash
git remote add upstream https://github.com/linkml/linkml
git fetch upstream --tags
```
We could add a note to the docs on the importance of tags for calculating the dynamic version.

iQuxLE:
I see your point.
Don't you think this overcomplicates things for community devs?

But I would be ok with adding a clear note in the test output if it fails.

dalito (approved):
Although I would not add it, I don't have a very strong opinion on skipping the version 0.0.0 case. Maybe someone else from the team can also comment.

This looks really nice now! The only thing that you could still do from my point of view is to add a note on fetching the tags to contributing.md. But I'll approve as is.

sierra-moxon (approved):

PR #2796, opened 2025-07-03
Clean-up `emit_metadata` as alias for `metadata`

### LINKED ISSUE #2795: Clean-up `emit_metadata` as alias for `metadata`
We have in [some of the generators](https://github.com/linkml/linkml/blob/d2bd7fa267bd3e645941bea4457accf742e693d2/linkml/generators/jsonldcontextgen.py#L46) the field `emit_metadata` as an alias, apparently, for the [`Generator` field `metadata`](https://github.com/linkml/linkml/blob/d2bd7fa267bd3e645941bea4457accf742e693d2/linkml/utils/generator.py#L128). [This to-do](https://github.com/linkml/linkml/blob/d2bd7fa267bd3e645941bea4457accf742e693d2/linkml/utils/generator.py#L205): `remove aliasing` seems to confirm it.

BTW, no only the generators are affected. [Also](https://github.com/linkml/linkml/blob/d2bd7fa267bd3e645941bea4457accf742e693d2/linkml/utils/schemaloader.py#L46) `SchemaLoader` [uses `emit_metadata` instead of `metadata`](https://github.com/linkml/linkml/blob/d2bd7fa267bd3e645941bea4457accf742e693d2/linkml/utils/schemaloader.py#L46). If I get it right, we would simply need to consistently replace `emit_metadata` with `metadata` everywhere... Or do I oversee something?

### LINKED ISSUE #2798: Python generator ignoring `emit_metadata` flag
**Describe the bug**

[The `PythonGenerator` class](https://github.com/linkml/linkml/blob/d2bd7fa267bd3e645941bea4457accf742e693d2/linkml/generators/jsonldcontextgen.py#L31) is ignoring the value of `emit_metadata`.

**Version of LinkML you are using** v1.9.1

[`test_issue_344`](https://github.com/linkml/linkml/blob/d2bd7fa267bd3e645941bea4457accf742e693d2/tests/test_issues/test_issue_344.py#L10) is instantiating `PythonGenerator` with `emit_metadata=False`, but this test is passing although [the snapshot being used for verification has metadata](https://github.com/linkml/linkml/blob/d2bd7fa267bd3e645941bea4457accf742e693d2/tests/test_issues/__snapshots__/issue_344_context.json#L2)!

But, why? Because the value of that flag is being ignored...

### LINKED ISSUE #2799: Python generator ignoring `emit_metadata` flag
**Describe the bug**

[The `PythonGenerator` class](https://github.com/linkml/linkml/blob/d2bd7fa267bd3e645941bea4457accf742e693d2/linkml/generators/pythongen.py#L41) is ignoring the value of `emit_metadata`.

**Version of LinkML you are using** v1.9.1

[`test_issue_344`](https://github.com/linkml/linkml/blob/d2bd7fa267bd3e645941bea4457accf742e693d2/tests/test_issues/test_issue_167.py#L27) is instantiating `PythonGenerator` with `emit_metadata=False`, but this test is passing although [the snapshot being used for verification has metadata](https://github.com/linkml/linkml/blob/d2bd7fa267bd3e645941bea4457accf742e693d2/tests/test_issues/__snapshots__/issue_167b.py#L1)!

But, why? Because the value of that flag is being ignored...

Silvanoc:
Probably due to historical reasons some generators were using `emit_metadata` as an alias for `metadata`. This PR removes that alias.

Due to bug #2799, the argument `emit_metadata` of `PythonGenerator` did not have any effect. Since the CLI flag `--(no-)head` sets that argument, that flag does not have any effect either. Therefore CLI flag `--(no-)head` is not needed and is removed in this PR too.

<details>
<summary>Click here to see the detailed explanation</summary>

[SchemaLoader](https://github.com/linkml/linkml/blob/main/linkml/utils/schemaloader.py) uses only `emit_metadata` with [`True` as default](https://github.com/linkml/linkml/blob/3e7a7845e96a501ebff8abd35813db38ed866399/linkml/utils/schemaloader.py#L46).

[`generator.py`](https://github.com/linkml/linkml/blob/main/linkml/utils/generator.py) uses mostly `metadata` with [`True` as default](https://github.com/linkml/linkml/blob/3e7a7845e96a501ebff8abd35813db38ed866399/linkml/utils/generator.py#L128), but then [switches to `emit_metadata` to invoke `SchemaLoader`](https://github.com/linkml/linkml/blob/3e7a7845e96a501ebff8abd35813db38ed866399/linkml/utils/generator.py#L251). The generators inherit this behavior, but start mixing up `emit_metadata`, `metadata` and, in some cases, even `head`.

For example, [pythongen.py](https://github.com/linkml/linkml/blob/main/linkml/generators/pythongen.py) uses [`head` with `False` as CLI default](https://github.com/linkml/linkml/blob/main/linkml/generators/pythongen.py) with the very same meaning. It then [passes it as `emit_metadata`](https://github.com/linkml/linkml/blob/3e7a7845e96a501ebff8abd35813db38ed866399/linkml/generators/pythongen.py#L1303), but uses [`True` as default on the Dataclass initialization](https://github.com/linkml/linkml/blob/3e7a7845e96a501ebff8abd35813db38ed866399/linkml/generators/pythongen.py#L60). But remember that `generator.py` uses `metadata` on the public interface, but exposes also `emit_metadata` publicly when mapping `metadata`. So in the end, due to all the changes between `head`, `metadata` and `emit_metadata` the default behavior is hard to predict and changes depending how the Python generator is being invoked, although the same default behavior can be expected.

</details>

Fixes #2795
Fixes #2798 
Fixes #2799


Silvanoc:
This PR changes code created a very long time ago. Therefore assigning @cmungall for review, because he was involved in that code.

ialarmedalien:
What is the reason for changing the current behaviour in a couple of the generators (jsonld, python, etc.)?

Silvanoc:
> What is the reason for changing the current behaviour in a couple of the generators (jsonld, python, etc.)?

@ialarmedalien as mentioned in the description, issue #2799 shows that there is a mixture of `metadata` and `emit_metadata` for the same purpose (at least I could not find any reason to have both, I think it simply has historical reasons). That worsen the understandability of the code and is error prone. This PR simply unifies them throughout the code base and by doing that also fixes issue #2799.

ialarmedalien:
I'm talking specifically about the generators where the default setting for metadata has changed from `True` to `False`. You can see in the `__snapshots__` files that the metadata section has been removed. Is there a reason for changing the behaviour?

Silvanoc:
> I'm talking specifically about the generators where the default setting for metadata has changed from `True` to `False`. You can see in the `__snapshots__` files that the metadata section has been removed. Is there a reason for changing the behaviour?

@ialarmedalien The existence of both `metadata` and `emit_medata` was provoking that the specified value for getting metadata embedded in the generated file was being ignored. As a consequence generated snapshots were capturing the erroneous behavior as correct. Fixing the behavior requires fixing the snapshots accordingly.

Notice that the JSON-LD and JSON-LD Context generators are specifying that the context should not have any metadata (see [here for `jsonldgen.py`](https://github.com/linkml/linkml/blob/3e7a7845e96a501ebff8abd35813db38ed866399/linkml/generators/jsonldgen.py#L167) and [here for `jsonldcontextgen.py`](https://github.com/linkml/linkml/blob/3e7a7845e96a501ebff8abd35813db38ed866399/linkml/generators/jsonldcontextgen.py#L46)), but the snapshots were capture with the failing behavior and therefore have metadata.

The Python and Pydantic Generators are even worse because they additionally have `head` for exactly the same purpose! In this case the default does not change (produce metadata header). Some of the tests were requiring not to generate metadata, but due to the erroneous behavior this configuration is being ignored. Observe, for example, that `test_issue_167b_python` [requests no metadata](https://github.com/linkml/linkml/blob/3e7a7845e96a501ebff8abd35813db38ed866399/tests/test_issues/test_issue_167.py#L27), but due to the bug hereby being fixed [the snapshot](https://github.com/linkml/linkml/blob/3e7a7845e96a501ebff8abd35813db38ed866399/tests/test_issues/__snapshots__/issue_167b.py#L1) do have a metadata header. Fixing the bug requires also fixing the erroneous snapshots.

I have to admit, though, that I haven't validated that every single modified snapshot is correct 😓 I haven't fixed testing either to have better coverage of all the different options to specify the metadata generation. That would be probably a meaningful addition.

As @sneakers-the-rat has already pointed out a couple of times, snapshooting has the major drawback that you don't describe the expected behavior and validate it, but rather capture something under the assumption that it's correct (although it might be wrong on the areas not being validated).

Silvanoc:
@ialarmedalien I've added to the description a "collapsible" detailed explanation of the inconsistent behavior being fixed.

ialarmedalien:
@Silvanoc Thank you for the in-depth explanation -- that makes perfect sense.

ialarmedalien (approved):
LGTM!

Silvanoc:
I've right now realized something, that I'm breaking the public interface without any kind of announcement or similar! 🤦 

I'm [removing one option from the PythonGen CLI](https://github.com/linkml/linkml/pull/2796/files#diff-baca17b6d0f9403fdc99778a5fb1eb858b910da0fddd4e77dd9d9ce2d8aff214L1265) and additionally removing the fields `head` and `emit_metadata` from multiple generator Dataclasses.

Fixing the removal of the PythonGen CLI option should be easy adding the option back and rewiring it to `metadata`, dropping a deprecation warning. I don't know how to fix the DataClass fields though. @ialarmedalien any idea?

Possibly I should have used the module [utils/deprecation.py](https://github.com/linkml/linkml/blob/main/linkml/utils/deprecation.py).

PR #2801, opened 2025-07-05
Replace NMDC schema with a tiny file

vladistan:


ialarmedalien:
Suggestions:

- Move this test to `test/test_utils/test_schema_fixer.py` so that all tests that apply to the SchemaFixer class are in one place.
- Optional: the example schema is still very large and complicated (although far, far better than before!). Ideally you would step through the function under test, understand what transformation is taking place, and then pare down the input data to the very minimum required -- take out unneeded fields, slots, classes, etc., so that you only have the parts of the schema that are actively being manipulated by the code.

ialarmedalien (approved):
Looks great!

PR #2803, opened 2025-07-10
2776  Pandera support for inlined slots

tfliss:
This PR implements most of the inlining compliance tests for the Pandera generator (except for foreign key associations). Although there is partial support for inlining as dictionary/simple-dict I'm considering this not truly supported. 

[The documentation page](https://github.com/linkml/linkml/blob/e68481099493855e7551530e77db3fd18832d2ab/docs/generators/pandera.rst) reflects the new functionality and next steps.
[This collab notebook gist](https://gist.github.com/tfliss/a42b096235d00ceba4ee4dfd26490156) demonstrates the inlined validation, similar to the unit tests.

Please let me know if you want to be a reviewer, especially if you have use cases or can recommend documentation improvements.

A key point in this feature is inlining as dictionary and inlining as simple dictionary forms (see https://linkml.io/linkml/schemas/inlining.html) are inefficient in PolaRS (though inlining structs that represent nested models rather than collections are fine). This implementation does a transform at validation time. In the near future this needs to be better separated out into a model transformer that can be called from a loader or other situations, and only inlined-as-list will be supported by default at validation time. The simple-dict compliance tests are skipped more because of challenges with loading the test fixtures from JSON into PolaRS than with the Pandera validator proper.

Because of the additional code in the validator that handles nesting with Pandera custom checks, the runtime code is no longer inlined in the generated Pandera model. This creates a runtime dependency on LinkML that I'd like to remove in the future.

Much of the core functionality uses PolaRS expressions The user guide is helpful: https://docs.pola.rs/user-guide/concepts/  Nesting in particular relies on Pandera custom check methods: https://pandera.readthedocs.io/en/stable/extensions.html#registering-custom-check-methods
In the future it may be better to use the Narwhals  API wrapper which generalizes other dataframe expression libraries.
Note that Pandera, PolaRS and Narwhals are configured as extras / optional dependencies for LinkML.

There is a [colab notebook gist](https://gist.github.com/tfliss/a42b096235d00ceba4ee4dfd26490156) that demonstrates much of the functionality. It is closely based on the test_panderagen.py module.


Silvanoc:
@tfliss I have no idea about Pandera, therefore I can look at other aspects of your patch, but not correctness WRT Pandera.

docs/generators/pandera.rst
```diff
@@ -16,118 +16,234 @@ The Pandera Generator produces Pandera models using the class-based API
...
+the first priority is implementing models of literal and nested data types and checks for single tables as shown below.
 `tests/test\_generators/test\_panderagen.py` also has an example using supported LinkML features.
 
-Currently supported features are:
+Currently supported LinkML features are:
 
 - literal slot ranges: string, integer, float, boolean, date, datetime
 - enums
 - constraints: required, pattern, minimum_value, maximum_value, multivalued
+- inlining: nested non-multivalued objects, lists of literals, lists of objects
```
Silvanoc: What does `nested non-multivalued objects` mean? `single object`?

docs/generators/pandera.rst
```diff
@@ -16,118 +16,234 @@ The Pandera Generator produces Pandera models using the class-based API
...
-
-        ontology_enum_column: Enum= pla.Field(dtype_kwargs={"categories":('fiction','non fiction',)})
-        """test enum column with ontology values"""
-
-        multivalued_column: List[int]= pla.Field()
-        """one-to-many form"""
+  class PanderaSyntheticTable(pla.DataFrameModel, _LinkmlPanderaValidator):
+      """
+      A flat table with a reasonably complete assortment of datatypes.
+      """
```
Silvanoc: In future PRs, please try to put these kind of code-formatting changes into ["style commits"](https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/). You don't need to use conventional commit format, but it illustrates what I mean.

If I see it right, all changes in this "diff hunk" until the `multivalued_column` definition are only code-formatting changes. Knowing that a code-formatter has done the changes simplifies the review.

If you are running a code formatter that is changing existing files (apart from those changed in this PR) and you think those changes make sense, I think I'd be meaningful having a PR just to "beautify" those files.

docs/generators/pandera.rst
```diff
@@ -16,118 +16,234 @@ The Pandera Generator produces Pandera models using the class-based API
...
+      test column with another class inlined as a list
+      """
+
+      inlined_simple_dict_column: Struct = pla.Field()
+      """
+      test column inlined using simple dict form
+      """
+
+      @pla.check("cardinality_column")
+      def check_cardinality_cardinality_column(cls, data: PolarsData):
```
Silvanoc: This name looks weird due to the double `_cardinality`. Looking at the code seems to be correct and if I get it right, it's a function to check the cardinality of a "cardinality" column, right?

tests/test_generators/test_panderagen.py
```diff
@@ -21,6 +22,30 @@ def cli_runner():
...
+@pytest.fixture(scope="module")
+def pl():
+    """The PolaRS package is optional, so use fixtures and importorskip to only run tests when it's installed"""
+    return pytest.importorskip("polars", minversion="1.0", reason="Polars >= 1.0 not installed")
+
+
+@pytest.fixture(scope="module")
+def pandera():
+    """The pandera package is optional, so use fixtures and importorskip to only run tests when it's installed"""
+    return pytest.importorskip("pandera.polars", reason="Pandera not installed")
```
Silvanoc: Does it make sense to run this test when `pandera` package is not installed?

tests/test_generators/test_panderagen.py
```diff
@@ -36,59 +61,132 @@ def synthetic_flat_dataframe_model():
...
-        multivalued: True
-        inlined_as_list: True
+        required: true
+        multivalued: true
+        inlined_as_list: true
+      # multivalued_one_many_column:
+      #   description: list form
+      #   range: integer
+      #   required: true
+      #   multivalued: true
```
Silvanoc: Why are these commented-out lines here? Are they needed?

linkml/generators/panderagen/panderagen.py
```diff
@@ -28,6 +30,7 @@
     "panderagen_class_based": {
```
Silvanoc, line 29: Wouldn't it make sense to merge this data types equivalence with that [available in linkml-runtime](https://github.com/linkml/linkml-runtime/blob/faac277936228d65318966ac5f5f3974a91566b5/linkml_runtime/processing/referencevalidator.py#L60) into a somehow dedicated module? Of course, not as part of this PR, but possibly as a follow-up issue.

linkml/generators/panderagen/panderagen.py
```diff
@@ -156,18 +229,25 @@ def render(self) -> OODocument:
...
                 source_class=c,
+                annotations=annotations,
             )
             classes.append(ooclass)
             if c.mixin:
                 ooclass.mixin = c.mixin
             if c.mixins:
                 ooclass.mixins = [(x) for x in c.mixins]
+            # if c.abstract:
+            #    ooclass.abstract = c.abstraccamelcased
```
Silvanoc: Can it be clean-up? Or is there a reason for keeping it?

linkml/generators/panderagen/panderagen.py
```diff
@@ -184,6 +264,7 @@ def render(self) -> OODocument:
         return oodoc
 
 
+# @shared_arguments(PanderaGenerator)
```
Silvanoc: Can it be clean-up? Or is there a reason for keeping it?

Silvanoc (commented):
Huge and very complex PR. I lack knowledge about Pandera to be able to assess the functionality. Additionally many code-formatting changes that make the review harder.

I've had a superficial look at it and mostly raised only nitpicks that should not block it from being merged. But I haven't been able to dig deep enough to give my approval, I might try to dig deeper at a later point if time, if I get to it. @tfliss feel free to "Resolve conversation" for those comments of mine whenever you consider them addressed (either discarded or accepted).

tfliss:
@Silvanoc much appreciated review comments! Many thanks for taking on such a large PR. I agree with most/all of your comments and can apply them in a day or two. It has been an interesting challenge navigating the interactions between inlinining / multivalued / and dataframe vs. json design.

Silvanoc:
> It has been an interesting challenge navigating the interactions between inlinining / multivalued / and dataframe vs. json design.

@tfliss I'm also currently having fun with multivalued, and inlining:
- https://github.com/linkml/linkml/pull/2807#issuecomment-3073470967
- https://github.com/linkml/linkml/issues/2813

docs/generators/pandera.rst
```diff
@@ -16,118 +16,234 @@ The Pandera Generator produces Pandera models using the class-based API
...
+        identifier_column:
+          description: identifier
+          identifier: true
+          range: integer
+          required: true
+        bool_column:
+          description: test boolean column
+          range: boolean
+          required: true
+          integer_column:
```
ialarmedalien: looks like this lost its nesting

tests/test_generators/test_panderagen.py
tfliss: I agree it doesn't make sense to run the test when the package isn't installed. This prevents the pytest collector from erroring out on missing imports when scanning the test directories for which tests to run.

Thank you for this reminder. I had intended to include this in a brief section of testing notes to the panderagen documentation. The next time I touch the file I'll also try using global importerskip calls. The fixture wrapper is a remnant of my earlier solution and may no longer be needed.

https://docs.pytest.org/en/latest/how-to/skipping.html#skipping-on-a-missing-import-dependency

```
# at top of file in place of import pandera
pandera = pytest.importorskip("pandera.polars", reason="Pandera not installed")
```

docs/generators/pandera.rst
tfliss: Yes, the LinkML documentation also uses the non-multivalued phrasing, but I'll change this to "single-valued objects", which is both clearer and appears to be the preferred terminology. Thank you.

https://linkml.io/linkml/schemas/inlining.html#inlining-a-single-valued-object

docs/generators/pandera.rst
tfliss: Thanks for pointing this out. I'll be touching this file for your other comments, and should be able to revert the style changes to make the functionality changes clear.

The changes in this specific file came from a cut/paste of formatted code rather than a direct formatter run. I'll review the rest of this PR and see if formatting noise is more widespread. I agree formatting updates should be isolated.

docs/generators/pandera.rst
tfliss: Thanks. I'm actually going to remove the cardinality functionality to a separate PR. It is distinct functionality from inlining.

To address your comment, this is a generated pandera custom check method, of the form `check_cardinality_{{column_name}}`. As you point out, the slot name `cardinality_column` in the synthetic test model results in the doubling. This is a good reminder to both document the custom checks better and change the column name to improve the example in the future PR.

tests/test_generators/test_panderagen.py
tfliss: Thanks, I will remove them.

linkml/generators/panderagen/panderagen.py
tfliss: I'll remove the lines for this PR and add a note in my roadmap around abstract handling.

docs/generators/pandera.rst
Silvanoc: If the name makes sense for whatever reason, adding a comment with something like "the name looks weird, but is right" suffice 😄

linkml/generators/panderagen/panderagen.py
tfliss: Yes, going forward I plan to both modularize this functionality and align with or reuse existing tables like the reference validator. Thank you for pointing it out; I was not familiar with it. For now, I will remove the dictionary entries that aren't specific to the class-based pandera generator for a future PR.

I've made recent progress in a local branch on next-steps to promote PolaRS schema generation (for tests and data tables to be validated) to a first-class target of this dataframe generator. I'll open a feature request issue soon (and comment here with the issue number) to capture that work. The design decisions and implementation are not finalized but generally align with what you suggest. Your feedback on the direction for this is very helpful.

tfliss:
@cosmicbboy I chatted with you a while back about nesting Pandera models.
This PR implements much of the LinkML recursive inlined nesting functionality using Pandera custom checks to recursively call validate. The description links to a colab notebook with an example.

I'm planning to further refine the technique and documentation, but the technique should be useful in its current form.

cosmicbboy:
Nice! Can you share links to the notebook and docs?

tfliss:
> Nice! Can you share links to the notebook and docs?

@cosmicBboy Yes, the colab notebook: https://gist.github.com/tfliss/a42b096235d00ceba4ee4dfd26490156
and doc page:
https://github.com/linkml/linkml/blob/2776-inlining-compliance-for-pandera-generator/docs/generators/pandera.rst

Both cover similar material.  The `inlined_class_column` and `inlined_as_list_column` are nested models. `inlined_as_simple_dict_column` is mostly a convenience form for JSON that is not well-supported by dataframe libraries.

I am still documenting the internals. In the generated Pandera code the nested models are defined as Structs with an empty Field().  A custom check then calls into a helper method from a validator mixin `_LinkmlPanderaValidator`.
```
class PanderaSyntheticTable(pla.DataFrameModel, _LinkmlPanderaValidator):
...
    inlined_class_column: Struct = pla.Field()
...
    @pla.check("inlined_class_column")
    def check_nested_struct_inlined_class_column(cls, data: PolarsData):
        return cls._check_collection_struct(data)
```

Depending on nesting form, the helper method unpacks a column using [PolaRS expressions](https://docs.pola.rs/user-guide/expressions/), which are generally efficient for nested single-valued objects and lists. At present this is done at validation time, but in the future the transform will be optionally performed at load time.

The snippet below shows the handling of nested lists of objects. The unnested column resulting from whatever form is then sent to the `validate()` from the nested pandera model. The nested class is currently stored in the _NESTED_RANGES array.

https://github.com/linkml/linkml/blob/155748cc27d8cb01e39f6e1c3a2059acd91cf340/linkml/generators/panderagen/transforms/list_dict_model_transform.py#L27C1-L33C10
```
  unnested_column = (
      df
      .select(column_name)
      .filter(pl.col(column_name).list.len() > 0) # see: https://github.com/pola-rs/polars/issues/14381
      .explode(column_name)
      .unnest(column_name)
        )
```

Thank you for taking the time to look at this! If you have recommendations about this approach or documentation I'm glad to work on it and the the linkml team I think would be interested as well.

tfliss:
@amc-corey-cox I'm back from my trip and completed the merge from main and fixing the issues from Silvano's review like we talked about. Are you able to recommend next steps or reviewers (specifically with inlining, generator, or dataframe interest) for this PR? Many thanks!

linkml/generators/panderagen/panderagen.py
tfliss: @Silvanoc this is the PolaRS feature that is driving more modularization in this generator. https://github.com/linkml/linkml/issues/2856

The fork that I am working in is here: https://github.com/tfliss/linkml/tree/polars_schema

Additional work on reorganizing tables like the type map and inlined form table I'd like to do in that fork or smaller sub-features broken out from it.

tfliss:
Thanks for the merge from main @iQuxLE

linkml/generators/panderagen/dataframe_field.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+
+    def inline_form(self):
+        return self.source_slot.annotations._get("inline_form", None)
+
+    def reference_class(self):
+        try:
+            return self.source_slot.annotations._get("reference_class", None)
+        except Exception:
+            return None
```
ialarmedalien: you could use a context manager to suppress the exception here -- I find it marginally nicer than try/except but YMMV.

ialarmedalien:
@tfliss are you happy with merging this now?

tfliss:
> @tfliss are you happy with merging this now?

@ialarmedalien yes, please.  The inlining functionality is usable and it will make it easier to start staging the next work increment as smaller PRs. Thanks!

PR #2807, opened 2025-07-11
Set container in json ld context for multi valued attributes

### LINKED ISSUE #181: JSON-LD Context generator needs container specification
The JSON-LD context generator needs to emit `"@container": "@set"` for any multi-valued attribute

Silvanoc:
Fixes #181


Silvanoc:
@alexbrdn could you give this fix a try?

I've added the flag `--fix-multivalue-containers` to enable the new context generation behavior, by default disabled (therefore `--no-fix-multivalue-containers`). I've decided to go for a default that doesn't break the current behavior. We can then in a 2nd step change the default, if it is meaningful.

alexbrdn:
@Silvanoc, the fix doesn't quite work as I expected and I don't know why.
I have (among others) these two slots in my linkml yamls:
```yaml
 hasRoomSegment:
    multivalued: true
    range: RoomSegment
    description: room segment
    slot_uri: bxlo:hasRoomSegment
 hasTag:
    multivalued: true
    range: Tag
    slot_uri: bx:hasTag
```
And the resulting generated json-ld context has these entries:
```json
      "hasTag": {
        "@container": "@index",
        "@id": "bx:hasTag",
        "@type": "@id"
      },
      "hasRoomSegment": {
        "@container": "@set",
        "@id": "bxlo:hasRoomSegment",
        "@type": "@id"
      },
```

I can't figure out why "hasTag" has `@index`. I've expected `@set`.

Silvanoc:
> @Silvanoc, the fix doesn't quite work as I expected and I don't know why. I have (among others) these two slots in my linkml yamls:
> 
> [...]
> 
> I can't figure out why "hasTag" has `@index`. I've expected `@set`.

@alexbrdn would it be possible for you to extend/modify the [data being used for testing](https://github.com/linkml/linkml/pull/2807/files#diff-72d0ef9d9a98c86760e2f4fa7f172bb06b297361bc2d2774d08e41090f20dc56) to reproduce the issue that you are facing?

Silvanoc:
> @Silvanoc, the fix doesn't quite work as I expected and I don't know why. I have (among others) these two slots in my linkml yamls:
> 
> [...]
> 
> I can't figure out why "hasTag" has `@index`. I've expected `@set`.

After getting a more complete schema from @alexbrdn to reproduce the issue and some analysis, I now know the root-cause of this unexpected behavior.

Although `inlined` is not specified, the fact that the referenced objects (`RoomSegment` and `Tag`) don't have an [identifier](https://linkml.io/linkml-model/latest/docs/identifier/) results in implicit inlining (see last line [here](https://linkml.io/linkml/schemas/inlining.html#no-inlining-reference-by-key)). Although the [default of `inlining_as_list` is undefined](https://github.com/linkml/linkml/issues/2813), it results in a `None` that is being interpreted as `False` and therefore results effectively in `inlined: True` and `inlined_as_list: False` (meaning an implicit dictionary that requires an `@index` container).

alexbrdn:
> After getting a more complete schema from @alexbrdn to reproduce the issue and some analysis, I now know the root-cause of this unexpected behavior.
> 
> Although `inlined` is not specified, the fact that the referenced objects (`RoomSegment` and `Tag`) don't have an [identifier](https://linkml.io/linkml-model/latest/docs/identifier/) results in implicit inlining (see last line [here](https://linkml.io/linkml/schemas/inlining.html#no-inlining-reference-by-key)). Although the [default of `inlining_as_list` is undefined](https://github.com/linkml/linkml/issues/2813), it results in a `None` that is being interpreted as `False` and therefore results effectively in `inlined: True` and `inlined_as_list: False` (meaning an implicit dictionary that requires an `@index` container).

Thanks for the explanation. I'm going to need to read up some more from the linkml spec and the JSON-LD spec and think through this thoroughly.

alexbrdn:
In https://linkml.io/linkml/schemas/inlining.html#no-inlining-reference-by-key, it says
> Note that if the range class does not declare an [identifier](https://w3id.org/linkml/identifier) then inlined is always true

A bit further down, in https://linkml.io/linkml/schemas/inlining.html#inlining-as-a-dictionary, it says
> Note this is only possible if the range of the slot is a class that has an identifier slot

So, it looks to me that if a range class doesn't have an identifier, it must get implicitly both `inlined=True` and `inlined_as_list=True`

Silvanoc:
> So, it looks to me that if a range class doesn't have an identifier, it must get implicitly both `inlined=True` and `inlined_as_list=True`

I agree on that. I've extended issue #2813 to fix this and to remark that is unclear what should happen if `inlined: True`, an [identifier](https://linkml.io/linkml-model/latest/docs/identifier/) has been specified and no value has been specified for `inlined_as_list`. I suppose that's as of now the only unspecified scenario.

Silvanoc:
@alexbrdn would you agree that this implementation is correct and that the issue is with the default for `inlined_as_list`? I would then extend the tests to have an `xfail` for the defaults.

alexbrdn:
Yes, I would guess that, provided a fix for the default `inlined_as_list`, this implementation will work correctly. As it stands right now, it's a bit hard to ascertain

alexbrdn:
FYI: this fix, together with https://github.com/linkml/linkml/pull/2820, lets me generate a JSON-LD context just as I expect it.

nlharris:
Can @Silvanoc resolve the conflicts or does one of the core linkml devs need to do that?

Silvanoc:
> Can @Silvanoc resolve the conflicts or does one of the core linkml devs need to do that?

@nlharris I'll take care of it.

Silvanoc:
@nlharris @amc-corey-cox conflicts resolution done and tests are passing successfully. Ready for merging, I'd say.

amc-corey-cox:
@Silvanoc I'm sorry I missed the window when this was passing tests and we could get it merged in. As it is now, I see what looks like a single simple test failing. Do you think you could patch that and I'll get it in? I could probably take it across the line but this seems central to what you were working on and I don't want to do something you didn't intend by accident.

PR #2808, opened 2025-07-14
style: clean-up whitespaces

Silvanoc:
Improve code-format removing unneeded whitespaces (blanks, newlines, tabs,...).

### DIFF
--- a/notebooks/tmp/personinfo.ddl.sql
+++ b/notebooks/tmp/personinfo.ddl.sql
@@ -1,266 +1,266 @@
 -- # Class: "NamedThing" Description: "A generic grouping for any identifiable entity"
---     * Slot: id Description: 
---     * Slot: name Description: 
---     * Slot: description Description: 
---     * Slot: image Description: 
+--     * Slot: id Description:
+--     * Slot: name Description:
+--     * Slot: description Description:
+--     * Slot: image Description:
 -- # Class: "Person" Description: "A person (alive, dead, undead, or fictional)."
---     * Slot: primary_email Description: 
---     * Slot: birth_date Description: 
---     * Slot: age_in_years Description: 
---     * Slot: gender Description: 
---     * Slot: id Description: 
---     * Slot: name Description: 
---     * Slot: description Description: 
---     * Slot: image Description: 
+--     * Slot: primary_email Description:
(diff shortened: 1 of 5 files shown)


notebooks/tmp/personinfo.ddl.sql
vladistan, line 1: Why do we check this is to begin with?

tests/test_notebooks/output/inheritance.txt
vladistan, line 1: Why is the output of the notebook is checked in into the repo?

tests/test_biolink_model/input/biolink-model.yaml
vladistan, line 1: Is this a local fix or did we fix biolink upstream an vendored it here?

notebooks/tmp/personinfo.py
vladistan, line 1: I think the right way to do this is to fix the templates / generators

tests/test_biolink_model/input/biolink-model.yaml
vladistan: Also, you can't just update biolink,  you have to fix the snapshots too

notebooks/tmp/personinfo.ddl.sql
Silvanoc: No idea, I had the hope that I could at least do some dumb code-formatting clean-up without activating the brain 😄 That's why it's still in draft mode. After it has failed I have realized that I'll need to boot-up my brain.

notebooks/tmp/personinfo.py
Silvanoc: Same as above, just a try to run `ruff check --fix` and merge the result. If this is a generated file, then you are right.

Silvanoc:
@vladistan I've created this PR after running `ruff check --fix` with the hope I could fix some code-formatting issues that are appearing in other PRs. I hoped to get it done without thinking much about it, that's why it's only a draft.

vladistan:
Yeah,  I tried this about a month ago, this resulted in mega giant PR that was impossible to review.   We have been chipping away at it with @sierra-moxon for the last few weeks.  This is why your change set much smaller.

ialarmedalien:
Is this superseded by @vladistan's set of PRs and can it thus be closed?

PR #2811, opened 2025-07-15
Use Pathlib consistently when calling generators

vladistan:


kevinschaper (approved):
Much nicer!

linkml/generators/projectgen.py
```diff
@@ -102,11 +102,12 @@ class ProjectGenerator:
...
         if config.directory is None:
             raise Exception("Must pass directory")
-        Path(config.directory).mkdir(parents=True, exist_ok=True)
+        output_dir = Path(config.directory)
+        output_dir.mkdir(parents=True, exist_ok=True)
         if config.mergeimports:
             all_schemas = [schema_path]
         else:
-            all_schemas = get_local_imports(schema_path, os.path.dirname(schema_path))
+            all_schemas = get_local_imports(schema_path, str(Path(schema_path).parent))
```
dalito: Instead of passing a string to `get_local_imports`, maybe `get_local_imports` can also be changed to accept path-objects? (I did not check if the function is used elsewhere.)

Silvanoc (approved):

linkml/generators/projectgen.py
vladistan: Yes.  `get_local_imports` is only used by the projectgen.  I converted it to use `pathlib` as well

PR #2815, opened 2025-07-16
Fix json schema generator

vladistan:


linkml/generators/jsonschemagen.py
```diff
@@ -694,8 +694,8 @@ def serialize(self, **kwargs) -> str:
         if self.materialize_patterns:
             logger.info("Materializing patterns in the schema before serialization")
             self.schemaview.materialize_patterns()
-
-        return self.generate().to_json(sort_keys=True, indent=self.indent if self.indent > 0 else None)
+        result = self.generate().to_json(sort_keys=True, indent=self.indent if self.indent > 0 else None)
+        return result.rstrip("\n") + "\n"
```
ialarmedalien: Are there other whitespace characters at the end of the string that we might want to keep?

ialarmedalien (approved):

PR #2816, opened 2025-07-16
Fix SQL Schema generator to create linter  compliant files

vladistan:


examples/PersonSchema/personinfo/sqlschema/personinfo.sql
```diff
@@ -1,107 +1,107 @@
...
 --     * Slot: Person_id Description: Autocreated FK slot
---     * Slot: diagnosis_id Description:
---     * Slot: procedure_id Description:
+--     * Slot: diagnosis_id
+--     * Slot: procedure_id
 -- # Class: "WithLocation" Description: ""
---     * Slot: id Description:
---     * Slot: in_location Description:
+--     * Slot: id
+--     * Slot: in_location
```
ialarmedalien, line 90: It would be nice to make the representation of the name and description in the class and the slot uniform -- i.e. either both should be quoted or both should not be. Similarly, if there is no description given for the class, omit the "Description" field (or add it back for the slot and have the contents as `""`). It looks rather haphazard as-is.

vladistan: Changed.  How does it look now?

ialarmedalien (approved):

PR #2817, opened 2025-07-16
Fix markdown generator

vladistan:


linkml/generators/markdowngen.py
```diff
@@ -149,6 +149,7 @@ def visit_schema(
             items = [i for i in items if i is not None]
             out = "\n".join(items) + "\n"
             out = pad_heading(out)
+            out = out.rstrip("\n") + "\n"
```
ialarmedalien: are there other whitespace characters that we might want to keep?

vladistan: Fixed

PR #2818, opened 2025-07-16
Fix uml generators

vladistan:


vladistan:
Closing because updated https://github.com/linkml/linkml/pull/2814 will take care of this as well.

Issue #2813, opened 2025-07-15
No clear default for `inlined_as_list`


Silvanoc:
**Describe the bug**

No default is neither documented nor implemented for `inlined_as_list`. In the end it gets the default Python value `None`. The robust conditional `<value> is not True` will make it equivalent to `False`, but the not so robust `<value> == False` would make that `None` equivalent to `True` 🤯 

Neither the schema loader nor the viewer provide such a default.

Since [the documentation says](https://linkml.io/linkml/schemas/inlining.html#inlining-as-a-dictionary) that "inlining as a dictionary [...] is only possible if the range of the slot is a class that has an identifier slot", we should:
1. Specify that the default depends on the attributes of the range class.
2. Fix the implementation so that `inlined_as_list` gets a default value.
     * If the class does not have an [identifier](https://linkml.io/linkml-model/latest/docs/identifier/), then `inlined: True` and `inlined_as_list: True`.
     * The question is what happens if `inlined: True` (explicitly set), it has an [identifier](https://linkml.io/linkml-model/latest/docs/identifier/) and no value has been specified for `inlined_as_list`. Should we have by default set `inlined_as_list: False` whenever the range classes have an `identifier`?

**Version of LinkML you are using** v1.9.2

**Please provide a schema (and if applicable, a data file) that replicates the issue**

alexbrdn:
If I am allowed to voice an opinion, I'd say that the default for `inlined_as_list` should always be `True`, as that's the safer option. One can assume that a list of objects always has clear semantics. With dictionaries, there's always the question of what semantics, if any, does the key have.

tfliss:
> The question is what happens if inlined: True (explicitly set), it has an [identifier](https://linkml.io/linkml-model/latest/docs/identifier/) and no value has been specified for inlined_as_list. Should we have by default set inlined_as_list: False whenever the range classes have an identifier?

My understanding is when id is present, inlined==True, inlined_as_list==None, then multivalued needs to be checked to distinguish between inlining as a list of instance structs and inlining as a single valued object struct.

Below is the table I came up for the Pandera generator to pass the inlining compliance tests that is relevant to this issue. The calculation of the form key addresses the defaults. Both inlined and inlined_as_list can be treated as None defaults to false for the lookup. @Silvanoc Thanks, I realized based on this I can remove the commented line in my PR.

[...]

PR #2819, opened 2025-07-16
Fix defaults, inconsistencies and issue detection for inlining in loader

Silvanoc:
Related to #2813


ialarmedalien:
Is this behaviour consistent (or intended to be consistent) between schema loader and schema viewer?

Should there be linter checks implemented to warn about insufficiently-specified inlining issues?

Silvanoc:
> Is this behaviour consistent (or intended to be consistent) between schema loader and schema viewer?

@ialarmedalien good point. I've analyzed the schema viewer and realized:
- The existing (before this PR) behavior WRT to the value of `inlined` when `inlined_as_list: True` is the same: set `inlined: True` (see [here](https://github.com/linkml/linkml-runtime/blob/346f4151e019591ef25b32af4ca0c93eb7115a0b/linkml_runtime/utils/schemaview.py#L1569) how the viewer does it).
- That very same logic is applied in the viewer [function `is_inlined`](https://github.com/linkml/linkml-runtime/blob/346f4151e019591ef25b32af4ca0c93eb7115a0b/linkml_runtime/utils/schemaview.py#L1698). And in general that function applies the same logic as the loader too.
- The viewer does not make any assumptions WRT to those unspecified combinations that this PR is addressing, it simply ignores them. And since there is no such a function as `is_inlined_as_list`, the viewer does not give any hint at all in that direction... It simply returns `None` (like the loader).
- Therefore the viewer tries to behave as the loader whenever `inlined_as_list: True` is found, but it fails... I have created #2822 for this issue.

I would say that handling of `inlined` and `inlined_as_list` in the viewer is even messier than in the loader.

> Should there be linter checks implemented to warn about insufficiently-specified inlining issues?

Having the linter pointing out those issues is absolutely meaningful, but an optimization IMO.

In the sense of "get it working, then optimize it" I would leave the linter for a follow-up PR.

Silvanoc:
@ialarmedalien issue #2822 shows that we have similar issues in the loader and the viewer, so having a unified resolution logic (in a sort of library) would be probably the most meaningful approach. But adapting this PR and #2820 too in that direction is an even bigger effort and I'm not sure I have the bandwidth to address it as of now.

As with the linter, I would prefer to take a 2-step approach, because this PR does not really change the behavior. First get it in here and then get it into a module that both the loader and viewer load. I see this PR as an intermediate step (with the hope that it can get quickly merged) before getting the "real fix" that provides PR #2820 

I would approach PR #2820 differently because it changes the behavior. Putting that logic into a module to be imported by the loader and viewer would be preferable to a 2-step approach.

Silvanoc:
Apparently [SchemaLoader should have been removed](https://github.com/linkml/linkml/issues/923) from the face of the LinkML earth long time ago... Getting rid of it will make this PR needless.

ialarmedalien:
Unfortunately I don't think SchemaLoader is going to be deprecated any time soon as there are too many places where it is still used ==> this work is definitely necessary and useful.

PR #2820, opened 2025-07-16
Set inlining defaults

### LINKED ISSUE #2813: No clear default for `inlined_as_list`

Silvanoc:
Fixes #2813


tests/test_issues/__snapshots__/issue_84.py
```diff
@@ -124,7 +124,9 @@ def __post_init__(self, *_: str, **kwargs: Any):
         if self.name is not None and not isinstance(self.name, str):
             self.name = str(self.name)
 
-        self._normalize_inlined_as_dict(slot_name="annotations", slot_type=Annotation, key_name="has raw value", keyed=False)
+        if not isinstance(self.annotations, list):
+            self.annotations = [self.annotations] if self.annotations is not None else []
+        self.annotations = [v if isinstance(v, Annotation) else Annotation(**as_dict(v)) for v in self.annotations]
```
Silvanoc, line 129: The reason for this change is that the default is different to the one being applied by the Python generator. 

The Python generator is implicitly casting `inlined` and `inlined_as_list` to booleans in [this comparison](https://github.com/linkml/linkml/blob/53f55260bce9934dc4301a0f5f3e242e6015d261/linkml/generators/pythongen.py#L946C1-L949C1), therefore if no value is assigned to `inlined_as_list` (which gets `None`), then it's being cast to `False`. [This comment](https://github.com/linkml/linkml/blob/53f55260bce9934dc4301a0f5f3e242e6015d261/linkml/generators/pythongen.py#L942) above the mentioned comparison explains how unreliable it is (notice the text _"[...] this may not always work [...]"_), because it will try to inline objects without identifiers in a dictionary. By assigning the more meaningful `True` value as a default when nothing is specified, we get a list (what will always work, as opposed to a dictionary).

Silvanoc:
~See [this comment](https://github.com/linkml/linkml/pull/2819#issuecomment-3082980916) for the reason why I've set this PR to `Draft`. Summarized, this changes the behavior and should therefore be applied on both the schema loader and the schema viewer.~

Blocker fixed with merge of PR #2819, so marking this PR ready for review.

alexbrdn:
FYI: this fix, together with https://github.com/linkml/linkml/pull/2807, lets me generate a JSON-LD context just as I expect it.

amc-corey-cox (approved):
Okay, this makes sense to me. Seems pretty straightforward.

Silvanoc:
> Okay, this makes sense to me. Seems pretty straightforward.

@amc-corey-cox am I right to assume that with your approval I can merge it myself?

amc-corey-cox:
I was planning to come back and merge it after I reconciled the branches and everything passed but yes, it's just fine if you merge it since I approved.

PR #2825, opened 2025-07-18
Convert invalid short form JSON to standard YAML

vladistan:


vladistan:
So the changed yaml files were not valid yaml.  Yaml allows inclusion of JSON fragments, but that has to be valid JSON.   Formal JSON spec says all strings should be quoted.  https://www.json.org/json-en.html

List elements,  object keys have to be valid strings  https://www.json.org/json-en.html

tests/test_issues/input/Issue_6_fixed.yaml
```diff
@@ -7,11 +7,14 @@ prefixes:
...
-imports: [metat:types]
+imports:
+  - metat:types
 default_range: string
 
 slots:
-  created: {range: datetime, slot_uri: dct:created}  # slot_uri: "dct:created"
+  created:
+    range: datetime
+    slot_uri: dct:created  # slot_uri: "dct:created"
```
ialarmedalien: I would just remove the commented-out part unless there is a very compelling reason to keep it.

vladistan: Yep

PR #2830, opened 2025-07-22
Regenerate personinfo sql ddl

vladistan:


vladistan:
Ahh.. the diffs in the DDL are caused by #696,  so until it's fixed there is no need to merge this.  Closing now.

PR #2836, opened 2025-07-23
Deprecate generators flags `emit_metadata` and `head`

Silvanoc:
Commit 3e5b8a495a657dca6ca05eccbdeb8f4d806fd73f has removed the flags `head` and `emit_metadata` without any deprecation grace time. This patch reintroduces those flags, but deprecate them.


packages/linkml/src/linkml/generators/pythongen.py
```diff
@@ -1312,7 +1316,7 @@ def cli(
     if validate:
         mod = gen.compile_module()
         logger.info(f"Module {mod} compiled successfully")
-    print(gen.serialize(emit_metadata=head, **args))
+    print(gen.serialize(**args))
```
Silvanoc, line 1319: `metadata` passed as `args["metadata"]` (see change above this one)

dalito:
So as a result of this PR the un-loved timestamps would be removed (for all generators?). That would be great! @vladistan was asking for this on slack.

Silvanoc:
> So as a result of this PR the un-loved timestamps would be removed (for all generators?). That would be great! @vladistan was asking for this on slack.

@dalito you are right, but requires those tests using the generators to specify that no metadata should be generated. Possibly it's already so and it wasn't working up-to-now due to the bugs being hereby fixed.

linkml/utils/deprecation.py
```diff
@@ -258,6 +260,20 @@ def warn(self, **kwargs):
...
+        name="metadata-flag",
+        deprecated_in=SemVer.from_str("1.9.2"),
+        removed_in=SemVer.from_str("1.10.0"),
+        message=(
+            "Use of flags `head` or `emit_metadata` to get a metadata header "
+            "on some generators is no longer supported. "
+            "Flags `head`, `emit_metadata` and `metadata` were being used for "
+            "the same purpose. "
+            "They have been unified, remaining only `metadata`."
+        ),
```
ialarmedalien, line 235: Suggested replacement text (shorter, grammar fixes):

"The flags `head` and `emit_metadata` have been deprecated; the `metadata` flag should be used instead."

linkml/utils/generator.py
```diff
@@ -76,6 +78,61 @@ def _resolved_metamodel(mergeimports):
...
 
+T = TypeVar("T")
+
+
+def deprecated_fields(deprecated_map: dict[str, str]):
+    """
+    Decorator to handle deprecated fields in dataclasses.
+
+    Args:
+        deprecated_map: Mapping from old field names to new field names
```
ialarmedalien: Should be in the form

```py
:param deprecated_map: Mapping from ...
```

to allow docs to be autogenerated

linkml/utils/rawloader.py
```diff
@@ -36,7 +39,8 @@ def load_raw_schema(
     source_file_size: Optional[int] = None,
     base_dir: Optional[str] = None,
     merge_modules: Optional[bool] = True,
-    emit_metadata: Optional[bool] = True,
+    metadata: Optional[bool] = True,
+    emit_metadata: Optional[bool] = _NOT_PROVIDED,
```
ialarmedalien: if `emit_metadata` can be a bool or None, why is the default `object()`?

linkml/utils/rawloader.py
```diff
@@ -53,6 +57,10 @@ def load_raw_schema(
     def _name_from_url(url) -> str:
         return urlparse(url).path.rsplit("/", 1)[-1].rsplit(".", 1)[0]
 
+    if emit_metadata is not _NOT_PROVIDED:
+        deprecation_warning("metadata-flag")
+        metadata = emit_metadata
```
ialarmedalien: Could you just have the default for `emit_metadata` be `None`, so then if `emit_metadata` does have a value (T or F), you can detect it via `emit_metadata is not None`?

linkml/utils/schemaloader.py
```diff
@@ -85,7 +90,13 @@ def __init__(
...
         self.schema_defaults: dict[str, str] = {}  # Map from schema URI to default namespace
         self.merge_modules = mergeimports
-        self.emit_metadata = emit_metadata
+        self.metadata = metadata
+        if emit_metadata is not _NOT_PROVIDED:
+            deprecation_warning("metadata-flag")
+            warnings.warn(
+                f"Value of flag `emit_metadata` (='{emit_metadata}') will overwrite value of flag `metadata`!"
+            )
+            self.metadata = emit_metadata
```
ialarmedalien: same comment applies here with `_NOT_PROVIDED` / None

tests/linkml/test_generators/test_pythongen.py
```diff
@@ -102,7 +102,7 @@ def test_head():
     assert output.startswith(f"# Auto generated from None by pythongen.py version: {PythonGenerator.generatorversion}")
 
     output = PythonGenerator(yaml, format="py", metadata=False).serialize()
-    assert output.startswith("\n# id: https://w3id.org/biolink/metamodel")
+    assert output.startswith("# id: https://w3id.org/biolink/metamodel")
```
ialarmedalien, line 105: 👍

tests/utils/compare_jsonld_context.py
```diff
@@ -9,7 +9,9 @@ def compare_with_snapshot(jsonld_context: str, snapshot: Path):
             actual = json.loads(jsonld_context)
             expected = json.loads(snapshot.read())
 
-            del actual["comments"]["generation_date"]
-            del expected["comments"]["generation_date"]
+            if "comments" in actual:
+                del actual["comments"]["generation_date"]
+            if "comments" in expected:
+                del expected["comments"]["generation_date"]
```
ialarmedalien:
```py
for whatever in [actual, expected]:
    if "comments" in whatever:
        del whatever["comments"]["generation_date"]
```

Do you also need to check for the presence of `generation_data`, just to be on the safe side? e.g. `if whatever.get("comments", {}).get("generation_date"):`

linkml/utils/deprecation.py
Silvanoc: Using GitHub "Add a suggestion" feature would have simplified my life 🙂 

This is your suggestion, right?

```suggestion
        message=(
            "The flags `head` and `emit_metadata` have been deprecated; the `metadata` flag should be used instead."
        ),
```

linkml/utils/generator.py
Silvanoc: This is what you suggest, right?

```suggestion
    :param deprecated_map: Mapping from old field names to new field names
```

BTW, I've taken over the use of `Args:` from somewhere else. I don't remember exactly which one of the [multiple places using it](https://github.com/search?q=repo%3Alinkml%2Flinkml+%22+Args%3A%22&type=code).

ialarmedalien: The format used is a bit all over the place, but I have reliable info [here](https://github.com/linkml/linkml-runtime/pull/457#discussion_r2396046240) that rst-style strings are the way to go.

linkml/utils/rawloader.py
Silvanoc: I can't tell you now how I came to such a weird idea. I suppose that in an early version of the patch it was something like `emit_metadata = _NOT_PROVIDED` until I realized that `Optional[bool]` was the right way to express it. But then didn't replace it correctly.

Anyway, it's clearly wrong. Fixing it.

linkml/utils/rawloader.py
Silvanoc: Once I've fixed `emit_metadata: Optional[bool] = _NOT_PROVIDED,` to be `emit_metadata: Optional[bool] = None,`, this change is also needed.

linkml/utils/schemaloader.py
Silvanoc: Once I've fixed `emit_metadata: Optional[bool] = _NOT_PROVIDED,` to be `emit_metadata: Optional[bool] = None,`, this change is also needed.

tests/utils/compare_jsonld_context.py
Silvanoc: Good point, I like it. Fixed.

@ialarmedalien please resolve the conversation if you agree with the fix.

ialarmedalien:
Not sure what has happened but it looks like some changes got overwritten -- e.g. the message about the deprecated arguments.

Silvanoc:
> Not sure what has happened but it looks like some changes got overwritten -- e.g. the message about the deprecated arguments.

I will check it... I might have shoot myself in the foot on the rebase...

tests/utils/compare_jsonld_context.py
```diff
@@ -12,8 +12,13 @@ def compare_with_snapshot(jsonld_context: str, snapshot: Path):
             actual = json.loads(jsonld_context)
             expected = json.loads(snapshot.read())
 
-            del actual["comments"]["generation_date"]
-            del expected["comments"]["generation_date"]
+            for context in [actual, expected]:
+                if context.get("comments", {}).get("generation_date", ""):
```
ialarmedalien: `something.get("key")` returns None if the key is absent, so no need for the `""` in `get("generation_date", "")`

Silvanoc: I know, but I thought that having it explicit would improve readability. But I don't really care, so I'll change it as you propose.

Silvanoc:
> Not sure what has happened but it looks like some changes got overwritten -- e.g. the message about the deprecated arguments.

@ialarmedalien could you please check if I've recovered what you were missing?

ialarmedalien (dismissed):
Looks great!

Silvanoc:
Misbehavior detected on manual test. Better test coverage needed! Moved back to draft until fixed.

Silvanoc:
@ialarmedalien I'm sorry, but I had to fix a couple of issues that I have detected. I have added the tests for the manual test that has helped me detect the issue.

You only need to review the commits that I've added today:
image

amc-corey-cox:
@Silvanoc thanks for the conscientious dismissal of the previous review. I'll re-review the PR and make sure we get it merged if @ialarmedalien is unable to get to this soon. Feel free to ping me if they don't get to this soon. I'd rather let them re-review but I don't want you to have to wait unnecessarily.

Silvanoc:
> @Silvanoc thanks for the conscientious dismissal of the previous review. I'll re-review the PR and make sure we get it merged if @ialarmedalien is unable to get to this soon. Feel free to ping me if they don't get to this soon. I'd rather let them re-review but I don't want you to have to wait unnecessarily.

@amc-corey-cox I have a high standard for quality. Although I try hard, I don't always reach it. That's what reviews are for 😄

Silvanoc:
@ialarmedalien could you please review my latest commits so that this PR can be merged before new changes creates conflicts? Thanks!

linkml/generators/javagen.py
```diff
@@ -153,20 +154,29 @@ def cli(
...
+    head=None,
+    emit_metadata=None,
     genmeta=False,
     classvars=True,
     slots=True,
     **args,
 ):
     """Generate java classes to represent a LinkML model"""
+    # default is adding metadata to the generated code
+    if "metadata" not in args.keys():
```
ialarmedalien: can be
```py
if "metadata" not in args:
```
(no need for `.keys()`)

Silvanoc: Will fix it

linkml/generators/javagen.py
```diff
@@ -9,7 +9,8 @@
 
 from linkml._version import __version__
 from linkml.generators.oocodegen import OOCodeGenerator
-from linkml.utils.generator import shared_arguments
+from linkml.utils.deprecation import deprecation_warning
+from linkml.utils.generator import deprecated_fields, shared_arguments
```
ialarmedalien: `deprecated_fields` originally comes from `linkml.utils.deprecation`, so it's better to import it from there

linkml/utils/deprecation.py
```diff
@@ -258,6 +260,20 @@ def warn(self, **kwargs):
...
+    Deprecation(
+        name="metadata-flag",
+        deprecated_in=SemVer.from_str("1.9.5"),
+        removed_in=SemVer.from_str("1.10.0"),
+        message=(
+            "Use of flags `head` or `emit_metadata` to get a metadata header "
+            "on some generators is no longer supported. "
+            "Flags `head`, `emit_metadata` and `metadata` were being used for "
+            "the same purpose. "
+            "They have been unified, remaining only `metadata`."
```
ialarmedalien: extremely minor:

"They have been unified, remaining only `metadata`."

==>

"They have been unified, leaving only the flag `metadata`."

linkml/utils/generator.py
```diff
@@ -76,6 +78,60 @@ def _resolved_metamodel(mergeimports):
...
+        # Store the original __init__ method
+        original_init = cls.__init__
+
+        # Create a new __init__ that handles deprecated fields
+        @functools.wraps(original_init)
+        def new_init(self, *args, **kwargs):
+            # Process kwargs to handle deprecated fields
+            for old_field, new_field in deprecated_map.items():
+                if old_field in kwargs:
+                    deprecation_warning("metadata-flag")
```
ialarmedalien: completely optional, but I'd make a "constant" for this value (e.g. `METADATA_FLAG = "metadata-flag"`) and use that in the various places that you mention `"metadata-flag"` -- makes it a lot quicker and easier to update the value or change it, e.g. if you wanted to make the value configurable and supply a variable instead of specifying a literal.

linkml/generators/javagen.py
Silvanoc: Strictly speaking you're wrong here 😄 I've declared `deprecated_fields` in `linkml.utils.generator`, so I have to import it from there.

Let me anticipate to your answer:
> `deprecated_fields` should be declared in `linkml.utils.deprecation` and imported from there.

What would be correct 😄

tests/test_scripts/test_gen_python.py
```diff
@@ -76,3 +76,35 @@ def test_gen_classvars_slots(input_path, snapshot):
     gen_and_comp_python(
         input_path("inheritedid.yaml"), snapshot("genpython/inheritedid_ncvs.py"), ["--no-classvars", "--no-slots"]
     )
+
+
+def test_metadata_flag():
+    "Test that metadata is generated, when flag active"
```
ialarmedalien: Should this be a standard doc string, i.e.
```py
    """Test that metadata is generated when flag is active."""
```

tests/test_scripts/test_gen_python.py
```diff
@@ -76,3 +76,35 @@ def test_gen_classvars_slots(input_path, snapshot):
...
+    assert result.output.splitlines()[2].startswith("# Schema: ")
+    assert result.output.splitlines()[4].startswith("# id: ")
+
+
+def test_no_metadata_flag():
+    "Test that metadata is generated, when flag inactive"
+    runner = CliRunner()
+    result = runner.invoke(cli, ["--no-metadata", KITCHEN_SINK_PATH])
+    assert result.exit_code == 0
+    assert result.output.splitlines()[0].startswith("# id: ")
```
ialarmedalien: Could also check that the metadata lines are not present -- i.e. ensure no lines start with `# Generation date: ` or other metadata.

tests/test_scripts/test_gen_python.py
Silvanoc: Habit from having learn Python long time ago, when single quotation marks were ok for docstrings. Will fix it.

tests/test_scripts/test_gen_python.py
Silvanoc: More robust. But since metadata are always in the header, not strictly needed, I'd say.

linkml/utils/generator.py
```diff
@@ -76,6 +78,60 @@ def _resolved_metamodel(mergeimports):
     return metamodel
 
 
+T = TypeVar("T")
+
+
+def deprecated_fields(deprecated_map: dict[str, str]):
```
ialarmedalien: Can you test this function directly in the `test_utils/test_generator.py` file to ensure that it's doing everything you expect it to?

Silvanoc: As mentioned [here](https://github.com/linkml/linkml/pull/2836#discussion_r2452999263), I think that this function should be in `linkml.utils.deprecation`. I will have to check how I can test it, but then in `test_utils/test_deprecation.py`.

linkml/generators/javagen.py
ialarmedalien: Serves me right for making assumptions based on the names of functions... 😁

tests/test_scripts/test_gen_python.py
ialarmedalien: I think I've misinterpreted the flag / tests. Basically, I would want the tests to show the difference between running the command with the `--no-metadata` flag and with the `--metadata` flag. If metadata gets generated regardless of the status of the metadata flag, you could just have a single parametrised test, e.g.

```py

@pytest.mark.parametrize("flag", ["metadata", "no-metadata"])
def test_gen_python_metadata_flag(flag: str) -> None:
    runner = CliRunner()
    result = runner.invoke(cli, [f"--{flag}", KITCHEN_SINK_PATH])
    assert result.exit_code == 0
    assert result.output.splitlines()[0].startswith("# Auto generated from ")
    assert result.output.splitlines()[1].startswith("# Generation date: ")
    assert result.output.splitlines()[2].startswith("# Schema: ")
    assert result.output.splitlines()[4].startswith("# id: ")
```

Right now, I can't tell from looking at the tests whether there are lines that appear only in one file and not in the other because there are no assertions to ensure that certain lines _do not_ appear in the results.

Silvanoc: In the end I've added it too.

Silvanoc:
@ialarmedalien hopefully I've addressed all your concerns and we are ready to go

tests/test_scripts/test_gen_python.py
```diff
@@ -76,3 +76,32 @@ def test_gen_classvars_slots(input_path, snapshot):
...
+@pytest.mark.parametrize("flag", ["metadata", "no-metadata"])
+def test_metadata_flag(flag: str) -> None:
+    """Test that metadata is only generated, when flag active"""
+    runner = CliRunner()
+    result = runner.invoke(cli, [f"--{flag}", KITCHEN_SINK_PATH])
+    assert result.exit_code == 0
+
+    # metadata lines only present if flag active
+    for line_head in ["# Auto generated from ", "# Generation date: ", "# Schema: "]:
+        matched_lines = [True for line in result.output.splitlines() if line.startswith(line_head)]
```
ialarmedalien: run `splitlines()` and then search for the `line_head`s in the results

Silvanoc: I don't fully get what you mean with *search*.

That's an optimization to avoid splitting into lines 3 times, right? If it's true, I'd leave as it is.

You can always provide a suggestion that i can simply apply. Or a commit on this branch fixing it. Or I create an issue to fix it afterwards.

Silvanoc: I think I got it now.

ialarmedalien: Yup, that looks great.

I'll check back again in a bit -- feel free to tag me when you have debugged the issue with the warnings so I can check the remaining code.

Silvanoc:
@ialarmedalien do you have expertise on warning filters? I'm almost sure that I'm failing to reset the warning registries and since the deprecation warning has been raised somewhere else, I'm not getting it anymore. Locally the tests succeed.

In my timezone it's already too late. I give up for today 🤷

ialarmedalien:
It looks like the problem is in the Deprecation class:


line 201-ish:

```py
    def warn(self, stack_level=3, **kwargs):
        if self.deprecated:
            # ensure filter has expected value
            warnings.filterwarnings("default", category=DeprecationWarning)
            warnings.warn(message=str(self), category=DeprecationWarning, stacklevel=stack_level, **kwargs)
```

`self.deprecated` is false, so the warning is not emitted.

`self.deprecated`:

```py
   @property
    def deprecated(self) -> bool:
        return SemVer.from_package("linkml") >= self.deprecated_in
```

SemVer for linkml in my local repo is `1.9.4`, `metadata-flag` has SemVer `1.9.5`.

What do you have locally for linkml semver?

amc-corey-cox:
CI is currently failing on an `assert False` in tests/test_utils/test_deprecation.py::test_dataclass_fields . Is that intentional since there are other issues or is this almost ready to go?

packages/linkml/src/linkml/utils/generator.py
```diff
@@ -292,6 +291,14 @@ def serialize(self, **kwargs) -> str:
         """
         out = ""
 
+        deprecation_map = {"emit_metadata": "metadata", "head": "metadata"}
+        for flag in deprecation_map:
+            if flag in kwargs:
+                deprecation_warning(METADATA_FLAG)
+                new_flag = deprecation_map[flag]
+                kwargs[new_flag] = flag
```
amc-corey-cox: Should this be
```
kwargs[new_flag] = kwargs[flag]
```
instead?

amc-corey-cox:
I think this is ready to go if we can fix the CI failures here. I think test_dataclass_fields is failing because the EMITTED set in deprecation.py isn't cleared between tests. Once any code triggers metadata-flag earlier in the test run, it won't fire again here. The reset_warnings fixture clears Python's  warning registries but not EMITTED.

packages/linkml/src/linkml/utils/generator.py
Silvanoc: You're fully right! Very good catch, thanks!

Silvanoc: Fixed

Silvanoc:
> I think this is ready to go if we can fix the CI failures here. I think test_dataclass_fields is failing because the EMITTED set in deprecation.py isn't cleared between tests. Once any code triggers metadata-flag earlier in the test run, it won't fire again here. The reset_warnings fixture clears Python's warning registries but not EMITTED.

@amc-corey-cox you nailed it! Now all the tests are passing!

Issue #2861, opened 2025-08-14
Mask Python keywords in pydantic bindings


EOltmanns:
---
name: Feature request
about: Suggest an idea for this project
title: 'Mask Python keywords in pydantic bindings'
labels: feature
assignees: ''

---

**What is your feature request?**
Let pydanticgen handle enum values and slot names gracefully even if they happen to be Python keywords. This lifts a restriction on the kind of data that can be processed using generated bindings. Here is a small JSON snippet which demonstrates the challenge:

```json
{"translation_dict": {"from": "del", "to": "eng"}}
```

**How important is this feature to you?** Select from the options below:
• Medium - we got bitten by the enum part of it and would like to see our solution merged upstream

**Additional context**
Please kinely review our pullrequest to be submitted shortly.

cmungall:
See also:

- #104
- #349
- #2479

PR #2862, opened 2025-08-14
Mask keywords

EOltmanns:
Test case and fix for issue #2861

### DIFF
--- a/linkml/generators/pydanticgen/template.py
+++ b/linkml/generators/pydanticgen/template.py
@@ -1,9 +1,15 @@
 import sys
 from importlib.util import find_spec
+from keyword import iskeyword
 from typing import Any, ClassVar, Literal, Optional, Union, get_args
 
+try:
+    from typing import Self
+except ImportError:
+    from typing_extensions import Self
+
 from jinja2 import Environment, PackageLoader
-from pydantic import BaseModel, Field, field_validator
+from pydantic import BaseModel, Field, field_validator, model_validator
 from pydantic.version import VERSION as PYDANTIC_VERSION
 
 from linkml.generators.common.template import (
@@ -124,8 +130,18 @@ class EnumValue(BaseModel):
     """
 
     label: str
+    alias: Optional[str] = None
     value: str
     description: Optional[str] = None
+
+    @model_validator(mode="after")
+    def alias_python_keywords(self) -> Self:
+        """Mask Python keywords used for `label` by appending `_`"""
+        if iskeyword(self.label):
+            if self.alias is None:
+                self.alias = self.label
+            self.label = self.label + "_"
+        return self
 
 
 class PydanticEnum(PydanticTemplateModel):
@@ -213,6 +229,15 @@ class PydanticAttribute(PydanticTemplateModel):
             return "..."
         else:
             return "None"
+
+    @model_validator(mode="after")
+    def alias_python_keywords(self) -> Self:
+        """Mask Python keywords used for `name` by appending `_`"""
+        if iskeyword(self.name):
+            if self.alias is None:
+                self.alias = self.name
+            self.name = self.name + "_"
+        return self
 
 
 class PydanticValidator(PydanticAttribute):
--- a/linkml/generators/pydanticgen/templates/base_model.py.jinja
+++ b/linkml/generators/pydanticgen/templates/base_model.py.jinja
@@ -1,5 +1,7 @@
 class {{ name }}(BaseModel):
     model_config = ConfigDict(
+        serialize_by_alias = True,
+        validate_by_name = True,
         validate_assignment = True,
         validate_default = True,
         extra = "{{ extra_fields }}",


pyproject.toml
```diff
@@ -98,7 +98,10 @@ dev = [
   "requests-cache >= 1.2.0",
   "myst-nb >= 1.0.0; python_version >= '3.9'",
   "sphinx-design >= 0.5.0",
-  "rich >= 13.7.1"
+  "rich >= 13.7.1",
+  "docker>=7.1.0",
```
cmungall: maybe this needs to go in a separate PR?

EOltmanns: Thanks for the hint. As it turns out, I just missed to include the tests-extra group in my uv sync call.

EOltmanns: Dropped this patch from the series and all checks have passed - locally as well as online.

pyproject.toml
```diff
@@ -53,7 +53,7 @@ dependencies = [ # Specifier syntax: https://peps.python.org/pep-0631/
     "parse",
     "prefixcommons >= 0.1.7",
     "prefixmaps >= 0.2.2",
-    "pydantic >= 1.0.0, < 3.0.0",
+    "pydantic >= 2.11.0, < 3.0.0",
```
ialarmedalien: Are your changes incompatible with pydantic 1.0? There is a deprecation process for upgrading big dependencies like pydantic (I think there may be an existing PR to remove support for v1), so if your fixes still work on v1, it's better to remove keep the version updates in a separate PR.

sneakers-the-rat: specifically, this deprecation has already been scheduled for `1.10.0`, so we should not do it sooner than that:

https://github.com/linkml/linkml/blob/765123cf3455c582c7c9bc9087fa2d8f38c824bb/linkml/utils/deprecation.py#L214-L224

and when we do it, we would probably set a range like `>2.0.0` because there are valid reasons to be using `<2.11`

linkml/generators/pydanticgen/templates/attribute.py.jinja
```diff
@@ -1,4 +1,5 @@
-{{name}}: {{ range }} = Field(default={{ field }}
+{{name}}{{"_" if (name | is_keyword)}}: {{ range }} = Field(default={{ field }}
+{%- if (name | is_keyword) %}, alias="{{name}}"{% endif -%}
```
sneakers-the-rat: the purpose of the template models is to avoid template soup where a ton of computation that mutates values happens in the templates that makes the outcome opaque. this should be an `after` validator in the `PydanticAttribute` model

```python
class PydanticAttribute(PydanticTemplateModel):
    # ...
    name: str
    alias: str | None = None
    # ...
    
    @model_validator(mode="after")
    def alias_python_keywords(self) -> Self:
        """
        Doing work like this in the template models lets us document it
        instead of having things turn into template soup
        """
        if keyword.iskeyword(self.name):
            if self.alias is None:
                self.alias = self.name
            self.name = self.name + "_"
        return self
```

so that by the time we hit the template, all the values are what they're expected to be, and the only logic we need there is switching logic for presence. coupling models and templates and moving work into the models is why the templates look like [this](https://github.com/linkml/linkml/blob/765123cf3455c582c7c9bc9087fa2d8f38c824bb/linkml/generators/pydanticgen/templates/class.py.jinja) instead of [this](https://github.com/linkml/linkml/blob/a1821046191b38a1271d8a9a904eebc14a7ac959/linkml/generators/pydanticgen.py#L41-L278)

linkml/generators/pydanticgen/templates/base_model.py.jinja
```diff
@@ -1,5 +1,6 @@
 class {{ name }}(BaseModel):
     model_config = ConfigDict(
+        serialize_by_alias = True,
```
sneakers-the-rat, line 3: likely also want `validate_by_name` true as well, since this may be surprising if the field is `_for` and i can't do `MyModel(_for="value")`

linkml/generators/pydanticgen/templates/enum.py.jinja
```diff
@@ -6,7 +6,7 @@ class {{ name }}(str{% if values %}, Enum{% endif %}):
 {% endif %}
 {% if values %}
     {% for pv in values.values() %}
-    {{pv.label}} = "{{pv.value}}"
+    {{pv.label}}{{"_" if (pv.label | is_keyword)}} = "{{pv.value}}"
```
sneakers-the-rat: same as above - avoid mutation in templates

tests/test_generators/test_pydanticgen.py
```diff
@@ -242,6 +243,16 @@ def test_pydantic_enum_descriptions():
         )
 
 
+def test_pydantic_unmasked_keywords(input_path):
+    gen = PydanticGenerator(input_path("unmasked_python_keywords_example.yaml"), package=PACKAGE)
+    code = gen.serialize()
+    mod = compile_python(code, PACKAGE)
+    translation_dict = {"from": "del", "to": "eng"}
+    translation_inst = json_loader.loads(translation_dict, mod.Translation)
+    assert translation_inst.from_ == mod.LanguageEnum("del")
```
sneakers-the-rat: since the enum template is modified too, we'll want to add a keyword enum option there too and test that it behaves as expected

pyproject.toml
sneakers-the-rat: i don't think changing the dep range is necessary here - one can access `model_fields` from the class rather than the instance in all versions of pydantic 2 (just tested in 2.0.3)

tests/test_generators/test_pydanticgen.py
EOltmanns: Actually, the enum is checked by this assertion too, and, indeed, it does that it is supposed to do. Please note that del is a Python keyword and the assertion would fail on two accounts. Writing this makes me realise, though, that this may be exactly your point. So, I will swap "eng" and "del" in the dict in order to be able to assert the slot and enum part separately.

linkml/generators/pydanticgen/templates/attribute.py.jinja
EOltmanns: Thanks for the explanation, I will look into this and resubmit accordingly.

EOltmanns:
Bumping up the pydantic version was related to adding to environment filters. This feature apparently has not been around for all to long. But since I am not supposed to meddle with the templates anyway, this may not be required after all.

linkml/generators/pydanticgen/templates/base_model.py.jinja
EOltmanns: As it happens, validate_by_name has been introduced in pydantic 2.11. In order to preserve compatibility, I suggest to use populate_by_name: True. It does exactly what we want, except that it is scheduled for deprecation. All the same, it will likely stay around for quite some time yet.

EOltmanns: Then again, serialize_by_alias has only been introduced in 2.11 too. So, I agree that we just use serialize_by_alias in conjunction with validate_by_name. This poses the problem that both options till have no effect on pydantic < 2.11 and I do not know how to account for that in the test function. Is the test suite supposed to pass on older versions of pydantic too? Is there a (preferred) way to run this test conditionally?

sneakers-the-rat: I think the old behavior was to serialize and validate by alias and the new options just gave finer control? I'll test with old pydantic versions. We used to test with pydantic 1, but stopped once we deprecated pydantic v1.

There's a bit of subtlety in the distinction between the version of pydantic that the code *generates for*, the version of pydantic that is *used during generation* (which should be latest), and the allowable pydantic versions in a project that depends on linkml (which is the thing set by the lower bound in the linkml package, and we want to be as wide as possible while still working). 

We could add back in a test that just runs the pydanticgen tests on some old version of pydantic that then becomes the floor, but we don't have a formal lower bound on supported pydantic versions for generated modules.

linkml/generators/pydanticgen/template.py
```diff
@@ -113,9 +119,19 @@ class EnumValue(BaseModel):
     """
 
     label: str
+    alias: Optional[str] = None
     value: str
     description: Optional[str] = None
 
+    @model_validator(mode="after")
+    def alias_python_keywords(self) -> Self:
+        """Mask Python keywords"""
```
sneakers-the-rat:
```suggestion
        """Mask Python keywords used for `label` by appending a `_`"""
```

linkml/generators/pydanticgen/template.py
```diff
@@ -202,6 +219,15 @@ def field(self) -> str:
         else:
             return "None"
 
+    @model_validator(mode="after")
+    def alias_python_keywords(self) -> Self:
+        """Mask Python keywords"""
```
sneakers-the-rat:
```suggestion
        """Mask Python keywords used for `name` by appending a `_`"""
```

EOltmanns:
> nice, revert the pydantic dep increase and looks good to me :)

Sorry, yes, I totally forgot about that. Fixed now. :)

EOltmanns:
Reopening after accidental close. Need to get some sleep.

pyproject.toml
EOltmanns: The new implementation does not rely on fiddling with template environments and should therefore work for older versions of pydantic too. The version change has been dropped accordingly. However, the new implementation relies on the model_validator decorator and from the pydantic migration guide I draw the conclusion that this one has been introduced with pydantic v2 replacing root_validator from v1. So, unless you have another idea, we may have to taghten the requirement to >=2.0.0 after all and cannot merge before linkml 1.10. What do you think?

linkml/generators/pydanticgen/template.py
EOltmanns: Unfortunately, I believe the use of @model_validator requires at least pydantic v2.0.0. I have split this into a separate patch. Can we play tricks with a conditional `from pydantic import root_validator as model_validator` or write a wrapper? Or do we have to wait for linkml 1.10 after all?

pyproject.toml
sneakers-the-rat: There are already pydantic 1 incompatible parts of the code in the lib, the lower bound is to allow people to still have pydantic 1 as a dep in the package that depends on linkml. See my prior comment on subtle distinction of what different dep ranges mean in the context of a code generator.

EOltmanns: Alright, in that case I am all for dropping the patch and have just done so. Thanks for clarifying.

sneakers-the-rat (approved):

PR #2873, opened 2025-08-18
Merge linkml-runtime into linkml

iQuxLE:
## Merge linkml-runtime into linkml

This PR brings the full history and codebase of linkml-runtime into the linkml repository, preserving commit history and placing the runtime package under its own subdirectory.

### What was done:

- Imported all branches and commit history from linkml-runtime.
- Moved the runtime codebase into the linkml_runtime/ subdirectory.
- Moved all test files that were part of linkml-runtime into tests/linkml_runtime/.
- Preserved the complete chronological commit history of both repositories
- adjusted gh-actions + package and release config 

Repo Structure:
```
linkml/
├── pyproject.toml  (top level toml will contain workspaces)
├── linkml
├── packages/
│   └── linkml_runtime/   
            └── linkml_runtime/   (needed for uv cause python standard)
                   └──  pyproject.toml (will hold dependencies of seperate package)    
└── tests/
    └── linkml/
    └── linkml_runtime/
```

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -197,17 +197,15 @@ jobs:
         working-directory: packages/linkml_runtime
         run: uv build
 
-      # Verify the built packages can be installed and imported
       - name: Test installation of built wheels
         run: |
-          # Create a fresh test environment
           uv venv test-env
-
-          # Install the linkml wheel we just built
-          uv pip install --python test-env dist/linkml-*.whl
-
-          # Install the linkml-runtime wheel we just built
-          uv pip install --python test-env packages/linkml_runtime/dist/linkml_runtime-*.whl
-
-          # Verify both packages can be imported
+          for wheel in dist/linkml-*.whl; do
+            uv pip install --python test-env "$wheel"
+            break
+          done
+          for wheel in dist/linkml_runtime-*.whl; do
+            uv pip install --python test-env "$wheel"
+            break
+          done
           uv run --python test-env python -c "import linkml; import linkml_runtime; print('✓ Both packages installed and imported successfully')"


sujaypatil96:
Using this document to check common use cases in the repo to test and make sure the linkml-runtime merge is working as expected: https://docs.google.com/document/d/1Z5Ln3Lm3hlRhcu2Chpn-JKl_mkE9sXUJVtbuxMt11Jw/edit?usp=sharing

Silvanoc:
The merging accomplished by this PR is really needed to make the developer experience much better! But it is probably tough as hell to be reviewed...
@iQuxLE any hints on which strategy any reviewer should follow?

pyproject.toml
```diff
@@ -116,6 +122,7 @@ tests-extra = [
   "mock >= 5.1.0",
   "testcontainers == 3.7.1",
   "jsonpatch >= 1.33",
+  "docker >= 6.1.0",
```
Silvanoc, line 125: Should not be needed explicitly, because `docker` is a dependency of `testcontainers`.

pyproject.toml
```diff
@@ -305,6 +320,11 @@ select = [
 
 # Notebooks can have unsorted imports
 "tests/test_notebooks/input/*" = ["E402"]
+"tests/linkml/test_notebooks/input/*" = ["E402"]
+
+# Runtime test files have different conventions
```
Silvanoc, line 325: I'd add an issue to address it. After merging them, we should evolve to have the same conventions throughout the whole repository.

Silvanoc (approved):

iQuxLE:
@Silvanoc thanks for your comments and apologies for the late answer. 
@sujaypatil96 and @sierra-moxon are testing these changes with current implementations and have a testing agenda.

We most likely need to redo parts of the PR as in the process we are rewriting history of runtime so that it looks as it was always inside the main repo. Otherwise the history can't be transferred.

Due to changes in runtime - as long as there is no general stop on PRs - we can't necessarily merge without running this again.

However there is a pipeline we can go through pretty fast and as soon as we know everything works on biolink etc this (the history merge) can be repeated and we should be able to go forward.

I know it's going a bit slowly but everyone needs to make sure we are not missing something.

Silvanoc:
@iQuxLE don't worry, I just wanted to avoid this effort stalling because nobody is looking at this PR. When someone invests the huge effort that you've probably invested here in the thought that it will be highly welcomed, it's hard the feeling that nobody cares.

I think that this merging is very important to improve the developers experience and therefore I wanted to give you the attention that this effort is worth and help moving forward.

I'm glad to learn that this topic is moving on, even if we don't have any observable manifestations of that movement.

sujaypatil96:
Thanks for reviewing this @Silvanoc. We reviewed this PR on call last Friday, 10/3. There is an internal document that describes the procedure that @iQuxLE followed to make this PR, and we tried to replicate the same in #2928 and I think we were able to do it successfully. @kevinschaper made a bash script that has all the git commands needed to replicate this monorepo merge process, and he will be putting the script in it's own repo in the linkml org soon, so others can replicate too. We have a testing document for this PR as well: https://docs.google.com/document/d/1Z5Ln3Lm3hlRhcu2Chpn-JKl_mkE9sXUJVtbuxMt11Jw/edit?usp=sharing. If you have any ideas, or you think we've missed something while testing this PR, please feel free to make suggestions on this document.

However, the plan for us is to resolve all the merge conflicts on this PR and make a release candidate from this branch. Then, some of us have volunteered to test this release candidate in downstream schema repos and then once we've confirmed that everything is working as expected, we can make an official release.

dalito:
A crazy question: Would it make sense to include also the [linkml-model ](https://github.com/linkml/linkml-model) repository into the monorepo? - If so, now would be the time. 😉

amc-corey-cox:
I'll add the consideration of adding the model to the mono-repo to the LinkML developers call. I do want to raise the concern that doing so might slow us down in getting this done. I think it is too important to wait much longer. I'd like to resolve the conflicts, re-run the workflow, and get this merge finished.

dalito:
I extended the [bash script](https://gist.github.com/kevinschaper/eced60608e84d46ce3e86c29cd2bbb34) from @kevinschaper to also include linkml_model [linkml-runtime-model-migration.sh](https://gist.github.com/dalito/d58758f9a8d1be05c86c0b97ccedae56)

This does not yet work well. There are several conflicts. The script needs more work.

dalito:
After thinking and experimenting more with merging linkml-model as well, I think that it should be done later and as a separate step. It has other challenges for example handling of the documentation which is not relevant for linkml-runtime.

I wonder about the final structure a bit. Why is it like it is (which somewhat deviates from uv-managed monorepo recommendations)? It would be great to have an issue for the monorepo plan. --> Made one: #2936

sneakers-the-rat:
I would strongly recommend structuring these as [PEP-420](https://peps.python.org/pep-0420/) namespace packages.

packaging docs: https://packaging.python.org/en/latest/guides/packaging-namespace-packages/
example: https://github.com/pypa/sample-namespace-packages/tree/master/native

so then they could be distributed as separable packages like `linkml-runtime`, `linkml-pydantic`, `linkml-core`, and so on (i would propose `linkml` as a distribution becomes just an alias package that refers to all the packages in the namespace, and that's what namespace packages imply) but then still share the same `linkml` namespace at import time: `from linkml.runtime import ...`, `from linkml.generators.pydantic import...`

the `uv` workspaces-only route may work for `linkml_runtime`, but it would make it more challenging to split off the generators and the other parts of linkml, since each of those would have to become `from linkml_pydantic import ...`. uv workspaces are also not a standard, but if this project is already locked in to `uv` then that bullet has already been bitten. it looks like they can be used together: https://github.com/libranet/demo-uv-workspace/tree/main (from this issue: https://github.com/astral-sh/uv/issues/10628 )

.github/workflows/main.yaml
```diff
@@ -1,7 +1,7 @@
-# Built from:
-# https://docs.github.com/en/actions/guides/building-and-testing-python
+# Unified CI workflow for linkml monorepo
+# Tests both linkml and linkml-runtime packages together
```
sneakers-the-rat, line 2: combining tests like this, imo, loses some of the advantages of separate packages within a monorepo, and has a pretty strong potential for making this and any other sub-packages non-independent (tests pass when both are installed together, but `linkml-runtime` can no longer be used by itself, accidentally). 

i think it would probably be good to have separate workflows for each of the sub-packages (and to reduce boilerplate one could make reusable local actions), and to use `path` filters so that, e.g. `linkml-runtime` tests are only run when `linkml-runtime` files are changed, and so on: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore

dalito:
This task was finally completed in #2987

PR #2872, opened 2025-08-16
Remove erroneous @click.option definition (closes #2871)

EOltmanns:
Act on a click warning indicating that the same option has been defined multiple times.

PR #2877, opened 2025-08-23
Make optional lists empty instead of none (closes #2876)

### LINKED ISSUE #2876: Do not require users of pydantic bindings to initialise optional lists before appending
---
name: Feature request
about: Suggest an idea for this project
title: 'Do not require users of pydantic bindings to initialise optional lists before appending'
labels: feature
assignees: ''

---

The Python bindings based on dataclasses provide a convenient feature: Users of those bindings can simply append values to list range fields, even the optional once, without bothering to initialise them first. It would be nice to make pydanticgen bindings behave similarly especially for users migrating existing code from pythongen to pydanticgen.

• Medium - can do work without it; but it would be nice to get our local solution merged upstrem

**Additional context**

Pull request to be submitted shortly.

EOltmanns:
Suggestion with regard to issue #2876.
This will initialise optional lists with `[]` instead of `None`, allowing users to append to them without checking whether they have been initialised yet. Additionally, a model serialiser is added to `ConfiguredBaseModel` making sure that empty lists are removed when `exclude_none=True` has been passed to `.model_dump`.

### DIFF
--- a/tests/test_compliance/test_array_compliance.py
+++ b/tests/test_compliance/test_array_compliance.py
@@ -111,7 +111,7 @@ def test_array(framework, description, ndim, object, is_valid):
                         "exact_number_dimensions": ndim,
                     },
                     "_mappings": {
-                        PYDANTIC: f"s1: Optional[{expected_range[ndim]}] = Field(default=None",
+                        PYDANTIC: f"s1: Optional[{expected_range[ndim]}] = Field(default=[]",
                     },
                 },
             },
--- a/tests/test_compliance/test_core_compliance.py
+++ b/tests/test_compliance/test_core_compliance.py
@@ -528,7 +528,7 @@ def test_cardinality(framework, multivalued, required, data_name, value):
     choices = {
         (PYDANTIC, False, False): "Optional[str] = Field(default=None",
         (PYDANTIC, False, True): "str = Field(default=...",
-        (PYDANTIC, True, False): "Optional[list[str]] = Field(default=None",
+        (PYDANTIC, True, False): "Optional[list[str]] = Field(default=[]",
         (PYDANTIC, True, True): "list[str] = Field(default=...",
         # TODO: values
         (PYTHON_DATACLASSES, False, False): "",
--- a/tests/test_generators/test_pydanticgen.py
+++ b/tests/test_generators/test_pydanticgen.py
@@ -307,8 +307,8 @@ slots:
     gen = PydanticGenerator(schema_str, package=PACKAGE)
     code = gen.serialize()
     assert "inlined_things: Optional[dict[str, Union[A, B]]] = Field(default=None" in code
-    assert "inlined_as_list_things: Optional[list[Union[A, B]]] = Field(default=None" in code
-    assert "not_inlined_things: Optional[list[str]] = Field(default=None" in code
+    assert "inlined_as_list_things: Optional[list[Union[A, B]]] = Field(default=[]" in code
+    assert "not_inlined_things: Optional[list[str]] = Field(default=[]" in code
 
 
 @pytest.mark.parametrize(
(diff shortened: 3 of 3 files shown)


EOltmanns:
Having another go at proper linking: This pull request is supposed to resolve #2876.

kervel:
i had this in the rust generator and i eventually reverted that, and i went back to optional list. The problem i had was then it was not possible anymore to see the difference between an uninitialised list (none) and an empty but initialized list ([]). i don't know if the same thing plays with the pydantic generator, but eg dict.update might react differently for nothing vs empty.

EOltmanns:
Yes, this change would make uninitialised lists indistinguishable from empty lists. This is a trade off for the added convenience that downstream code does not have to check before every append whether the list has been initialised yet. In fact, this is the current behaviour of pythongen as opposed to pydanticgen. We are in the process of migrating our code from pythongen based bindings to pydanticgen based ones, which is why this change would make our lives easier. If pydanticgen actually is supposed to make a difference between empty and uninitialised lists, would it perhaps be acceptable to make this behaviour configurable somehow? I am not quite sure how to do this, perhaps by subclassing pydantic.ConfigDict, adding an `init_empty_lists` option and using that subclass in ConfiguredBaseModel. What do you think?

kevinschaper:
pydantic does track __fields_set__, and I think it won't include defaults in that. This is actually a bummer when your type designator is set as a default and you want to export to json and include explicitly set nulls, but I think it does make it possible to differentiate an unset empty list and an explicitly set empty list (though, arguably, you might explicitly set to None to really affirm the choice). 

I generally like the convenience of being able to add to empty lists more than I worry about the performance, so my inclination is that we bring this in, and if we find that some people really need to keep lists as None, we solve that problem after it actually comes up.

kevinschaper:
@EOltmanns it looks like I can't update the branch in the GitHub UI to make merging easier, so you'll need to update on your end so that I can merge the change.

EOltmanns:
Thanks for the thumbs up. I just merged current main into the branch but I am unsure whether you would prefer a rebase onto main instead. Please let me know either way, so I know for future contributions as well.

kevinschaper:
@EOltmanns I'm not totally sure where to change the setting (not sure if it's per PR or global), but can you set it so the upstream maintainer can update the PR? otherwise we need to make sure nothing else goes in between when you update with main and I can press the merge button.

dalito:
Since the PR is from an organisation "AV-EFI" instead of an user account it is not possible to set "allow edits by maintainer", see https://github.com/orgs/community/discussions/5634

@EOltmanns - please rebase the PR on current main.

EOltmanns:
@dalito: Thank you very much for the link explaining the issue with pull requests from organisations. I have just rebased onto main. The same obstacle will most likely get in the way with my other pull requests #2862 and #2872, so I am whether I can make things easier for you guys by transferring those to my personal GitHub account somehow. The situation is far from optimal, but as far as future pull requests are concerned, I should probably submit them from my personal account, right? What about the existing pull requests then?

dalito:
Since linkml uses the GitHub setting "Require branches to be up to date before merging" to only merge proven code, rebasing is frequently needed here in contrast to many other repos. This is fine as long as the PR authors enable "allow edits by maintainers" and do not use an organization. For your PRs from the AV-EFI organization you have to help us with the rebases - sorry.

PRs from individual accounts are clearly more convenient for contributors.

EOltmanns:
Right, I rebased onto main and checks appear to have passed. Let me know if there is anything to be done or discussed.

kevinschaper:
I was very excited to see the green button! @EOltmanns I'll open up your other PRs in their own windows and try to keep an eye out them to be both up to date and and for the tests to be done so that I can merge, if you can keep them open and regularly hit the button to keep them up to date, I think we can manage to merge them

kervel:
this caused regressions for us.

```
  AssociatedNetElement:
    description: >-
      A RSM Associated Net Element
    attributes:
      netElement:
        range: PositioningNetElement
        inlined: false
      bounds:
        range: float
        multivalued: true
        minimum_cardinality: 1
        maximum_cardinality: 2
      appliesInDirection:
        range: string
      lateralPosition:
        range: string # FIXME custom class LateralPosition instead of str in RSM
```

now we can't create AssociatedNetElement() like we did before (validation fails because altough its not required, when its filled in, the cardinality is mandatory). this is a behavioural change for sure , whether intended or not

Issue #2882, opened 2025-08-30
SQL DDL generator produces non-deterministic output. CREATE INDEX statements get reordered on  each run


vladistan:
---
name: Sql generator produces deterministic output
title: 'SQL DDL generator produces non-deterministic output. CREATE INDEX statements get reordered on  each run'
labels: bug
assignees: ''
---

**Describe the bug**

The LinkML SQL DDL generator produces non-deterministic output for CREATE INDEX statements when tables have multiple indexes. Each generation run reorders the indexes differently, causing file content to change even when the schema hasn't changed.

**Version of LinkML you are using**
1.9.3.post27.dev0 

**Please provide a schema (and if applicable, a data file) that replicates the issue**

PersonInfo schema included in the project.  `examples/PersonSchema/person.yaml`

vladistan:
#696 is a parent issue for this

VictoriaSavage1NNL:
Hello! can you try this again? The recently closed pull request #2904 streamlined a bit and this may have remedied this issue.

PR #2883, opened 2025-08-30
SQL DDL generator to produce deterministic output

vladistan:
See #2882 for detailed explanation


matentzn:
@vladistan the PR is still in draft state - can you let me know if you are still seeking to get this merged / willing to see it through?

PR #2884, opened 2025-08-30
Bump left out checkout action to v5

vladistan:


.github/workflows/test-with-unreleased-runtime.yaml
```diff
@@ -76,15 +76,15 @@ jobs:
           echo "upstream_repo=${{ inputs.upstream_repo }}" >> "$GITHUB_ENV"
 
       - name: checkout upstream linkml-runtime
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
```
dalito, line 79: If dependabot should apply minor and patch updates (and not just major), we need to give the full version number "v5.0.0".

vladistan: I put it this way because the other workflows use only major number https://github.com/linkml/linkml/blob/9323b73ba548debe5015c10f445ca95c0c5066bb/.github/workflows/main.yaml#L18

This not only for checkout action,  for example `uv` action also uses only major number.   We should decide which way to use and change all the workflows to use the right pattern.

My thoughts about this:

1. I believe specifying only major number will always pull the latest version not the one `.0.0` 
2. If above is true this might not be what we want, specifying exact version will make more reproducible builds, as we'll be relying on the exact version of the action. 
3. Some projects go even step further and use commit hashes instead of version tags.  This protects from the potential security problem if the action was compromised by malicious party.   I am not sure if dependabot can handle this renovate definitely does:  https://docs.renovatebot.com/modules/manager/github-actions/

dalito: Dependabot could also work with hashes. But it is somewhat inconvenient (it was used here shortly). 

But we definitely need to give the full version number to get minor & patch updates via dependabot (see #2888 vs #2885). The actions files had full version numbers a while ago but recently too many contributors changed and merged action-files without paying attention so we are back at a messy state.

vladistan: Ok,  so you are saying we need to use full versions everywhere.  Currently
it's a mix of just major versions and full version numbers:

https://github.com/linkml/linkml-runtime/blob/main/.github/workflows/main.yaml#L101
https://github.com/linkml/linkml-runtime/blob/5ef72d483f28fd86685c19a254bb335792923a70/.github/workflows/main.yaml#L69
https://github.com/linkml/linkml-runtime/blob/5ef72d483f28fd86685c19a254bb335792923a70/.github/workflows/main.yaml#L26
https://github.com/linkml/linkml/blob/9323b73ba548debe5015c10f445ca95c0c5066bb/.github/workflows/main.yaml#L18
https://github.com/linkml/linkml/blob/9323b73ba548debe5015c10f445ca95c0c5066bb/.github/workflows/main.yaml#L24

vladistan: @sierra-moxon @ialarmedalien what's your opinion on this?

ialarmedalien: The main reason to go with specific versions--apart from wanting to ensure that there's a steady flow of PRs to update GHAs--is to avoid cases where the latest version of an action may be broken or compromised (e.g. to steal GitHub tokens). That happened with one of the GHAs on a repo I work on, although luckily we had not run or updated the actions recently.

vladistan: Ok you have the point... closing this in favor of  the two below:

https://github.com/linkml/linkml-runtime/pull/450
https://github.com/linkml/linkml/pull/2890

@ialarmedalien @dalito

vladistan:
Use full versions instead:

https://github.com/linkml/linkml-runtime/pull/450
https://github.com/linkml/linkml/pull/2890

PR #2364, opened 2024-10-09
`rustgen` - Initial draft

sneakers-the-rat:
Related to (but doesn't close): https://github.com/linkml/linkml/issues/2360

Opening this early as a draft to see if i can get some people who know rust better than me to help out, and also to track progress :).

My eyes are burning atm and need to stop computering so i will write a fuller description later, but currently we have a generator that creates PyO3-enabled rust classes. Lots of work to be done, they are literally the simplest possible version and don't do any more advanced things yet.

We get something like this:

```
├── Cargo.lock
├── Cargo.toml
├── pyproject.toml
└── src
    └── lib.rs

```

lib.rs
```rust
#![allow(non_camel_case_types)]
use pyo3::prelude::*;
use pyo3::types::{PyDate};

type id = String;
type name = String;
type description = String;
type started_at_time = Py<PyDate>;
type ended_at_time = Py<PyDate>;
// ...

enum FamilialRelationshipType {
    SIBLINGOF,
    PARENTOF,
    CHILDOF,
}

// ...


#[pyclass(unsendable)]
#[derive(Clone)]
struct Activity {
    #[pyo3(get, set)]
    id: String,
    #[pyo3(get, set)]
    started_at_time: Py<PyDate>,
    #[pyo3(get, set)]
    ended_at_time: Py<PyDate>,
    #[pyo3(get, set)]
    was_informed_by: Py<Activity>,
    #[pyo3(get, set)]
    was_associated_with: Py<Agent>,
    #[pyo3(get, set)]
    used: String,
    #[pyo3(get, set)]
    description: String
}

#[pymethods]
impl Activity {
    #[new]
    pub fn new(id: String,started_at_time: Py<PyDate>,ended_at_time: Py<PyDate>,was_informed_by: Py<Activity>,was_associated_with: Py<Agent>,used: String,description: String) -> Self {
        Activity{ id,started_at_time,ended_at_time,was_informed_by,was_associated_with,used,description }
    }
}

// ...
```

pyproject.toml
```toml
[project]
name = "kitchen_sink"
version = "0.0.0"

[build-system]
requires = ["maturin>=1.0,<2.0"]
build-backend = "maturin"

[tool.maturin]
features = ["pyo3/extension-module"]
```

Cargo.toml
```toml
[package]
name = "kitchen_sink"
version = "0.0.0"
edition = "2021"

[lib]
name = "kitchen_sink"
crate-type = ["cdylib"]

[dependencies]

[dependencies.pyo3]
version = "0.21.1"
[...]

Issue #2360, opened 2024-10-08
scoping out a rust generator


sneakers-the-rat:
**What is your feature request?**
Id like to make a rust generator, because eventually I will need to write schemaview in rust to get perf in line with p2p applications and swapping and mutating schemas back and forth, and if we have rustgen then we can generate the metamodel and return fully induced classes via pyO3. That also should let us compile schemaview to wasm and directly use linkml in the browser instead of via pyodide.

Making this issue to a) see what we would want as an MVP for rustgen, and b) see if there are any other ppl interested. I think we dont need a full implementation of the metamodel, since the metamodel is comparatively simple vs. What can be expressed with it, as long as we can express and induce classes with it then that's plenty for my purposes.

Initial sketching
- [x] basics: generating structs with fields and types - definitely possible.
- [ ] slots? metadata? Can we do something like Python's annotated types to make slots with metadata?
- [ ] pyO3: optional pyo3 generation
- [ ] dynamic compilation? Is this possible and reasonable? We would need it for testing
- [ ] union types? I think you need to make an enum for this but they dont quite behave like python unions for maybe obvious reasons
- [ ] vec and hashmap: what are the limitations vs. python lists and dicts?
- [ ] can we generate classes that mimick dataclasses?
- [ ] constraints on fields: cardinality, regex patterns, etc.
- [ ] ..?

**How important is this feature to you?** Select from the options below:
• Medium - can do work without it; but it's important (e.g. to save time or for convenience).

cmungall:
Some of this may be useful but it looks like you are implicitly following this:
https://linkml.io/linkml/howtos/port-linkml.html

cmungall:
Putting high level design issues here rather than the PR, which is liable to get in the weeds:

- what is the high level mapping? Do classes, enums map to structs, traits, enums, impls?

sneakers-the-rat:
ya ya ya - 

* Classes -> Structs ([generation](https://github.com/sneakers-the-rat/linkml/blob/91d513d380d0ea35c64c521d49e1a33d2b02e41c/linkml/generators/rustgen/rustgen.py#L254-L263), [template](https://github.com/sneakers-the-rat/linkml/blob/91d513d380d0ea35c64c521d49e1a33d2b02e41c/linkml/generators/rustgen/templates/struct.rs.jinja))
[...]

PR #2889, opened 2025-09-01
Rust generator for LinkML

kervel:
see https://github.com/linkml/linkml/issues/2360

this PR supersedes:
* https://github.com/linkml/linkml/pull/2364 
* https://github.com/linkml/linkml/pull/2722 

it adds an extra test workflow that pulls in the rust toolchain for the rust related tests.


kervel:
> ## [Codecov](https://app.codecov.io/gh/linkml/linkml/pull/2889?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=linkml) Report
> ❌ Patch coverage is `0%` with `1138 lines` in your changes missing coverage. Please review. ✅ Project coverage is 0.31%. Comparing base ([`9323b73`](https://app.codecov.io/gh/linkml/linkml/commit/9323b73ba548debe5015c10f445ca95c0c5066bb?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=linkml)) to head ([`84f37ed`](https://app.codecov.io/gh/linkml/linkml/commit/84f37ed46d34fb7889158f7ae860ac9c3b57ab83?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=linkml)).

i guess this has to do with my new tests being disabled still.. but i have no idea how to enable them for codecov.

kervel:
hi @sneakers-the-rat could you look at the common changes here ? i'm not sure if i did the rebasing correctly. i don't know the newer pydantic stuff, so i don't know if the below stuff makes sense.

Questionable diffs

- linkml/generators/common/build.py: SkipValidation rewrite looks broken.
    - Replaced a working custom SkipValidation type (using get_pydantic_core_schema with any_schema) with a type alias:
    - from pydantic import GetPydanticSchema
    - SkipValidation = Annotated[Tsv, GetPydanticSchema(lambda tp, handler: core_schema.any_schema())]
- Issues:
    - GetPydanticSchema is not a public pydantic v2 API (expected types are GetCoreSchemaHandler / GetJsonSchemaHandler for custom classes, or metadata markers like
pydantic.SkipValidation for Annotated).
    - Using a TypeVar inside Annotated as a module-level alias (Tsv) and then subscripted later is not how Annotated/generics are intended to be used; the original class-based approach
supported SkipValidation[Foo] properly.

sneakers-the-rat:
> SkipValidation rewrite looks broken. 

Do you mean the version in this PR is broken, or was the original broken? It was working fine afaik, why change it?

> GetPydanticSchema is not a public pydantic v2 API 

it is!
https://docs.pydantic.dev/latest/concepts/types/#using-getpydanticschema-to-reduce-boilerplate

> Using a TypeVar inside Annotated as a module-level alias (Tsv) and then subscripted later is not how Annotated/generics are intended to be used

It is!
https://typing.python.org/en/latest/spec/qualifiers.html#annotated

linkml/generators/common/template.py
```diff
@@ -42,7 +43,7 @@ def render(self, environment: Optional[Environment] = None, **kwargs) -> str:
             environment (:class:`jinja2.Environment`): Template environment - see :meth:`.environment`
         """
         if environment is None:
-            environment = TemplateModel.environment()
+            environment = type(self).environment()
```
sneakers-the-rat, line 46: This got a little busted in refactoring, but we do want this to be written in such a way that the environment is shared among template models (at least among template models of the same kind, like for pydanticgen etc.) for caching/perf purposes.

tests/test_generators/test_rustgen.py
```diff
@@ -0,0 +1,43 @@
...
+
+from linkml.generators.rustgen import RustGenerator
+from linkml.generators.rustgen.rustgen import (
+    SlotContainerMode,
+    SlotInlineMode,
+    determine_slot_mode,
+)
+
+
+@pytest.mark.rustgen
```
sneakers-the-rat: probably want to set this at the module level `pytestmark = pytest.mark.rustgen`

tests/test_generators/test_rustgen.py
```diff
@@ -0,0 +1,43 @@
...
+def test_generate_crate(kitchen_sink_path, temp_dir):
+    gen = RustGenerator(kitchen_sink_path, mode="crate", output=temp_dir)
+    _ = gen.serialize(force=True)
+
+
+def test_determine_slot_mode(kitchen_sink_path):
```
sneakers-the-rat: doesn't seem like a rustgen test?

tests/test_generators/test_rustgen_maturin.py
```diff
@@ -0,0 +1,128 @@
...
+from linkml_runtime.utils.introspection import package_schemaview
+
+from linkml.generators.rustgen import RustGenerator
+
+
+@pytest.mark.rustgen
+def test_rustgen_personschema_maturin(input_path, temp_dir):
+    """
+    Generate a Rust crate from personinfo.yaml and try to build
+    the Python bindings with maturin. Skips if maturin/cargo are absent.
```
sneakers-the-rat: i think we shouldn't double-skip here: you already are skipping in the conftest, so if have reached this test we are explicitly intending the rustgen tests to run. so if we reach here and still don't have the appropriate dependencies, we should fail.

tests/test_generators/test_rustgen_maturin.py
```diff
@@ -0,0 +1,128 @@
...
+                f"exit: {offline_res.returncode}\n"
+                f"stdout:\n{offline_res.stdout}\n"
+                f"stderr:\n{offline_res.stderr}\n"
+            )
+
+    # If build succeeded, confirm a wheel exists
+    wheels_dir = out_dir / "target" / "wheels"
+    assert wheels_dir.exists(), "maturin did not produce target/wheels"
+    wheels = list(wheels_dir.glob("*.whl"))
+    assert wheels, "no wheel produced by maturin"
```
sneakers-the-rat: can we try and import the wheel here?

tests/test_generators/test_rustgen_maturin.py
```diff
@@ -0,0 +1,128 @@
...
+    out_dir.mkdir(parents=True, exist_ok=True)
+
+    # Generate the crate with pyO3 bindings enabled
+    gen = RustGenerator(metamodel_sv.schema, mode="crate", pyo3=True, serde=True, output=str(out_dir))
+    gen.serialize(force=True)
+
+    cargo_toml = out_dir / "Cargo.toml"
+    pyproject_toml = out_dir / "pyproject.toml"
+    assert cargo_toml.exists(), "Cargo.toml not generated"
+    assert pyproject_toml.exists(), "pyproject.toml not generated"
```
sneakers-the-rat: sort of duplicating tests here. probably would be good to turn "build a rustgen schema" into a fixture so that we don't end up needing to copy/paste for every rustgen test.

examples/PersonSchema/rust/Cargo.toml
sneakers-the-rat, line 1: in general i feel like any examples should get auto-built or else they're guaranteed to go out of date. at least a test that regenerates and confirms that the examples are correct. out of date examples are sometimes worse than no examples.

sneakers-the-rat (commented):
just commenting on common/testing things. dont have time to review actual rust gen atm. pr is pretty humongous

tests/test_generators/test_rustgen_maturin.py
kervel: tx, indeed!

examples/PersonSchema/rust/Cargo.toml
kervel: you are right. There was a Makefile under personschema so i added the rust generation there. but it seems the Makefile is just sitting there, when i run "make" it seems also the pydantic and other examples were outdated. so probably this needs some extra thinking.

linkml/generators/common/template.py
kervel: ok.. actually on this whole "common" part i just wanted to verify with you if i didn't botch my merge/rebase ... i'm going to assume i didn't based on your feedback :-)

kervel:
> > SkipValidation rewrite looks broken.
> 
> Do you mean the version in this PR is broken, or was the original broken? It was working fine afaik, why change it?
> 
> > GetPydanticSchema is not a public pydantic v2 API
> 
> it is! https://docs.pydantic.dev/latest/concepts/types/#using-getpydanticschema-to-reduce-boilerplate
> 
> > Using a TypeVar inside Annotated as a module-level alias (Tsv) and then subscripted later is not how Annotated/generics are intended to be used
> 
> It is! https://typing.python.org/en/latest/spec/qualifiers.html#annotated

sorry about this! i used AI help to look at common.py because i was not deep into it enough myself to understand everything, and i tought i removed all AI rubbish before posting the comment, obviously i didn't...

tests/test_generators/test_rustgen_maturin.py
kervel: ok, i'll try. its not as simple to import a binary wheel. but it can be done.

kervel:
Hello @sneakers-the-rat

Thanks for the review! I think i addressed most of your issues, i don't know if the `common.py` now is good enough (see my comment). But if you think it is, i guess the PR is ready for another review (i'll ask who could be a candidate). Indeed reviewing properly will be tough because of the size of the PR..

linkml/generators/common/template.py
sneakers-the-rat: I'll take a closer look, I only took a skim yesterday, but i can do a more complete check. I won't have time for a full review of the rustgen and would hope we could get a lot of the validation by plugging it into the core spec tests, though I don't know how we would make that not be ridiculously expensive to test since we would need to compile every test and there are hundreds/thousands.

tfliss:
@kervel Ok, I checked out and ran/passed the test. I've wanted to ramp up on rust, so am planning to to review in some detail. Is there a particular timeline / deadline you have in mind?

kervel:
> @kervel Ok, I checked out and ran/passed the test. I've wanted to ramp up on rust, so am planning to to review in some detail. Is there a particular timeline / deadline you have in mind?

hey,

not particulary.. my main concern is that if it stays too long in a separate branch it will become unmergeable/bitrotted.. 
also, i was planning to add wasm bindings too to the rust generated crate, but that would make the already big PR even bigger, so maybe i should do that in a separate PR.

if you want to check the rust side too, it might be handy to look at a crate that actually uses the generated classes. For this, https://github.com/Kapernikov/rust-linkml-core/tree/new-rustgen is a candidate ("new-rustgen" branch!). its a rust schemaview and some tools with python bindings.

thanks already!
frank

tests/test_generators/test_rustgen.py
```diff
@@ -0,0 +1,142 @@
...
+    schema_path = Path(input_path("personinfo.yaml"))
+    out_dir = _generate_rust_crate(str(schema_path), Path(temp_dir) / "personinfo_rust")
+
+    # Try to build a wheel via maturin. This may require network access
+    # to fetch crates unless dependencies are cached.
+    _build_rust_crate(out_dir)
+    # Try importing the generated wheel/module and sanity-check a class
+    mod = _import_built_wheel(out_dir, module_name="personinfo")
+    assert hasattr(mod, "Person"), "Expected class 'Person' not found in module"
+
```
tfliss, line 117: I adapted Person data I had (see below) to check that it loaded/validated. Thanks for writing this test module, the additional check was quite easy to add.

I noticed the enumeration with values seem to be generatingd the incorrect value, and also that I needed to pass in an object for an associated address. Maybe there's a different approach that would accept nested dicts or JSON.

This will be the form of most of the compliance tests, and this test body should be easy to adapt for that (especially if you can round-trip back to a dict/JSON)

```python
    person_data = {
        "id": "1",
        "name": "P. One",
        "description": "Person One",
        "depicted_by": "http://example.org/image_one.jpg",
        "primary_email": "one@example.org",
        "birth_date": "1900-01-01",
        "age_in_years": 125,    
        "gender": "CisgenderMan",  # should be "cisgenger man"
        "current_address": mod.Address( # compliance tests will have a dict here.
            **{"street": "1 Maple Street", "city": "Springfield, AZ", "postal_code": "12345"}
        ),
        "telephone": "800-555-1111",
        "has_employment_history": None,
        "has_familial_relationships": None,
        "has_interpersonal_relationships": None,
        "has_medical_history": None,
        "has_news_events": None,
        "aliases": None
    }

    person = mod.Person(**person_data)
    assert person.name == "P. One"

```

tfliss:
@kervel As far as the compliance tests that @sierra-moxon recommended, it might be worth working on that as a separate issue. I'll comment in my review how you may want to document the current state of implementation.

This [draft PR in a fork](https://github.com/linkml/linkml/pull/2895/files) contains my current work-in-progress PolaRS generator. The relevant part for you would be tests/test_compliance/ (please ignore the rest).

- in helper.py:
    - add a constant for your generator: `RUST_GEN = "rust_gen"`
    - add an entry with the class in the GENERATORS dict: `RUST_GEN: generators.RustGenerator,`
    - in check_data() delegate the check to your generator:
    ```
    if isinstance(gen, GENERATORS[RUST_GEN]):
            check_data_rustgen(.....)
    ```
    - add the definition of check_data_rustgen()
        - note that I'm using `pl = pytest.importorskip("polars", minversion="1.0", reason="Polars >= 1.0 not installed")` to conditionally import an optional package and skip the tests if it's not present.
        - I check `schema['name']` to skip individual tests rather than touching every test_compliance/test_*.ply file.
            - could also just skip anything except specifically enabled tests to start.
        - The body of this check is similar to your `test_rustgen_personschema_maturin()`
            - target_class is the name of your model
            - object_to_validate will be a dict
            - schema is the linkml yaml
            - expected_behavior indicates whether your generator implements the feature or not
            - valid is whether the object_to_validate is expected to pass or fail validation.
            - my simplistic approach is to try to round-trip the object_to_validate back to a dict and then compare them
            - note: I'm not necessarily recommending this approach, it's just my current work-in-progress and is probably something you could use to get a basic compliance test running.
- in test_compliance.py:
    - hook up the constant you defined in helper.py to a pytest mark in CORE_FRAMEWORKS: `pytest.param(RUST_GEN), marks=[pytest.mark.with_panderagen]),`
 - in test_core_compliance.py:
     - look at the `_mapping` parameters for examples of comparing your generated code to what is expected for the input schema (most of them are very small schemas)
     - I found stepping through with a debugger as helpful for some of these tests, they are heavily parametrized (allso see --with-output).
 
 At this point I generally invoke using something like the following (you should need to use your pytests mark):
 
 ```
uv run pytest -s -vv --log-cli-level=INFO tests/test_compliance -m with_rustgen --with-output -k test_attributes | tee test.out
```

The `--with-output` flag will drop detailed information about each test under tests/test_compliance/output/ in a directory structure reflecting the test_compliance modules, individual tests and the various generators. 
        
There is also a [test_compliance/README.md](https://github.com/linkml/linkml/blob/main/tests/test_compliance/README.md)  

I hope this helps get you started. Feel free to reach out if anything doesn't make sense or doesn't work. I wouldn't expect it to take much to get these tests hoooked up (probably not all passing, but some should). If you get stuck I may have missed a detail. I'm also very open to suggestions for other approaches as I am also learning my way around the compliance tests.

**Comment by tfliss** [comment]
@kervel Here's a quick patch with mostly the same changes as described above, in this gist.: https://gist.github.com/tfliss/75462c9d5284973ce06f7f7773afc233

The run_one_test.sh script will run one parametrized enum test and generate output. 12 pass and 12 fail, which is great for the first try. I think some of the errors are due to needing name cleaning in a couple of places (schema name). The check is mostly a copy/paste from one of your tests, then instantiate the object and compare the top-level key/value pairs. Not production quality, but somewhere to start. Speed seems ok. Once the schema name thing is fixed we can run more of the suite.

Hope this helps, I can also push a branch or share the logs and generated code.

I'll continue the review next week. Have a great weekend. So far this has been great to work with. Thank you.

**Comment by sneakers-the-rat** [comment]
I think this would be great to break up into smaller PRs. I would recommend no new features in this one, and focus on testing.

I suspect that the rust tests requiring compilation for each schema will make the tests take an enormous amount of time, though we should measure that, so they will likely need to be split out into their own job like the slow tests. 

The compliance tests are definitely the target we need to eventually hit, but it may require a significant amount of work, since there currently isn't a good way to do "partial compliance" without manually editing the skip conditions in the body of each test. In the meantime, the checklist in the scoping issue may be a useful guide for what would be good to have in rustgen-specific tests. Without those it's hard to properly review, since its hard to know what is implemented, what isn't, and whether what is written works.

The compliance tests are designed so there is one schema per test (mostly, except for some parameterizations), which would mean one compilation per test, but for many of the base-case rustgen boilerplate tests, we could probably get away with a module-scoped fixture that compiles the kitchen sink schema once and use that for everything.

**Comment by kervel** [comment]
> Hope this helps, I can also push a branch or share the logs and generated code.

Hello! i just pushed a branch with just your gist in it. I'll try to look at it in more detail soon. Thanks!
Frank

File: linkml/generators/rustgen/template.py [hunk 8]
```diff
@@ -0,0 +1,797 @@
...
+    """
+
+    type_getter: str
+    needs_lifetime: bool
+
+
+def _needs_lifetime(sig: str) -> bool:
+    """Heuristic to detect if a signature string requires an explicit lifetime.
+
+    We consider borrowed types (`&T`), container views (`SeqRef`/`MapRef`), and
```
tfliss, line 243: I'm reading that explicit lifetimes are only necessary for multiple borrowed references. I'm not a rust programmer, so maybe this is ok since it's a code generator, or maybe it's worth noting in the documentation / comment.

docs/generators/rust.rst
```diff
@@ -0,0 +1,194 @@
...
+
+    pub trait MedicalEvent: Event {
+        fn in_location<'a>(&'a self) -> Option<&'a str>;
+        fn diagnosis<'a>(&'a self) -> Option<&'a crate::DiagnosisConcept>;
+        fn procedure<'a>(&'a self) -> Option<&'a crate::ProcedureConcept>;
+    }
+
+    impl Event for crate::MedicalEvent {
+            fn started_at_time(&self) -> Option<&NaiveDate> {
+            return self.started_at_time.as_ref();
```
tfliss: I'm seeing that `return` is considered 'redundant' or not recommended rust style in this situation. I understand if it's needed in some cases in the generator, but if not, could it be removed from the template?

linkml/generators/rustgen/rustgen.py
```diff
@@ -0,0 +1,1037 @@
...
+        has_default=not (s.required or False) or (s.multivalued or False),
+        containerType=(
+            ContainerType.LIST
+            if container_mode == SlotContainerMode.LIST
+            else ContainerType.MAPPING
+            if container_mode == SlotContainerMode.MAPPING
+            else None
+        ),
+        child_ranges=sub_ranges if len(sub_ranges) > 1 else None,
+        box_needed=inline_mode == SlotInlineMode.INLINE and can_contain_reference_to_class(s, cls, sv),
```
tfliss, line 315: I understand the boxing might have significant overhead compared to an enum approach. But it seems appropriate for a generally open-world assumption.

Consider in the future a user might want to have an option/flag to assume a closed model and generate code accordingly. I don't know how realistic that is to implement, just a comment, not a requested change for this PR.

docs/generators/rust.rst
```diff
@@ -0,0 +1,194 @@
...
+
+The structs:
+`lib.rs <https://github.com/linkml/linkml/tree/main/examples/PersonSchema/rust/src/lib.rs>`_
+
+The traits:
+`poly.rs <https://github.com/linkml/linkml/tree/main/examples/PersonSchema/rust/src/poly.rs>`_
+
+Overview
+--------
+
```
tfliss, line 19: Please note that the generator is still under development and that a feature compliance list is below (or your preferred wording).

docs/generators/rust.rst
```diff
@@ -0,0 +1,194 @@
...
+.. autoclass:: RustGenerator
+    :members:
+
+Features
+--------
+
+- Serde: Code that depends on Serde is behind the Cargo feature ``serde`` (``#[cfg(feature = "serde")]``).
+- PyO3: Python bindings are behind the Cargo feature ``pyo3`` (``#[cfg(feature = "pyo3")]``).
+- Enable features when building your crate (e.g., ``--features serde,pyo3``) to include the corresponding code paths.
+
```
tfliss, line 230: This may be a good spot to indicate which major parts of the compliance suite are fully or mostly passing and which are skipped. I think for an initial implementation a hand-compiled high-level list is ok, I don't know the state of automatic generation of that report, and the rust helper would need to be improved in any case.

tests/test_generators/test_rustgen.py
```diff
@@ -0,0 +1,142 @@
...
+from linkml_runtime.utils.introspection import package_schemaview
+
+from linkml.generators.rustgen import RustGenerator
+
+pytestmark = [
+    pytest.mark.rustgen,
+]
+
+
+def _generate_rust_crate(schema_input, out_dir: Path) -> Path:
```
tfliss, line 18: consider promoting or factoring out some of this functionality so it could be also used in a compliance test helper.  Also consider putting some of the compilation steps into fixtures. You can `yield` a generated artifact and then do cleanup after the yield.

linkml/generators/rustgen/templates/pyproject.toml.jinja
```diff
@@ -0,0 +1,10 @@
+[project]
+name = "{{ name }}"
+version = "{{ version }}"
+
+[build-system]
+requires = ["maturin>=1.0,<2.0"]
```
tfliss, line 6: Can this version be paramtrized?

linkml/generators/rustgen/rustgen.py
```diff
@@ -0,0 +1,1037 @@
...
+    * If ``mode == "file"`` , a file with a ``.rs`` extension
+
+    If output is not provided at object instantiation,
+    it must be provided on a call to :meth:`.serialize`
+    """
+
+    _environment: Optional[Environment] = None
+
+    def __post_init__(self):
+        # TODO: consider moving up a level
```
tfliss: Is this TODO still an issue?

linkml/generators/rustgen/rustgen.py
```diff
@@ -0,0 +1,1037 @@
...
+            type_=RustTypeAlias(
+                name=get_name(type_), type_=get_rust_type(type_.base, self.schemaview, self.pyo3), pyo3=self.pyo3
+            ),
+            imports=self.get_imports(type_),
+        )
+        slot = self.after_generate_type(res, self.schemaview)
+        return slot
+
+    def generate_enum(self, enum: EnumDefinition) -> EnumResult:
+        # TODO: this
```
tfliss: Is this TODO still an issue?

linkml/generators/rustgen/rustgen.py
```diff
@@ -0,0 +1,1037 @@
...
+
+        imports = DEFAULT_IMPORTS.model_copy()
+        imports += PYTHON_IMPORTS
+        imports += SERDE_IMPORTS
+        if need_merge_crate:
+            imports += MERGE_IMPORTS
+        for result in [*enums, *slots, *classes]:
+            imports += result.imports
+
+        # TODO: get imports from all results
```
tfliss: Is this TODO still needed?

linkml/generators/rustgen/rustgen.py
```diff
@@ -0,0 +1,1037 @@
...
+        else:
+            raise ValueError(f"Invalid generation mode: {mode}")
+
+        return output
+
+    @property
+    def template_environment(self) -> Environment:
+        if self._environment is None:
+            self._environment = RustTemplateModel.environment()
+        # if self.template_dir is not None:
```
tfliss: Please remove commented code unless needed.

linkml/generators/rustgen/template.py
```diff
@@ -0,0 +1,797 @@
...
+        For class ranges, setters use a type parameter `E` constrained as
+        `Into<T>` to allow ergonomic conversions. Non-class ranges do not
+        require a bound and return None.
+        """
+        if self.is_class_range:
+            tp = self.type_
+            return f"Into<{tp}>"
+        return None
+
+    def element_type_value(self, crateref: Optional[str]) -> str:
```
tfliss: I'm not seeing coverage on element_type_value or element_type_borrowed.
That might be ok, but it surprised me, I haven't looked further at this time.

tfliss: Actually I did a quick grep and suspect they're currently unused. If that's the case and you don't plan to use them in the future, please remove.

tfliss (changes requested):
Overall to me this looks like a solid and useful initial generator implementation. Thank you @kervel 

I've looked over pretty much everything, with an assist from an LLM on the generated PersonSchema rust. Ran the makefile, ran a lot of compliance tests in a branch (some passed, some failed), coverage looks quite good. I did not go over the jinja2 templates in detail, but had eyes on it all and the style and readability are high quality.

If anything is split into a separate PR I'd like to see the linkml/generators/common changes be reviewed more widely, and that would be fairly small.

Other than that I'd want to see a note in the .rst with respect to general compliance testing being a work-in-progress. Also please look over the remaining TODO comments to check if they are still relevant, some commented out code fragments and the two functions that might be unused per coverage. If those changes are made I will approve this PR (but still recommend a final approval by a core dev).

Have a good weekend!

kervel:
Hello @tfliss , thanks for your review! im going to start working on your feedback now!

linkml/generators/rustgen/template.py
kervel: hi,

This is to solve a new warning introduced with the latest rust release: when you already have a lifetime parameter in a function, then everything that is bound by that lifetime should have that parameter explicitly mentioned.

the code compiles without but you'll get tons of warnings.

greetings,
Frank

linkml/generators/rustgen/rustgen.py
kervel: hello, the boxing is needed to keep structs sized, and unfortunately it doesn't make the generated code more open-world ...

basically you can't have a Mystruct with a member of type Mystruct in rust, because that would lead to infinite size. The inner Mystruct needs to be allocated on the heap (hence the box)

see this: https://stackoverflow.com/questions/25296195/why-are-recursive-struct-types-illegal-in-rust

tests/test_generators/test_rustgen.py
kervel: argh good catch.
* i have data normalisation implemented but only for json/yaml -> rust, not for python -> rust.
* enums --> real issue, will fix!

while waiting for the real compliance testing i'll add your testcase as a test.

kervel:
> If anything is split into a separate PR I'd like to see the linkml/generators/common changes be reviewed more widely, and that would be fairly small.

I actually did that, here https://github.com/linkml/linkml/pull/2722  which then sat there for 3 months until it was no longer mergeable, so i closed it created this version rebased on a newer main branch ...

tests/test_generators/test_rustgen.py
kervel: its fixed (i hope). unfortunately i can't share all code between the serde->rust and the python->rust paths, so the latest commit is a bit bigger than i'd want.

linkml/generators/rustgen/template.py
kervel: tx!

docs/generators/rust.rst
kervel: i have put something but i'm not sure if it is what you want.. my best "view" on the completeness of this is the list @sneakers-the-rat originally made in #2360

tfliss: Yes, including the checklist from 2360 in some form would work for me.

If any of the features are implemented but fail a lot of compliance edge cases, maybe some 'partial' mark would be informative for users.

kervel:
hello @tfliss i think i treated your feedback now! the most important point was the normalisation of values coming from pyo3 + the rust.rst doc.

cmungall (approved):

sneakers-the-rat:
it's already merged but fwiw changes to common/pydanticgen look fine. 

thanks for carrying this through @kervel :)

PR #2893, opened 2025-09-04
Add Solr schema generator

kevinschaper:
Claude generated (warning!) solr schema generator, prompted to take the schema generation from linkml-solr and convert it to using SchemaView

### DIFF
--- a/linkml/generators/solrgen.py
+++ b/linkml/generators/solrgen.py
@@ -1,0 +1,203 @@
+"""
+Generate Solr schema definitions from a LinkML schema.
+
+The output is a JSON document representing Solr operations such as add-field.
+Note that this makes a flattened 'union' schema for all classes, as Solr does not 
+have the concept of classes/records.
+"""
+
+import json
+import os
+from dataclasses import dataclass
+from typing import Dict, List, Optional, Union
+
+import click
+from linkml_runtime.linkml_model.meta import ClassDefinitionName, SlotDefinition
+from linkml_runtime.utils.schemaview import SchemaView
+from pydantic import BaseModel
+
+from linkml._version import __version__
+from linkml.utils.generator import Generator, shared_arguments
+
+
+# Solr Schema Model Classes (inline Pydantic models)
+class Field(BaseModel):
+    """Represents a Solr field definition."""
+    name: str
+    type: str
+    stored: Optional[bool] = None
+    indexed: Optional[bool] = None
+    multiValued: Optional[bool] = None
+
+
+class DynamicField(BaseModel):
+    """Represents a Solr dynamic field definition."""
+    name: str
+    type: str
+    stored: Optional[bool] = None
+
+
+class CopyField(BaseModel):
+    """Represents a Solr copy field definition."""
+    source: str
+    dest: Union[str, List[str]]
+    maxChars: Optional[int] = None
+
+
+class FieldType(BaseModel):
+    """Represents a Solr field type definition."""
+    name: str
+    class_: Optional[str] = None
+    positionIncrementGap: Optional[str] = None
+
+
+class Transaction(BaseModel):
+    """Represents a Solr schema transaction with multiple operations."""
+    add_field: List[Field] = []
+    delete_field: List[str] = []
+    replace_field: List[str] = []
+    add_field_type: List[FieldType] = []
+    delete_field_type: List[str] = []
+    replace_field_type: List[str] = []
+    add_dynamic_field: List[DynamicField] = []
+    delete_dynamic_field: List[str] = []
+    replace_dynamic_field: List[str] = []
+
+
+# Map from underlying python data type to solr equivalent
+# Note: The underlying types are a union of any built-in python datatype + any type defined in
+#       linkml-runtime/utils/metamodelcore.py
+# Note the keys are all lower case
+solr_schema_types: Dict[str, str] = {
+    "int": "int",
+    "integer": "int",
+    "str": "string",
+    "string": "string",
+    "bool": "boolean",
+    "boolean": "boolean",
+    "float": "pfloat",
+    "double": "pdouble",
+    "decimal": "pdouble",
+    "xsddate": "date",
+    "xsddatetime": "date",
+    "xsdtime": "time"
+}
+
+
+@dataclass
+class SolrSchemaGenerator(Generator):
+    """
+    Generates a Solr schema from a LinkML schema.
+
+    The output is a JSON document representing Solr operations such as add-field.
+    Note that this makes a flattened 'union' schema for all classes, as Solr does not
+    have the concept of classes/records.
+    """
+    
+    # ClassVars
+    generatorname = os.path.basename(__file__)
+    generatorversion = __version__
+    valid_formats = ["json"]
+    uses_schemaloader = False
+    visit_all_class_slots = True
+    
+    # Instance variables
+    schemaview: SchemaView = None
+    transaction: Transaction = None
+
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        self.schemaview = SchemaView(self.schema)
+        self.transaction = Transaction()
+
+    def _transaction_json(self, transaction: Transaction) -> str:
+        """Convert transaction to JSON format expected by Solr."""
+        # Convert to dict and transform underscore keys to hyphen keys for Solr
+        obj = transaction.model_dump(exclude_unset=True, exclude_none=True)
+        obj = {k.replace('_', '-'): v for k, v in obj.items()}
+        return json.dumps(obj, indent=2)
+
+    def class_schema(self, class_name: Union[str, ClassDefinitionName]) -> str:
+        """Generate schema for a single class."""
+        transaction = self.get_transaction(class_name)
+        return self._transaction_json(transaction)
+
+    def serialize(self, **kwargs) -> str:
+        """Generate schema for all classes."""
+        transaction = self.get_transaction()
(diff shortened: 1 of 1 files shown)


packages/linkml/src/linkml/generators/solrgen.py
```diff
@@ -0,0 +1,203 @@
...
+    add_field: List[Field] = []
+    delete_field: List[str] = []
+    replace_field: List[str] = []
+    add_field_type: List[FieldType] = []
+    delete_field_type: List[str] = []
+    replace_field_type: List[str] = []
+    add_dynamic_field: List[DynamicField] = []
+    delete_dynamic_field: List[str] = []
+    replace_dynamic_field: List[str] = []
+
```
amc-corey-cox, line 65: **Dead code: ~60% of models are unused scaffolding**

`DynamicField`, `CopyField`, `FieldType`, and all the `delete_*`/`replace_*` fields on `Transaction` are never used — not in the generator, not in tests. Only `Field` and `Transaction.add_field` are actually exercised. This is ~40 lines of dead code in a 203-line file.

If these are needed later they can be added then — right now they just add noise and maintenance surface.

packages/linkml/src/linkml/generators/solrgen.py
```diff
@@ -0,0 +1,203 @@
...
+    def class_schema(self, class_name: Union[str, ClassDefinitionName]) -> str:
+        """Generate schema for a single class."""
+        transaction = self.get_transaction(class_name)
+        return self._transaction_json(transaction)
+
+    def serialize(self, **kwargs) -> str:
+        """Generate schema for all classes."""
+        transaction = self.get_transaction()
+        return self._transaction_json(transaction)
```
amc-corey-cox, line 128: **Bypasses Generator's visitor pattern entirely**

This `serialize()` override replaces the base class's visitor-pattern infrastructure with a from-scratch implementation. The base `Generator.serialize()` walks the schema elements and calls `visit_*`/`end_*` hooks — that's the contract generators are expected to use (see `golrgen.py`, `graphqlgen.py`, etc.).

As a consequence, `__post_init__` also creates a redundant `self.schemaview = SchemaView(self.schema)` on line 110, shadowing the one the base class already provides. And `self.transaction` (line 111) is set but never read — both `serialize()` and `get_transaction()` create their own local `Transaction()`.

This looks like AI-generated code that didn't understand the base class rather than a deliberate design choice. I'd recommend either:
1. Refactoring to use the visitor pattern (`visit_class`, `visit_class_slot`), or
2. At minimum, removing the redundant `schemaview`/`transaction` instance vars and adding a comment explaining why the visitor pattern doesn't fit here (if there's a real reason).

packages/linkml/src/linkml/generators/solrgen.py
```diff
@@ -0,0 +1,203 @@
...
+        
+        if solr_type not in solr_schema_types.values():
+            # Unknown type, default to string
+            solr_type = "string"
+        
+        return Field(
+            name=slot.name,
+            type=solr_type,
+            multiValued=slot.multivalued
+        )
```
amc-corey-cox, line 182: **`multiValued: false` emitted for every non-multivalued field**

`slot.multivalued` is `False` (not `None`) for non-multivalued slots, so `exclude_none=True` in `_transaction_json` won't filter it out. Every field will include `"multiValued": false` in the output, which is unnecessary noise.

Should be:
```python
multiValued=True if slot.multivalued else None
```

packages/linkml/src/linkml/generators/solrgen.py
```diff
@@ -0,0 +1,203 @@
...
+        return self._transaction_json(transaction)
+
+    def get_transaction(self, class_name: Union[str, ClassDefinitionName] = None) -> Transaction:
+        """Build transaction with fields from specified class or all classes."""
+        transaction = Transaction()
+        sv = self.schemaview
+        field_dict = {}
+        
+        for cn in sv.all_classes():
+            if class_name is None or str(cn) == str(class_name):
```
amc-corey-cox, line 137: **`--top-class` comparison is case-sensitive and may silently produce empty output**

`str(cn) == str(class_name)` is a case-sensitive comparison, but the CLI test passes `"employee"` (lowercase) while the schema defines `Employee`. This works only because `SchemaView.all_classes()` happens to return snake_case/lowercase keys — but that's an implementation detail, not a contract.

If a user passes `--top-class Employee` (matching the schema definition name), it may not match depending on how SchemaView normalizes. Consider using `sv.get_class(class_name)` to resolve the name properly, then comparing canonical names.

packages/linkml/src/linkml/generators/solrgen.py
```diff
@@ -0,0 +1,203 @@
...
+            # Enum - treat as string
+            solr_type = "string"
+        elif range_name is None:
+            # Use default range
+            default_range = sv.schema.default_range or "string"
+            solr_type = solr_schema_types.get(str(default_range).lower(), "string")
+        elif range_name in sv.all_types():
+            # LinkML type
+            type_def = sv.get_type(range_name)
+            base_type = type_def.typeof or type_def.base
```
amc-corey-cox, line 165: **Type resolution doesn't follow the `typeof` chain**

LinkML types can typedef other types (e.g., `uri` → `uriorcurie` → `string`). This code only looks one level deep with `type_def.typeof or type_def.base`. If the chain is deeper, it won't resolve to the base type and will fall through to the default `"string"`.

Other generators handle this by following `typeof` in a loop until reaching a type with a `base`. `SchemaView` may have helpers for this — worth checking.

packages/linkml/src/linkml/generators/solrgen.py
```diff
@@ -0,0 +1,203 @@
...
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        self.schemaview = SchemaView(self.schema)
+        self.transaction = Transaction()
+
+    def _transaction_json(self, transaction: Transaction) -> str:
+        """Convert transaction to JSON format expected by Solr."""
+        # Convert to dict and transform underscore keys to hyphen keys for Solr
+        obj = transaction.model_dump(exclude_unset=True, exclude_none=True)
+        obj = {k.replace('_', '-'): v for k, v in obj.items()}
```
amc-corey-cox, line 117: **`replace('_', '-')` is fragile for key transformation**

This blanket replacement will break if any future Pydantic field name legitimately contains an underscore (e.g., `add_field_type` → `add-field-type` works today, but it's a coincidence that all Solr API keys use hyphens). A safer approach would be an explicit mapping dict or using Pydantic's `alias` feature to define the Solr key names directly on the model fields.

packages/linkml/src/linkml/generators/solrgen.py
```diff
@@ -0,0 +1,203 @@
...
+    generator = SolrSchemaGenerator(yamlfile, **kwargs)
+    if top_class:
+        output = generator.class_schema(top_class)
+    else:
+        output = generator.serialize(**kwargs)
+    print(output)
+
+
+if __name__ == '__main__':
+    cli()
```
amc-corey-cox, line 203: Nit: missing trailing newline.

tests/linkml/test_generators/test_solrgen.py
```diff
@@ -0,0 +1,293 @@
...
+    schema = json.loads(output)
+    fields = {f["name"]: f for f in schema["add-field"]}
+
+    # String fields should map to 'string' type
+    assert fields["id"]["type"] == "string"
+    assert fields["name"]["type"] == "string"
+
+    # Integer fields should map to 'int' type
+    if "age_in_years" in fields:
+        assert fields["age_in_years"]["type"] == "int"
```
amc-corey-cox, line 57: **Tests silently skip assertions behind `if` guards**

This pattern appears in several tests (`test_solr_field_types`, `test_solr_multivalued_fields`, `test_solr_class_references`):

```python
if "age_in_years" in fields:
    assert fields["age_in_years"]["type"] == "int"
```

If the field doesn't exist, the assertion is silently skipped and the test passes without testing anything. These should be unconditional — if the field is expected, assert it exists. If the test schema might not have it, the test needs a schema that does.

tests/linkml/test_generators/test_solrgen.py
```diff
@@ -0,0 +1,293 @@
...
+    all_fields = {f["name"] for f in schema_all["add-field"]}
+
+    # Generate for Employee class only
+    generator_employee = SolrSchemaGenerator(schema_path)
+    output_employee = generator_employee.class_schema("employee")
+    schema_employee = json.loads(output_employee)
+    employee_fields = {f["name"] for f in schema_employee["add-field"]}
+
+    # Employee-specific fields should be present
+    assert "first_name" in employee_fields or "first name" in employee_fields
```
amc-corey-cox, line 115: This `or` assertion is testing two different things and will mask failures. If neither `first_name` nor `first name` is present, it fails — but if the wrong one is present, it passes silently. The test should know which form to expect from the schema and assert that specific one.

packages/linkml/src/linkml/generators/solrgen.py
amc-corey-cox: I'm not certain about this one - I didn't dig any deeper and this may be incorrect.

packages/linkml/src/linkml/generators/solrgen.py
amc-corey-cox: This one I would also defer to expert knowledge. There may be good reason to treat this as a contract but I don't have the depth to know.

PR #2894, opened 2025-09-05
fix(rdfgen): binary-safe output with -o

lapkinvladimir:
Binary RDF serializers (e.g., [Jelly](https://w3id.org/jelly)) could raise a UnicodeDecodeError when writing to stdout.

With -o/--output, rdfgen now lets RDFLib write directly to the file (Graph.serialize with destination).
Binary serializers write to the file, leaving stdout empty.

Behaviour without -o is unchanged.

Tests:
Unit: mock Graph.serialize; verify file writing, empty stdout, and ttl still maps to turtle.
Integration: run with Jelly if available; check empty stdout and binary file output.

### DIFF
--- a/linkml/generators/rdfgen.py
+++ b/linkml/generators/rdfgen.py
@@ -67,12 +67,21 @@ class RDFGenerator(Generator):
             base=str(self.namespaces._base),
             prefix=True,
         )
-        out = self._data(graph)
+        fmt = "turtle" if self.format == "ttl" else self.format
         if output:
+            # Binary-safe when -o/--output is used:
+            # delegate to RDFLib (Graph.serialize(destination=..., format=...)).
+            # Serializers that produce bytes write directly to the file; stdout stays empty.
+            try:
+                out = graph.serialize(format=fmt)
+            except UnicodeDecodeError:
+                graph.serialize(destination=output, format=fmt)
+                return ""
             with open(output, "w", encoding="UTF-8") as outf:
                 outf.write(out)
+            return out
 
-        return out
+        return self._data(graph)
 
 
 @shared_arguments(RDFGenerator)
--- a/tests/test_generators/test_rdfgen_binary.py
+++ b/tests/test_generators/test_rdfgen_binary.py
@@ -1,0 +1,97 @@
+from pathlib import Path
+from typing import Dict, Any
+
+import pytest
+from rdflib import Graph
+from linkml.generators.rdfgen import RDFGenerator
+
+
+def _write_min_schema(p: Path) -> Path:
+    p.write_text(
+        "id: http://example.org/min\n"
+        "name: min\n"
+        "prefixes:\n"
+        "  ex: http://example.org/\n"
+        "default_curi_maps:\n"
+        "  - semweb_context\n"
+        "classes:\n"
+        "  A: {}\n",
+        encoding="utf-8",
+    )
+    return p
+
+
+def test_with_output_binary_path_on_decode_error(monkeypatch, tmp_path):
+    """On UnicodeDecodeError: write via destination, keep stdout empty."""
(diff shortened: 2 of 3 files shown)


Ostrzyciel (approved):
Makes sense to me and it works.

It's a bit of a hack, but I can't see how to make this better with how currently these methods are laid out. Maybe this could use a small refactor? Easiest way to clean it up would be to return `bytes` instead of `str`. But that's beyond the scope of this PR, I think.

linkml/generators/rdfgen.py
```diff
@@ -67,12 +67,21 @@ def end_schema(self, output: Optional[str] = None, context: str = None, **_) ->
             base=str(self.namespaces._base),
             prefix=True,
         )
-        out = self._data(graph)
+        fmt = "turtle" if self.format == "ttl" else self.format
         if output:
```
Silvanoc: Variable `fmt` is only needed `if output == True`, right? Then I'd rather put it inside of the `if` block.

```suggestion
        if output:
            fmt = "turtle" if self.format == "ttl" else self.format
```

tests/test_generators/rdf/test_rdfgen_binary.py
Silvanoc, line 1: *Same comment applies to the other test file being added in this PR, therefore I'll simply copy this same comment.*

The directory `tests/test_generators/` IMO already has too many files and the current pattern is sort of 1 file per generator. Therefore I'd
1. either integrate the tests in this file into `tests/test_generators/test_rdfgen.py`
2. or create the directory `tests/test_generators/rdf` and place this file into it (similar to the way it's done for the Python and SHACL generators)

tests/test_generators/rdf/test_rdfgen_binary_integration.py
Silvanoc, line 1: *Same comment applies to the other test file being added in this PR, therefore I'll simply copy this same comment.*

The directory `tests/test_generators/` IMO already has too many files and the current pattern is sort of 1 file per generator. Therefore I'd
1. either integrate the tests in this file into `tests/test_generators/test_rdfgen.py`
2. or create the directory `tests/test_generators/rdf` and place this file into it (similar to the way it's done for the Python and SHACL generators)

linkml/generators/rdfgen.py
lapkinvladimir: Good point, moved `fmt` inside the `if output:` block.

tests/test_generators/rdf/test_rdfgen_binary.py
lapkinvladimir: Yep, makes sense. I moved the binary-related tests into new folder `tests/test_generators/rdf/`.

lapkinvladimir:
Several tests failed due to network errors, as I noticed.  
This looks unrelated to my changes, so as I understand, it just needs to be rerun and everything should pass.

Silvanoc:
> Several tests failed due to network errors, as I noticed. This looks unrelated to my changes, so as I understand, it just needs to be rerun and everything should pass.

That's right. But please take into account that it's probably some throttling pulling some resources. So retrying too many and too quickly might end up with the same result. Just be patient and retry after some time, not straightaway.

Silvanoc:
> Makes sense to me and it works.
> 
> It's a bit of a hack, but I can't see how to make this better with how currently these methods are laid out. Maybe this could use a small refactor? Easiest way to clean it up would be to return `bytes` instead of `str`. But that's beyond the scope of this PR, I think.

I would also go for small-scoped PRs adding improvements/clean-ups/optimizations on a separate PR.
Discussions about clean-ups/improvements/optimizations on changing external interfaces (refactoring in general) are frequently more code-focused and in some cases don't require any expertise about the functionality, but rather about the Python.
Whereas PRs adding/changing functionality typically require people with more expertise about the functionality (e.g. RDF in this case).

PR #2897, opened 2025-09-08
[draft] compliance testing for rustgen

kervel:
for now this is nothing but Tim's patch (https://gist.github.com/tfliss/75462c9d5284973ce06f7f7773afc233), on top of the rustgen-rebased PR (which needs to go in first)

see also https://github.com/linkml/linkml/pull/2889#issuecomment-3260086651

more work to come


sneakers-the-rat:
Looks like this either runs indefinitely or takes a long time to run, probably want to disable this from running in CI to avoid taking up runner time until it works/runs in a reasonable amount of time: https://github.com/linkml/linkml/actions/runs/17554754670/job/49855880125?pr=2897

PR #2778, opened 2025-06-19
Revert commit 9c0bfee6 "Fix how inlined types are managed by the JSON-LD context generator"

hadjian:
This reverts commit 9c0bfee65707fa564591e377ab71ae16c8fe19f3.

Bug reported in issue #2772

PR #2900, opened 2025-09-08
Correct misconception about use of `@type` in JSON-LD `@context`

### LINKED ISSUE #2772: Inlined Classes: Misinterpretation of the JSON-LD standard in PR #2461
---
name: Inlined Classes generate wrong JSON-LD context
title: ''Misinterpretation of the JSON-LD standard in PR #2461"
labels: bug

---

**Describe the bug**
I think PR #2461 misinterpreted the JSON-LD standard, which resulted in using the class IRI specified in the range of a slot as the ```@type``` in the JSON-LD context. 

1.9.2

I added the ```spouse``` slot to the personinfo example:
```
id: https://w3id.org/linkml/examples/personinfo
name: personinfo
prefixes:
  personinfo: https://w3id.org/linkml/examples/personinfo#
imports:
  - linkml:types
default_range: string
default_prefix: personinfo

classes:
  Person:
    attributes:
      id:
      full_name:
      aliases:
      phone:
      age:
    slots:
      - spouse

slots:
  spouse:
    range: Person
    inlined: true
```
The generated context below (the ```spouse``` term has an ```@type``` of ```personinfo:Person``` ):
```
{
   "@context": {
      "personinfo": "https://w3id.org/linkml/examples/personinfo#",
      "@vocab": "https://w3id.org/linkml/examples/personinfo#",
      "age": {
         "@id": "age"
      },
      "aliases": {
         "@id": "aliases"
      },
      "full_name": {
         "@id": "full_name"
      },
      "id": {
         "@id": "id"
      },
      "phone": {
         "@id": "phone"
      },
      "spouse": {
         "@type": "personinfo:Person",
         "@id": "spouse"
      },
      "Person": {
         "@id": "Person"
      }
   }
}
```
**Replicate the Bug**

Using the generated context, I prepared an example on the JSON-LD playground: https://tinyurl.com/5n75yjwc

If you inspect the "n-quads" tab, you will see that the object of the spouse triple reads:

```
"personinfo:f1922f1d-409f-402d-b44b-9248633ac9f6"^^<https://w3id.org/linkml/examples/personinfo#Person>
```

This is a typed string and should be an IRI. IMHO, this is the confusion between [typing node objects][0] and [typing value objects][1] in the JSON-LD standard. Both use the keyword ```@type```, but one designates the type of a resource as in RDF:

```
:john rdf:type :Person
```
While the other designates datatypes for strings to enable the mapping of JSON strings to RDF literals, as in:

```
"3"^^^xsd:integer
"ISBN 978-3-8668-0192-9"^^^schema:isbn
```
As IRIs in RDF are identified in lexical space, they would not appear as quoted strings, but with angle brackets:

```
<http://www.example.com>
```
You can also observe in the JSON-LD playground that the spouse is not embedded anymore, because the information that it is an IRI is lost. 

To rectify the example, switch to the "framed" tab in the playground example and change in the context the ```@type``` in the ```spouse``` term to ```@id``` again, but both in the input document and in the frame document or open this link, where I applied the change: https://tinyurl.com/2p7hjvxc

Now you can see that the spouse object is embedded and the "n-quads" tab shows the ID of John correctly as an IRI literal.

Also mind the note in the standard:

---
**Note**
The use of @type to define a type for both [node objects](https://www.w3.org/TR/json-ld11/#dfn-node-object) and [value objects](https://www.w3.org/TR/json-ld11/#dfn-value-object) addresses the basic need to type data, be it a literal value or a more complicated resource. Experts may find the overloaded use of the @type keyword for both purposes concerning, but should note that Web developer usage of this feature over multiple years has not resulted in its misuse due to the far less frequent use of @type to express typed literal values.
---

I strongly disagree with this observation ;-)

@vincentkelleher 

[0]: https://www.w3.org/TR/json-ld11/#node-objects 
[1]: https://www.w3.org/TR/json-ld11/#value-objects

Silvanoc:
PR #2461 introduced the error raised by issue #2772: term `@type`s in the `@context` cannot specify the type of an attribute object.

It's taken me some time discussing offline with @hadjian to correct my misconceptions in this area. Therefore I initially was not convinced that @hadjian was right. But now that I finally got to the depth needed to understand the issue, I can confirm the need for reverting the changes introduced by #2461.

We might need to get the people motivating PR #2461 involved to find out how to address their issues in a correct JSON-LD way.

The error probably originates from the erroneous expectation to get from the `@context` more than it really provides...

The [`@context`'s goal is according the specification](https://www.w3.org/TR/json-ld11/#the-context)
> Simply speaking, a [context](https://www.w3.org/TR/json-ld11/#dfn-context) is used to map [terms](https://www.w3.org/TR/json-ld11/#dfn-term) to [IRIs](https://tools.ietf.org/html/rfc3987#section-2).

JSON-LD is only an alternative representation of RDF. Therefore it is about triples: subject -> predicate -> object.

The `@type` inside of `@context` basically allow you to tell if the "triple object" is
* a leaf in the graph (AKA [external vertex](https://en.wikipedia.org/wiki/Tree_(graph_theory)#Tree)), which will be a string that you might get informed on how to interpret it (e.g. `xsd:integer`)
* the identifier of a node in the graph (AKA [internal vertex](https://en.wikipedia.org/wiki/Tree_(graph_theory)#Tree) without any information about it's type (and here appears the misconception driving to the current issue)

The only place to give information about node typing or shaping the JSON document is in the data (outside of the `@context` part). In the data you can have [embedded objects](https://www.w3.org/TR/json-ld11/#embedding), explicit typing,... You can also use [framing](https://www.w3.org/TR/json-ld11-framing/) to change the shape/representation of your data. But it has nothing to do with the `@context`!

This PR overrides #2778 because `main` has evolved that far from the commit introducing the bug, that reverting the commit does not work anymore.

Fixes #2772

### DIFF
--- a/tests/test_base/__snapshots__/annotations.context.jsonld
+++ b/tests/test_base/__snapshots__/annotations.context.jsonld
@@ -9,7 +9,7 @@
       "linkml": "https://w3id.org/linkml/",
       "@vocab": "https://w3id.org/linkml/",
       "annotations": {
-         "@type": "linkml:Annotation",
+         "@type": "@id",
          "@id": "annotations"
       },
       "tag": {
@@ -17,11 +17,11 @@
          "@id": "tag"
       },
       "value": {
-         "@type": "linkml:Any",
+         "@type": "@id",
          "@id": "value"
       },
       "extensions": {
-         "@type": "linkml:Extension",
+         "@type": "@id",
          "@id": "extensions"
       },
       "@base": "https://w3id.org/linkml/",
--- a/tests/test_base/__snapshots__/annotations.json
+++ b/tests/test_base/__snapshots__/annotations.json
@@ -498,7 +498,7 @@
       "linkml": "https://w3id.org/linkml/",
       "@vocab": "https://w3id.org/linkml/",
       "annotations": {
-        "@type": "linkml:Annotation",
+        "@type": "@id",
         "@id": "annotations"
       },
       "tag": {
@@ -506,11 +506,11 @@
         "@id": "tag"
       },
       "value": {
-        "@type": "linkml:Any",
+        "@type": "@id",
         "@id": "value"
       },
       "extensions": {
-        "@type": "linkml:Extension",
+        "@type": "@id",
         "@id": "extensions"
       },
       "Annotatable": {
--- a/tests/test_base/__snapshots__/extensions.context.jsonld
+++ b/tests/test_base/__snapshots__/extensions.context.jsonld
@@ -13,11 +13,11 @@
          "@id": "tag"
       },
       "value": {
-         "@type": "linkml:Any",
+         "@type": "@id",
          "@id": "value"
       },
       "extensions": {
-         "@type": "linkml:Extension",
+         "@type": "@id",
          "@id": "extensions"
       },
       "@base": "https://w3id.org/linkml/",
--- a/tests/test_base/__snapshots__/extensions.json
+++ b/tests/test_base/__snapshots__/extensions.json
@@ -445,11 +445,11 @@
         "@id": "tag"
       },
       "value": {
-        "@type": "linkml:Any",
+        "@type": "@id",
         "@id": "value"
       },
       "extensions": {
-        "@type": "linkml:Extension",
+        "@type": "@id",
         "@id": "extensions"
       },
       "AnyValue": {
--- a/tests/test_base/__snapshots__/meta.context.jsonld
+++ b/tests/test_base/__snapshots__/meta.context.jsonld
@@ -61,11 +61,11 @@
          "@id": "skos:altLabel"
       },
       "all_members": {
-         "@type": "linkml:AnonymousSlotExpression",
+         "@type": "@id",
          "@id": "all_members"
       },
       "all_of": {
-         "@type": "linkml:Expression",
+         "@type": "@id",
          "@id": "all_of"
       },
       "source": {
@@ -76,15 +76,15 @@
          "@id": "description"
       },
       "alt_descriptions": {
-         "@type": "linkml:AltDescription",
+         "@type": "@id",
          "@id": "alt_descriptions"
       },
       "annotations": {
-         "@type": "linkml:Annotation",
+         "@type": "@id",
          "@id": "annotations"
       },
       "any_of": {
-         "@type": "linkml:Expression",
+         "@type": "@id",
          "@id": "any_of"
       },
       "apply_to": {
@@ -92,7 +92,7 @@
          "@id": "apply_to"
       },
       "array": {
-         "@type": "linkml:ArrayExpression",
+         "@type": "@id",
          "@id": "array"
       },
       "asymmetric": {
@@ -100,7 +100,7 @@
          "@id": "asymmetric"
       },
       "attributes": {
-         "@type": "linkml:SlotDefinition",
+         "@type": "@id",
          "@id": "attributes"
       },
       "base": {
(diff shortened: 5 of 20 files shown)


Silvanoc:
@lapkinvladimir, @ostrzyciel you both seem to be deep enough into JSON-LD and RDF to be good reviewers for this PR. I see that you are involved in PR #2912, which is also related to the code being changed here. Could any of you please have a look at it? Thanks!

lapkinvladimir:
@Silvanoc Im not very deep into JSON-LD, but from what I see, the switch from `@type` to `@id` looks correct according to the spec. Honestly, didnt spot anything obviously wrong, but I cant guarantee I fully get it.

Ostrzyciel:
We just tested this patch in combination with the frame generation feature added in #2912

I'm happy to report that with this patch, inlining nested JSON objects finally works correctly :)

So we would very much appreciate if this was merged.

dalito (approved):
The fix agrees with the discussion and analysis in the issue which I found plausible and correct. So I approve. But I have to note that I am not a json-ld specialist so a second review may be good.

Silvanoc:
> We just tested this patch in combination with the frame generation feature added in #2912
> 
> I'm happy to report that with this patch, inlining nested JSON objects finally works correctly :)
> 
> So we would very much appreciate if this was merged.

@Ostrzyciel liking a PR (giving a "thumb up") helps in the prioritization of the maintainers. AFAIK the more "thumbs up" a PR gets, the sooner it will be considered for merging.

amc-corey-cox:
Oops, sorry @Silvanoc I didn't realize you were working on this. I tried a rebase to make the branch merge-able. I hope I didn't mess anything up with what you were doing.

amc-corey-cox:
Oh, it looks like that was just activity from my rebase. Sorry for the confusion.

dalito:
@amc-corey-cox you and I reviewed and @Silvanoc knows what he does. :-) - Is there anything from your side preventing a merge?

Silvanoc:
> Oops, sorry @Silvanoc I didn't realize you were working on this. I tried a rebase to make the branch merge-able. I hope I didn't mess anything up with what you were doing.

@amc-corey-cox rebasing is in fact my favored workflow, so you can simply rebase my PRs whenever you want. I would appreciate you letting me know to avoid that I can override it with a forced push, what I do to beautify the git history.

Silvanoc:
> @amc-corey-cox you and I reviewed and @Silvanoc knows what he does. :-) - Is there anything from your side preventing a merge?

@dalito thanks for the appraisal 😄 I can say that in this case it took me a while to fully grasp the original issue, but I do know pretty well what I've done here.

I have once again the feeling that the lack of clarity about the merging process is the main issue here. I have the rights to merge, but I don't do it unless it's water-clear, that I am allowed/expected to do so.

amc-corey-cox:
@dalito I forgot to get back to this. It would have been fine for @Silvanoc to merge it but I understand the hesitancy; I'm often the same way. I just noticed I had not merged it and hit the button. Cheers!

Silvanoc:
> @dalito I forgot to get back to this. It would have been fine for @Silvanoc to merge it but I understand the hesitancy; I'm often the same way. I just noticed I had not merged it and hit the button. Cheers!

First things first: thanks for merging!

@amc-corey-cox as I remark whenever I see a fitting chance 😉 , IMO the [processes are not clear](https://github.com/orgs/linkml/discussions/2518). At least in the past I had the impression that I was not expected to merge, although I have the rights for some time now. That's why I simply don't do it before I break something. I follow the "don't ask for permission, ask for forgiveness" principle rarely.

I think I'll make a PR documenting my understanding of the processes probably following the [proposal from](https://github.com/linkml/linkml/issues/2382#issuecomment-2623825809) @dalito

Issue #1699, opened 2023-11-06
"Decimal": Integer() should be Float() in sqltablegen.py


kappe-c:
**Describe the bug**
In the RANGEMAP dict in https://github.com/linkml/linkml/blob/main/linkml/generators/sqltablegen.py Decimal should be mapped to Float() if I am not much mistaken. [xsd:decimal](https://www.w3.org/TR/xmlschema-2/#decimal) is defined to represent "a subset of the real numbers, which can be represented by decimal numerals"
Probably this is just a typo or copy and paste mistake. (I would create a pull request fixing this myself but that seems overkill for half a line of code.)

cmungall:
See

https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.DECIMAL
https://en.wikipedia.org/wiki/Double-precision_floating-point_format

Decimal has higher precision for base10 numbers. Of course, some DBMSs may map the SQLA type to a float anyway....

kappe-c:
Just to clarify, my concern is not whether `Decimal` is represented by a 4- or 8-Byte floating point number (or some other construct to hold rational/real numbers (to some precision)), just that that data type should definitely be able to hold more than just integers.

PR #2905, opened 2025-09-14
fix Decimal sqla mapping

pizzashi:
Addresses #1699


pizzashi:
Looks like it's all good now.

I changed the mapping of LinkML's `decimal` to SQLA's `Numeric`. Previously it was mapped to `Integer`. Also changed the mapping in the test under tests/test_generators/test_sqltablegen.py.

The [bug](https://github.com/linkml/linkml/issues/1699) that this PR resolves suggested `decimal` be mapped as `Float` but I figure it'd be better to use Numeric since by default, it returns Python `decimal.Decimal` objects and can fallback to float as needed.

More info here: https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.Numeric

PR #2908, opened 2025-09-15
fix: Handle special characters in slot names

### LINKED ISSUE #2901: SyntaxError using `PydanticValidationPlugin` with attributes starting with at sign '@'
**Describe the bug**

When trying to use the `validate` method of a validator that has been created with:

- the `PydanticValidationPlugin`
- a LinkML `.yaml` that contains attributes or slots that start with the character `@`, such as `@id`

I get a syntax error. Trying to play with aliases didn't help.

- Python 3.10.18
- linkml 1.9.4

`script.py`:
```
from linkml.validator import Validator
from linkml.validator.plugins import PydanticValidationPlugin

valid_simple_jsonld = {
    "@graph": [
        {
            "@id": "./",
            "@type": "Dataset",
            "hasPart": [
                {"@id": "README.md"},
            ],
        },
        {
            "@id": "README.md",
            "@type": "File",
        },
    ]
}

ontology_path = "linkml_simple_ontology.yaml"

linkml_validator = Validator(
    schema=ontology_path,
    validation_plugins=[
        PydanticValidationPlugin(),
    ],
)

validation_report = linkml_validator.validate(valid_simple_jsonld)
print(validation_report)
```

`linkml_simple_ontology.yaml`:
```
id: linkml_simple_ontology.yaml
name: linkml_simple_ontology
imports:
  - linkml:types
default_range: string

classes:
  Container:
    tree_root: true
    attributes:
      "@context":
      "@graph":
        multivalued: true
        inlined_as_list: true
        range: Entity
  Entity:
    abstract: true
    slots:
      - "@id"
      - "@type"
  Dataset:
    is_a: Entity
    attributes:
      "@type":
        pattern: "^Dataset$"
      hasPart:
        multivalued: true
        range: File
        inlined_as_list: true
  File:
    is_a: Entity
    attributes:
      "@type":
        pattern: "^File$"

slots:
  "@id":
    identifier: true
    required: true
  "@type":
    designates_type: true
```

emmanuel-ferdman:
## PR Summary
This PR fixes a SyntaxError in `PydanticValidationPlugin` when LinkML schemas contain slots with special characters like `@id`, `@type`, or `@context`. Previously, these characters would create invalid Python identifiers in the generated Pydantic code, causing validation to fail. The fix converts invalid identifiers to valid Python field names while preserving the original names through Pydantic field aliases, ensuring both JSON-LD style data (`@id`) and Python-style data (`id`) validate correctly.

Fixes #2901.

### DIFF
--- a/.github/workflows/rustgen.yaml
+++ b/.github/workflows/rustgen.yaml
@@ -1,0 +1,64 @@
+name: Build and test linkml (rustgen)
+
+env:
+  UV_VERSION: "0.7.13"
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  test-rustgen:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+
+      - name: Ensure tags if not run from main repo
+        if: github.repository != 'linkml/linkml'
+        run: |
+          git remote add upstream https://github.com/linkml/linkml
+          git fetch upstream --tags
+
+      - name: Install uv and setup uv caching
+        uses: astral-sh/setup-uv@v5
+        with:
+          version: ${{ env.UV_VERSION }}
+          enable-cache: true
+
+      - name: Set up Python
+        uses: actions/setup-python@v5.6.0
+        with:
+          python-version: '3.12'
+
+      - name: Install Rust toolchain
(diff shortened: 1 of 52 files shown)


kevinschaper:
Perfect! thanks so much @emmanuel-ferdman!

Potentially we should share code between this and `generate_enums` in oocodegen.py, but there are some slightly different requirements (enum values like `+` and `-` hopefully wouldn't also be slot names!), so it probably does make sense to keep them both on their own.

PR #2912, opened 2025-09-18
Feature: JSON-LD framing support

lapkinvladimir:
This PR adds JSON-LD framing support to the JsonLdContextGenerator.

### What it does
New --emit-frame CLI flag.

When enabled, generator now writes:
- <schema>.context.jsonld - as before
- <schema>.frame.jsonld - JSON-LD frame

Slots with inlined: true get @embed: `@always`.
Slots with inlined: false get @embed: `@never`.

Root class for the frame is chosen by:
- first class marked tree_root: true, or
- the first defined class as fallback.

### Tests
Added unit tests covering:
- correct @embed generation for inlined and non-inlined slots
- tree_root selection
- CLI --emit-frame behavior
- ensuring no frame is written if there are no inlined slots

### No changes when --emit-frame is not used. Existing behavior is preserved.

### DIFF
--- a/.github/workflows/rustgen.yaml
+++ b/.github/workflows/rustgen.yaml
@@ -1,0 +1,64 @@
+name: Build and test linkml (rustgen)
+
+env:
+  UV_VERSION: "0.7.13"
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  test-rustgen:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+
+      - name: Ensure tags if not run from main repo
+        if: github.repository != 'linkml/linkml'
+        run: |
+          git remote add upstream https://github.com/linkml/linkml
+          git fetch upstream --tags
+
+      - name: Install uv and setup uv caching
+        uses: astral-sh/setup-uv@v5
+        with:
+          version: ${{ env.UV_VERSION }}
+          enable-cache: true
+
+      - name: Set up Python
+        uses: actions/setup-python@v5.6.0
+        with:
+          python-version: '3.12'
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Show rust/cargo version
+        run: |
+          rustc -V
+          cargo -V
+
+      - name: Install Python deps (rustgen tests)
+        run: uv sync --group dev --group tests-rustgen --group tests-extra
+
+      - name: Run rustgen tests with coverage
+        run: |
+          uv run --group dev --group tests-rustgen coverage run -m pytest -m rustgen --with-rustgen
+          uv run coverage xml
+          uv run coverage report -m
+
+      - name: Upload coverage report
+        if: github.repository == 'linkml/linkml' && github.actor != 'dependabot[bot]'
+        uses: codecov/codecov-action@v5.4.3
+        with:
+          name: codecov-results-rustgen
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: coverage.xml
+          fail_ci_if_error: true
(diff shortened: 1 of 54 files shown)


linkml/generators/rdfgen.py
```diff
@@ -67,12 +67,21 @@ def end_schema(self, output: Optional[str] = None, context: str = None, **_) ->
             base=str(self.namespaces._base),
             prefix=True,
         )
-        out = self._data(graph)
+        fmt = "turtle" if self.format == "ttl" else self.format
```
Ostrzyciel: This is carried over from #2894 

please untangle it – one PR should be just about one feature!

docs/data/rdf.md
```diff
@@ -21,7 +21,7 @@ When this JSON-LD context is combined with JSON it generates valid RDF
 Note there are some features missing from JSON-LD generation:
 
  * `@type` information is not generated at all levels
- * `@embed` is not generated for JSON-LD framing, limiting the ability to translate from RDF to JSON
+ * `@embed` is generated for JSON-LD framing only when `--emit-frame` is used
```
Ostrzyciel: This is inconsistent with line 21. Remove this entirely?

Ostrzyciel (changes requested):

lapkinvladimir:
@Silvanoc can you have a look at this?

linkml/generators/jsonldcontextgen.py
```diff
@@ -125,6 +131,40 @@ def end_schema(
             with open(output, "w", encoding="UTF-8") as outf:
                 outf.write(as_json(context))
 
+        # If --emit-frame: always write the context file next to the YAML;
```
Silvanoc: The flag `--emit-frame` is changing the output pattern of all the generators too much. I don't have a good solution though...

What I mean is that the output of most (probably even all) the generators goes directly to stdout. And the moment that you add this flag, suddenly you get the context both on stdout and on file. And also the frame. However you cannot specify a target path.

I do not like that standard behavior. But I you want to change it (at least in this generator), then I think that it must be done better.

I would add a first commit in this patch serie adding an option to get the output to a file. I'd use the same [pattern used for RDF](https://github.com/linkml/linkml/blob/1baf7ad19cf2f1a52cbdd7ae90524cce167ed48e/linkml/generators/rdfgen.py#L80):  `--output xyz`. Over time `--output xyz` could become a `linkml generate` option that automatically applies to all generators. I mean, the output of the specific generator `stdout` could be redirected to a file.

I'd then add all your other changes. But with following changes:
* I'd reject `--emit-frame` if `--output` has not been specified.
* I'd add the possibility of supporting both `--emit-frame` and `--emit-frame xyz`. Just in case the frame should go to a completely different place.

Silvanoc:
After writing [here](https://github.com/linkml/linkml/pull/2912#discussion_r2378794332)

> Over time --output xyz could become a linkml generate option that automatically applies to all generators. I mean, the output of the specific generator stdout could be redirected to a file.

I have realized that such a pattern is [already foreseen](https://github.com/linkml/linkml/blob/1baf7ad19cf2f1a52cbdd7ae90524cce167ed48e/linkml/utils/generator.py#L152). It only needs to get embraced! @lapkinvladimir simply add the CLI option `--output` and it should work out of the box.

lapkinvladimir:
@Silvanoc thanks for the review, I've enforced --output for --emit-frame and updated docs/tests. Please have another look.

Silvanoc (approved):

tests/test_generators/test_jsonldcontextgen.py
```diff
@@ -50,7 +49,226 @@ def test_inlined_external_types(input_path, snapshot_path):
...
+    attributes:
+      id: {identifier: true, range: string}
+      friend: {range: Person, inlined: true}
+""",
+        encoding="utf-8",
+    )
+    runner = CliRunner()
+    result = runner.invoke(jsonld_context_cli, [str(schema), "--emit-frame"])
+    assert result.exit_code != 0
+    assert "--emit-frame requires --output" in result.output
```
Silvanoc, line 274: CLI tests are typically in `tests/test_scripts`. But I'm not really sure that that separation makes so much sense...

PR #2916, opened 2025-09-22
[rustgen] generate stubs for python type hints

kervel:
This PR adds the pyo3-stub-gen crate to the python3 generator, so that we are able to generate python3 stubs so we can use typed python with the extension.

first https://github.com/linkml/linkml/pull/2889 needs to be merged ...


linkml/generators/rustgen/rustgen.py
```diff
@@ -386,12 +408,70 @@ def __post_init__(self):
         self.schemaview: SchemaView = SchemaView(self.schema)
         super().__post_init__()
 
+    def _select_root_class(self, class_defs: list[ClassDefinition]) -> Optional[ClassDefinition]:
```
kervel, line 412: this now is overengineered rubbish. i'll fix it

linkml/generators/rustgen/templates/enum.rs.jinja
```diff
@@ -52,3 +52,19 @@ impl<'py> FromPyObject<'py> for {{ name }} {
         }
     }
 }
+
+#[cfg(feature = "stubgen")]
+impl ::pyo3_stub_gen::PyStubType for {{ name }} {
+    fn type_output() -> ::pyo3_stub_gen::TypeInfo {
+        {% set literal_values = items | map(attribute='text_literal') | list %}
+        {% set escaped_values = literal_values | map('replace', "'", "\\'" ) | list %}
+        {% if escaped_values %}
```
tfliss: Is this intended to be a length check for > 0 entries?

something like {% if escaped_values and escaped_values|length > 0 %}

linkml/generators/rustgen/templates/lib_shim.rs.jinja
```diff
@@ -0,0 +1,51 @@
...
+
+pub mod generated;
+
+pub use generated::*;
+pub use chrono::NaiveDate;
+pub use chrono::NaiveDateTime;
+
+{% if root_struct_name %}
+#[cfg(feature = "serde")]
+/// Example helper that loads a YAML document into the root class. Edit or extend as needed.
```
tfliss, line 12: Is a 'yaml' feature needed here specifically in addition to regular serde?  Maybe this being an example it isn't that important.

linkml/generators/rustgen/templates/lib_shim.rs.jinja
```diff
@@ -0,0 +1,51 @@
+//! Auto-generated shim: re-export generated code and host the PyO3 module entry point.
+
```
tfliss, line 3: Is there a better name than shim, or is that a rust convention? I can't suggest anything better so maybe this is ok. The comment does explain what it's for.

linkml/generators/rustgen/templates/struct_or_subtype_enum.rs.jinja
```diff
@@ -106,3 +106,6 @@ impl serde_utils::InlinedPair for {{enum_name}} {
     }
 }
 {% endif %}
+
+#[cfg(feature = "stubgen")]
+::pyo3_stub_gen::impl_stub_type!({{ enum_name }} = {{ struct_names | join(' | ') }});
```
tfliss, line 111: This might be an edge case, or even not allowed in linkml, but if the enum is empty the join will probably generate incorrect syntax. Maybe caught earlier? I saw it in a couple of places, and maybe also the case for some for loops. Won't hold up the PR on this, though. I'm curious if there's a compliance case for empty enums.

linkml/generators/rustgen/templates/stub_gen.rs.jinja
```diff
@@ -0,0 +1,68 @@
...
+
+#[cfg(feature = "stubgen")]
+fn check_stubs(stub: &pyo3_stub_gen::StubInfo) -> pyo3_stub_gen::Result<()> {
+    use std::fs;
+    use std::io::ErrorKind;
+
+    let mut issues = Vec::new();
+
+    for (name, module) in &stub.modules {
+        let normalized_name = name.replace('-', "_");
```
tfliss: Is this name-cleaning going to be robust enough? Consider at least pulling the normalize into a helper function that is a clear place for future improvements.

tfliss: Should also check for i.e. * in names that are converted to paths.

linkml/generators/rustgen/rustgen.py
```diff
@@ -989,18 +1058,52 @@ def write_crate(
...
+            for rel_path, template in rendered.bin_files.items():
+                rendered_bin = template.render(self.template_environment)
+                rendered_bin = rendered_bin.rstrip("\n") + "\n"
+                bin_path = (src_dir / rel_path).with_suffix(".rs")
+                bin_path.parent.mkdir(exist_ok=True, parents=True)
+                with open(bin_path, "w") as bf:
+                    bf.write(rendered_bin)
+
+        if self.handwritten_lib:
+            shim_path = src_dir / "lib.rs"
```
tfliss, line 1081: Can the path elements be checked for i.e. '..' and '/' and the construction of the path, mkdir, etc. be generalized somewhat? It seems to be repeated. I actually need to look at this in panderagen soon, so wonder if that could go into a shared module.

tfliss:
Ok, I read through all the code today, with an AI assist to explain some of the rust / crate bits. rustgen.py feels like it's getting a little long and I left a few other comments, probably not anything I'll hold up the PR on. Tomorrow I'll try running the tests and a cli example and maybe try it out in a colab notebook. For the most part this looks good and very useful. Thank you.

kervel:
hi @tfliss .. is it possible you forgot to submit your PR review ? i don't see the comments ..

docs/generators/rust.rst
```diff
@@ -29,6 +29,62 @@ It additionally generates a trait for every struct to provide polymorphic access
...
+
+   ``maturin build`` is an alternative when you want distributable wheels in ``target/wheels`` instead.
+
+#. **Import the module from Python** and try the generated YAML loader.
+
+   .. code-block:: python
+
+      import personinfo
+      # Update the path below to point at your data file.
+      container = personinfo.load_yaml_container("path/to/data.yaml")
```
tfliss: can you point this to say one of the tutorial person files or some other specific file?  I got None when I used examples/tutorial/tutorial07/personinfo.yaml

I probably could find or make one to use, but a specific file would be easier for new users.

tfliss: Interesting, examples/PersonSchema/data/example_personinfo_data.yaml loads the file of course, but prints [<builtins.Person object at 0x10cabd110>, <builtins.Person object at 0x10cabd2c0>]
I guess this is ok for now, seems odd that Person is defined in builtins.

linkml/generators/rustgen/templates/struct_or_subtype_enum.rs.jinja
kervel: tis is about the OrSubtype enums not about regular linkml enums ... these enums are only generated when a class has at least one subtype.

It is needed to model slot range where a slot value can be something or a subtype of that something. rust has no struct inheritance ...

i'll treat the other comments by changing / clarifying code .

docs/generators/rust.rst
tfliss: Ok, I just changed my version of this script fragment to the following and to me it feels better for a quick-start example. Consider using something like this:

```
container = personinfo.load_yaml_container("examples/PersonSchema/data/example_personinfo_data.yaml")
print(container.persons[0].name)
```

I ran the generator and coverage and it all works fine. I'm marking my review 'changes requested' because you said you're making minor documentation / comment edits and code clarification. Please request a new review when that's done and I'll approve.

docs/generators/rust.rst
```diff
@@ -29,6 +29,62 @@ It additionally generates a trait for every struct to provide polymorphic access
...
+   ``cargo run --bin stub_gen --features stubgen`` from the crate directory; add ``-- --check`` when you want to verify
+   existing stubs instead of overwriting them.
+
+#. **Build and install the wheel** with ``maturin``. Running ``maturin develop`` compiles the extension and puts it in the
+   active virtual environment.
+
+   .. code-block:: bash
+
+      cd personinfo_rust
+      maturin develop
```
tfliss, line 75: I had to `uv pip install --group tests-rustgen` before the maturin command would work. I don't remember if uv is the default linkml environment manager yet, but some command that installs maturin would be helpful in a 'quick start' context. I recognize most rust developers might know how to do this, but python devs maybe not.

amc-corey-cox:
@kervel and @tfliss are these comments completed? Is this ready for me to review?

kervel:
Not yet! Will try next week...

docs/generators/rust.rst
kervel: > Interesting, examples/PersonSchema/data/example_personinfo_data.yaml loads the file of course, but prints [<builtins.Person object at 0x10cabd110>, <builtins.Person object at 0x10cabd2c0>]

i just found out that you can tweak the module name (so that its no longer builtins.) for generated classes. however, its a bit involved (change every #pyclass if i see) and then i have hardcoded module names all over the place. so i'd rater leave it as is for now

the "fix" would be to write:

```rust
#[pyclass(module = "mymodule")]
```
instead of just pyclass.. but then for instance renaming the module becomes next to impossible (having to manually edit generated code). so i'd argue that the fix is worse than the problem.

linkml/generators/rustgen/templates/lib_shim.rs.jinja
kervel: well the "serde" feature flag actually brings in a serde_yml dependency:

```python
SERDE_IMPORTS = Imports(
    imports=[
        Import(
            module="serde",
            version="1.0",
            features=["derive"],
            objects=[
                ObjectImport(name="Serialize"),
                ObjectImport(name="Deserialize"),
                ObjectImport(name="de::IntoDeserializer"),
            ],
            feature_flag="serde",
        ),
        Import(module="serde-value", version="0.7.0", objects=[ObjectImport(name="Value")]),
        Import(module="serde_yml", version="0.0.12", feature_flag="serde", alias="_"),
        Import(module="serde_path_to_error", version="0.1.17", objects=[], feature_flag="serde"),
    ]
)
```

so i think that's fine..

linkml/generators/rustgen/templates/stub_gen.rs.jinja
kervel: ok the helper is there now. its in the rust code not the python code, and its a bit more involved than i originally tought. but its there.

linkml/generators/rustgen/rustgen.py
kervel: i have done something, not sure if it is what you wanted.

kervel:
hello @tfliss , i think i implemented your feedback.

tfliss (approved):
@kervel I really appreciate your thoughtful work to implement the suggestions. This looks really great and I will look at re-using your file system methods.

@amc-corey-cox If you or another core dev are able to merge I reviewed this PR extensively, used the python API successfullly, and Frank went above and beyond implementing the comment suggestions.

PR #2921, opened 2025-09-30
OnePerClass linter rules for `identifier` and `key` attributes

ialarmedalien:
Adding linter rules to enforce rules that ensure that `identifier` and `key` only appear once per class

### DIFF
--- a/linkml/linter/config/datamodel/config.yaml
+++ b/linkml/linter/config/datamodel/config.yaml
@@ -37,32 +37,17 @@ classes:
       Each attribute of this class represents a rule that can be enabled and possibly
       configured by a configuration file.
     attributes:
+      canonical_prefixes:
+        range: CanonicalPrefixesConfig
+        description: >-
+          Enforce canonical prefixes by verifying that the mappings defined in the `prefixes`
+          slot agree with those provided by the [prefixmaps](https://github.com/linkml/prefixmaps)
+          package. Not auto-fixable.
       no_empty_title:
         range: NoEmptyTitleConfig
         description: >-
           Disallow empty titles on schema elements. Autofix will transform the element's
           name into a title.
-      permissible_values_format:
-        range: PermissibleValuesFormatRuleConfig
-        description: >-
-          Enforce consistent formatting of enum permissible values. This rule may conflict
-          with the standard_naming rule. Not auto-fixable.
-      tree_root_class:
-        range: TreeRootClassRuleConfig
-        description: >-
-          Require a single class with `tree_root: true` and optionally verify that class's
-          name. Autofix will create a new class with `tree_root: true`, a name based on
-          the rule configuration, and slots for existing classes.
-      recommended:
-        range: RecommendedRuleConfig
-        description: >-
-          Require any slot in the metamodel with `recommended: true` (e.g. description)
-          to be provided. Not auto-fixable.
-      no_xsd_int_type:
-        range: RuleConfig
-        description: >-
-          Disallow use of `uri: xsd:int` in type definitions. Autofix will change the
-          uri to xsd:integer.
       no_invalid_slot_usage:
         range: RuleConfig
         description: >-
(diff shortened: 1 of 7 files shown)


linkml/linter/config/datamodel/config.yaml
```diff
@@ -37,32 +37,17 @@ classes:
       Each attribute of this class represents a rule that can be enabled and possibly
       configured by a configuration file.
     attributes:
+      canonical_prefixes:
```
ialarmedalien, line 40: I put these in alphabetical order, hence the large diff in this and the generated file.

linkml/linter/linter.py
```diff
@@ -45,8 +45,7 @@ def get_metamodel_validator() -> Validator:
     meta_json_gen = JsonSchemaGenerator(LOCAL_METAMODEL_YAML_FILE, not_closed=False)
     meta_json_schema = meta_json_gen.generate()
     validator_cls = jsonschema.validators.validator_for(meta_json_schema, default=jsonschema.Draft7Validator)
-    validator = validator_cls(meta_json_schema, format_checker=validator_cls.FORMAT_CHECKER)
-    return validator
```
ialarmedalien, line 49: No need to create a variable if you return it immediately.

linkml/linter/rules.py
```diff
@@ -319,6 +319,43 @@ def check(self, schema_view: SchemaView, fix: bool = False) -> Iterable[LinterPr
                 yield LinterProblem(f"Root type '{t}' is missing the required 'uri' attribute")
 
 
+class OnePerClass(LinterRule):
+    """Ensures that there is only one slot with the specified attribute per class."""
```
ialarmedalien, line 323: Both checks are very similar so can create a more generic function to test.

linkml/linter/rules.py
```diff
@@ -319,6 +319,43 @@ def check(self, schema_view: SchemaView, fix: bool = False) -> Iterable[LinterPr
...
+    id = "one_per_class"
+
+    def _check(self, schema_view: SchemaView, attr_name: str) -> Iterable[LinterProblem]:
+        for cn in schema_view.all_classes():
+            slots_with_attr = [
+                slot
+                for slot in schema_view.class_induced_slots(cn)
+                if hasattr(slot, attr_name) and getattr(slot, attr_name)
+            ]
+            if len(slots_with_attr) > 1:
```
dalito, line 334: You could simplify to `if slots_with_attr:` since empty lists are false.

tests/test_linter/test_one_per_class.py
```diff
@@ -0,0 +1,126 @@
...
+
+    for attr_type, val in exp_slots.items():
+        if len(val) > 1:
+            assert len(problems[attr_type]) == 1
+            msg = f"Class 'TestClass' has more than one '{attr_type}' slot: "
+            msg += ", ".join(sorted(val))
+            assert {p.message for p in problems[attr_type]} == {msg}
+
+
+def check_schema(test_schema: str, check_identifier: bool) -> list[LinterProblem]:  # noqa: FBT001
```
dalito: Maybe move this helper function above the test?

linkml/linter/rules.py
ialarmedalien: The length of the list has to be greater than one -- i.e. there are two or more identifiers or keys in a class. Zero or one keys is fine.

ialarmedalien:
LMK if there is extra documentation or more comments that I can add to the tests to make it clearer how I'm generating the schemas. It makes perfect sense when you're writing the code but isn't always easy to decode when reading it.

linkml/linter/rules.py
dalito: Of course. It was too late obviously.

tests/test_linter/test_one_per_class.py
```diff
@@ -0,0 +1,161 @@
...
+    classes:
+      TestClass:
+        slots:
+          - slot_a
+          - slot_b
+          - id_slot_a
+          - id_slot_false
+          - key_slot_false
+
+    This allows the testing of a large number of combinations without having to generate them all manually.
```
dalito, line 126: To help grasp the pytest trick you could be explicit: "This and the stacked parametrize decorators allow testing of all combinations (2**6=64) ..."

dalito (approved):
Thanks for adding the additional comments. They helped to understand what will happen in the test. I first read the code and then ran it with some print-statements to verify my understanding.

PR #2927, opened 2025-10-03
Feature: Remote Schema Support in CLI

infispiel:
A project of mine would like to validate against NMDC's schema without fetching the schema and all of its dependencies to ensure the most recent version of the schema is being used. This functionality appears to be in the runtime but not supported by the CLI because of Click's `File` type validation. This PR is a proposal/implementation of a new parameter, `-r`/`--schema-remote` that accepts a URL and bypasses pre-validation.

### What it does:

Adds the `-r`/`--schema-remote` option to the CLI.

Since the linkml runtime can handle URLs for schemas, I just made `config_args` take `schema_remote` if `schema` is `None`, and continue the existing `None` checking logic from there.

Also added logic to make sure that both `-r` and `-s` are not specified.

### Other Notes

This code doesn't do any logic to validate whether the remote URL exists and depends entirely on bubbling up the runtime's errors in response to an invalid URL. Is this acceptable? Or should I add some validation logic to make sure the remote URL exists and have the CLI fail early? Should I also add checks to ensure a user doesn't put a file path in the `-r` field?

### DIFF
--- a/packages/linkml/src/linkml/validator/cli.py
+++ b/packages/linkml/src/linkml/validator/cli.py
@@ -176,7 +176,7 @@ def cli(
             fg="yellow",
         )
 
-    if schema is not None and schema_remote is not None :
+    if schema is not None and schema_remote is not None:
         raise click.ClickException(
             "Both a remote and a local schema were provided. Please use either the -s/--schema "
             "option or the -r/--schema-remote option to define a schema, not both."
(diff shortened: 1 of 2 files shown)


matentzn:
@pkalita-lbl could you create a corresponding issue to track this work and bring it to conclusion?

pkalita-lbl:
@matentzn all that was needed was to get the development branch updated with `main` and fix a few formatting issues. I did that, but now that I am an author on this PR I can't also be a reviewer. But I think the changes here are good and this is now ready to be merged.

matentzn:
@amc-corey-cox please review and merge, thank you!

PR #2934, opened 2025-10-10
Feature: YARRRML generator with CLI

lapkinvladimir:
This PR adds a basic YARRRML generator to LinkML.

### What it does
New CLI entrypoint: gen-yarrrml
Generates one YARRRML mapping per LinkML class
Subject from identifier or key slot (safe fallback if none)
Defaults: JSON source (<file>.json~jsonpath, $.items[*])

### Tests
End-to-end tests covering:
- identifier / key slots and subject fallback
- slot alias expansion
- multiple classes and prefixes
- inlined objects skipped if inlined: true
- negative case (missing target instances) marked xfail

### DIFF
--- a/tests/test_generators/test_yarrrmlgen.py
+++ b/tests/test_generators/test_yarrrmlgen.py
@@ -1,7 +1,8 @@
 import json
+import sys
 from pathlib import Path
 
-import sys, pytest
+import pytest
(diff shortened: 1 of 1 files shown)


matentzn:
Aweeeesome!!!

Issue #2939, opened 2025-10-13
Metamodel out of date re: `extra_slots`


sneakers-the-rat:
---
name: Metamodel out of date re: `extra_slots`
title: Metamodel out of date re: `extra_slots`
labels: bug

---

**Describe the bug**

`extra_slots` was merged into the metamodel nearly a year ago - https://github.com/linkml/linkml-model/pull/205

yet it is not present in the `linkml_runtime` metamodel version of `ClassDefinition`: https://github.com/linkml/linkml-runtime/blob/ff268a3f2ef3e3f78399d0c4fbe6d595c6b5dd53/linkml_runtime/linkml_model/meta.py#L3072

can we get an updated metamodel plz? 

blocking resolution of: 
- https://github.com/linkml/linkml/issues/2241
- https://github.com/linkml/linkml/issues/2238

gouttegd:
Can we know what is the blocker here? What needs to happen for the meta model to be updated?

The absence of `extra_slots` from the LinkML-runtime’s `meta.py` module makes that slot in effect unusable, so I would think that updating the meta model in the runtime should be uncontroversial?

sneakers-the-rat:
Hopefully its just regenerating the metamodel, no? Should be more or less automated/automatable? I still dont know exactly how its done but esp with the monorepo now should b easy to test

gouttegd:
>  I still dont know exactly how its done

My understanding is that it is done by running

```sh
$ make update_model MODEL_DIR=/path/to/linkml-model/repository
```

within `packages/linkml_runtime`.

Happy to have a try at it but I wonder if there’s a reason why LinkML devs have not done it already…

gouttegd:
Related issue about the lack of updates to the meta model (not specifically about `extra_slots`): #2718

matentzn:
> Happy to have a try at it but I wonder if there’s a reason why LinkML devs have not done it already…

If you _could_ do it I will prioritise and review. But feel free to just tell me to do it myself.

gouttegd:
I would first like to know what is LinkML’s policy (if there is one!) regarding the frequency of updates of the meta model in the LinkML-runtime.

Looking at the history of both the [meta model](https://github.com/linkml/linkml-model) and of the [`meta.py` module in the runtime](https://github.com/linkml/linkml-runtime/blob/main/linkml_runtime/linkml_model/meta.py):

| LinkML model release | Date the model was released | Date the model was updated in the runtime |
[...]

PR #2940, opened 2025-10-13
implement `extra_slots` in jsonschemagen

### LINKED ISSUE #2238: JSON Schema "additionalProperties" default value not set correctly
**Describe the bug**
In the context of LinkML schema, currently (to the best of my knowledge) we cannot set a boolean value for `additionalProperties`. 
By default, the LinkML JSON Schema generator assigns ```json"additionalProperties": false``` for object type schemas. 

However, according to the [JSON Schema Specification Draft 00 ](https://tools.ietf.org/html/draft-wright-json-schema-validation-00#page-10):

`If "additionalProperties" is absent, it may be considered present with an empty schema as a value.`

This means to conform to the spec, by default `additionalProperties` should be set to `{}` for object type schemas and not to `false`.

**To reproduce**
The following LinkML 
```
classes:
  ExpectedResponse:
    attributes:
      contentType:
        required: true
```
produces the JSON Schema: 

```json
"ExpectedResponse": {
            "properties": {
                "contentType": {
                    "type": "string"
                }
            },
            "required": [
                "contentType"
            ],
            "title": "ExpectedResponse",
            "type": "object"
        }
```

**Expected behavior**
I would expect the default value of `"additionalProperties"` is `true` when it cannot be even specified.
```json
"ExpectedResponse": {
            "additionalProperties": true,
            "properties": {
                "contentType": {
                    "type": "string"
                }
            },
            "required": [
                "contentType"
            ],
            "title": "ExpectedResponse",
            "type": "object"
        }
```

We could introduce a new term to the LinkML vocabulary on the `class` level called `"additionalProperties"` of type `boolean` to customise the default behaviour. For instance, 
```
classes:
  ExpectedResponse:
    attributes:
      contentType:
        additionalProperties: false
        required: true
```

sneakers-the-rat:
Does what it says.

Fix: https://github.com/linkml/linkml/issues/2238

THIS SHOULD NOT BE MERGED UNTIL 
- https://github.com/linkml/linkml-model/pull/227
- https://github.com/linkml/linkml/issues/2939

it contains a very crappy monkeypatch that is just there for the sake of implementation, but after the metamodel change is merged and `linkml_runtime.linkml_model` is regenerated, then the monkeypatch can be removed.

Also:
- removed a test that confirms that a deprecation warning is emitted for markdowngen. this fails on local installs because the local version is not always 0.0.0 when installed in development environment, and this behavior is already tested here: https://github.com/linkml/linkml/blob/main/tests/test_utils/test_deprecation.py


sneakers-the-rat:
oh my precious sub-5 minute tests, what has become of you.

also holy warnings batman, when i last i left it i had battled the warnings to zero and now they are back up to... thousands???

sneakers-the-rat:
truly no idea why the renamer test is failing, i am gonna chalk it up to "just jsonobj2 things for no reason" and wait until the above changes are made and we can take the monkeypatch out to rule that out as the source of the problem. very little time for digging out mystery bugs these days

matentzn:
@sneakers-the-rat I moved this to draft as it conflicts quite a bit and the dependencies are also not yet deal with. Can you tell me whether this is still a priority to get in, and whether you want someone else to take the PR over or you are happy to see it through?

sneakers-the-rat:
I can handle this, doesnt look too hard to deconflict

sneakers-the-rat:
ok rebased this. reminder on the status, it includes a crappy monkeypatch of the `extra_slots` behavior that should be present in the metamodel, and should not be merged until https://github.com/linkml/linkml-model/pull/227 is merged and the metamodel is regenerated (resolving https://github.com/linkml/linkml/issues/2939 ). then i'll remove the monkeypatch stuff.

edit: [this](https://github.com/linkml/linkml/pull/2940#issuecomment-3399492585) remains true, no idea why renamer test is failing, but i am assuming that resolving the crappy monkeypatch will resolve this problem

yarikoptic:
FWIW : conflicts emerged again, might be worth updating @sneakers-the-rat and see if test still fails...?

matentzn:
>  FWIW : conflicts emerged again, might be worth updating @sneakers-the-rat and see if test still fails...?

In addition @sneakers-the-rat - can you set the PR to non-draft state if you think its ready to merge so I can add it to the review queue

sneakers-the-rat:
Ah yes, has the metamodel been regenerated after we merged https://github.com/linkml/linkml-model/pull/227 ? I'll remove the monkeypatch and refresh

sneakers-the-rat:
ah looks like the metamodel has not yet been regenerated, @matentzn ping me when that happens :)

matentzn:
@sneakers-the-rat I can see the `extra_slots` now: https://github.com/linkml/linkml/blob/8b7bfa4de4322f797dbb37dfb2e98be867c50a7a/packages/linkml_runtime/src/linkml_runtime/linkml_model/meta.py#L6227

Or was there something else missing?

sneakers-the-rat:
we're so close - what we need specifically is this change: https://github.com/linkml/linkml-model/pull/227

so this: https://github.com/linkml/linkml/blob/8b7bfa4de4322f797dbb37dfb2e98be867c50a7a/packages/linkml_runtime/src/linkml_runtime/linkml_model/meta.py#L5081

needs to be `AnonymousSlotExpression`

amc-corey-cox:
I will try to make sure linkml/linkml-model#227 is in the next metamodel release. As soon as we have that we should be able to get this in.

sneakers-the-rat:
Just @ me when thats done and ill clear out the monkeypatching!

matentzn:
I hope we are closing in here. @sneakers-the-rat can you confirm that with this recently updated meta model we can set this PR to ready and review? https://github.com/linkml/linkml/pull/3388

gouttegd:
Testing this PR, I find that it does not seem to work correctly when the class that carries the `extra_slots` slot happens to be the _root_ class of the LinkML schema?

First a test when the class with `extra_slots.allowed` is _not_ the root class:

```yaml
id: https://example.org/extra-sub
prefixes:
imports:
  - linkml:types
default_range: string

classes:

  Foo:
    attributes:
      foo_name:
      bar:
        range: Bar
    tree_root: true

  Bar:
    attributes:
      bar_name:
    extra_slots:
      allowed: true
```

The root class is `Foo`, one of its attributes should be an instance of `Bar` and `Bar` allows extra slots.

This seems to work fine, for example the following instance data validates correctly:

```yaml
foo_name: the name of foo
bar:
  bar_name: the name of bar
```

Now almost the same schema, but this time it is the root class `Foo` that is supposed to accept extra slots:

```yaml
prefixes:
imports:
  - linkml:types
default_range: string

classes:

  Foo:
    attributes:
      foo_name:
      bar:
        range: Bar
    tree_root: true
    extra_slots:
      allowed: true

  Bar:
    attributes:
      bar_name:
```

I would expect the following instance data to be found valid:

```yaml
foo_name: the name of foo
bar:
  bar_name: the name of bar
```

but it is not:

```console
[ERROR] [samples/extra/extra-top.yaml/0] Additional properties are not allowed ('extra' was unexpected) in /
```

gouttegd:
> Testing this PR, I find that it does not seem to work correctly when the class that carries the extra_slots slot happens to be the root class of the LinkML schema?

So in fact the **JSON-Schema generator** is working as it should. With the second example in my previous comment, it produces the following schema (showing only the relevant portions):

```json
{
    "$defs": {
        "Bar": {
[ ... OUTPUT TRUNCATED FOR BREVITY ... ]
        },
        "Foo": {
            "additionalProperties": true,
[ ... OUTPUT TRUNCATED FOR BREVITY ... ]
    },
    "$id": "https://example.org/extra-top",
    "$schema": "https://json-schema.org/draft/2019-09/schema",
    "additionalProperties": true,
[ ... OUTPUT TRUNCATED FOR BREVITY ... ]
}
```

where

* the top-level `additionalProperties` is always set to `true`,
* the `additionalProperties` of any given class is set according to the value of the `extra_slots` slot for that class (so in this example, `false` for `Bar` and `true` for `Foo`).

I believe this is correct, and Python’s `jsonschema` validator correctly validates the example given in my previous message.

But the problem is with LinkML’s `jsonschema_validation_plugin` (used under the hood by the `linkml validate` command). The JSON schema generated internally by this plugins (from the same source LinkML schema) is identical to the one produced by `linkml generate json-schema`, _except_ that the top-level `additionalProperties` is set to `false`.

And apparently (not sure whether this is intended behaviour or a bug in `jsonschema`?), **the top-level `additionalProperties` takes precedence over the `additionalProperties` of the root class**.

sneakers-the-rat:
It looks like the root of that inconsistency is just this default:

https://github.com/linkml/linkml/blob/36b35d2cbcbb782f3404384b1f3d698aabe75f97/packages/linkml/src/linkml/validator/cli.py#L23

which sets the `closed` to `True` where it usually defaults to `False`. 

Agreed that's pretty counterintuitive, it seems like we should honor the schema - if we didn't have to worry about backwards compatibility it seems like that `not_closed` param should be removed and we just use the declarations in the schema now that it's possible to specify if things allow extra props or not. neither option of a) ignoring the schema and using the `not_closed` value or b) ignoring the `not_closed` value and using the value from the schema seems good to me. I guess maybe it is useful in cases where there isn't a top class?

So i don't know what the best fix for that problem is, if it was up to me i would put a deprecation notice on the `not_closed` property, emit a warning when the values conflict, and eventually remove it and just use the `extra_slots` param. i would be happy to write that if that sounds good. edit: or maybe just make the default `None`, and then if no top class is present and it's unset flip it to `True`, and if both `not_closed` and `top_class` are set then emit a warning?

gouttegd:
> It looks like the root of that inconsistency is just this default:

That’s what I thought at first, but while changing this default would allow to _work around_ the issue, I don’t think this is the correct fix because I don’t think this default is the root cause.

From what I understand, the root cause is within the JSON-Schema generator (despite what I said in my previous comment, “in fact the JSON-Schema generator is working as it should” – in fact it isn’t), and fixing the problem there would automatically fix the `jsonschema_validation_plugin`.

The root cause, I believe, is that the JSON-Schema generator sets the “top-level” `additionalProperties` to a value that is solely defined by the global `not_closed` property.

For a LinkML schema where the root class has `extra_slots.allowed=True`, and assuming `not_closed=False`, this results in a JSON-Schema like this:

```json
{
    "$defs": {
        "Bar": {
[ ... OUTPUT TRUNCATED FOR BREVITY ... ]
        },
        "Foo": {
            "additionalProperties": true,
[ ... OUTPUT TRUNCATED FOR BREVITY ... ]
    },
    "$id": "https://example.org/extra-top",
    "$schema": "https://json-schema.org/draft/2019-09/schema",
[ ... OUTPUT TRUNCATED FOR BREVITY ... ]
}
```

and I no longer believe this is correct, because with such a schema, the fact that the `Foo` object is defined with `additionalProperties: true` **does not matter**. Since `Foo` is the `tree_root` of the LinkML schema, what matters here is **solely** the top-level `additionalProperties`.

Changing the default value of `not_closed` would _appear_ to fix the issue, but only because it would _unconditionally_ set the top-level `additionalProperties` to `true`, thereby always allowing additional properties in the root class – even when the root class is defined in the LinkML model with `extra_slots.allowed=False`!

So I believe the _correct_ fix here is to have the JSON-Schema generator to set the value of the top-level `additionalProperties` **based on the extra_slots of the root class**. That’s what my proposed commit https://github.com/linkml/linkml/pull/3399/changes/fcae01f2059b2de65efe3de0aa2215b1407a4481 does.

gouttegd:
As for what to do with `not_closed`:

I agree that this global parameter will no longer really be useful now that we have support for `extra_slots`. However, for backwards compatibility I believe it should be kept around (after all LinkML claims to follow semantic versioning rules).

What I propose is to use `not_closed` as the default value for `additionalProperties` whenever a class does not have an explicit `extra_slots` slot (my proposed change https://github.com/linkml/linkml/pull/3399/changes/73930c7413f9a26883dc32df4187cec6cb7b4238).

This way, when a schema does not make use of `extra_slots`, then the existing behaviour is preserved: whether additional properties are allowed or not will depend on the global `not_closed` parameter. When a class does make use of `extra_slots`, then this takes precedence over the global `not_closed`.

sneakers-the-rat:
> the root of that inconsistency 

Yep! I was just describing why using the validator is inconsistent with using the json schema generator directly

> or maybe just make the default None, and then if no top class is present and it's unset flip it to True, and if both not_closed and top_class are set then emit a warning?

Also yes, thats what i was thinking too. I was concerned with the implicit behavior of ignoring the `not_closed` property when the top class has extra slots defined - if someone explicitly set that to `False` then they might be surprised that the outcome is open if the top class allows extra slots. I would think emitting some warning there would be good, but then that would mean we emit a warning every time someone runs the validator on a schema with an open top class, which would not be great.

So i think we're agreed on the basic handling.

- `not_closed` unset, `top_class`&`extra_slots` unset: default `not_closed` to true
- `not_closed` set, `top_class`&`extra_slots` unset: use `not_closed`
- `not_closed` unset, `top_class`&`extra_slots` set: use top class's extra_slots
- `not_closed` set, `top_class`&`extra_slots` set: use top class's extra_slots 

And i guess the only question i have is if we should warn in case 4 if `not closed` != `extra_slots`.

(Also as a side note, can we add on a driveby change deprecating `not_closed` and changing that to `closed` lol)

gouttegd:
> So i think we're agreed on the basic handling.
> 
>     * `not_closed` unset, `top_class`&`extra_slots` unset: default `not_closed` to true
> 
>     * `not_closed` set, `top_class`&`extra_slots` unset: use `not_closed`
> 
>     * `not_closed` unset, `top_class`&`extra_slots` set: use top class's extra_slots
> 
>     * `not_closed` set, `top_class`&`extra_slots` set: use top class's extra_slots

Yes, that looks good to me.


> And i guess the only question i have is if we should warn in case 4 if `not closed` != `extra_slots`.

I am inclined to say yes, but I have no strong opinion here.


> (Also as a side note, can we add on a driveby change deprecating `not_closed` and changing that to `closed` lol)

Seconded! :D

sneakers-the-rat:
alright, i missed the notif that the updated metamodel got merged bc i was following the PR that got superceded, but just was clearing my unreads and remembered this and noticed it was done. 

so rebased, removed the monkeypatch, the test failures are gone, so this should be good to go. 

since @gouttegd identified and wrote the patch for the validator bug and it's orthogonal to the feature implementation, i'll let them handle that fix in a separate PR so they get credit

gouttegd:
> since @gouttegd identified and wrote the patch for the validator bug and it's orthogonal to the feature implementation, i'll let them handle that fix in a separate PR so they get credit

Fine with me. I’ll do that once that PR is in. :)

amc-corey-cox (approved):
This looks good to me. I think it is finally ready to go. Thanks for the ongoing effort on this @sneakers-the-rat - I really appreciate the work!

Issue #2856, opened 2025-08-07
Support PolaRS schema generation as a target of the pandera generator


tfliss:
**What is your feature request?**
Extend the panderagen generator to optionally produce PolaRS schema objects that correspond to the Pandera validation models it already produces.
see: https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.schema.html

**How important is this feature to you?** Select from the options below:
• Medium - can do work without it; but it's important (e.g. to save time or for convenience)

The PolaRS schema is particularly helpful for models with specific datatypes and nesting. As more Pandera features are implemented, it becomes increasingly difficult to manage the existing method of generating a PolaRS schema to by introspecting the Pandera model.

**Additional context**

- Pandera is not a stand-alone dataframe library, it validates data stored in various back-end libraries including Pandas, PolaRS, PySpark.
- The Pandera generator already has some features to support multiple dataframe targets because of the multiple-backend nature of Pandera.
- A local proof-of-concept branch already exists that can generate a large subset of the personinfo model. It will need to be checked in and cleaned up in the context of this issue.
- Some of the work may be broken into smaller separate PRs to modularize utility code and tables, such as type mapping.

bollwyvl:
Another tool in the `polars` space is [dataframely](https://github.com/Quantco/dataframely). It has the killer feature of multi-dataframe [`Collections`](https://dataframely.readthedocs.io/en/latest/_api/dataframely.collection.html), which brings this kind of definition closer to an RDMBS/SHACL level of validation.

tfliss:
> Another tool in the `polars` space is [dataframely](https://github.com/Quantco/dataframely). It has the killer feature of multi-dataframe [`Collections`](https://dataframely.readthedocs.io/en/latest/_api/dataframely.collection.html), which brings this kind of definition closer to an RDMBS/SHACL level of validation.

Thanks @bollwyvl I was not previously aware of this Dataframely.  I'm reading over the documentation and real-world example. Do you have a use case or a favorite tutorial?

Once the multiple-target support is working for panderagen, basic Dataframely support should not be difficult to add. I actually just got a first draft of foreign key associations running in PolaRS generation, so the multi-dataframe Collections feature is interesting. Panderagen's support of inlined models does support multi-dataframe models in some sense.

bollwyvl:
Sheepishly, I don't have anything to share.

I've got an in-house HR system that builds reports on:
- ground-truth data from a time tracking system from a well-formed API
[...]

PR #2946, opened 2025-10-16
2856 Pandera generator association dependencies

tfliss:
This is part of a group of PRs that implements independent PolaRS support 
in the Pandera dataframe generator. https://github.com/linkml/linkml/issues/2856

This PR specifically adds dependency ordering that accounts for both the class inheritance hierarchy and inlined associations.

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -52,8 +52,7 @@ The LinkML registry collects some existing schemas and allows for basic browsing
 
 ## Presentations about LinkML
 
+ * Intro to LinkML, with LLM help for basic setup: presentation for Automated Image Management and Metadata Annotation (AMIM) User Group, BioImaging North America ([video, 47min](https://www.youtube.com/watch?v=NJcd8b5BP50&t=144s)) 2025-11-07
+ * Tutorial (ICBO 2025): Annotating Data with Ontologies: LinkML Can Help ([slides](https://go.lbl.gov/ICBO-LinkML-slides); [video, 2hr 25min](https://youtu.be/ijCXDGO-NTU?si=xO32Ed5bVe1LQrsa)) 2025-11-05
  * [AI-assisted workflows for integrating data using LinkML](https://zenodo.org/records/15186182) (slides) 2025-04-10
  * [Introduction to LinkML by Sierra Moxon](https://www.youtube.com/watch?v=nyNp09WYLzw&t=183s) (video, 47 min) 2023-10-04
- * [An introduction to LinkML](https://zenodo.org/records/7778641) (slides) 2022-07-01
- * [Yosemite Project Webinar](https://www.youtube.com/watch?v=1cZu4kZ2ufQ) (video) - April 2021
- * [ICBO 2021 talk](http://ceur-ws.org/Vol-3073/paper24.pdf) (proceedings) Sept 2021
--- a/docs/faq/why-linkml.md
+++ b/docs/faq/why-linkml.md
@@ -31,7 +31,6 @@ classes:
       email:
       age:
       occupation:
-      ...
 ```
 
 this is already useful as it states which column headers are expected in your data.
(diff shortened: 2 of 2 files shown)


linkml/generators/panderagen/class_generator_mixin.py
```diff
@@ -1,20 +1,51 @@
...
 
 
 class ClassGeneratorMixin:
+    def enum_or_class(self, cn):
+        if cn in self.schemaview.all_enums():
+            return self.schemaview.get_enum(cn, strict=True)
+        elif cn in self.schemaview.all_classes():
+            return self.schemaview.get_class(cn, strict=True)
+        else:
+            raise Exception(f"Unknown class or enum {cn}")
```
ialarmedalien: This can be simplified to:

```py
        if cn in self.schemaview.all_enums():
            return self.schemaview.get_enum(cn, strict=True)
        if cn in self.schemaview.all_classes():
            return self.schemaview.get_class(cn, strict=True)
        raise Exception(f"Unknown class or enum {cn}")

File: linkml/generators/panderagen/dependency_sorter.py [hunk 2]
```diff
@@ -0,0 +1,75 @@
...
+
+    def expand_dependencies(self) -> None:
+        """
+        Expand dependencies to include all dependencies of the dependencies.
+        """
+        all_deps = set()
+
+        for deps in self.dependency_dict.values():
+            for dep in deps:
+                all_deps.add(dep)
```
ialarmedalien: Can you use `all_deps.update` to add them in one go instead of iterating through and adding each one by one?

tests/test_generators/panderagen/test_panderagen_dependency_sorter.py
```diff
@@ -0,0 +1,40 @@
...
+    for node, dependencies in dependency_dict.items():
+        for dep in dependencies:
+            if sorted_list.index(dep) > sorted_list.index(node):
+                pytest.fail(f"Dependency '{dep}' should appear before its dependent '{node}'")
+
+
+def test_sort_dependencies():
+    ds = DependencySorter({"A": ["B", "C"], "B": [], "C": ["D"], "D": ["E"], "E": []})
+    result = ds.sort_dependencies()
+    # assert result == ['B', 'C', 'D', 'E', 'A']
```
ialarmedalien: either uncomment or delete this -- it is confusing to have commented-out lines in greenfield code

linkml/generators/panderagen/dependency_sorter.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def add_dependency(self, node: str, dependency: str = None):
+        """
+        Add a single dependency to the dependency dictionary.
+        If dependency is not provided or is None the node is added with an empty
+        dependency list.
+
+        Args:
+            node: The node to which the dependency is added.
+            dependency: The dependency to be added.
```
ialarmedalien: should be in rst format

linkml/generators/panderagen/dependency_sorter.py
```diff
@@ -0,0 +1,75 @@
...
+                visited.add(node)
+                result.append(node)
+
+        for node in self.dependency_dict:
+            if node not in visited:
+                visit(node)
+
+        return result
+
+    def add_dependency(self, node: str, dependency: str = None):
```
ialarmedalien, line 71: What do `node` and `dependency` refer to here? Without further documentation, it's difficult to work out what kind of objects we're dealing with and what this code actually does. `dependency` suggests py package deps, but the usage of `add_dependency` suggests these are links between classes in a schema. Help!

linkml/generators/panderagen/dependency_sorter.py
```diff
@@ -0,0 +1,75 @@
...
+        Raises:
+            ValueError: If there are cyclic dependencies
+        """
+        self.expand_dependencies()
+
+        result = []
+        visited = set()
+        temp_visited = set()
+
+        def visit(node: str):
```
ialarmedalien: This looks familiar... I just added a cycle detection algorithm to SchemaView in linkml-runtime!

tests/test_generators/panderagen/test_panderagen_dependency_sorter.py
```diff
@@ -0,0 +1,40 @@
...
+    assert_sorted_order(result, ds.dependency_dict)
+
+
+def test_cyclic_dependency():
+    ds = DependencySorter({"A": ["B"], "B": ["A"]})
+    with pytest.raises(ValueError) as cm:
+        ds.sort_dependencies()
+    assert "Cyclic dependency detected" in str(cm.value)
```
ialarmedalien:
```py
with pytest.raises(ValueError, match="Cyclic dependency detected"):
    ds.sort_dependencies()
```

linkml/generators/panderagen/dependency_sorter.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+        all_deps = set()
+
+        for deps in self.dependency_dict.values():
+            for dep in deps:
+                all_deps.add(dep)
+
+        for dep in all_deps:
+            if dep not in self.dependency_dict:
+                self.dependency_dict[dep] = []
```
ialarmedalien: Just thinking about it, this could be done with a comprehension and then dictionary merge:

```py
    expanded_dep_dict = {
        # add the dependency with an empty list as its dependencies
        # if it isn't a key in the dependency dict
        dep: [] for dep_list in self.dependency_dict.values() for dep in dep_list if dep not in self.dependency_dict
    }
    self.dependency_dict.update(expanded_dep_dict)
```

linkml/generators/panderagen/dependency_sorter.py
tfliss: @ialarmedalien I saw that! but this was independent of it.  I was going to ask you to review this PR for that reason but you beat me to it. I'm glad to discuss using or merging with your schema view dependency feature. There are several other things I kept in this generator while I worked out the details that could possibly be moved into SchemaView, either before these PRs are merged or as follow-up. I'll see about listing them out specifically in the next few days. Thanks again for reviewing so soon.

tfliss: Ok, I looked through schema_view detect-cycles() and _closure() for a while today, though I'm not super concerned with efficiency I wasn't able to re-use either of them without walking the graph a second time anyway.

It may be possible to extend detect_cycles to also return the dependency order, but I'd prefer doing that in a separate PR. I am modifying my implementation for now to hopefully make it clearer, and possibly closer to detect_cycles().

tfliss: An example of modifying detect_cycles() for use in place of visit():

```patch
diff --git a/linkml_runtime/utils/schemaview.py b/linkml_runtime/utils/schemaview.py
index f3ee003..5035bc2 100644
--- a/linkml_runtime/utils/schemaview.py
+++ b/linkml_runtime/utils/schemaview.py
@@ -96,7 +96,7 @@ GREY = 1
 BLACK = 2
 
 
-def detect_cycles(f: Callable[[Any], Iterable[Any] | None], x: Any) -> None:
+def detect_cycles(f: Callable[[Any], Iterable[Any] | None], x: Any) -> list:
     """Detect cycles in a graph, using function `f` to walk the graph, starting at node `x`.
 
     Uses the classic white/grey/black colour coding algorithm to track which nodes have been explored. In this
@@ -122,6 +122,7 @@ def detect_cycles(f: Callable[[Any], Iterable[Any] | None], x: Any) -> None:
     # processed_flag == True means all neighbours (nodes generated by running `f(node)`)
     # have been added to the todo stack and the node can be marked BLACK.
     todo: list[tuple[Any, bool]] = [(x, False)]
+    result = []
 
     while todo:
         node, processed_flag = todo.pop()
@@ -129,6 +130,7 @@ def detect_cycles(f: Callable[[Any], Iterable[Any] | None], x: Any) -> None:
         if processed_flag:
             # all neighbours have been processed
             processing_state[node] = BLACK
+            result.append(node)
             continue
 
         # check the state of this node
@@ -150,6 +152,8 @@ def detect_cycles(f: Callable[[Any], Iterable[Any] | None], x: Any) -> None:
 
         # push the neighbours on to the processing stack
         todo.extend((child, False) for child in f(node) or [])
+    
+    return result
 
 
 def _closure(
```

ialarmedalien: This can definitely be punted until after the merge of linkml and linkml-runtime, but having some generic graph functions in their own file would be great.

tfliss: Great! I'll open a low-priority issue on schemaview with notes from this conversation.
Otherwise I resolved the conversations where I accepted your suggestions verbatim.
The rest are doc/organization updates on dependency sorter that are ready to re-review.

This is the first PR in the chain so it is safe to merge after it's approved.

linkml/generators/panderagen/dependency_sorter.py
tfliss: @ialarmedalien I expanded the documentation on this file.  Hopefully that clarifies that I'm using string labels to represent the nodes in a dependency tree.  It's mostly an abstract data structure, though the purpose was to be a helper for the generator rather than full generality. Let me know if there's anything else I can clarify.

linkml/generators/panderagen/dependency_sorter.py
tfliss: When I looked at doing this change, I ended up doing a slightly larger refactor that hopefully make the algorithm and intent clearer. What it's really doing is just taking the 'leaf' nodes in the dependency tree and recording them at the dictionary top level as having an empty dependency list. I did use sum() to flatten the list of dependency lists, hopefully that's not too obscure.

linkml/generators/panderagen/dependency_sorter.py
tfliss: Opened https://github.com/linkml/linkml/issues/2997 for adding functionality to detect_cycles() in the future.

tfliss:
@amc-corey-cox Thanks for merging up. One notebook/example CI test unrelated (I think) to this PR is now failing though.  Is this a known/expected issue or should I investigate?

> FAILED tests/test_notebooks/test_examples.py::test_distributed_models - FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/linkml/linkml/attributes.yaml'

linkml/generators/panderagen/class_generator_mixin.py
```diff
@@ -1,20 +1,51 @@
-from linkml_runtime.linkml_model import ClassDefinitionName
+from .dependency_sorter import DependencySorter
 
 
 class ClassGeneratorMixin:
+    def enum_or_class(self, cn):
+        if cn in self.schemaview.all_enums():
+            return self.schemaview.get_enum(cn, strict=True)
+        elif cn in self.schemaview.all_classes():
```
ialarmedalien:
```py
elif cn in self.schemaview.all_classes():
```

can just be

```py
if cn in self.schemaview.all_classes():
```

since the previous statement was a `return`

linkml/generators/panderagen/transforms/simple_dict_loader.py
```diff
@@ -0,0 +1,54 @@
...
+        nested_tx=lambda x: x,
+    ):
+        self.struct_schema = struct_schema
+        self.id_col: str = id_col
+        self.other_col: str = other_col
+        self.id_dtype = id_dtype
+        self.other_dtype = other_dtype
+        self.nested_tx = nested_tx
+        self.polars_schema_keys = set(self.struct_schema.keys())
+        self.ordered_schema_keys = list(self.struct_schema.keys())
```
ialarmedalien, line 21: If `self.struct_schema` is a dict, you can get the keys with just `list(dictionary)` or `set(dictionary)` -- no need to call `keys()`.

linkml/generators/panderagen/transforms/simple_dict_loader.py
```diff
@@ -0,0 +1,54 @@
...
+            if isinstance(range_value, dict) and (set(range_value.keys()) <= self.polars_schema_keys):
+                base_dict = {k: None for k in self.ordered_schema_keys}
+                base_dict.update(self.nested_tx(range_value))
+                base_dict[self.id_col] = id_value
+                yield base_dict
+            else:
+                base_dict = {k: None for k in self.ordered_schema_keys}
+                base_dict[self.id_col] = id_value
+                base_dict[self.other_col] = range_value
+                yield base_dict
```
ialarmedalien: Minor simplification here:
```py
        for id_value, range_value in linkml_simple_dict.items():
            base_dict = {k: None for k in self.ordered_schema_keys}
            if isinstance(range_value, dict) and (set(range_value.keys()) <= self.polars_schema_keys):
                base_dict.update(self.nested_tx(range_value))
            else:
                base_dict[self.other_col] = range_value
            base_dict[self.id_col] = id_value
            yield base_dict
```

ialarmedalien: (Thought I had posted this comment when I reviewed previously, but I must not have clicked the button.)

PR #2947, opened 2025-10-16
2856 pandera dict to list transforms

tfliss:
This is part of a group of PRs that implements independent PolaRS support 
in the Pandera dataframe generator. https://github.com/linkml/linkml/issues/2856

This PR specifically adds improved transform classes that convert the inlined dict forms of linkml slots into inlined list forms that are better supported by dataframe libraries.  There is a unit test, but the new classes will not be used until a later PR.
Older forms of the transform will also be removed at that time.


tfliss:
@amc-corey-cox can we chat briefly Monday/next week? These PRs are split from a large branch to make reviews easier. If they are merged out of order, they will progressively return to the large form. I should have gotten the list of that order out sooner. Maybe I should also mark them all draft so they are merged in order.

PR #2951, opened 2025-10-17
2856 refactor dataframe helper out of compliance test helper

tfliss:
This is part of a group of PRs that implement multiple generation targets
in the Pandera dataframe generator. https://github.com/linkml/linkml/issues/2975
This is to enable PolaRS support https://github.com/linkml/linkml/issues/2856.
(see the description of 2856 for the full list of PRs)

The main tests to re-run on the specific functionality are:
uv run pytest -s -vv --cli-log-level=INFO -m panderagen tests/test_compliance/

This PR specifically: 

- refactors the `check_data_pandera()` function *mostly* verbatim out of helper.py in the test_compliance suite.
- adds a skip-list that will call pytest skip list if the schema name starts with a list entry.
- parametrizes the importerskip version for pandera that was hardcoded to 1.0
- adds a backing-form parameter to the panderagen constructor that is used in helper.py (to simplify next PR in the series)
- panderagen.py also had a few ruff fixes that I previously missed.

Note that the logic in `check_data_pandera()` will be improved in a future PR in this series once the PolaRS schema generator is enabled. A few additional inlining tests will also be skipped until that time.


tests/test_compliance/dataframe_helper.py
```diff
@@ -0,0 +1,74 @@
...
+    "test_class_boolean",
+    "test_non_standard",
+    "test_equals_string",
+    "test_value_presence",
+    "test_cardinality-MVTrue_REQFalse",
+    "test_cardinality-MVTrue_REQTrue",
+    "test_cardinality-ClassNameEQ_C__SlotNameEQ_sSPACE1__TypeNameEQ_tSPACE1",
+    "test_cardinality-ClassNameEQ_C__SlotNameEQ_1s__TypeNameEQ_T1",
+    "test_cardinality-ClassNameEQ_C__SlotNameEQ_1s__TypeNameEQ_T1",
+]
```
amc-corey-cox, line 38: Are we skipping these because they are not implemented or unnecessary? Will we be dealing with these later somehow?

tfliss: I think those tests fail because I need to improve slot name handling in edge cases, so not addressed in these PRs, but I may have a fix in work-in-progress code.

I'm moving toward tracking the skip list here rather than touching many individual test/test_compliance/*.py files just to skip or not skip individual tests while a generator is under development.  Very open to suggestions or questions on this approach.

Most of the other tests that are turned off in this PR that deal with inlining will be re-enabled by the end of these PRs, expanding the functionality overall.

PR #2973, opened 2025-10-25
Fix: Clean YARRRML generator output & CLI availability

lapkinvladimir:
This PR fixes and refines the YARRRML generator behavior and output.

### What it does
Restores proper YAML structure (mappings - sources - s - po)  
Ensures classes without identifiers are included with fallback subject  
Removes invalid prefixes and !!python/object garbage from YAML  
Adds CLI alias `gen-yarrrml` for easier use  
Keeps JSON and CSV/TSV generation consistent  

### Tests
Regression tests for:
- inclusion of classes without identifiers (fallback subject)  
- YAML serialization cleanup (no !!python/object)  
- CLI behavior  
Updated e2e tests for JSON / CSV / TSV  
All previous tests pass (1 xfail preserved intentionally)

### DIFF
--- a/tests/test_generators/test_yarrrmlgen.py
+++ b/tests/test_generators/test_yarrrmlgen.py
@@ -273,7 +273,7 @@ default_prefix: ex
 default_range: string
 
 classes:
-  Notification:
+  Note:
     attributes:
       text: {}
 """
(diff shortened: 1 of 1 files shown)


amc-corey-cox (approved):
Yes, this looks good!

PR #2977, opened 2025-10-27
Merge linkml-runtime into linkml, creating a uv monorepo

kevinschaper:
Re-do of #2873, merging runtime into linkml as part of a uv monorepo

Initially, this PR has the git work done and some minimal changes to the top level and runtime pyproject.toml, but fails on uv sync. Pushing to a draft PR to share the work of fixing it.

### DIFF
--- a/packages/linkml_runtime/pyproject.toml
+++ b/packages/linkml_runtime/pyproject.toml
@@ -250,6 +250,7 @@ fallback-version = "0.0.0"
 source = "uv-dynamic-versioning"
 
 [tool.hatch.build.targets.wheel]
+packages = ["."]
 
 [tool.hatch.build.targets.sdist]
 include = ["**/*"]


sneakers-the-rat:
Reupping suggestion of using namespace package structure https://github.com/linkml/linkml/pull/2873#issuecomment-3412951676

dalito:
@sneakers-the-rat - where would you see the advantage of PEP420 namespaces (which would put linkml and linkml-runtime under the same src/ directory with a single pyproject.toml) vs. uv-workspaces where each package has its own pyproject.toml in addition to a root-pyproject.toml? As far as I know one goal was to still be able to realease linkml and linkml-runtimes independently which is not possible with PEP420 (or is it?).

Regarding tests I would also prefer to have them per package as you suggested.

sneakers-the-rat:
> PEP420 namespaces (which would put linkml and linkml-runtime under the same src/ directory with a single pyproject.toml

not so! [here](https://github.com/libranet/demo-uv-workspace/tree/main) is an example repo using uv workspaces and namespace packages.

The directory structure would be something like this

```
.
├── packages
│   ├── linkml-pydantic
│   │   ├── pyproject.toml
│   │   └── src
│   │       └── linkml
│   │           └── generators
│   │               └── pydantic
│   ├── linkml-runtime
│   │   ├── pyproject.toml
│   │   └── src
│   │       └── linkml
│   │           └── runtime
│   └── linkml-schemaview
│       ├── pyproject.toml
│       └── src
│           └── linkml
│               └── schemaview
└── pyproject.toml
```

where each subpackage can be independent:
- e.g. distributed as `linkml-runtime` , `linkml-schemaview` , etc. 
- have independent deps (including within-monorepo deps as per [the core/modular install issue](https://github.com/linkml/linkml/issues/1784#issuecomment-2031033942) )

but then each becomes part of a shared namespace, so one might install `linkml-runtime` and then do `from linkml import runtime` rather than `from linkml_runtime import ...`.

the reason this is preferable to just piling the existing packages into a single repository is that it gives us room to further modularize the package in a way that has a single design, rather than making this huge change and still being stuck with "well there are these legacy packages that behave differently but..." - specifically i want to split off the pydantic generator into its own package that i can install separately, but in order to avoid breaking everything, ideally it would still be installed and used like `from linkml.generators.pydantic` ...

edit: and also we could keep the tests separate too, i just didn't put that in the example file tree

edit2: we could do both, but i figure having all the subpackages behave similarly would probably be a good change instead of having a few "special" packages that are not part of the `linkml` namespace and that becoming confusing in a few years

dalito:
Thanks for the detailed reply @sneakers-the-rat . This structure is also what I meant with "uv-workpaces" solution. I was just not sure if you implied something different with the PEP-240 reference.

Issue #2983, opened 2025-10-30
Snapshot diff display broken, calling private pytest function with incorrect signature


sneakers-the-rat:
---
name: Bug report
about: Create a report to help us improve
title: ''
labels: bug
assignees: ''

---

**Describe the bug**

when snapshots don't match, instead of showing the diff, pytest goes like

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/conftest.py:104: in compare_to_snapshot
    self.eq_state = "\n".join(_diff_text(actual, expected, self.config.getoption("verbose")))
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

left = '/ns/shex#"\n    },\n    {\n      "@base": "https://w3id.org/linkml/"\n    }\n  ]\n}\n'
right = '/ns/shex#",\n      "@vocab": "https://w3id.org/linkml/"\n    },\n    {\n      "@base": "https://w3id.org/linkml/"\n    }\n  ]\n}\n', highlighter = 0, verbose = 0

    def _diff_text(
        left: str, right: str, highlighter: _HighlightFunc, verbose: int = 0
    ) -> list[str]:
        """Return the explanation for the diff between text.
    
        Unless --verbose is used this will skip leading and trailing
        characters which are identical to keep the diff minimal.
        """
        from difflib import ndiff
    
        explanation: list[str] = []
    
        if verbose < 1:
            i = 0  # just in case left or right has zero length
            for i in range(min(len(left), len(right))):
                if left[i] != right[i]:
                    break
            if i > 42:
[...]

PR #2984, opened 2025-10-30
use a kwarg when calling pytest's `_diff_text`

sneakers-the-rat:
fix: #2983

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -101,7 +101,7 @@ class SnapshotFile(Snapshot):
             if not is_eq:
                 # TODO: probably better to use something other than this pytest
                 # private method. See https://docs.python.org/3/library/difflib.html
-                self.eq_state = "\n".join(_diff_text(actual, expected, self.config.getoption("verbose")))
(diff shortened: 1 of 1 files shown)


dalito:
Interesting point: Why were no tests run on this PR? It broke main.

sneakers-the-rat:
i think they did? https://github.com/linkml/linkml/actions/runs/18956818397/job/54135494144

no way this PR could have broken that.

what's breaking is that the schema in that notebook imports biolink (somehow, the URI it's using 404's), and then biolink imports `attributes`: https://github.com/biolink/biolink-model/blob/02953e948a8574d925ec3ba24bdd77499e47e028/biolink-model.yaml#L208

which is relative to it, but the SchemaLoader doesn't resolve it relative to the biolink URI. The question really seems to be how that ever would have worked, because in the 'working' case it would try and import `https://w3id.org/biolink/attributes` which also 404's. 

maybe biolink updated their site and broke the w3id uris? idk

dalito:
> no way this PR could have broken that.

True. I just saw main fails with this. But looking at the commits in main a lot of the recent commits failed: https://github.com/linkml/linkml/commits/main/ (so the problem was introduced before this PR)

PR #2988, opened 2025-10-31
conflict fixes, conftest changes and pre-commit config for runtime combination PR

kevinschaper:
Merge Conflicts Resolved:
  - Root pyproject.toml: Added tool.* configs from main while preserving workspace structure
  - packages/linkml/pyproject.toml: Added missing license-files field and gen-yarrrml script
  - Accepted pydantic_ifabsent_processor.py at new monorepo location

  Pytest Fixes:
  - Created root conftest.py to register pytest CLI options (--with-slow, --with-network, etc.)
  - Updated packages/linkml/tests/conftest.py to use getattr() instead of getoption() (workaround for monorepo scoping issue)
  - Added testpaths = ["packages/linkml/tests"] to pytest config

  Pre-commit:
  - Excluded packages/linkml_runtime/ from root pre-commit to preserve its separate standards


pyproject.toml
```diff
@@ -1,68 +1,8 @@
...
-  | __snapshots__
-  | docs
-  | examples
-  | notebooks
-  | linkml/linter/config/datamodel
-)/
-'''
-
 [tool.pytest.ini_options]
+testpaths = ["packages/linkml/tests"]
```
kevinschaper: We probably want this to include the runtime tests as well, rather than having an action to run the runtime tests in addition to the top level te

PR #3001, opened 2025-11-11
Add links to new LinkML tutorial video and associated slides

nlharris:
and remove three old (2021) presentations

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -52,6 +52,7 @@ The LinkML registry collects some existing schemas and allows for basic browsing
 
 ## Presentations about LinkML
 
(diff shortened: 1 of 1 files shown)


nlharris:
Some tests (looks like the windows and ubuntu ones) are failing; I assume that has nothing to do with this PR, which only affects documentation.

Test results: https://github.com/linkml/linkml/actions/runs/19279776345

sierra-moxon:
this is the biolink test failure I was talking about in our meeting today; working on a fix.  I wouldn't block on this failure, but if I can get it fixed today, we can make this green and pretty before merge.

PR #3006, opened 2025-11-14
FAQ alternative frameworks for linkml paper revisions

sierra-moxon:


docs/faq/why-linkml.md
```diff
@@ -401,6 +401,98 @@ semantically describe JSON files with LinkML too.
 
 We are currently planning on writing a generator for CSVW JSON-LD.
 
+## Why should I use LinkML over Frictionless Data
```
nlharris:
```suggestion
## Why should I use LinkML over Frictionless Data?
```
(Just adding missing question mark.)
TBH, I would probably have phrased all of these with "rather than" or "instead of" rather than "over" (because I worry that "over" could be misinterpreted as "above" (in a technical sense, like above it in the stack, or upstream of it) but I guess we can leave them all "over"

docs/faq/why-linkml.md
```diff
@@ -401,6 +401,98 @@ semantically describe JSON files with LinkML too.
...
+LinkML offers sophisticated validation rules including patterns, ranges, and dynamic enumerations linked to ontologies, 
+and can generate Python dataclasses, TypeScript interfaces, JSON-Schema, SQL DDL, and more from a single schema.
+
+If you're specifically working with ML datasets and need schema.org compatibility, Croissant may be appropriate.
+
+## Why should I use LinkML over Schema.org?
+
+[Schema.org](https://schema.org) is a collaborative initiative that provides a collection of schemas (structured data vocabularies) 
+used to mark up web pages so that search engines can better understand their content. It's supported by major search 
+engines including Google, Microsoft, Yahoo, and Yandex, and is used by over 45 million websites.
```
nlharris: Yahoo seems past relevance and I've never heard of Yandex. This seems like marketing copy.
```suggestion
engines including Google and Microsoft and is used by over 45 million websites.
```

docs/faq/why-linkml.md
```diff
@@ -401,6 +401,98 @@ semantically describe JSON files with LinkML too.
...
+
+However, Schema.org is a vocabulary that you can use to annotate data for use on the web.  LinkML provides rich
```
nlharris:
```suggestion
Schema.org is, at heart, a vocabulary that you can use to annotate data for use on the web. LinkML, in contrast, provides rich 
```

nlharris (changes requested):
Looks great! I made a few minor suggestions.

PR #3012, opened 2025-11-26
pydanticgen: make the default for optional list configurable

kervel:
Historical behaviour for optional lists in pydantic was to initialise them with None. However, https://github.com/linkml/linkml/pull/2877 changed this behaviour to empty list. This caused regressions for us (see example below). This patch makes the behaviour of pydantic for optional lists configurable, so we can restore the old behaviour.

Test case:

```yaml
  AssociatedNetElement:
    description: >-
      A RSM Associated Net Element
    attributes:
      netElement:
        range: PositioningNetElement
        inlined: false
      bounds:
        range: float
        multivalued: true
        minimum_cardinality: 1
        maximum_cardinality: 2
      appliesInDirection:
        range: string
      lateralPosition:
        range: string # FIXME custom class LateralPosition instead of str in RSM
```

After PR #2877 AssociatedNetElement() will raise an error (validation fails because altough `bounds` is not required, when it is filled in, the cardinality is mandatory).

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -24,7 +24,7 @@ jobs:
 
       - name: Docker metadata
         id: meta
-        uses: docker/metadata-action@v5.8.0
+        uses: docker/metadata-action@v5.10.0
         with:
           images: linkml/linkml
           tags: |
@@ -41,7 +41,7 @@ jobs:
           echo "Ref: ${{ github.ref }}"
 
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v3.6.0
+        uses: docker/setup-qemu-action@v3.7.0
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3.11.1
--- a/.github/workflows/pypi-publish.yaml
+++ b/.github/workflows/pypi-publish.yaml
@@ -32,7 +32,7 @@ jobs:
         uv build --all-packages
 
     - name: Store built distribution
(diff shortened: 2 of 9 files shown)


sierra-moxon (approved):

sierra-moxon:
@kervel - thanks a million - we are current "frozen" for PR merging until Monday (the monorepo), but I will merge in ASAP. I think we should be able to get this one in before the monorepo merge, but I have to check with @kevinschaper (monorepo dev) who is off for holiday this week.

kervel:
hello @sierra-moxon ! for me totally fine if we merge this after the monorepo transition. in the mean time i'll try to see if all compat issues i had are solved with this PR.

kevinschaper:
Thanks for being ok with waiting on this, and sorry for letting the regression in. I wonder if we could find some appropriate model semantics to trigger empty vs null? I’m ok with the PR, though I actually think it might be more appropriate to default to False - @EOltmanns would you be ok adding a flag when you build models after the next release?

EOltmanns:
@kevinschaper thanks for the heads up. Yes, I am fine with adding a flag after the next release. In fact, since #2877 turned out to introduce a regression after all, I actually agree that a default of False rather than True might be more appropriate. Last,but not least, thanks to @kervel for rewolving the issue in this way.

kervel:
hi! what shall we do now ? review behaviour ? should i change anything to this PR ?

kervel:
@kevinschaper i updated the test to reflect the new default

PR #3030, opened 2025-12-08
Add initial CLAUDE.md and coverage

amc-corey-cox:
This adds a basic CLAUDE.md based on recommendations from @vladistan and ignores CLAUDE.local.md. I've also added a sub-file Claude can parse for understanding how to pull the coverage report and identify issues with low coverage because this was something I struggled to get it to do correctly.

Vlad, I hope I understood your recommendations correctly please take a look and give me feedback on how to make this better.

### DIFF
--- a/tests/linkml/test_utils/test_deprecation.py
+++ b/tests/linkml/test_utils/test_deprecation.py
@@ -4,7 +4,6 @@ from operator import eq, ge, gt, le, lt
 from unittest.mock import patch
 
 import pytest
-
 from linkml.utils import deprecation as dep_mod
 from linkml.utils.deprecation import DEPRECATIONS, EMITTED, Deprecation, SemVer, deprecation_warning
 


kevinschaper:
I think we'll add lots more, but I'm happy with this as a start.

sierra-moxon:
@amc-corey-cox - I added a few more best practices copied from other repos (from Chris) that are great (imo).  let me know what you think.

CLAUDE.md
```diff
@@ -0,0 +1,37 @@
...
+| `linkml` | `packages/linkml/src/linkml/` | `tests/linkml/` | [linkml](https://pypi.org/project/linkml/) |
+| `linkml-runtime` | `packages/linkml_runtime/src/linkml_runtime/` | `tests/linkml_runtime/` | [linkml-runtime](https://pypi.org/project/linkml-runtime/) |
+
+All commands use `uv run` prefix (e.g., `uv run pytest`).
+
+## Task-Specific Guides
+
+| Task | Guide |
+|------|-------|
+| Code coverage reports | [.claude/coverage.md](.claude/coverage.md) |
```
cmungall: use a subagent or skill

AGENTS.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Task-Specific Guides
+
+| Task | Guide |
+|------|-------|
+| Code coverage reports | [.claude/coverage.md](.claude/coverage.md) |
+
+## Best Practices
+
+* Check CI first - When something doesn't work locally, check `.github/workflows/main.yaml` to see how CI handles it. CI often uses external services or has workarounds for local limitations.
```
cmungall: this is a bit ambiguous - does this mean push then check CI via `gh` commands?

CLAUDE.md
```diff
@@ -0,0 +1,37 @@
+# Claude Code Notes for LinkML
+
+## Project Structure
+
+This is a UV workspace monorepo that publishes **two PyPI packages**:
+
+| Package | Source | Tests | PyPI |
+|---------|--------|-------|------|
```
vladistan: Just chiming in.  This line of dashes is a lot of tokens,  and it doesn't actually add any info for the agent to act on.   I suggest dropping it.

amc-corey-cox:
This is all great feedback. Thank you! I will work these in.

CLAUDE.md
amc-corey-cox: I have removed this with the change to skills like cmungall suggested. Thanks!

amc-corey-cox:
@cmungall I think this should address all of the points made. I'd like to get this merged soon as there is an annoying dependency issue that @sierra-moxon fixed here that I would like to get into main so we can merge other PRs.

sierra-moxon (approved):

Issue #3002, opened 2025-11-12
Integrate community meeting summaries into main docs


cmungall:
Currently if I want to ask a question such as "has anyone tried modeling CDISC in LinkML", I need to search the main site, issues, discussions, slack, and also external google docs such as the community meeting notes google doc.

The latter is particularly hard to find, you need to find this FAQ entry: https://linkml.io/linkml/faq/contributing.html#how-do-i-stay-involved

It would be easy to have a github action that regularly downloads the community meeting google doc as markdown and makes a PR placing it in the repo. Optionally there could be some lightweight human or AI curation to remove administrivia, but not sure if that is necessary.

sierra-moxon:
we also have the auto-generated registry: [](https://linkml.io/linkml-registry/registry/)

<img width="1388" height="945" alt="Image" src="https://github.com/user-attachments/assets/ffafee46-0170-47a2-a20b-a20656e25638" />

we have the open item though, to increase coverage of the autogeneration of these registry items to include a search/retrieval step for GitLab in addition to the existing GitHub scraping it does...

sierra-moxon:
This is separate from the idea that we should have a pointer to community meeting notes on the main site - this would be awesome!  (esp with links to youtube content for presentations and a link to all the previous presentation topics as you suggest here).

nlharris:
Maybe @sagehrke and I can look into this...

sagehrke:
We also have this: https://linkml.io/community
Which I could see actually making more sense moving to the linkml.io/linkml instead and then we can just have a link to the docs (or do this fancy version with GH actions 😄 )

It seems like maybe we should make a new link in the left navigvation of linkml.io/linkml that is "Get involved" and that can have things like this. What do you think?

cmungall:
oh I didn't know that site existed! Yes this should go to linkml.io/linkml!

sagehrke:
Totally! I had Carlo make this for me in the summer before I realized that is a splash page vs the documentation site. I will add the migration to my to-do list. I will plan to work on it during DDD!

sagehrke:
Related to this: Add the LinkML ICBO2025 workshop to the documentation – maybe making an “educational resources” page? TBD. I will think about a nice way for organize this so it's intuitive to find all the community opprotunities.

sagehrke:
As of today, we added a new "Get Involved" navigatoin option in linkml.io/linkml. You can find that here: https://linkml.io/linkml/get-involved/index.html
[...]

PR #3036, opened 2025-12-12
Update get-involved section of the documentation on linkml.io/linkml

sagehrke:
- Update the get-involved index.rst to now include information about mailing lists, slack, linkedin, mastodon.
- Create a new workshops-and-presenations.md under the get-involved section of the documentation where we can list past outreach the community may be interested in watching
- Update the get-involved index.rst to show the new workshops-and-presenations.md in the navigation page

Addresses #3002


docs/get-involved/workshops-and-presentations.md
```diff
@@ -0,0 +1,21 @@
+# Workshops and Presentations
```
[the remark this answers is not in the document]
sagehrke: This is a good point, but I am not concerned about this. There are files with both upper and lower case names in the documentation.

sagehrke:
@amc-corey-cox I believe both errors that popped up are not a concern. But let me know if something should be changed before we can push this. thanks!

nlharris (approved):
Looks good! Let's merge and then we can check to make sure there aren't any formatting issues.

Are you posting on Mastodon for Monarch? I think I'm ready to give up on it for BOSC; too little uptake.

nlharris:
I see the check flagged two pretty trivial things:
> /home/runner/work/linkml/linkml/docs/get-involved/workshops-and-presentations.md.rst:6: WARNING: Non-consecutive header level increase; H1 to H3 [myst.header]

@sagehrke do you want to fix this so it stops complaining? (I don't see why it's a huge issue)

> /home/runner/work/linkml/linkml/docs/get-involved/workshops-and-presentations.md:21: ERROR: Document may not end with a transition. [docutils]

I am not sure what this means.

docs/get-involved/workshops-and-presentations.md
```diff
@@ -0,0 +1,20 @@
...
+
+This page is regularly updated to share links to previous LinkML workshops and presentations. Happy viewing! 
+_____
+
+### Annotating Data with Ontologies: LinkML Can Help
```
nlharris: > /home/runner/work/linkml/linkml/docs/get-involved/workshops-and-presentations.md.rst:6: WARNING: Non-consecutive header level increase; H1 to H3 [myst.header]

if you change this to ## rather than ### that should fix this warning. (check if there are other ###s in the document)

nlharris (changes requested):

docs/get-involved/workshops-and-presentations.md
sagehrke: True. I purposefully made the H1 to H3 just because I like the look better, but I won't die on this hill 😄 I changed it so the error should be cleared.

nlharris:
I don't understand why "[Build and test linkml / quality-checks (pull_request)](https://github.com/linkml/linkml/actions/runs/20282717564/job/58248719045?pr=3036) is failing, but probably that doesn't matter so I guess it's ok to merge?

PR #3045, opened 2025-12-18
Sort type declaration classes in `pythongen` according to inheritence graph

### LINKED ISSUE #2600: Cannot use derived class as `range` for a `key` slot
I am trying to do something similar to `linkml` annotations:

```yaml
  Annotation:
    description: >-
      A tag/value pair with the semantics of OWL Annotation.
    slots:
      - annotation_tag
      - annotation_value
    slot_usage:
      annotation_tag:
        key: true
```

things are fine when `annotation_tag` is something like this

```yaml
  annotation_tag:
    description: A tag identifying an annotation.
    range: uriorcurie
```

things are also fine, when the `range` of `annotation_tag` is a class with an `identifier:true` slot.

things are no longer fine, when `range` is a subclass of such a class. In other words. The following `AnnotationTag` class is not acceptable, whenever its base class `Thing` is.

```yaml
  AnnotationTag:
    is_a: Thing
```

The error I am getting from `gen-python` is

```
  File "<stdin>", line 179, in <module>
NameError: name 'AnnotationTagPid' is not defined
```

christian-monch:
Fixes #2600

### DIFF
--- a/packages/linkml/src/linkml/generators/common/ifabsent_processor.py
+++ b/packages/linkml/src/linkml/generators/common/ifabsent_processor.py
@@ -73,7 +73,7 @@ class IfAbsentProcessor(ABC):
     URI_SPECIAL_CASES = ("class_uri", "slot_uri")
     CURIE_SPECIAL_CASES = ("class_curie", "slot_curie")
     DEFAULT_RANGE_SPECIAL_CASE = "default_range"
-    UNIMPLEMENTED_DEFAULT_VALUES = ("default_ns",)
+    UNIMPLEMENTED_DEFAULT_VALUES = ("bnode", "default_ns")
 
     def __init__(self, schema_view: SchemaView):
         self.schema_view = schema_view
--- a/packages/linkml/src/linkml/generators/jsonschemagen.py
+++ b/packages/linkml/src/linkml/generators/jsonschemagen.py
@@ -439,9 +439,21 @@ class JsonSchemaGenerator(Generator, LifecycleMixin):
 
         if cls.none_of is not None and len(cls.none_of) > 0:
             subschema["not"] = {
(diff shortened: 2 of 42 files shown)


kevinschaper:
This one makes me a little nervous, because it exposes the circular dependency that we still have. The core tooling in linkml_runtime uses generated python from linkml-model, so we would need to test this against a metamodel branch that uses this branch to build the pythongen models used by schemaview etc.

kevinschaper:
@christian-monch Sorry that we've been sitting on this one, pythongen changes are potentially high stakes, because they're really at the core of how schemaview etc operate.  

1. Right now the snapshots are out of date, which is causing test failures, they can be updated with
```
uv run pytest tests/linkml/test_scripts/test_gen_python.py --snapshot-update
```

2. It would be good, though maybe not strictly necessary, to add a test that demonstrates pythongen generated python failing without this change, so that we have a regression test in place.  

3. The big challenge is that we need to demonstrate that pythongen generated classes for the metamodel will work properly with this change. @matentzn just worked through the complex circular dependency to create a new linkml-model release, so he will have the strongest understanding here. My understanding is that the meta.py used in this repository is a local copy, rather than imported. I think that means that we can actually test the whole circle of the dependency in this repo by updating the meta.py using 

```
uv run gen-python --genmeta packages/linkml_runtime/src/linkml_runtime/linkml_model/model/schema/meta.yaml > packages/linkml_runtime/src/linkml_runtime/linkml_model/meta.py
```

This is all pretty deep in the core of linkml though, so you can also feel free to hand this off to somebody on the core dev team to handle the additional work.

kevinschaper (approved):
This is ready to go, thanks @christian-monch!  I made a couple of commits to get updated snapshots in place so that tests will pass.

PR #3046, opened 2025-12-18
fix linting errors on main

sierra-moxon:
This just removes trailing whitespace - it causes code quality checks to fail on main atm, so important to get it in for other PRs.  If the linting checks fail on PR, run `make lint-fix` from the command line and it will automatically fix the things it can.


nlharris (approved):
Thanks! I didn't realize trailing spaces were viewed as errors!

PR #3061, opened 2026-01-07
Fix SHACL generator to use rdfs:comment for NodeShape descriptions

### LINKED ISSUE #3059: SHACL generator improperly adds `sh:description` to node shapes
**Describe the bug**

When generating a SHACL shape from a LinkML model like this:

```yaml
name: People
id: https://example.org/shapes#People
imports:
  - linkml:types
prefixes:
  ex: https://example.org/shapes#
  foaf: http://xmlns.com/foaf/0.1/
  schema: http://schema.org/
default_prefix: ex

slots:
  knows:
    slot_uri: foaf:knows
    multivalued: true

classes:
  Person:
    class_uri: schema:Person
    description: 'A person!'
    slots:
      - knows
```

I get:

```turtle
@prefix foaf: <http://xmlns.com/foaf/0.1/> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix schema: <http://schema.org/> .
@prefix sh: <http://www.w3.org/ns/shacl#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

schema:Person a sh:NodeShape ;
    sh:closed true ;
    sh:description "A person!" ;
    sh:ignoredProperties ( rdf:type ) ;
    sh:property [ sh:order 0 ;
            sh:path foaf:knows ] ;
    sh:targetClass schema1:Person .
```

Note that `classes.Person.description` was translated to `sh:description` applied to the root `sh:NodeShape`.

This is unfortunately inconsistent with the SHACL vocabulary which says that the `rdfs:domain` of `sh:description` is `sh:PropertyShape`: https://www.w3.org/ns/shacl.ttl

If the SHACL validator applies RDFS inferencing on top of the shapes graph (Eclipse RDF4J does this, for example), we get:

```
schema:Person a sh:NodeShape , sh:PropertyShape .
```

...and the validator goes haywire because a NodeShape can't be also a PropertyShape.

Latest (1.9.6).

**Suggested fix**

Just don't use sh:description for this. Instead we can use `rdfs:comment`, a solution that's used by the SHACL spec itself: https://www.w3.org/TR/shacl/#shacl-shacl

kevinschaper:
## Summary

- Fix SHACL generator to use `rdfs:comment` instead of `sh:description` for NodeShape (class) descriptions
- PropertyShape (slot) descriptions continue to use `sh:description` as before (correct per SHACL spec)
- Add regression test to verify the fix

Fixes #3059

## Problem

Per the SHACL specification, `sh:description` has `rdfs:domain` of `sh:PropertyShape`. When used on a `sh:NodeShape`, RDFS-aware validators (like Eclipse RDF4J with `inference="rdfs"`) incorrectly infer that the NodeShape is also a PropertyShape, causing validation failures.

Notably, LinkML's own `ShaclValidationPlugin` uses RDFS inferencing, so this bug could affect LinkML users validating schemas where classes have descriptions.

## Solution

Use `rdfs:comment` for NodeShape descriptions instead of `sh:description`. This follows the pattern used by the SHACL specification itself (see [shacl-shacl](https://www.w3.org/TR/shacl/#shacl-shacl)).

## ⚠️ Downstream Impact (Breaking Change)

**If you consume generated SHACL output**, be aware of the following change:

| Element | Before | After |
|---------|--------|-------|
| Class descriptions | `sh:description` | `rdfs:comment` |
| Slot descriptions | `sh:description` | `sh:description` (unchanged) |

### Example

**Before:**
```turtle
schema:Person a sh:NodeShape ;
    sh:description "A person, living or dead" ;
    ...
```

**After:**
```turtle
schema:Person a sh:NodeShape ;
    rdfs:comment "A person, living or dead" ;
    ...
```

### Who is affected?

- Anyone querying generated SHACL for `sh:description` on NodeShapes will need to query for `rdfs:comment` instead
- Projects that commit generated SHACL files will see diffs when regenerating
- The `linkml-model` repo's `meta.shacl.ttl` may need regeneration

### Who benefits?

- Users of RDFS-aware SHACL validators will no longer see incorrect validation failures
- Generated SHACL is now spec-compliant

## Test plan

- [x] Added `test_nodeshape_description_uses_rdfs_comment` test
- [x] Verified NodeShapes use `rdfs:comment` for descriptions
- [x] Verified NodeShapes do NOT have `sh:description`
- [x] Verified PropertyShapes still use `sh:description`
- [x] All existing SHACL generator tests pass

---

cc @Ostrzyciel - As the issue submitter, could you please review this fix and confirm it addresses the problem you encountered? Specifically, does using `rdfs:comment` for NodeShape descriptions resolve the validation issues you saw with Eclipse RDF4J?

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/linkml/src/linkml/generators/shaclgen.py
+++ b/packages/linkml/src/linkml/generators/shaclgen.py
@@ -91,7 +91,9 @@ class ShaclGenerator(Generator):
             else:
                 shape_pv(SH.closed, Literal(False))
             if c.title is not None:
-                shape_pv(SH.name, Literal(c.title))
+                # Use rdfs:label for NodeShape titles per SHACL spec.
+                # sh:name has rdfs:domain of sh:PropertyShape. See issue #3059.
+                shape_pv(RDFS.label, Literal(c.title))
             if c.description is not None:
                 # Use rdfs:comment for NodeShape descriptions per SHACL spec.
                 # sh:description has rdfs:domain of sh:PropertyShape, so using it
--- a/tests/linkml/test_generators/test_shaclgen.py
+++ b/tests/linkml/test_generators/test_shaclgen.py
@@ -592,12 +592,12 @@ def test_exclude_imports(input_path):
     assert "https://example.org/baseProperty" in property_paths
 
 
-def test_nodeshape_description_uses_rdfs_comment(kitchen_sink_path):
-    """Test that NodeShape descriptions use rdfs:comment, not sh:description.
+def test_nodeshape_uses_rdfs_predicates(kitchen_sink_path):
+    """Test that NodeShapes use rdfs:label and rdfs:comment, not sh:name and sh:description.
 
-    Per the SHACL spec, sh:description has rdfs:domain of sh:PropertyShape,
-    so using it on NodeShapes causes RDFS-aware validators to incorrectly
-    infer the NodeShape is also a PropertyShape. See issue #3059.
+    Per the SHACL spec, sh:name and sh:description both have rdfs:domain of sh:PropertyShape,
+    so using them on NodeShapes causes RDFS-aware validators to incorrectly infer the
+    NodeShape is also a PropertyShape. See issue #3059.
     """
     shacl = ShaclGenerator(kitchen_sink_path, mergeimports=True).serialize()
     g = rdflib.Graph()
@@ -616,6 +616,11 @@ def test_nodeshape_description_uses_rdfs_comment(kitchen_sink_path):
     # Verify NodeShape does NOT have sh:description (this was the bug)
     nodeshape_sh_descriptions = list(g.objects(person_uri, SH.description))
     assert len(nodeshape_sh_descriptions) == 0, "NodeShapes should not use sh:description; use rdfs:comment instead"
+
+    # Verify no NodeShape has sh:name (sh:name also has rdfs:domain sh:PropertyShape)
+    for node_shape in g.subjects(RDF.type, SH.NodeShape):
+        sh_names = list(g.objects(node_shape, SH.name))
+        assert len(sh_names) == 0, f"NodeShape {node_shape} should not use sh:name; use rdfs:label instead"
 
     # Verify PropertyShapes still use sh:description (this is correct per spec)
     # Check that at least one property shape (BNode) uses sh:description


packages/linkml/src/linkml/generators/shaclgen.py
```diff
@@ -93,7 +93,11 @@ def shape_pv(p, v):
             if c.title is not None:
                 shape_pv(SH.name, Literal(c.title))
```
Ostrzyciel: I just noticed that sh:name also has the same domain as sh:description. See:

```turtle
sh:name
	a rdf:Property ;
	rdfs:label "name"@en ;
	rdfs:comment "Human-readable labels for the property in the context of the surrounding shape."@en ;
	rdfs:domain sh:PropertyShape ;
	# range: xsd:string or rdf:langString
	rdfs:isDefinedBy sh: .
```

So this should also be replaced with something else, e.g., rdfs:label.

Ostrzyciel (changes requested):

Ostrzyciel:
Thanks for the PR! I had a look at the code and noticed one more issue.

kevinschaper:
Thanks for the review help @Ostrzyciel, I'm not too familiar with SHACL so it's very much appreciated.

packages/linkml/src/linkml/generators/shaclgen.py
kevinschaper: Good catch\! I've updated the PR to also replace `sh:name` with `rdfs:label` for NodeShape titles. The test has been expanded to verify that no NodeShape uses either `sh:name` or `sh:description`.

kevinschaper:
## Updated PR

Based on feedback from @Ostrzyciel, I've expanded this PR to also fix the `sh:name` issue:

### Additional Changes (commit 8ec060897)

| Element | Before | After |
|---------|--------|-------|
| Class titles | `sh:name` | `rdfs:label` |
| Class descriptions | `sh:description` | `rdfs:comment` |
| Slot titles | `sh:name` | `sh:name` (unchanged) |
| Slot descriptions | `sh:description` | `sh:description` (unchanged) |

The test has been renamed to `test_nodeshape_uses_rdfs_predicates` and expanded to verify that no NodeShape uses either `sh:name` or `sh:description`.

PR #3069, opened 2026-01-11
Add support for "true" enumerated types in Java

### LINKED ISSUE #3068: Enums are rendered as String-typed fields in Java
**Describe the bug**
The code generated by the Java generator has absolutely no support for enumerations, which results in:

(1) no code being generated at all for the enumerations, and

(2) enum-typed slots being rendered as String-typed fields.

This makes the Java generator next to useless when working on any schema that makes use of enumerations.

Tip of the main branch.

Any schema that contains at least one enumeration and a class with a slot whose range is set to that enumeration.

gouttegd:
Fixes #3068

This PR allows the Java code generator to represent LinkML enumerations as _bona fide_ Java `enum` objects (#3068).

It adds a new boolean option `true_enums` to the `OOCodeGenerator` class; that option enables two distinct behaviours in the generator:

(a) the list of `OODocument` objects returned by the `create_documents()` will contain one document for each enumeration, in addition to one document for each class (it is up to the callers of that method to know what to do with those documents; in this PR, only the Java code generator is updated accordingly – part of the reason why `true_enums` defaults to `False`, so as to avoid breaking the other generators);

(b) enum-typed slots are assigned a range that corresponds specifically to their enum, instead of being assigned a generic string type.

The templates used by the Java code generator are updated to cope with the possibility that the document they are supposed to generate may contain an enumerated type rather than a class; the template used in the former case is based on what has been proposed in #3068.

### DIFF
--- a/.github/workflows/check-external-links.yaml
+++ b/.github/workflows/check-external-links.yaml
@@ -16,12 +16,12 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Set up Python 3.
-        uses: actions/setup-python@v6.0.0
+        uses: actions/setup-python@v6.1.0
         with:
           python-version: "3.12"
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v7.1.6
+        uses: astral-sh/setup-uv@v7.2.0
         with:
           version: ${{ env.UV_VERSION }}
 
@@ -29,7 +29,7 @@ jobs:
         run: uv pip install --system requests
 
       - name: Restore link cache
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: .github/link-cache.csv
           key: link-cache-${{ github.run_id }}
--- a/.github/workflows/doc-pages.yaml
+++ b/.github/workflows/doc-pages.yaml
@@ -24,13 +24,13 @@ jobs:
           git fetch upstream --tags
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v7.1.6
+        uses: astral-sh/setup-uv@v7.2.0
         with:
           version: ${{ env.UV_VERSION }}
           enable-cache: true
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v6.0.0
+        uses: actions/setup-python@v6.1.0
         id: setup-python
         with:
           python-version: ${{ matrix.python-version }}
@@ -47,7 +47,7 @@ jobs:
           cp -r _build/* ../gh-pages/
       - name: Deploy documentation.
         if: ${{ github.event_name == 'push' }}
-        uses: JamesIves/github-pages-deploy-action@v4.7.3
+        uses: JamesIves/github-pages-deploy-action@v4.8.0
         with:
           branch: gh-pages
           force: true
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -44,7 +44,7 @@ jobs:
         uses: docker/setup-qemu-action@v3.7.0
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3.11.1
+        uses: docker/setup-buildx-action@v3.12.0
 
       - name: Login to DockerHub
         if: startsWith(github.ref, 'refs/tags/v')
--- a/.github/workflows/docs-test.yaml
+++ b/.github/workflows/docs-test.yaml
@@ -29,13 +29,13 @@ jobs:
           git fetch upstream --tags
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v7.1.6
+        uses: astral-sh/setup-uv@v7.2.0
         with:
           version: ${{ env.UV_VERSION }}
           enable-cache: true
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v6.0.0
+        uses: actions/setup-python@v6.1.0
         id: setup-python
         with:
           python-version: ${{ matrix.python-version }}
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -17,11 +17,11 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - name: Install uv
-        uses: astral-sh/setup-uv@v7.1.6
+        uses: astral-sh/setup-uv@v7.2.0
         with:
           version: ${{ env.UV_VERSION }}
 
-      - uses: actions/setup-python@v6.0.0
+      - uses: actions/setup-python@v6.1.0
         with:
           python-version: 3.13
       - name: Check pyproject.toml and uv.lock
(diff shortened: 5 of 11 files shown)


packages/linkml/src/linkml/generators/javagen.py
matentzn, line 1: I notice there are between 50 - 60 lines of almost identical Jinja code in javagen.py and the template file - is the one generated from the other or are these really redundant (sorry, still need to get my bearing, will be better once I review 2-3 PRs and get a better sense).

packages/linkml/src/linkml/generators/javagen.py
```diff
@@ -170,6 +242,7 @@ def cli(
         genmeta=genmeta,
         gen_classvars=classvars,
         gen_slots=slots,
+        true_enums=true_enums,
```
matentzn, line 245: I now the request is annoying from a noob, but I would like it if we could add a few minimal unit tests for at least:

- OOCodeGenerator.generate_enum_objects()
- OOCodeGenerator.create_documents() with true_enums=True
- JavaGenerator.serialize() with true_enums=True

packages/linkml/src/linkml/generators/oocodegen.py
```diff
@@ -134,8 +160,29 @@ def generate_enum_label(self, value: str) -> str:
 
             return safe_label
 
+    def generate_enum_objects(self, all_enums: dict[EnumDefinitionName, EnumDefinition]) -> dict[str, OOEnum]:
```
matentzn: Can we add a doctstring to explain this method?

matentzn: Also: why is this returning `dict[str, OOEnum]` ? If it is iterating through a `dict[EnumDefinitionName, EnumDefinition]` then it seems the return type should also be `dict[EnumDefinitionName, OOEnum]`?

packages/linkml/src/linkml/generators/javagen.py
gouttegd: Happy to do that now that there is an interest in the PR. :)

packages/linkml/src/linkml/generators/oocodegen.py
gouttegd: > Can we add a doctstring to explain this method?

Sure can.

> it seems the return type should also be dict[EnumDefinitionName, OOEnum]?

`EnumDefinitionName` is merely a glorified string, through several layers of ~~obfuscation~~ indirection:

```python
class extended_str(str, TypedNode):
    def concat(self, *items) -> "extended_str":
        rval = extended_str(str(self) + "".join([str(item) for item in items]))
        for item in items[::-1]:
            if isinstance(item, TypedNode):
                rval._s = item._s
                rval._len = item._len
                break
        return rval

class ElementName(extended_str):
    pass

class DefinitionName(ElementName):
    pass

class EnumDefinitionName(DefinitionName):
    pass
```

I am not frankly convinced of the interest of these types, even for static type checking…

But “in Rome, do as the Romans”, as the saying goes, so right, this method should return a `dict[EnumDefinitionName, OOEnum]`.

packages/linkml/src/linkml/generators/javagen.py
gouttegd: > is the one generated from the other or are these really redundant

Neither. One is not generated from the other and they are not redundant.

A bit of background may be necessary here.

The Java code generator can use two distinct templates:

* one that is embedded directly within the `javagen.py` module;
* one that is provided as the separate “resource” file `linkml/generators/javagen/java_record_template.jinja2`.

The generator uses the embedded template by default, unless the `--generate-records` option is used, in which case it uses the `java_record_template.jinja2` template (that’s the only thing this option does – switching the default template). (Of course if an explicit template is specified using the `--template-file` option, then the generator will use neither the embedded template nor the `java_record_template.jinja2` template.)

The reason for these two templates is that the generator wants to support two different ways of representing LinkML classes in Java (similarly to how there are different ways of representing LinkML classes in Python, e.g. using `dataclasses` or using Pydantic – except that for Python this is done via two distinct commands, `linkml generate python` and `linkml generate pydantic`, whereas in Java this is all handled by the same `linkml generate java` command):

* `Lombok`-annotated “standard” Java classes (that’s what the default template does), or
* Java “records” (a special kind of class intended to hold immutable data; that’s what the external template does).

But the way to represent enumerations is the same in both case, which is why both templates contain the same ~50 lines to generate the `enum` code. I admit this should be avoided, but this would require reorganising the way the generator works with its templates and I don’t think this would belong to this PR.

kevinschaper: I agree about postponing that refactor so that it's not in this PR - but if you have the bandwidth, it would be awesome to have an issue proposing the refactor

kevinschaper:
My first reaction when I saw this issue was that true enums was potentially new aspect of Java that I'm not caught up with, like that we were already generating java enums, but then a true enum is some newer language feature or something. If this PR is adding enum support to java like we have for other languages, because we totally lacked it before, I'm inclined to take the flag off and just make it the standard behavior.

gouttegd:
> I'm inclined to take the flag off and just make it the standard behavior.

It could certainly be the _default_ behaviour, because I’d expect most Java developers to want actual Java `enum` objects to represent LinkML enumerations.

But I can think of at least two cases where they _might_ want to keep representing them as `String` objects:

(A) When a slot has a range made of an union of several enums, as in (from the LinkML documentation):

```yaml
slots:
  vital_status:
    required: true
    range: Any
    any_of:
      - range: MissingValueEnum
      - range: VitalStatusEnum
```

This might be easier to work with in Java (or with any statically typed language for that matter) with strings, because in fact right now I cannot even imagine how that would work with real `enum` objects – you can’t just say that `vital_status` is “either a `MissingValueEnum` or a `VitalStatusEnum`”, Python might be fine with that but Java will definitely not be…

(B) With dynamic enums, depending on when you want to resolve the list of permissible values:

```
enums:
  NeuronTypeEnum:
    reachable_from:
      source_ontology: obo:cl
      source_nodes:
        - CL:0000540 ## neuron
      include_self: false
      relationship_types:
        - rdfs:subClassOf
```

If you’re happy with a _compile-time_ resolution (e.g. with `vskit expand`), then an actual `enum` object will be fine, but if you want a really “dynamic” enum, where the list of permissible values is only discovered (by querying an ontology source) _at runtime_, then almost by definition you can’t use a `enum` (which must be defined at compile-time) – you would have to use a standard `String` and then implement some custom logic to validate that the value of the `String` is one of the permissible values.

packages/linkml/src/linkml/generators/javagen.py
gouttegd: @kevinschaper See #3073. :)

kevinschaper:
I agree about both of the complicated enums!

matentzn:
@gouttegd I was told the sphinx thing was fixed - can you update your PR from main branch?

gouttegd:
> But I can think of at least two cases where [Java developers] might want to keep representing [enum] as String objects:

Make that three cases.

There is also the case where a String-typed slot in a base class is redefined as a Enum-typed slot in a derived class, as in:

```yaml
default_range: string

enums:
  MyEnum:
    permissible_values:
      Alice:
      Bob:
      Charlie:

slots:
  type:

classes:
  BaseClass:
    slots:
      - type

  DerivedClass:
    is_a: BaseClass
    slot_usage:
      type:
        range: MyEnum
```

Again, this might be fine in Python, but definitely not in Java, and in fact in many other languages. I’d file that in the category “Python giving developers some very bad habits”…

kevinschaper:
I think the third case would just be handled in schemaview when you call induced_slot, but it might actually throw an error about being non-monotonic inheritance.

PR #3078, opened 2026-01-16
Cache ShEx JSON-LD context to avoid 429 errors from w3.org

kevinschaper:
## Summary
- Embeds the ShEx JSON-LD context directly in `shexgen.py` to avoid remote fetches from w3.org
- Fixes HTTP 429 "Too Many Requests" errors when generating ShEx in RDF format
- Removes `@pytest.mark.network` from `test_meta` since it no longer needs network access

## Problem
The `ShExGenerator`'s RDF output format (`-f rdf`) parses JSON-LD using rdflib, which triggers a fetch of `http://www.w3.org/ns/shex.jsonld`. This remote service rate-limits requests, causing test failures:

```
FAILED tests/linkml/test_scripts/test_gen_shex.py::test_meta[arguments2-metashex.ttl] - AssertionError: assert 1 == 0
 +  where 1 = <Result <HTTPError 429: 'Too Many Requests'>>.exit_code
```

## Solution
The ShEx JSON-LD context is a stable W3C specification that rarely changes. This PR:
1. Adds a `SHEX_CONTEXT` constant containing the cached context
2. Modifies `end_schema()` to inline the context before JSON-LD parsing
3. Removes the network test marker since the dependency is eliminated


matentzn:
@kevinschaper which issue does that relate to?

kevinschaper:
> which issue does that relate to? 
@matentzn No issue for this, just trying to deal with a common network failure on other PRs - worth making a tracking issue for it?

matentzn:
Generally it would help me a little tracking everything if we always had issues that include a strong motivation :)

PR #3090, opened 2026-01-22
Fix issue #3011 test and add documentation for SHACL naming modes

### LINKED ISSUE #3011: Unintentional SHACL Class merge
---
name: Bug report
about: Shacl Generator
title: Distinct SHACL Classes Bug and SHACL-shape native names Feature Request
labels: bug 
assignees: ''

**Describe the bug**
Currently distinct linkML classes get merged into a single SHACL shape if they have nondistinct class_uri's which are their target class.

In SHACL it is good consensus that your SHACL shapes and target class URI's must not match. So for example we would be able to have two shapes against which we want to check data against. One being the simple_shape and one being the complex_shape, which either extends on the simple shape or even being different in their validation rules. The same is basically possible in LinkML.

We Implemented such an approach here:
https://github.com/nfdi-de/dcat-ap-plus/blob/main/src/dcat_ap_plus/schema/dcat_ap_plus.yaml
with e.g. the LinkML class `AgenticEntity` and `Device` both using `class_uri: prov:Agent` and the LinkML class `Entity`,`AnalysisDataset`, and `EvaluatedEntity` using `class_uri: prov:Entity`. 
This causes issues after the shacl generator, which can be seen here:
https://github.com/nfdi-de/dcat-ap-plus/blob/main/project/shacl/dcat_ap_plus.shacl.ttl
Here you can easily spot, that the description of both `AgenticEntity` and `Device` are merged into the `prov:Agent` Nodeshape

```
prov:Agent a sh:NodeShape ;
    sh:description "A material instrument that is designed to perform a function primarily by means of its mechanical or electrical nature.",
                   "An entity that is somehow responsible for an Activity to take place." ;
```
and similar with `prov:Entity`
```
prov:Entity a sh:NodeShape ;
    sh:description "A physical, digital, conceptual, or other kind of thing with some fixed aspects; entities may be real or imaginary.",
                   "A piece of information that is attributed to an Entity, Activity or AgenticEntity.",
                   "An Entity that is being evaluated in a DataGeneratingActivity.",
                   "Information that was evaluated within a DataAnalysis." ;
```

v1.9.5

Schema: `https://github.com/nfdi-de/dcat-ap-plus/blob/main/src/dcat_ap_plus/schema/dcat_ap_plus.yaml`
Output file of the generator: `https://github.com/nfdi-de/dcat-ap-plus/blob/main/project/shacl/dcat_ap_plus.shacl.ttl`

kevinschaper:
## Summary

Fixes #3011

The test for issue #3011 was marked `xfail` due to a bug in the test itself, not because the functionality was broken. PR #3020 already fixed the underlying issue by adding the `use_class_uri_names` parameter. This PR fixes the test and adds comprehensive documentation.

## Problem

The test schema was missing the `linkml:` prefix, causing a `ValueError: Unknown CURIE prefix: linkml` error:

```yaml
# Before (broken)
prefixes:
  prov: http://www.w3.org/ns/prov#
# linkml: prefix was missing!

imports:
  - linkml:types  # This failed to resolve
```

Additionally, the assertion logic had a bug where `"Entity"` matched all three class names:

```python
# Before (broken) - "Entity" is a substring of all names
for cname in ["Entity", "EvaluatedEntity", "ThirdEntity"]:
    matching = [s for s in shapes if cname in str(s)]  # Matches all 3!
```

## Solution

**Fixed the test schema:**
```yaml
prefixes:
  prov: http://www.w3.org/ns/prov#
  linkml: https://w3id.org/linkml/  # Added
```

**Fixed the assertion logic:**
```python
# After (correct) - exact suffix match
for cname in ["Entity", "EvaluatedEntity", "ThirdEntity"]:
    matching = [s for s in shapes if str(s).endswith(f":{cname}")]
```

## New Tests Added

Added edge case tests to ensure the naming mode works correctly with:

1. **`test_shacl_target_class_correct_with_native_names`** - Verifies `sh:targetClass` is correctly set to the `class_uri` even when using native names
2. **`test_shacl_inheritance_with_shared_class_uri`** - Tests `is_a` inheritance with shared `class_uri`
3. **`test_shacl_mixin_with_shared_class_uri`** - Tests mixins with shared `class_uri`

## Documentation Added

Added comprehensive class docstring to `ShaclGenerator` explaining the naming modes:

```python
@dataclass
class ShaclGenerator(Generator):
    """
    Generate SHACL (Shapes Constraint Language) shapes from a LinkML schema.

    Shape Naming Modes
    ------------------
    The generator supports two naming modes controlled by `use_class_uri_names`:

    **Default mode (use_class_uri_names=True):**
        Shapes are named using the class_uri. If multiple LinkML classes share
        the same class_uri, their properties are merged into a single shape.

    **Native names mode (use_class_uri_names=False):**
        Shapes are named using the native LinkML class name from the schema.
        Each LinkML class produces a distinct shape, even if they share a class_uri.
        The sh:targetClass still correctly points to the class_uri.

    Use native names mode when:
        - Multiple LinkML classes intentionally map to the same external ontology class
        - You need distinct validation shapes per LinkML class
        - You want to avoid unintentional shape merging (see issue #3011)
    """
```

### DIFF
--- a/packages/linkml/src/linkml/generators/jsonldcontextgen.py
+++ b/packages/linkml/src/linkml/generators/jsonldcontextgen.py
@@ -19,6 +19,7 @@ from linkml.utils.generator import Generator, shared_arguments
 from linkml_runtime.linkml_model.meta import ClassDefinition, SlotDefinition
 from linkml_runtime.linkml_model.types import SHEX
 from linkml_runtime.utils.formatutils import camelcase, underscore
+from linkml_runtime.utils.schemaview import SchemaView
 
 URI_RANGES = (SHEX.nonliteral, SHEX.bnode, SHEX.iri)
 
@@ -53,6 +54,10 @@ class ContextGenerator(Generator):
     prefixes: bool | None = True
     flatprefixes: bool | None = False
     fix_multivalue_containers: bool | None = False
+    exclude_imports: bool = False
+    """If True, elements from imported schemas won't be included in the generated context"""
+    _local_classes: set | None = field(default=None, repr=False)
+    _local_slots: set | None = field(default=None, repr=False)
 
     # Framing (opt-in via CLI flag)
     emit_frame: bool = False
@@ -64,6 +69,16 @@ class ContextGenerator(Generator):
         super().__post_init__()
         if self.namespaces is None:
             raise TypeError("Schema text must be supplied to context generator.  Preparsed schema will not work")
+        if self.exclude_imports:
+            if self.schemaview:
+                sv = self.schemaview
+            else:
+                source = self.schema.source_file or self.schema
+                if isinstance(source, str) and self.base_dir and not Path(source).is_absolute():
+                    source = str(Path(self.base_dir) / source)
+                sv = SchemaView(source, importmap=self.importmap, base_dir=self.base_dir)
+            self._local_classes = set(sv.all_classes(imports=False).keys())
+            self._local_slots = set(sv.all_slots(imports=False).keys())
 
     def visit_schema(self, base: str | Namespace | None = None, output: str | None = None, **_):
         # Add any explicitly declared prefixes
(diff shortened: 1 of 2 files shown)


amc-corey-cox:
Rebased on main and fixed the RST indentation in the ShaclGenerator docstring that was causing the docs build to fail (warnings-as-errors). Also tightened the examples slightly and fixed an `:issue:` role reference (not configured in our Sphinx setup).

PR #3096, opened 2026-01-23
Fix Excel generator handling of empty class lists

### LINKED ISSUE #3097: Excel generator fails on schemas with no concrete classes
The Excel generator crashes when processing schemas that have no concrete (non-abstract, non-mixin) classes:

```
IndexError: At least one sheet must be visible
```

## Steps to Reproduce

1. Create a schema with only abstract classes or mixins:

```yaml
id: https://example.org/abstract-only
name: abstract_only
prefixes:

classes:
  AbstractBase:
    abstract: true
    attributes:
      id:
        identifier: true

  SomeMixin:
    mixin: true
    attributes:
      name:
        range: string
```

2. Run: `gen-excel abstract_only.yaml` (with the current linkml-model version!)

3. Observe crash:
```
IndexError: At least one sheet must be visible
```

## Expected Behavior

The generator should handle this gracefully - either skip Excel generation with a warning, or produce an empty but valid workbook.

## Root Cause

In `excelgen.py`, the `create_workbook_and_worksheets` method:
1. Creates a workbook with a default sheet
2. Removes the default sheet
3. Iterates over classes to create sheets
4. If no classes, the workbook has no sheets
5. `workbook.save()` fails because openpyxl requires at least one visible sheet

## Affected Use Cases

- Processing the LinkML metamodel (all classes are abstract or mixins)
- Schemas that define only base classes meant to be extended
- Library schemas with reusable patterns but no concrete classes

matentzn:
Closes #3097

Return early with a warning when there are no concrete classes to process, instead of creating an empty workbook that fails to save.

## The Problem

The Excel generator failed when processing schemas with no concrete (non-abstract, non-mixin) classes:

```
IndexError: At least one sheet must be visible
```

This occurred because:
1. `create_workbook_and_worksheets` creates a workbook
2. Removes the default sheet (`workbook.remove(workbook.active)`)
3. Loops over classes to create sheets - but if `classes` is empty, no sheets are created
4. `workbook.save()` fails because openpyxl requires at least one visible sheet

## The Fix

```python
def create_workbook_and_worksheets(self, output_path: Path, classes: list[str]) -> None:
    if not classes:
        self.logger.warning(
            f"No classes to process for Excel generation. "
            f"Skipping workbook creation for {output_path}"
        )
        return
    # ... rest of method
```

## Why this is not a bandaid

The Excel generator creates **data entry templates** - each sheet represents a class where users enter instance data. The question is: what should happen when there are no concrete classes?

**Alternative approaches considered:**

| Approach | Why it's wrong |
|----------|----------------|
| Create placeholder sheet | Misleading - suggests there's data to enter when there isn't |
| Raise an error | Too harsh - schemas with only abstract/mixin classes are valid |
| Create empty file | openpyxl won't allow it, and an empty Excel file is useless |

**Why early return is correct:**

1. **Nothing to generate** - Abstract classes and mixins can't be instantiated, so there's no data entry template to create

2. **Valid schema state** - Schemas with no concrete classes are valid (e.g., the LinkML metamodel itself). This shouldn't cause failures.

3. **Informative warning** - Users are told why no file was created, not left with a cryptic error

4. **Consistent with other generators** - When there's nothing meaningful to output, generators should handle it gracefully

The fix addresses the root cause: the generator assumed there would always be classes to process, which isn't true for all valid schemas.

## Test Plan

- [x] `uv run pytest tests/linkml/test_generators/test_excelgen.py` passes (3 tests)

### DIFF
--- a/packages/linkml/src/linkml/generators/shacl/shacl_ifabsent_processor.py
+++ b/packages/linkml/src/linkml/generators/shacl/shacl_ifabsent_processor.py
@@ -13,6 +13,11 @@ class ShaclIfAbsentProcessor(IfAbsentProcessor):
                 return True, URIRef(class_uri)
             return True, ""
 
+        # TODO: The below conditional is a stop gap for a proper solution, see @sneakers-the-rat proposal in https://github.com/linkml/linkml/issues/2522
+        if default_value == "default_range":
+            default_range = self.schema_view.schema.default_range or "string"
+            return True, Literal(default_range, datatype=ShaclDataType.STRING.uri_ref)
+
         return False, None
 
     def map_enum_default_value(
--- a/tests/linkml/test_compliance/test_ifabsent_compliance.py
+++ b/tests/linkml/test_compliance/test_ifabsent_compliance.py
@@ -39,6 +39,7 @@ FUZZ_STR = "a b_c!@#$%^&*_+{}|:<>?[]()'\""
         ("boolT", "boolean", "true", "no_value", None, True, True, True, []),
         ("boolF", "boolean", "false", "no_value", None, False, True, True, []),
         ("class_curie", "uriorcurie", "class_curie", "no_value", None, "ex:C", True, True, []),
+        ("default_range", "string", "default_range", "no_value", None, "string", True, True, []),
         ("D", CLASS_D, "string(p1)", "no_value", None, "p1", False, True, []),
         # Skip Python, Pydantic and Shacl frameworks because this incompatibility is not possible with the processor
         (
--- a/tests/linkml/test_generators/shacl/test_shacl_ifabsent_processor.py
+++ b/tests/linkml/test_generators/shacl/test_shacl_ifabsent_processor.py
@@ -482,3 +482,35 @@ def test_process_sparql_path_ifabsent_attribute():
             schema_view.all_slots()[SlotDefinitionName("unimplementedSparqlPath")],
             schema_view.all_classes()[ClassDefinitionName("Student")],
         )
+
+
+def test_process_default_range():
+    """Test that default_range resolves to the schema's default range value."""
+    schema_with_default_range = """
+id: ifabsent_tests
+name: ifabsent_tests
+default_range: string
+
+prefixes:
+  ex: https://example.org/
+default_prefix: ex
+
+classes:
+  Student:
+    attributes:
+"""
+    schema_view = SchemaView(
+        schema_with_default_range
+        + """
+    - name: rangeSlot
+      range: string
+      ifabsent: default_range
+    """
+    )
+
+    processor = ShaclIfAbsentProcessor(schema_view)
+
+    assert processor.process_slot(
+        schema_view.all_slots()[SlotDefinitionName("rangeSlot")],
+        schema_view.all_classes()[ClassDefinitionName("Student")],
+    ) == Literal("string", datatype=ShaclDataType.STRING.uri_ref)
--- a/tests/linkml/test_metamodel_compat/test_metamodel_projectgen.py
+++ b/tests/linkml/test_metamodel_compat/test_metamodel_projectgen.py
@@ -36,7 +36,7 @@ def test_metamodel_python_compiles(metamodel_path, tmp_path):
     config = ProjectConfiguration()
     config.directory = tmp_path
     config.generator_args["owl"] = {"metaclasses": False, "type_objects": False}
-    config.excludes = ["shacl", "excel"]
+    config.excludes = ["excel"]
     gen = ProjectGenerator()
     gen.generate(metamodel_path, config)
     python_file = tmp_path / "meta.py"


packages/linkml/src/linkml/generators/excelgen.py
```diff
@@ -48,6 +48,12 @@ def create_workbook_and_worksheets(self, output_path: Path, classes: list[str])
         :param output_path: The path where the workbook should be created.
         :param classes: List of class names for which worksheets should be created.
         """
+        if not classes:
+            self.logger.warning(
+                f"No classes to process for Excel generation. Skipping workbook creation for {output_path}"
+            )
+            return
```
matentzn: @sierra-moxon might prepare an error - lets see if that fits with the current way the test suite works @sujaypatil96

sujaypatil96: I agree with @sierra-moxon. It's definitely more useful for debugging if we include a specific error message.

sujaypatil96: I would try to improve this logic by further checking the following within the `if not classes:`

- [x] if only abstract (`abstract: true`) classes are initialized, then log a specific error message that indicates that
- [x] if only mixin (`mixin: true`) classes are initialized, then log a specific error message that indicates that
  - [x] Note that there is also a `--include_mixins` optional argument in the generator logic that can override this assumption and create sheets in the Excel workbook even if it is a mixin class
- [x] if combination of abstract and mixin classes are initialized, then log a specific error message that indicates that
- [x] it is also possible to have syntax that looks like below where you simply leave the top level `classes` slot empty:

```yaml
id: https://example.org/abstract-only
name: abstract_only
prefixes:

classes:
```
So that would have it's own error message too

Besides the above 4 cases it is also possible for us to create "Container" classes (see [docs](https://linkml.io/linkml/howtos/ontologies-as-values.html#adding-a-container)) in a LinkML schema, and we need to decide how to handle the creation of those in the Excel output, but that's a separate issue of it's own.

sujaypatil96 (changes requested):
I've left a [comment](https://github.com/linkml/linkml/pull/3096/changes#r2739225667) requesting for more specific error handling/logging steps to be added to the generator to report a more specific error in the case when the top level `classes:` slot has not been "populated" appropriately for Excel generator output.

packages/linkml/src/linkml/generators/excelgen.py
matentzn: I moved the error handling to serialize() where filtering logic lives, providing
specific ValueError exceptions when no classes can be processed:

- No classes defined in schema
- Only abstract classes
- Only mixin classes (suggests --include-mixins only when flag not set)
- Abstract + mixin classes only (suggests --include-mixins only when flag not set)

Rename create_workbook_and_worksheets to _create_workbook_and_worksheets to make it private - this way we can be sure that "classes" is working as intended.

packages/linkml/src/linkml/generators/excelgen.py
```diff
@@ -134,18 +134,43 @@ def column_enum_validation(
 
     def serialize(self, **kwargs) -> str:
         sv = self.schemaview
+        all_classes = sv.all_classes(imports=self.mergeimports)
+
+        if not all_classes:
+            raise ValueError(
+                "No classes defined in the schema. "
+                "Excel generation requires at least one non-abstract class to create worksheets."
+            )
```
sujaypatil96, line 143: Note to self: handling the case where there are no classes defined under the top level `classes:` slot.

packages/linkml/src/linkml/generators/excelgen.py
```diff
@@ -134,18 +134,43 @@ def column_enum_validation(
...
-                if not cls.mixin and not cls.abstract
+                cls_name for cls_name, cls in all_classes.items() if not cls.mixin and not cls.abstract
             ]
 
+        if not classes_to_process:
+            if abstract_classes and not mixin_classes:
+                raise ValueError(
+                    f"Schema contains only abstract classes: {', '.join(sorted(abstract_classes))}. "
+                    "Excel generation requires at least one non-abstract class to create worksheets."
+                )
```
sujaypatil96, line 160: Note to self: handling the case when only abstract classes are initialized

packages/linkml/src/linkml/generators/excelgen.py
```diff
@@ -134,18 +134,43 @@ def column_enum_validation(
...
+            if mixin_classes and not self.include_mixins:
+                if not abstract_classes:
+                    raise ValueError(
+                        f"Schema contains only mixin classes: {', '.join(sorted(mixin_classes))}. "
+                        "Use --include-mixins flag to generate worksheets for mixin classes."
+                    )
```
sujaypatil96, line 166: Note to self: handling the case when mixin classes are initialized but `--include-mixins` is turned off

packages/linkml/src/linkml/generators/excelgen.py
```diff
@@ -134,18 +134,43 @@ def column_enum_validation(
...
+                else:
+                    raise ValueError(
+                        f"Schema contains only abstract classes ({', '.join(sorted(abstract_classes))}) "
+                        f"and mixin classes ({', '.join(sorted(mixin_classes))}). "
+                        "Use --include-mixins flag to generate worksheets for mixin classes."
+                    )
```
sujaypatil96, line 172: Note to self: handling the case when abstract classes and mixin classes are initialized

PR #3100, opened 2026-01-27
Add SKIP_KROKI_IO env var to skip Kroki in CI

### LINKED ISSUE #3099: Flaky CI failures due to Kroki API 403 errors
Thank you @gouttegd for the diagnostics.

The `test_datadict_personinfo` test intermittently fails in CI when Kroki returns HTTP 403:

```
WARNING  root:markdowndatadictgen.py:278 Failed to render 'class_event_local' with Kroki: HTTP Error 403: Forbidden
```

This causes snapshot mismatches since the output contains mermaid code blocks instead of SVG diagrams.

We need a way to skip Kroki calls in CI while still allowing local development and production builds to use it.

matentzn:
Closes #3099 

Adds `SKIP_KROKI_IO` environment variable to skip Kroki API calls in CI.

When set, the diagram renderer returns mermaid code blocks without attempting any network calls. This avoids flaky test failures from Kroki 403 errors.

## Changes:

- Add env var check in `DiagramRenderer.render()`
- Set `SKIP_KROKI_IO=1` in CI workflow
- Add test for the skip behavior

### DIFF
--- a/packages/linkml/src/linkml/generators/markdowndatadictgen.py
+++ b/packages/linkml/src/linkml/generators/markdowndatadictgen.py
@@ -214,6 +214,11 @@ class DiagramRenderer:
         if not self.kroki_server:
             return f"```{diagram_type}\n{clean_source}\n```"
 
+        # Allow skipping Kroki in test environments via environment variable
+        if os.environ.get("SKIP_KROKI_IO"):
+            logging.debug("Skipping Kroki rendering (SKIP_KROKI_IO is set)")
+            return f"```{diagram_type}\n{clean_source}\n```"
+
         try:
             svg_content = None
             cache = SvgCache(self.diagram_dir)
--- a/tests/linkml/test_generators/test_markdowndatadictgen.py
+++ b/tests/linkml/test_generators/test_markdowndatadictgen.py
@@ -37,6 +37,18 @@ def test_diagram_renderer_kroki_failure():
     assert "A --> B" in result
 
 
+def test_diagram_renderer_skip_kroki_env(monkeypatch):
+    """Test DiagramRenderer skips Kroki when SKIP_KROKI_IO environment variable is set."""
+    monkeypatch.setenv("SKIP_KROKI_IO", "1")
+    renderer = DiagramRenderer(kroki_server="https://kroki.io")
+    result = renderer.render("classDiagram\nA --> B", diagram_name="test")
+    assert result.startswith("```mermaid")
+    assert "classDiagram" in result
+    assert "A --> B" in result
+    assert "<svg" not in result
+    assert "</svg>" not in result
+
+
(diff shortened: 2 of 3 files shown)


tests/linkml/test_generators/test_markdowndatadictgen.py
```diff
@@ -37,6 +37,18 @@ def test_diagram_renderer_kroki_failure():
     assert "A --> B" in result
 
 
+def test_diagram_renderer_skip_kroki_env(monkeypatch):
+    """Test DiagramRenderer skips Kroki when SKIP_KROKI_IO environment variable is set."""
+    monkeypatch.setenv("SKIP_KROKI_IO", "1")
```
tfliss, line 42: When I change this line to set a nonsense variable, the test is still passing.

matentzn: Thats true.. if `SKIP_KROKI_IO` is false, then this test may or may not pass - this is really ONLY makeing sure that if its true, it always passes.

packages/linkml/src/linkml/generators/markdowndatadictgen.py
```diff
@@ -214,6 +214,11 @@ def render(self, diagram_source: str, diagram_type: str = "mermaid", diagram_nam
         if not self.kroki_server:
             return f"```{diagram_type}\n{clean_source}\n```"
 
+        # Allow skipping Kroki in test environments via environment variable
+        if os.environ.get("SKIP_KROKI_IO"):
```
tfliss, line 218: This will also skip if SKIP_KROKI_IO is set to 0

Something like os.environ.get("SKIP_KROKI_IO", "0").lower() in ["1", "true", "t"]
would get more 'truthy' values, or just comparing to 1 might avoid confusion if someone does set the value to 0.

tests/linkml/test_generators/test_markdowndatadictgen.py
```diff
@@ -37,6 +37,18 @@ def test_diagram_renderer_kroki_failure():
...
+    renderer = DiagramRenderer(kroki_server="https://kroki.io")
+    result = renderer.render("classDiagram\nA --> B", diagram_name="test")
+    assert result.startswith("```mermaid")
+    assert "classDiagram" in result
+    assert "A --> B" in result
+    assert "<svg" not in result
+    assert "</svg>" not in result
+
+
 def test_diagram_renderer_large_diagram_post(tmp_path):
```
tfliss, line 52: This test seems to be failing on the SKIP_KROKI_IO (because the post doesn't happen).
But I'm wondering if the other KROKI CI tests could be adapted to use a similar approach, mocking the return from the url open

matentzn:
Superceded by #3108

PR #3101, opened 2026-01-27
Add Kroki server ENV to CI

### LINKED ISSUE #3099: Flaky CI failures due to Kroki API 403 errors

matentzn:
Fixes #3099 

Alternative to #3100 - disadvantage, essentially skipping diagram rendering related tests on windows images.

Starts a local Kroki Docker container in CI on Ubuntu for more reliable diagram rendering. Updates the markdown data dictionary generator tests to use the KROKI_SERVER environment variable, skipping the test on Windows where Kroki is unavailable.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -62,6 +62,20 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Start Kroki server
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          docker run -d --name kroki -p 8000:8000 yuzutech/kroki
+          # Wait for Kroki to be ready
+          for i in {1..30}; do
+            if curl -s http://localhost:8000/health > /dev/null 2>&1; then
+              echo "Kroki is ready"
+              break
+            fi
+            echo "Waiting for Kroki... ($i/30)"
+            sleep 1
+          done
+
       - name: Ensure tags if not run from main repo
         if: github.repository != 'linkml/linkml'
         run: |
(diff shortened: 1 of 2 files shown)


matentzn:
I am now against this - the local kroki server produces slightly different results than kroki.io. I dont think CI should depend this way on the behaviour of an externally hosted service if I am honest. @dalito do you have an opinion on this?

dalito:
With "agianst this" you mean to test/validate an external API in the LinkML tests? Then I fully agree. Testing with plantuml directly would have much lower risk to break.

matentzn:
> With "agianst this" you mean to test/validate an external API in the LinkML tests? Then I fully agree. Testing with plantuml directly would have much lower risk to break.

No, sorry for not being clearer. This PR here suggests to delegate the KROKIO calls not to a mock, but to a self hosted KROKI  server (self hosted by the github action runner and deployed on localhost). The differences in the SVG generation are too large.

@tfliss suggested I should look into mocking instead of SKIPPING, so I will look at that as a third option. But this one hear is almost certainly not the way.

matentzn:
This is not working, so closing. The issue was patched here, but not solved. https://github.com/linkml/linkml/pull/3108

PR #3105, opened 2026-01-29
Clarify why PRs should be draft

matentzn:
Update contributing guidance to explain that PRs should remain in DRAFT until ready to reduce notification noise. The added note clarifies that GitHub sends notifications on PR creation and each automated review, which can overwhelm maintainers subscribed to “All Activity.” This makes the rationale for using draft PRs explicit in docs/maintainers/contributing.md.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -86,13 +86,13 @@ jobs:
         uses: ts-graphviz/setup-graphviz@v2.0.2
       - name: Test linkml package
         run: |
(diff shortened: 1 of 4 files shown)


matentzn:
Since you cant force people to create draft PRs, we need to think maybe of a better way to reduce chatter. Maybe a pull request template? Or maybe we can somehow enforce the codecoverage bot to report "as a check" rather than by writing a comment?

dalito (approved):

dalito:
I think adding a note is enough. Thanks!

Issue #3111, opened 2026-01-30
Add compliance tests for JSONSchema and Pydantic validators in their support of `any_of` range constraints combined with regular `range` constraints.


sierra-moxon:
Looking at differences between Pydantic and JSONSchema generator support for `any_of` range constraints and `range` range constraints on the same slot.

```
id: http://example.org/test                                                                                                                                                          
  name: test                                                                                                                                                                           
  prefixes:                                                                                                                                                                            
    linkml: https://w3id.org/linkml/                                                                                                                                                   
  imports:                                                                                                                                                                             
    - linkml:types                                                                                                                                                                     
                                                                                                                                                                                       
  slots:                                                                                                                                                                               
    my_slot:                                                                                                                                                                           
      range: string                                                                                                                                                                    
      any_of:                                                                                                                                                                          
        - range: integer                                                                                                                                                               
        - range: Letters                                                                                                                                                               
                                                                                                                                                                                       
  enums:                                                                                                                                                                               
    Letters:                                                                                                                                                                           
      permissible_values:                                                                                                                                                              
        ALPHA:                                                                                                                                                                         
        BRAVO:                                                                                                                                                                         
        CHARLIE:                                                                                                                                                                       
                                                                                                                                                                                       
  classes:                                                                                                                                                                             
    MyClass:                                                                                                                                                                           
      tree_root: true                                                                                                                                                                  
      slots:                                                                                                                                                                           
        - my_slot    
```


**JSONSchema:** Creates an intersection (AND) values must match both the direct range type AND one of the any_of options.  I think we could generate JSONSchemas like this: 

```
                                                                                                                                                                                       
  {                                                                                                                                                                                    
    "type": "string",        // from range                                                                                                                                             
    "anyOf": [               // from any_of                                                                                                                                            
      {"type": "integer"},   // can never match (blocked by type: string)                                                                                                              
      {"$ref": "#/$defs/Letters"}                                                                                                                                                      
    ]                                                                                                                                                                                  
  } 
```
  - [ ] Add a linter rule(s)/warning to notify of unsatisfiable schema.
  - [ ] Add compliance test to document this behavior

**Pydantic**: Creates a union (OR) from any_of, only the direct range is dropped.
  - [ ] Add a linter rule(s)/warning that range will be dropped if both range and any_of exist on a slot. 
  - [ ] Add compliance test to document this behavior

sierra-moxon:
Noting the same action items for `domain` and `any_of` domain constraints.

[...]

PR #3114, opened 2026-01-30
Test jsonschema anyof range conflict

sierra-moxon:
fixes the first half of https://github.com/linkml/linkml/issues/3111 
- adds compliance tests for JSONSchema generator for the use case of `any_of` range constraints and `range` constraints on the same slot, including test cases for inlined true/false, identifier range = integer vs. string/uri, and identifier true/false.
- adds linting rule to warn about generation of unsatisfiable JSONSchema as a result of the JSONSchema generator (AND) handling of `any_of` range and `range` range elements.

### DIFF
--- a/packages/linkml/src/linkml/linter/rules.py
+++ b/packages/linkml/src/linkml/linter/rules.py
@@ -515,7 +515,7 @@ class RangeAnyOfIncompatibleRule(LinterRule):
                 return "object"
             else:
                 # Not inlined, reference type is the identifier's type
-                return self.get_json_schema_type(id_slot.range, sv) if id_slot.range else "string"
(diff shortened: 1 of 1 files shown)


packages/linkml/src/linkml/linter/config/datamodel/config.py
```diff
@@ -1,5 +1,5 @@
 # Auto generated from config.yaml by pythongen.py version: 0.0.1
-# Generation date: 2025-09-30T13:24:19
+# Generation date: 2026-01-29T17:24:44
```
sierra-moxon, line 2: we have to regenerate because we added a new linting rule.

PR #3126, opened 2026-02-03
Revise PR guidelines for clarity and requirements

matentzn:
Fixes #3125

Updated PR guidelines to use 'MUST' for clarity and added requirements for linking PRs to issues and assigning reviewers.

PR #3131, opened 2026-02-03
Feature: Improve YARRRML generator object semantics & inline handling

lapkinvladimir:
This PR improves object slot handling and inline semantics in the YARRRML generator

Previously, `inlined: true` was not handled using proper joins, and multi-owner cases were not detected. This change uses the documented YARRRML join pattern ([spec section 9.7](https://rml.io/yarrrml/spec/)).

### What it does

- Emits object links (`inlined: false`) explicitly as IRIs (`type: iri`)
- Supports multivalued object links as lists of IRIs
- Implements join-based handling for `inlined: true` slots using YARRRML `mapping + condition`
- Requires inline classes to define an identifier (or key), as joins need a stable linking field
- Rejects multi-owner inline usage: a YARRRML mapping defines a single `sources`/iterator context, and different owners imply different data paths that cannot be represented within one mapping

### Tests

- Object links emitted as IRIs
- Multivalued object link generation
- Join-based inline behavior
- RDF materialization checks
- Explicit error for unsupported inline configurations

### DIFF
--- a/packages/linkml/src/linkml/generators/pydanticgen/pydanticgen.py
+++ b/packages/linkml/src/linkml/generators/pydanticgen/pydanticgen.py
@@ -776,7 +776,12 @@ class PydanticGenerator(OOCodeGenerator, LifecycleMixin):
             pyrange = f'Literal["{slot_def.equals_string}"]'
         elif slot_def.equals_string_in:
             pyrange = "Literal[" + ", ".join([f'"{a_string}"' for a_string in slot_def.equals_string_in]) + "]"
-        elif self.expand_subproperty_of and slot_def.subproperty_of:
+        elif (
+            self.expand_subproperty_of
+            and slot_def.subproperty_of
+            and slot_range not in sv.all_classes()
+            and slot_range not in sv.all_enums()
+        ):
             pyrange = self._generate_subproperty_constraint(slot_def)
         elif slot_range in sv.all_classes():
             pyrange = self.get_class_slot_range(
--- a/packages/linkml_runtime/src/linkml_runtime/processing/referencevalidator.py
+++ b/packages/linkml_runtime/src/linkml_runtime/processing/referencevalidator.py
@@ -737,14 +737,15 @@ class ReferenceValidator:
         # Induced slot
         for slot in target.attributes.values():
             # TODO: required slots MUST be present UNLESS this is a CompactDict
-            if slot.required and slot.alias not in input_object and not (slot.identifier or slot.key):
+            slot_key = slot.alias or slot.name
+            if slot.required and slot_key not in input_object and not (slot.identifier or slot.key):
                 report.add_problem(
                     ConstraintType.RequiredConstraint,
                     slot.name,
                     input_object,
                     predicate=target.name,
                 )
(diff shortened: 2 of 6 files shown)


docs/generators/yarrrml.rst
```diff
@@ -71,6 +71,15 @@ Besides JSON, CSV/TSV is supported. The key differences:
          value: $(employer)
```
Ostrzyciel, line 71: Please rebase the branch on top of main.

tests/linkml/test_generators/test_yarrrmlgen.py
```diff
@@ -186,13 +186,31 @@ def _validate_with_shacl(schema_path: Path, g: rdflib.Graph):
         inlined: true
```
Ostrzyciel, line 181: Please add negative tests, making sure your error handling works correctly – testing failure modes of a program is also important! :) 

See the Codecov report: https://app.codecov.io/gh/linkml/linkml/pull/3131?src=pr&el=tree

matentzn:
@lapkinvladimir amazing contribution, thanks! 

We are contemplating some changes to the governance processes; they are not enacted yet so you dont need to worry. But I would like to get your opinion on that: https://github.com/linkml/linkml/pull/3126/changes

The main reason for the issue first policy is to ensure that the community has a chance to respond/interact before actual work is done, and it makes it also a little more transparent to have issues with open discussions about a problem and keeping the PR focused on the implementation. Can you perhap review that suggestion and tell me your opinion? It will NOT apply to this PR here - but to future ones. Let me know if you think this is busywork.

lapkinvladimir:
@matentzn thanks for sharing the changes to the governance processes. The idea makes sense, especially when dealing with bigger or architectural changes where early discussion is really important. But in smaller changes, where the scope is already well understood, requiring the issue first can be a bit heavy, as the PR can already be the discussion space.

So maybe some flexibility based on the scope of the change could be a good idea.

lapkinvladimir:
Hi @sierra-moxon @kevinschaper, could we please get a review on this when you have a moment? 🙏
All checks are passing and there is already one approval

packages/linkml/src/linkml/generators/yarrrmlgen.py
```diff
@@ -52,11 +52,48 @@ def as_dict(self) -> dict[str, Any]:
...
+                if inlined is None:
+                    inlined = False
+
+                if inlined:
+                    inline_targets.add(range_cls.name)
+                    alias = decl.alias if decl and decl.alias else s.alias
+                    var = alias or s.name
+                    inline_owners.setdefault(range_cls.name, []).append((owner.name, var))
+                else:
+                    non_inline_targets.add(range_cls.name)
```
kevinschaper: Does anything happen with non_inline_targets? It seems to be populated but not referenced anywhere

lapkinvladimir:
@kevinschaper non_inline_targets (and inline_targets) were leftovers from an earlier iteration and are no longer used. I removed them in a cleanup commit

lapkinvladimir:
Hi @sierra-moxon @kevinschaper, could we please get a review on this when you have a moment? 🙏

matentzn:
I moved this PR up to top place now in the review priority order

Issue #2581, opened 2025-03-24
Make it possible to configure inlined multivalued strings syntax


matentzn:
---
name: Feature request
about: Suggest an idea for this project
title: ''
labels: feature
assignees: ''

---

**What is your feature request?**

I would like to be able to create schemas for TSV files which use a variety of different ways to represent inlined values. AFAIK right now, only a weird JSON list style syntax is supported, like `["a", "b"]`.

However, in order for us to be able to reflect much more common separators like:

- `a|b` (pipe separated)
- `a;b` (semicolon separated)
- `a,b` (csv separated)
-  `a, b` (csv+space separated)

etc, we need some way to configure / document the "delimiter" or "list-syntax" or however you like to call it.

**How important is this feature to you?** Select from the options below:

• Important - it's a blocker and can't do work without it

**Additional context**

This is also related to @tfliss work on pandera generators and in general at better support of dataframe validation.

matentzn:
> Is this most practical to implement as model annotations, (de)serializer configuration or as a transform (maybe different sides of a coin?).

@tfliss I am not sure.. I think I would like @sierra-moxon and @cmungall to give their intuition here..

turbomam:
see also
- https://github.com/orgs/linkml/discussions/1996

turbomam:
Presumably this problem is experienced in `linkml-validate` or `linkml-convert`, but is somewhat relevant to schemasheets too.

turbomam:
MIxS is a key use case for this feature. Related issues in the MIxS repo:

- GenomicsStandardsConsortium/mixs#952 ("specify how the LinkML `multivalued` metaslot should be used with MIxS terms")
- GenomicsStandardsConsortium/mixs#465 ("allow whitespace between delimiters in Value syntax patterns?")

MIxS has historically used inconsistent delimiters (`|`, `;`, `,`) in spreadsheet-based specifications without validation. The community is trying to standardize on conventions that align with LinkML's serialization behavior.

turbomam:
#3134 merged (Feb 27), adding configurable list formatting to the runtime loader/dumper. Missed v1.10.0 by two days — not in any release yet. Next release should include it in both packages.

Remaining gaps:
- `linkml-validate` uses a separate CSV/TSV loader with no list-splitting support (#3147)
- Installing linkml from git main + linkml-runtime from PyPI triggers a version skew crash (#3241) — this resolves once both packages are released together

turbomam:
[...]

Issue #491, opened 2021-11-30
Add citation format file to this repo, and the model repo


cmungall:
See https://citation-file-format.github.io/

sierra-moxon:
I see the citation link in this repo.  closing as done.

PR #3134, opened 2026-02-04
Add configurable list formatting for CSV/TSV serialization

### LINKED ISSUE #3041: CSV/TSV loader does not split brackets-free, multivalued primitive slots (with pipe delimiter)
## Summary
When loading CSV/TSV files, multivalued primitive slots (e.g., `slot_name: string*` or `multivalued: true` with a primitive range) are not properly handled. Values like `a|b|c` are not split into `["a", "b", "c"]`.

## Current Behavior
Given a schema with:
```yaml
slots:
  keywords:
    range: string
    multivalued: true
```

And a TSV row with `keywords: "apple|banana|cherry"`, the loader returns:
```yaml
keywords: "apple|banana|cherry"  # String, not a list
```

## Expected Behavior
The loader should recognize the multivalued nature and split on the pipe delimiter:
```yaml
keywords:
  - apple
  - banana
  - cherry
```

## Root Cause Analysis
In `linkml_runtime/utils/csvutils.py`, the `_get_key_config()` function only creates `KeyConfig` entries for **class-ranged** slots that are inlined:

```python
if range in all_cls and schemaview.is_inlined(slot):
    # Creates KeyConfig...
```

Multivalued **primitive** slots (string, integer, etc.) never get a `KeyConfig`, so `json-flattener` doesn't know to split them.

## Proposed Solution
Add handling for multivalued primitive slots in `_get_key_config()`:

```python
# After the existing class-range handling
elif slot.multivalued and range not in all_cls:
    # Primitive multivalued slot - tell json-flattener it's a list
    return KeyConfig(is_list=True)
```

And configure the loader/dumper to use pipe as the delimiter without brackets by setting `csv_list_markers=('', '')` in `GlobalConfig`.

## Complications Discovered
Initial attempts to implement this broke existing tests:
1. Empty values in enum-ranged multivalued slots became `[None]` instead of empty
2. Complex nested object tests failed with `IndexError: string index out of range`

These need careful handling, likely requiring:
- Special handling for empty strings
- Coordination with `json-flattener` library behavior
- Possibly separate configuration for simple vs complex list handling

## Related Issues
- #2580 - Boolean coercion (addressed in PR #3040)
- #2581 - Multivalued primitive handling in CSV/TSV
- MIxS [#952](https://github.com/GenomicsStandardsConsortium/mixs/issues/952) - multivalued metaslot usage
- MIxS [#1060](https://github.com/GenomicsStandardsConsortium/mixs/issues/1060) - cardinality and syntax

## Schemasheets Precedent
Schemasheets uses `internal_separator: |` (pipe) as the standard for multivalued fields. This is also the MIxS/GSC convention. The fix should align with this established pattern.

## Future Configurability
While the immediate fix should use the pipe `|` as the default delimiter (matching schemasheets and MIxS conventions), future enhancements could allow:

1. **Schema-level configuration** - `list_separator: "|"` in schema settings
2. **Loader parameter** - `csv_loader.load(..., list_separator="|")`  
3. **Slot-level annotation** - `list_separator: ";"` for slots needing different delimiters
4. **linkml-runtime config file** - Global configuration similar to schemasheets

For now, the single-paradigm approach (pipe delimiter, no brackets) is preferred to establish a working baseline.

## Files to Modify
- `packages/linkml_runtime/src/linkml_runtime/utils/csvutils.py` - Add primitive multivalued handling
- `packages/linkml_runtime/src/linkml_runtime/loaders/delimited_file_loader.py` - Configure list markers
- `packages/linkml_runtime/src/linkml_runtime/dumpers/delimited_file_dumper.py` - Configure list markers
- Tests for round-trip behavior

turbomam:
## New Summary from 2026-02-11

Adds configurable multivalued field formatting for CSV/TSV serialization, via schema-level annotations and CLI options.

**Before:** Multivalued fields always serialize with brackets: `[value1|value2|value3]`
**After:** With `list_wrapper: none`, fields serialize without brackets: `value1|value2|value3`

Closes #3041. Addresses the core of #2581 (filed by @matentzn as a blocker for supporting common delimited formats like pipe-separated, semicolon-separated, etc.).

## Origin and design

This follows the design @cmungall and I agreed on in our Dec 15 rolling meeting notes:

```yaml
annotations:
  list_wrapper: none   # square (default) | curly | paren | none
  list_delimiter: "; "     # any string; space must be explicit
```

With mapping to json-flattener: `list_wrapper: none` → `csv_list_markers=("", "")`, `list_delimiter` → `csv_inner_delimiter`.

### Deviation from spec: schema-level only

The Dec 15 spec discussed slot-level annotations overriding schema-level defaults via SchemaView. The implementation is **schema-level only**. json-flattener's `GlobalConfig` defines `csv_list_markers` and `csv_inner_delimiter` at the top level with no per-column configuration path, so slot-level overrides would require extending json-flattener itself. The primary use case (MIxS-style "semicolon-delimited, no brackets") is uniform across all multivalued fields in a schema, so this felt like the right scope for now.

### No changes to csvutils.py

Per Chris's guidance ("prefer no changes in csvutils.py"), configuration is handled in the loader and dumper rather than in the shared utility layer.

## SSSOM alignment

@matentzn suggested checking how SSSOM handles multivalued field packing. With `list_wrapper: none` and `list_delimiter: "|"`, our output matches SSSOM's TSV spec exactly (plain `a|b|c`, no brackets, strip whitespace). LinkML generalizes what SSSOM hardcodes — appropriate for a general-purpose modeling language where different schemas need different conventions.

The SSSOM ecosystem is actively working on delimiter-in-value escaping ([sssom#507](https://github.com/mapping-commons/sssom/pull/507), [sssom-java#17](https://github.com/gouttegd/sssom-java/pull/17)). This PR doesn't implement escaping either, but the annotation-based configuration provides the right foundation to add it later.

## What's not in scope

- **`linkml-validate` loader**: This PR modifies the `linkml_runtime` loader/dumper (used by `linkml-convert`), not the separate `linkml.validator.loaders.delimited_file_loader`. Filed #3147 to track unifying them.
- **Pandera / column-oriented data**: @tfliss and @sneakers-the-rat raised broader tabular concerns in [Discussion #1996](https://github.com/orgs/linkml/discussions/1996). This is row-oriented only — those feel like follow-up work for the tabular data library discussion.
- **Delimiter-in-value escaping**: Neither this PR nor SSSOM 1.0 implement escaping. Instead, the `refuse_delimiter_in_data` annotation/CLI flag raises a `ValueError` before serialization if any value contains the delimiter — preventing silent data corruption. Full escaping (e.g., SSSOM 1.1's backslash approach) can be added later.
- **RDF order preservation**: @gouttegd [clarified](https://github.com/linkml/linkml/pull/3134#issuecomment-3886155495) that multivalued slot order non-preservation is a LinkML-wide property (not SSSOM-specific), since the RDF translation rules use unstructured triples even when `list_elements_ordered: true`. Worth noting but orthogonal to this PR.

### Schema annotations

```yaml
id: https://example.org/myschema
name: myschema
annotations:
  list_wrapper: none
  list_delimiter: "|"
  list_strip_whitespace: "true"
  refuse_delimiter_in_data: "true"
```

| Annotation | Values | Default | Description |
|------------|--------|---------|-------------|
| `list_wrapper` | `square`, `curly`, `paren`, `none` | `square` | `square` uses `[a\|b]`, `curly` uses `{a\|b}`, `paren` uses `(a\|b)`, `none` has no wrapper `a\|b` |
| `list_delimiter` | any string | `\|` (pipe) | Character(s) used to separate list items |
| `list_strip_whitespace` | `true`, `false` | `true` | Strip whitespace around delimiters when loading and dumping |
| `refuse_delimiter_in_data` | `true`, `false` | `false` | Raise `ValueError` if any multivalued field value contains the delimiter, preventing silent data corruption |

### CLI options (override schema annotations)

```bash
linkml-convert -s schema.yaml -C Container -S items -t tsv \
  --list-wrapper none \
  --list-delimiter "|" \
  --list-strip-whitespace \
  --refuse-delimiter-in-data \
  input.yaml
```

| CLI Option | Default | Description |
|------------|---------|-------------|
| `--list-wrapper` | None (use schema) | `square`, `curly`, `paren`, or `none` |
| `--list-delimiter` | None (use schema) | Delimiter string |
| `--list-strip-whitespace` / `--no-list-strip-whitespace` | None (use schema) | Strip whitespace from list values |
| `--refuse-delimiter-in-data` / `--no-refuse-delimiter-in-data` | None (use schema) | Raise error if any value contains the delimiter |

## Review feedback addressed

From @cmungall's review (Feb 5):
- ✅ Converted all tests to pure idiomatic pytest (no unittest classes, no hybrid styles)
- ✅ Removed verbose agent-conversation-style comments
- ✅ Made helper functions public (dropped underscore prefix)

From Copilot:
- ✅ Fixed schema-level vs slot-level annotation mismatch in tests
- ✅ Removed unused variables
- ✅ Added warning log for invalid `list_wrapper` values

Coverage:
- ✅ Added CLI integration tests — 25 converter tests + 29 CSV/TSV runtime tests pass

## Files changed

- `docs/data/csvs.md` — documentation
- `packages/linkml/src/linkml/converter/cli.py` — CLI options
- `packages/linkml_runtime/src/linkml_runtime/dumpers/delimited_file_dumper.py` — output formatting
- `packages/linkml_runtime/src/linkml_runtime/loaders/delimited_file_loader.py` — input parsing
- `tests/linkml/test_utils/test_converter.py` — CLI tests
- `tests/linkml_runtime/test_loaders_dumpers/test_csv_tsv_loader_dumper.py` — runtime tests
- `tests/linkml_runtime/test_utils/test_csv_utils.py` — utility tests

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,4 +1,8 @@
 {
+  "access_right": "open",
+  "communities": [
+    {"identifier": "linkml"}
+  ],
   "creators": [
     {
       "name": "Moxon, Sierra A.T.",
@@ -69,7 +73,8 @@
       "name": "Kroll, Carlo"
     },
     {
-      "name": "Linke, David"
+      "name": "Linke, David",
+      "orcid": "0000-0002-5898-1820"
     },
     {
       "name": "Ly, Ryan"
@@ -139,5 +144,6 @@
       "relation": "isSupplementTo",
       "scheme": "url"
     }
-  ]
+  ],
+  "upload_type": "software"
 }


tests/linkml_runtime/test_utils/test_csv_utils.py
```diff
@@ -23,3 +23,127 @@ def test_get_configmap(self):
...
+    identifier: true
+  name:
+    range: string
+  tags:
+    range: string
+    multivalued: true
+    annotations:
+      list_syntax: plaintext
+      list_delimiter: "|"
+"""
```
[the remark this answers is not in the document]
turbomam: Fixed in commit 9a5f9d580. Moved annotations from slot-level to schema-level in the test schema to match the actual implementation, and updated the test methods to check `schemaview.schema.annotations` instead of slot annotations.

tests/linkml_runtime/test_loaders_dumpers/test_csv_tsv_loader_dumper.py
```diff
@@ -139,3 +143,439 @@ def test_tsvgen_unroundtrippable(self):
...
+        """Alias containing the delimiter character needs proper escaping."""
+        # This is a tricky edge case - what if someone's alias contains a pipe?
+        data = {
+            "persons": [
+                {"id": "1", "name": "Test", "aliases": ["Name|With|Pipes", "Normal"]},
+            ]
+        }
+        output_file = tmp_path / "delimiter_in_value.tsv"
+
+        # This test documents expected behavior - may need escaping strategy
```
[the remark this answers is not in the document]
turbomam: Fixed in commit 9a5f9d580. Removed the unused `data` and `output_file` variables, replaced with a comment describing the edge case being documented.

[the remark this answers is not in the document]
turbomam: Fixed in commit 9a5f9d580. Removed the unused `output_file` variable along with `data`, replaced with a comment describing the edge case.

turbomam:
**Re: patch coverage**

~~The 12 uncovered lines are in `cli.py` where the new CLI options are passed through to the loader/dumper.~~

**Update:** Added CLI integration tests in commit a1db0e5e9. The CLI options (`--list-syntax`, `--list-delimiter`, `--list-strip-whitespace`) are now tested directly via `CliRunner`.

Test coverage includes:
- 4 CLI tests in `test_converter.py` (linkml package)
- 33 tests in `test_csv_tsv_loader_dumper.py` and `test_csv_utils.py` (linkml_runtime package)

packages/linkml_runtime/src/linkml_runtime/loaders/delimited_file_loader.py
```diff
@@ -13,6 +13,144 @@
...
+    if not schemaview or not schemaview.schema:
+        return list_markers, inner_delimiter, strip_whitespace
+
+    # Check schema-level annotations
+    if schemaview.schema.annotations:
+        annotations = schemaview.schema.annotations
+        if "list_syntax" in annotations:
+            syntax = annotations["list_syntax"].value
+            if syntax == "plaintext":
+                list_markers = ("", "")
```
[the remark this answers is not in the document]
turbomam: Fixed in commit 971b3ecc5. Added a warning log when an invalid `list_syntax` value is provided. The code now warns users about typos and defaults to 'python' style.

tests/linkml_runtime/test_utils/test_csv_utils.py
```diff
@@ -23,3 +23,127 @@ def test_get_configmap(self):
 
 if __name__ == "__main__":
     unittest.main()
+
+
+# =============================================================================
+# pytest-style unit tests for annotation-based CSV configuration (issue #3041)
```
cmungall: This test file is now a hybrid of 3 styles:

1. UnitTest
2. pure pytest
3. non-idiomatic pytest (using classes)

The guidelines aren't clear what to do when contributing a new test to an existing UnitTest file:

https://linkml.io/linkml/maintainers/contributing.html#unittest-to-pytest-conversion

I favor consistency here, and I'd just convert this all to pure pytest

tests/linkml_runtime/test_utils/test_csv_utils.py
```diff
@@ -23,3 +23,127 @@ def test_get_configmap(self):
...
+# - list_syntax: python (default, brackets) or plaintext (no brackets)
+# - list_delimiter: character between values (default "|")
+# - Logic is in loader/dumper, NOT in csvutils.py
+# =============================================================================
+
+
+# -----------------------------------------------------------------------------
+# Inline test schemas for annotation testing
+#
+# We use inline schemas here because:
```
cmungall: This is a bit too much verbiage, sounds like the results of a conversation with an agent

tests/linkml_runtime/test_utils/test_csv_utils.py
```diff
@@ -23,3 +23,127 @@ def test_get_configmap(self):
...
+@pytest.fixture
+def schemaview_with_annotations():
+    """Schema with list_syntax and list_delimiter annotations on slot."""
+    return SchemaView(SCHEMA_WITH_LIST_ANNOTATIONS)
+
+
+# -----------------------------------------------------------------------------
+# Note on KeyConfig generation for multivalued primitive slots
+#
+# The _get_key_config() function in csvutils.py is NOT modified per Chris's
```
cmungall: Adding comments to tests is good if they help future maintenance or explain he purpose or function of the text, but this looks like a conversation that has lost its context

tests/linkml_runtime/test_loaders_dumpers/test_csv_tsv_loader_dumper.py
```diff
@@ -139,3 +143,428 @@ def test_tsvgen_unroundtrippable(self):
...
+    return SchemaView(SCHEMA_WHITESPACE_STRIP)
+
+
+@pytest.fixture
+def whitespace_preserve_schemaview():
+    """Schema with whitespace preservation enabled."""
+    return SchemaView(SCHEMA_WHITESPACE_PRESERVE)
+
+
+class TestWhitespaceStripping:
```
cmungall: this style of pytest seems non-idiomatic for this repo

tests/linkml_runtime/test_loaders_dumpers/test_csv_tsv_loader_dumper.py
```diff
@@ -8,6 +8,10 @@
 
 from linkml_runtime.dumpers import csv_dumper, json_dumper, tsv_dumper, yaml_dumper
 from linkml_runtime.loaders import csv_loader, tsv_loader, yaml_loader
+from linkml_runtime.loaders.delimited_file_loader import (
+    _get_list_config_from_annotations,
```
cmungall: consider making more clearly intended as public

(doctests might work better if it's intended as private)

tests/linkml_runtime/test_utils/test_csv_utils.py
turbomam: Fixed in 9b1e739 and 19b9006. Converted the entire file to pure pytest — both the old `CsvUtilTestCase` and the new annotation tests are now plain functions. Removed verbose comments and section headers.

tests/linkml_runtime/test_utils/test_csv_utils.py
turbomam: Fixed. Removed the verbose comment blocks.

tests/linkml_runtime/test_utils/test_csv_utils.py
turbomam: Fixed. Removed the KeyConfig note — it was stale context.

tests/linkml_runtime/test_loaders_dumpers/test_csv_tsv_loader_dumper.py
turbomam: Fixed. Converted all test classes to plain `test_*` functions, including the pre-existing `CsvAndTsvGenTestCase` (mechanical conversion, no behavior change).

tests/linkml_runtime/test_loaders_dumpers/test_csv_tsv_loader_dumper.py
turbomam: Made them public by dropping the underscore prefix: `get_list_config_from_annotations`, `enhance_configmap_for_multivalued_primitives`, `strip_whitespace_from_lists`.

Considered doctests, but the repo has no doctest infrastructure — there's no `--doctest-modules` in pyproject.toml and no doctest step in CI. The existing `>>>` examples in ~9 source files are documentation-only and never executed as tests. Making the functions public with dedicated pytest coverage seemed more practical.

turbomam:
## Consolidated unaddressed feedback — list formatting (PR #3134)
Gathering all outstanding feedback from multiple sources so nothing falls through the cracks.

### 1. Chris's CHANGES_REQUESTED review (Feb 5) — addressed in code, awaiting re-review
All 5 inline comments have been addressed:

* ✅ Converted test file from hybrid unittest/pytest to pure pytest ([9b1e739](https://github.com/linkml/linkml/commit/9b1e7390a3f42bba7773ffb8f5e5d711faf25300), [19b9006](https://github.com/linkml/linkml/commit/19b900688f942e61b2402820c7bd68b18d49c7da))
* ✅ Removed verbose agent-conversation-style comments
* ✅ Removed stale KeyConfig context note
* ✅ Converted test classes to plain `test_*` functions
* ✅ Made helper functions public (dropped underscore prefix)

**Status:** Code changes pushed, Chris has not re-reviewed yet.

### 2. Chris's design spec from Dec 15 rolling meeting notes
The agreed-upon design from our Dec 15 meeting (Chris & Mark rolling notes):

```yaml
# Schema annotation spec
attributes:
  name_list:
    multivalued: true
    annotations:
      list_syntax: python  ## allowed: python | plaintext
      list_delimiter: "; "  ## must include space explicitly. No effect if list_syntax == 'python'
```

**Mapping to json-flattener:**

* If `list_syntax == "python"` → use defaults
* Else → `csv_list_markers = ("", "")`, `csv_inner_delimiter = $list_delimiter`

**Cascading:** Use SchemaView — default to schema-level annotations, slot-level annotations override.

**Additional guidance from Chris:**

* "prefer no changes in `packages/linkml_runtime/src/linkml_runtime/utils/csvutils.py`"
* "remember that json-flattener isn't really schema aware"
* "pass csv list marker (a tuple) and inner delimiter (or a csv style syntax enum) **in schema**"

**Need to verify:** Does the current implementation match this spec exactly? Specifically:

* [ ]  Slot-level annotation override of schema-level annotations
* [ ]  Correct mapping to json-flattener's `csv_list_markers` and `csv_inner_delimiter`
* [ ]  No changes in `csvutils.py`

### 3. Chris's helper function visibility comment (PR inline)
Chris said "consider making more clearly intended as public (doctests might work better if it's intended as private)". I made them public and noted the repo has no doctest infrastructure — filed #3146 to track adding it. Chris hasn't responded to this.

### 4. Copilot review — invalid `list_syntax` values
Copilot flagged that invalid `list_syntax` values (e.g. "foobar") silently default to python style. Fixed in [971b3ec](https://github.com/linkml/linkml/commit/971b3ecc5cb49214d325c51e913f23008bb21f44) — added a warning log. ✅

### 5. Discussion #1996 context
My [progress update](https://github.com/orgs/linkml/discussions/1996#discussioncomment-15711681) in the "Improved ways of working with tabular data" discussion links this PR. Broader context from the discussion:

* @tfliss raised interaction with Pandera generator for inlined-as-simple-dict and range classes
* @sneakers-the-rat raised the two orientations of tabular data (column-oriented vs row-oriented)
* Chris's original post discusses whether we should have a separate tabular data library with plugin architecture

### 6. Feb 9 rolling notes
Chris noted "Finish linkml PRs" as a current action item.

### 7. Related issues
* [CSV/TSV loader does not split brackets-free, multivalued primitive slots (with pipe delimiter) #3041](https://github.com/linkml/linkml/issues/3041) — parent issue (Nico asked about ownership Jan 19, I confirmed Feb 4)
* [Make it possible to configure inlined multivalued strings syntax #2581](https://github.com/linkml/linkml/issues/2581) — original feature request (Nico asked for Chris/Sierra input Mar 2025)
* [specify how the LinkML `multivalued` metaslot should be used with MIxS terms GenomicsStandardsConsortium/mixs#952](https://github.com/GenomicsStandardsConsortium/mixs/issues/952) — MIxS multivalued field conventions
* [allow whitespace between delimiters in `Value syntax` patterns? GenomicsStandardsConsortium/mixs#465](https://github.com/GenomicsStandardsConsortium/mixs/issues/465) — whitespace between delimiters

### Next steps
1. Verify implementation matches Chris's Dec 15 spec (especially slot-level override cascade and json-flattener mapping)
2. Rebase if needed
3. Request re-review from @cmungall

turbomam:
@cmungall Heads up on one deviation from our Dec 15 spec. We discussed slot-level annotations overriding schema-level defaults, but the current implementation only supports schema-level annotations.

**Why:** json-flattener's [`GlobalConfig`](https://github.com/cmungall/json-flattener/blob/2b4f40d74287/src/json_flattener/flattener.py#L108-L127) defines `csv_list_markers` and `csv_inner_delimiter` at the top level, not on `KeyConfig`. These get applied uniformly to all columns — there's no per-column configuration path. So slot-level overrides would require extending json-flattener itself.

I think schema-level-only is the right call here. The main use case driving this (MIxS-style "semicolon-delimited, no brackets") is uniform across all multivalued fields in a given schema anyway. If a per-slot use case comes up later, we can add it via json-flattener at that point.

Also rebased onto current main and resolved the conflict with #3118's new converter tests. All 54 tests pass (25 converter + 29 CSV/TSV runtime). Ready for re-review when you get a chance.

turbomam:
## Scope and known limitations

### What this PR does and doesn't touch

This PR modifies the `linkml_runtime` loader/dumper (used by `linkml-convert`). It does **not** touch the separate `linkml.validator.loaders.delimited_file_loader` (used by `linkml-validate`), which is a simpler 79-line loader built on bare `csv.DictReader` without json-flattener.

That means after this merges, `linkml-convert` will correctly split `a|b` into `['a', 'b']`, but `linkml-validate` on the same CSV will still see it as the raw string `a|b`. I filed #3147 to track unifying these two loaders — that felt like a separate effort.

### Schema-level only annotations

Our Dec 15 spec discussed slot-level annotation overrides via SchemaView. The implementation is schema-level only — see [my earlier comment](https://github.com/linkml/linkml/pull/3134#issuecomment-3885720449) for why (json-flattener's `GlobalConfig` has no per-column delimiter support).

### Known edge cases

- **Delimiter-in-value**: If a value contains the delimiter character, round-tripping will break. No escaping mechanism yet. Tracked in a skipped test with a note.
- **Empty multivalued fields**: Skipped test due to a json-flattener `json_clean` issue — empty lists don't roundtrip cleanly.
- **Pandera / column-oriented data**: @tfliss and @sneakers-the-rat raised broader tabular concerns in Discussion #1996. This PR is row-oriented only and doesn't address those — they feel like follow-up work for the tabular data library discussion.

### `list_strip_whitespace` accepts only `true`/`false`

Tightened in e4d955e to only accept case-insensitive `"true"` or `"false"`. Previously accepted YAML 1.1 conventions (`yes`/`no`, `0`/`1`). Changed to stay consistent with the direction in #3144 per Chris's feedback about not mixing boolean conventions.

turbomam:
## SSSOM alignment — context from today's linkml-dev meeting

Nico suggested looking at how SSSOM handles multivalued field packing in TSV. Turns out the SSSOM ecosystem is actively debating the same problems we're solving here, literally today.

### How SSSOM does it

From the [SSSOM/TSV spec](https://github.com/mapping-commons/sssom/blob/master/src/docs/spec-formats-tsv.md):

> "Multi-valued slots MUST be serialised as a list of values separated by `|` characters."

- No brackets — plain `value1|value2|value3`
- Pipe is hardcoded in the spec, not configurable
- sssom-py strips whitespace: `[s.strip() for s in v.split("|")]`
- Schema-driven — checks `multivalued: true` to know which columns to split

### Active spec work happening right now

- [mapping-commons/sssom#507](https://github.com/mapping-commons/sssom/pull/507) (opened Feb 10) — proposes backslash escaping (`\|` for literal pipe) within multivalued values, targeting SSSOM 1.1
- [mapping-commons/sssom#429](https://github.com/mapping-commons/sssom/issues/429) — the underlying debate about whether to forbid pipe in values, escape it, or percent-encode it
- [gouttegd/sssom-java#17](https://github.com/gouttegd/sssom-java/pull/17) — concrete Java implementation of the escape mechanism
- [mapping-commons/sssom#504](https://github.com/mapping-commons/sssom/pull/504) — Nico is promoting sssom-java as the reference implementation, so its escaping approach will likely set the standard

### How PR #3134 compares

| Aspect | SSSOM | LinkML PR #3134 |
|--------|-------|-----------------|
| Format | No brackets, plain `a` + pipe + `b` + pipe + `c` | Configurable: brackets by default, no brackets with `list_syntax: plaintext` |
| Delimiter | Pipe, fixed by spec | Configurable via `list_delimiter` annotation |
| Whitespace | `strip()` on each value | Configurable via `list_strip_whitespace` |
| Escaping | None in 1.0; backslash escaping proposed for 1.1 | None (same gap) |
| Schema-driven | Checks `multivalued: true` | Same, via SchemaView + `enhance_configmap_for_multivalued_primitives()` |

With `list_syntax: plaintext` and `list_delimiter: "|"`, our output matches SSSOM's format exactly. LinkML generalizes what SSSOM hardcodes — which makes sense for a general-purpose modeling language where different schemas need different conventions (SSSOM uses pipe, MIxS has historically used semicolons, commas, etc.).

### Escaping

Neither SSSOM (1.0) nor this PR handle delimiter-in-value escaping. SSSOM is actively working on backslash escaping for 1.1 (backslash-pipe for literal pipe, double-backslash for literal backslash). If/when that lands and we want to support it in LinkML, it would be a follow-up — the annotation-based configuration in this PR provides the right foundation to build on.

### Related

- Nico filed [#2581](https://github.com/linkml/linkml/issues/2581) (the origin issue for this work), rating configurable delimiters as a "blocker"
- [SSSOM discussion #428](https://github.com/mapping-commons/sssom/discussions/428) found 4 of 5 published SSSOM datasets on Zenodo don't follow the spec — reinforcing why good serialization tooling matters
- [SSSOM issue #491](https://github.com/mapping-commons/sssom/issues/491) notes SSSOM does not guarantee order preservation in multivalued slots (for RDF simplicity) — something to be aware of for LinkML's semantics

gouttegd:
Not strictly related to the issue of list delimiters, but since this is noted here:

> SSSOM issue #491 notes SSSOM does not guarantee order preservation in multivalued slots (for RDF simplicity) — something to be aware of for LinkML's semantics

This is not a SSSOM-specific limitation. SSSOM does not guarantee the order of the values in a multi-valued slot, because LinkML itself does not guarantee that. The behaviour of SSSOM here was directly taken from the behaviour of the LinkML runtime.

The [rules for RDF translations](https://linkml.io/linkml-model/latest/docs/specification/06mapping/#translation-to-rdf) do not cover the case of multi-valued slots, so I don’t know what was the _intention_ here, but in effect the LinkML runtime translates multi-valued slots as simple unstructured list of triples (even if the slot is defined with `list_elements_ordered: true`). Such a translation cannot ensure that the order of values is preserved. So LinkML _may_ preserve the order of values in all other formats, but as soon as you convert to or from RDF the order of values cannot be expected to be preserved. _Ergo_, more generally, LinkML does not guarantee that the order of values in multi-valued slots is preserved.

turbomam:
Merged main and fixed two lint issues introduced by the merge (Ruff UP038 union syntax, unused `Union` import). The link checker failure is a transient 404 on an unrelated URL — re-run triggered and it's passing on main.

All review feedback from Feb 5 was addressed in commits 050f6d6–f360e7f. Ready for re-review.

turbomam:
Merge blocked: branch protection still requires test (ubuntu-latest, 3.9) and test (windows-latest, 3.9) but CI matrix no longer includes 3.9 — needs admin to update required checks.

docs/data/csvs.md
```diff
@@ -59,6 +59,203 @@ The [json-flattener/](https://github.com/cmungall/json-flattener/) library is us
...
+value1|value2|value3
+```
+
+You can customize the formatting using schema-level annotations:
+
+```yaml
+id: https://example.org/myschema
+name: myschema
+annotations:
+  list_syntax: plaintext    # removes brackets from all multivalued fields
```
kevinschaper: Would you be open to reframing this as list_wrapper with permissible values of square, curly, paren with a default value of none?

packages/linkml_runtime/src/linkml_runtime/dumpers/delimited_file_dumper.py
```diff
@@ -8,6 +8,12 @@
 from linkml_runtime.dumpers.dumper_root import Dumper
 from linkml_runtime.dumpers.json_dumper import JSONDumper
 from linkml_runtime.linkml_model.meta import SchemaDefinition, SlotDefinitionName
+from linkml_runtime.loaders.delimited_file_loader import (
+    check_data_for_delimiter,
+    enhance_configmap_for_multivalued_primitives,
+    get_list_config_from_annotations,
+    strip_whitespace_from_lists,
+)
```
kevinschaper, line 17: optional style feedback, but it might be worth splitting shared functions out into a separate 'utils' file, rather than putting them in the load and importing into the dumper (or the opposite).

packages/linkml_runtime/src/linkml_runtime/loaders/delimited_file_loader.py
```diff
@@ -12,6 +15,222 @@
 from linkml_runtime.utils.yamlutils import YAMLRoot
 
 
+def get_list_config_from_annotations(
+    schemaview: SchemaView,
+    index_slot: SlotDefinitionName = None,
```
kevinschaper: It looks like index_slot isn't actually used

docs/data/csvs.md
kevinschaper: Just to expand on this, I researched what this parameter is named anywhere else that I could find it, and found that SQLServer has a boolean flag `WITHOUT_ARRAY_WRAPPER` and Jackson, a java package for serialization has a `WRITE_SINGLE_ELEM_ARRAYS_UNWRAPPED` boolean.

turbomam:
Thanks @kevinschaper that was really helpful. I'm not very good at pulling out dead code or refactoring out sharable code. "python" and "plain text" styles were @cmungall's idea and I appreciated how he was trying to ground the distinction in terms that would be familiar to us, but I prefer your suggested style.

turbomam:
oops needs some tweaks still

turbomam:
ready for re-review, @kevinschaper

docs/data/csvs.md
```diff
@@ -59,6 +59,202 @@ The [json-flattener/](https://github.com/cmungall/json-flattener/) library is us
...
+
+TSV output: `[Alice|Bob|Charlie]`
+
+**No wrapper with pipe delimiter:**
+
+```yaml
+id: https://example.org/unwrapped
+name: unwrapped_example
+annotations:
+  list_wrapper: none
```
sneakers-the-rat, line 154: It seems like the wrapper being a keyed value (e.g. "square" rather than []) and the delimiter character being separate fields with different semantics adds a great deal of complexity. Why not just make `list_delimiters` a thruple like 

```yaml
list_delimiters: ["[", "|", "]"]
```

Or 

```yaml
list_delimiters: [null, "|", null]
```

then it seems like most of this could just be regex?

```python
re.findall(
  rf"{delim[0] or ''}"
  rf"(.*){re.escape(delim[1])}?"
  rf"{delim[2] or ''}"
)
```

(Or whatever the actual pattern is).

packages/linkml_runtime/src/linkml_runtime/dumpers/delimited_file_dumper.py
sneakers-the-rat: Yeah, and since most of the body of the functions is just resolving the config, this seems like something that should just be a class so that happens exactly once and the rest of the operations can use the settled config values

```python
formatter = ListFormatter.from_schemaview(sv, **kwargs)

delimited = formatter.to_delimited(value)
formatter.to_list(delimited)
```

packages/linkml_runtime/src/linkml_runtime/loaders/delimited_file_loader.py
```diff
@@ -62,11 +68,44 @@ def _get_json_str_to_load(
...
+        list_strip_whitespace: bool = None,
         **kwargs,
     ):
         if schemaview is None:
             schemaview = SchemaView(schema)
+
+        # Read list configuration from schema annotations
+        list_markers, inner_delimiter, strip_whitespace, _refuse = get_list_config_from_annotations(schemaview)
+
+        # CLI options override schema annotations
```
sneakers-the-rat: Yeah doing this in the function body seems messy. Seems like there should just be one `get_list_config(schemaview, **cli_kwargs)` function that retuens the settled config so that resolving the config happens in one place and that place is a different place than where it's applied, since there are already plenty of other things going on in this method?

packages/linkml_runtime/src/linkml_runtime/utils/list_utils.py
```diff
@@ -0,0 +1,235 @@
...
+            value = str(annotations["refuse_delimiter_in_data"].value).lower()
+            if value not in ("true", "false"):
+                logger.warning(
+                    f"Invalid refuse_delimiter_in_data value '{value}'. "
+                    "Expected 'true' or 'false'. Defaulting to false."
+                )
+            else:
+                refuse_delimiter_in_data = value == "true"
+
+    return list_markers, inner_delimiter, strip_whitespace, refuse_delimiter_in_data
```
sneakers-the-rat: Rather than these vars being slung around loose and free in a bunch of places, dataclass or typeddict

docs/data/csvs.md
kevinschaper: I think I would interpret a multivalued `list_delimiters` argument to be an OR of any character specified, meaning more like "split on any of these": 

```
list_delmiters: ["\t",  "," , "|"]  
```

packages/linkml_runtime/src/linkml_runtime/utils/list_utils.py
kevinschaper: yeah, good call

turbomam: Addressed in 8686ba4 — introduced a `ListConfig` dataclass that bundles the four settings. Config resolution (defaults → schema annotations → CLI overrides) now happens in a single `get_list_config()` function.

packages/linkml_runtime/src/linkml_runtime/loaders/delimited_file_loader.py
turbomam: Done in 8686ba4 — `get_list_config(schemaview, **cli_kwargs)` now returns a settled `ListConfig` dataclass. Both loader and dumper just call it and use the attributes.

packages/linkml_runtime/src/linkml_runtime/dumpers/delimited_file_dumper.py
turbomam: Agreed — addressed in the same commit (8686ba4). The dumper body is now just a `get_list_config()` call followed by the actual work.

docs/data/csvs.md
sneakers-the-rat: well, yes it would require the same kind of documentation that this PR has on usage. but yeah generally there are three characters involved, and two of those characters are specified with a string mapping. if we instead had three characters, then the operation could lose a step and just be a single operation while also being more general. not a strong opinion and i'm coming in late, so feel free to ignore.

turbomam: Thanks, @sneakers-the-rat . That's fine that you're speaking up late in the game. You have a better sw engineering perspective than I do. But I would prefer to merge this in now as close to as-is as possible

tests/linkml_runtime/test_loaders_dumpers/test_csv_tsv_loader_dumper.py
turbomam: Correction: I overstated this. Chris demonstrated in our meeting today that `pytest --doctest-modules` works fine on the linkml package — it found 9 doctests across 5 files, 6 passing and 3 failing due to stale expected output (not logic bugs). So the `>>>` examples are real doctests, not just documentation — they're just not wired into CI yet. I'll look into adding doctests for these functions.

PR #3135, opened 2026-02-04
Fix induced_slot setting alias equal to slot name

### LINKED ISSUE #2911: `SchemaView.induced_slot` produces a slot definition with an `alias` equal to its `name`
**Describe the bug**
When `SchemaView.induced_slot` produces an induced slot definition, if the underlying slot has no `alias` it will set the `alias` to an underscored version of the `name`. Often, this will be the same as the `name` itself. This causes problems if that induced slot definition goes through certain generators, for example `PythonGenerator`.

1.9.4

This toy example shows extracting an induced slot definition from a source schema, adding it to a target schema, and passing that target schema through `PythonGenerator`:

```python
source_schema = SchemaView("""
id: http://example.org/source
name: source
imports:

slots:
  slot_a:
    range: integer

classes:
  SourceClass:
    slots:
      - slot_a
""")

target_schema = SchemaView("""
id: http://example.org/target
name: target
imports:

classes:
  TargetClass:
""")

induced_slot = source_schema.induced_slot("slot_a", "SourceClass")
target_schema.add_slot(induced_slot)
target_schema.get_class("TargetClass").slots.append("slot_a")
target_schema.set_modified()

PythonGenerator(target_schema.schema).serialize()
```

The final line raises the following exception:
```
  File "tmp.py", line 197, in <module>
    PythonGenerator(target_schema.schema).serialize()
  File ".venv/lib/python3.12/site-packages/linkml/utils/generator.py", line 344, in serialize
    sub_out = self.end_schema(**kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/linkml/generators/pythongen.py", line 281, in end_schema
    return re.sub(r" +\n", "\n", self.gen_schema().replace("\t", "    ")).strip(" ")
                                 ^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/linkml/generators/pythongen.py", line 272, in gen_schema
    {self.gen_classdefs()}
     ^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/linkml/generators/pythongen.py", line 472, in gen_classdefs
    return "\n".join([self.gen_classdef(v) for v in clist if not v.imported_from])
                      ^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/linkml/generators/pythongen.py", line 478, in gen_classdef
    slotdefs = self.gen_class_variables(cls)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/linkml/generators/pythongen.py", line 587, in gen_class_variables
    initializers += [self.gen_class_variable(cls, slot, not is_root) for slot in slot_variables]
                                                                                 ^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/linkml/generators/pythongen.py", line 1005, in _slot_iter
    for slot in self.all_slots(cls):
                ^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/linkml/utils/generator.py", line 485, in all_slots
    sname_base = alias_root(self.schema, sname)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/linkml/utils/mergeutils.py", line 193, in alias_root
    raise ValueError("Error: Slot {slotname} is aliased to itself.")
ValueError: Error: Slot {slotname} is aliased to itself.
```

kevinschaper:
## Summary

- Fixes `SchemaView.induced_slot` setting `alias` to an underscored version of the slot name even when it equals the name itself
- This caused downstream errors like `ValueError: Error: Slot {slotname} is aliased to itself.` in PythonGenerator
- Now only sets `alias` when the underscored version differs from the slot name

Fixes #2911

### Core fix (`schemaview.py`)

**Before:**
```python
if not induced_slot.alias:
    induced_slot.alias = underscore(slot_name)
```

**After:**
```python
if not induced_slot.alias:
    underscored = underscore(slot_name)
    if underscored != induced_slot.name:
        induced_slot.alias = underscored
```

### Reference validator fix (`referencevalidator.py`)

The reference validator used `slot.alias` directly to check for required/recommended slots in input objects. With `alias` now correctly `None` when it equals the name, these lookups need to fall back to `slot.name`:

**Before:**
```python
if slot.required and slot.alias not in input_object and not (slot.identifier or slot.key):
```

**After:**
```python
slot_key = slot.alias or slot.name
if slot.required and slot_key not in input_object and not (slot.identifier or slot.key):
```

### Test changes (`test_schemaview.py`)

New parametrized test covering the fix:
```python
@pytest.mark.parametrize(
    ("slot_name", "expected_alias"),
    [
        ("my_slot", None),      # underscore("my_slot") == name → no alias
        ("mySlot", None),       # underscore("mySlot") == name → no alias
        ("my slot", "my_slot"), # underscore("my slot") != name → alias set
    ],
)
def test_induced_slot_alias_not_equal_to_name(...)
```

Updated `test_alias_slot` to reflect that alias is now `None` when it would equal the slot name, replacing the previous assertion that alias was always populated.

### DIFF
--- a/tests/linkml/test_issues/test_issue_2911.py
+++ b/tests/linkml/test_issues/test_issue_2911.py
@@ -8,7 +8,7 @@ from linkml_runtime.utils.schemaview import SchemaView
 @pytest.mark.parametrize(
     "slot_name,expected_alias",
     [
-        ("slot_a", "slot_a"),
+        ("slot_a", None),
         ("slot-a", "slot_a"),
     ],
 )


amc-corey-cox:
Hey @kevinschaper, I rebased this onto main. The only CI failure is `test_issue_2911.py` which landed on main recently — its `("slot_a", "slot_a")` parametrized case expects the old behavior where alias is always set. With your fix, that correctly returns `None` now. Just need to update line 11 to expect `None` instead of `"slot_a"` and it should go green.

PR #3159, opened 2026-02-11
Extract literal_form from StructuredAlias in all_aliases()

### LINKED ISSUE #3158: `SchemaView.all_aliases()` returns StructuredAlias objects instead of strings
`all_aliases()` appends raw `StructuredAlias` dataclass objects to the alias list instead of extracting `literal_form`. The method's return type claims `list[str]` but actually returns `dict[str, list[StructuredAlias]]`.

Callers doing `"alias_name" in aliases[element]` get `False` because `str != StructuredAlias`.

### To reproduce

```python
sv = SchemaView("kitchen_sink.yaml")
aliases = sv.all_aliases()
assert "dad" in aliases["Adult"]  # fails
```

### Expected

`all_aliases()` should return `dict[str, list[str]]` with `literal_form` extracted from structured aliases.

matentzn:
Fixes #3158 

- Extract `sa.literal_form` instead of appending raw `StructuredAlias` objects in `all_aliases()`

## Changed files
- `packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py` — `all_aliases()` method

## Test plan
- [x] `test_all_aliases` passes — string membership checks work
- [x] Full suite passes

### DIFF
--- a/docs/faq/tools.md
+++ b/docs/faq/tools.md
@@ -58,7 +58,7 @@ For small schemas with limited inheritance, it should be possible to mentally pi
 
 There are a few strategies:
 
-* Use [gen-doc](https://linkml.io/linkml/generators/markdown) to make markdown that can be viewed using mkdocs
+* Use [gen-doc](https://linkml.io/linkml/generators/docgen) to make markdown that can be viewed using mkdocs
     * note you get this "for free" if you set up your project using the LinkML project copier template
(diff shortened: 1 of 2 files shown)


matentzn:
@sierra-moxon I got the QC to pass on this one as well. Thanks for checking!

PR #3162, opened 2026-02-11
Fix pythongen to generate _normalize_inlined for inlined_as_list slots

### LINKED ISSUE #3160: pythongen skips `_normalize_inlined` for `inlined_as_list` slots without identifier
`gen_postinit` in `pythongen.py` has a fallback that searches for a required field to use as `key_name` when a range class has no identifier. But the condition guards this with `not slot.inlined_as_list`, so `inlined_as_list` slots never get the fallback.

Without a `key_name`, the generator emits an inline list comprehension instead of `_normalize_inlined_as_list`. The inline comprehension can't handle dict-form YAML input — it wraps the dict in a list and passes dict keys as `**kwargs`, crashing with unexpected keyword arguments.

### To reproduce

Load any schema with `structured_aliases` from dict-form YAML:

```
TypeError: StructuredAlias.__init__() got an unexpected keyword argument 'dad'
```

32 errors + 3 failures across the test suite.

### Expected

`_normalize_inlined_as_list(key_name="literal_form", keyed=False)` should be generated, same as the old (2024) `meta.py` which used `_normalize_inlined_as_dict(key_name="literal_form", keyed=False)`.

matentzn:
Fixes #3160 

- Remove `and not slot.inlined_as_list` guard from identifier fallback in `gen_postinit`
- Both `inlined_as_dict` and `inlined_as_list` need a `key_name` to handle dict input

## Changed files
- `packages/linkml/src/linkml/generators/pythongen.py` — condition change at line 955

## Test plan
- [x] `test_metamodel` snapshot updated — `has_familial_relationships` now uses `_normalize_inlined_as_list`
- [x] `test_issue_50` snapshot updated — `annotations` slot uses `_normalize_inlined_as_list`
- [x] Full suite passes — no `StructuredAlias.__init__` keyword argument errors

### DIFF
--- a/tests/linkml/test_issues/__snapshots__/issue_84.py
+++ b/tests/linkml/test_issues/__snapshots__/issue_84.py
@@ -124,9 +124,7 @@ class Biosample(YAMLRoot):
         if self.name is not None and not isinstance(self.name, str):
             self.name = str(self.name)
 
-        if not isinstance(self.annotations, list):
-            self.annotations = [self.annotations] if self.annotations is not None else []
-        self.annotations = [v if isinstance(v, Annotation) else Annotation(**as_dict(v)) for v in self.annotations]
+        self._normalize_inlined_as_list(slot_name="annotations", slot_type=Annotation, key_name="has raw value", keyed=False)
 
         if not isinstance(self.alternate_identifiers, list):
             self.alternate_identifiers = [self.alternate_identifiers] if self.alternate_identifiers is not None else []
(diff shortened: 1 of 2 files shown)


amc-corey-cox:
Hi @matentzn — sorry I missed this PR and duplicated the work in #3165. I've compared the two and #3165 is a superset of your fix, so I'm closing this one in favor of it.

Key differences in #3165:

- **Class-range guard**: The fallback loop now also checks `range_slot.range not in self.schema.classes`, skipping required fields whose range is a class (e.g. `type: TypeObj` in the issue #463 schema). Without this, `_normalize_inlined_as_list` hits `TypeError: unhashable type` when it tries to add the complex object to `cooked_keys` in yamlutils.
- **yamlutils hardening**: `cooked_keys.add(key)` is now guarded by `if keyed`, since the set is only used for duplicate detection in the `keyed=True` path. This prevents unhashable type errors as a defense-in-depth measure.
- **Tests**: 8 new tests covering the core fix, dict-form input, duplicate key values with `keyed=False`, and edge cases (no required fields, class-range required fields).
- **All 4 snapshots updated** (meta.py, biolink.py, issue_84.py, genpython/meta.py).

I've added you as a reviewer on #3165.

amc-corey-cox:
Closing in favor of #3165 which includes this fix plus additional safety guards and tests.

Issue #3129, opened 2026-02-03
Toolkit doesn't fully support bundled linkml-model schemas (arrays, extended_types)


matentzn:
linkml-model ships schemas like `arrays.yaml` and `extended_types.yaml` alongside the core metamodel. These are intended for schema authors to use via `imports` or `implements`. However, the current toolkit can't fully process them:

- **arrays.yaml**: `SchemaView` fails to load it because `slot_usage` on slots triggers `TypeError: SlotDefinition.__init__() got an unexpected keyword argument 'slot_usage'` in the runtime loader. `PythonGenerator` also fails for the same reason. #3389 
- **extended_types.yaml**: `SchemaView` loads it fine, but `PythonGenerator` fails with `ValueError: type "any_number" must declare a type base or parent (typeof)` — the `union_of` construct on types isn't handled by `SchemaLoader`.

Both schemas only use features that the metamodel itself defines (`slot_usage`, `union_of`), so the burden of fix is on linkml, not on linkml-model.

Tests have been added in the linked PR to cover these schemas alongside the existing metamodel compat tests.

matentzn:
arrays.yaml: add a note in the header and make sure the schema validator passes
extended_types.yaml: add note that its experimantal make sure schema validator passes

amc-corey-cox:
## Root cause for the `extended_types.yaml` failure

The `union_of` `ValueError` traces to `schemaloader.py:326` — the validation check requires every type to have either `base` or `typeof`, but doesn't account for `union_of` as a valid type definition:

```python
for typ in self.schema.types.values():
    if not typ.base and not typ.typeof:
        self.raise_value_error(
            f'type "{typ.name}" must declare a type base or parent (typeof)',
            typ.name,
        )
```

The `any_number` type in `extended_types.yaml` only has `union_of` members, which should be sufficient.

(Migrated from #3180, which is a duplicate of this issue.)

matentzn:
Closing in favour of https://github.com/linkml/linkml/issues/3406

The other sub issues in this super issue have been addressed

PR #3163, opened 2026-02-11
Update all linkml-model files to the latest 1.10.rc4

### LINKED ISSUE #3076: Update linkml to 1.10.0 and the  linkml-model to 1.10.0-rc1
Blocked by:

- #3093 
- #3088 
- #3097
- #3156
- #3158
- #3160
- #3161

Here is the process:

1. Rebuild linkml-model files (the version of LinkML used for the rebuild does not matter too much, but best something newer)
1. Publish linkml-model as RC, eg. `v1.10.0-rc1`
1. Create branch on linkml repo and test linkml-model `v1.10.0-rc1` using the new pytest system (make this a little more intuitive maybe with a Makefile rule? Note to self).
1. If all tests are ok, have a meeting between leads and sign off (this is because I am new to the project)
1. Publish linkml-model `v1.10.0` (no RC)
1. Publish linkml version `v1.10.0` with the inkml-model `v1.10.0`
1. Add a safeguard to ensure that future minor version increases in `linkml` are always accompanied by a linkml-model release (not patches)

@cmungall correct me if I am wrong.

matentzn:
Fixes #3076 

Blockers:

- https://github.com/linkml/linkml/pull/3159
- https://github.com/linkml/linkml/pull/3157

### DIFF
--- a/packages/linkml/src/linkml/__init__.py
+++ b/packages/linkml/src/linkml/__init__.py
@@ -30,11 +30,8 @@ LOCAL_MODEL_YAML_FILES = (
 
 # Local location of jsonld and context.jsonld files
 LOCAL_METAMODEL_LDCONTEXT_FILE = linkml_files.LOCAL_PATH_FOR(Source.META, Format.JSONLD)
-LOCAL_METAMODEL_JSONLD_FILE = linkml_files.LOCAL_PATH_FOR(Source.META, Format.JSON)
 LOCAL_TYPES_LDCONTEXT_FILE = linkml_files.LOCAL_PATH_FOR(Source.TYPES, Format.JSONLD)
-LOCAL_TYPES_JSONLD_FILE = linkml_files.LOCAL_PATH_FOR(Source.TYPES, Format.JSON)
 LOCAL_MAPPINGS_LDCONTEXT_FILE = linkml_files.LOCAL_PATH_FOR(Source.MAPPINGS, Format.JSONLD)
-LOCAL_MAPPINGS_JSONLD_FILE = linkml_files.LOCAL_PATH_FOR(Source.MAPPINGS, Format.JSON)
 
 # Local location of metamodel shex file
 LOCAL_SHEXJ_FILE_NAME = linkml_files.LOCAL_PATH_FOR(Source.META, Format.SHEXJ)
--- a/packages/linkml_runtime/src/linkml_runtime/linkml_model/linkml_files.py
+++ b/packages/linkml_runtime/src/linkml_runtime/linkml_model/linkml_files.py
@@ -37,7 +37,6 @@ class Format(_AutoName):
 
     EXCEL = auto()
     GRAPHQL = auto()
-    JSON = auto()
     JSONLD = auto()
     JSON_SCHEMA = auto()
     NATIVE_JSONLD = auto()
@@ -65,7 +64,6 @@ class _Path:
 
     EXCEL = _PathInfo("excel", "xlsx")
     GRAPHQL = _PathInfo("graphql", "graphql")
-    JSON = _PathInfo("json", "json")
     JSONLD = _PathInfo("jsonld", "context.jsonld")
     JSON_SCHEMA = _PathInfo("jsonschema", "schema.json")
     NATIVE_JSONLD = _PathInfo("jsonld", "model.context.jsonld")
@@ -222,10 +220,6 @@ class ModelFile:
     @property
     def graphql(self) -> ModelLoc:
         return ModelFile.ModelLoc(self._model, Format.GRAPHQL)
-
-    @property
-    def json(self) -> ModelLoc:
-        return ModelFile.ModelLoc(self._model, Format.JSON)
 
     @property
     def jsonld(self) -> ModelLoc:


matentzn:
I asked claude to do a highly critical, adverserial review of this PR for inspiration.

## CRITICAL ERRORS

### 1. `json/meta.json` was NOT regenerated — missing all new metamodel elements

**File:** `packages/linkml_runtime/src/linkml_runtime/linkml_model/json/meta.json`

This file is the JSON serialization of the full schema. It has 6,878 lines but contains **zero** occurrences of `extra_slots`, `ExtraSlotsExpression`, `extra_slots_expression`, or `allowed`. The only change in this file is a single description text update (`ifabsent` expanded description).

Every other generated artifact correctly includes the new elements:
- `meta.py` (Python) — YES
- `meta.schema.json` (JSON Schema) — YES (4 occurrences of ExtraSlotsExpression)
- `meta.graphql` (GraphQL) — YES (camelCase: extraSlots, ExtraSlotsExpression)
- `meta.proto` (Protobuf) — YES
- `meta.shex` (ShEx) — YES
- `meta.owl.ttl` (OWL) — YES (10 occurrences)
- `meta.shacl.ttl` (SHACL) — YES (7 occurrences)
- `meta.sql` (SQL Schema) — YES (6 occurrences)
- **`meta.json` (JSON) — NO (0 occurrences)**

**Impact:** Any downstream tooling that reads the schema from `json/meta.json` (e.g., schema loaders, validators, documentation generators) will not see the new `extra_slots` feature, the `ExtraSlotsExpression` class, the `allowed` slot, or the `alias` addition to ClassDefinition.

**Action required:** Regenerate `json/meta.json` from the updated `meta.yaml`.

### 2. `extended_types.py` is an empty file despite substantial source schema

**File:** `packages/linkml_runtime/src/linkml_runtime/linkml_model/extended_types.py`

This file is completely empty (0 bytes). However, its corresponding source schema `model/schema/extended_types.yaml` defines 152 lines of content including:
- Class `Any` ("An unconstrained type or object")
- 13 numeric types: `signedinteger`, `unsignedinteger`, `int8`, `int16`, `int32`, `int64`, `uint8`-`uint64`, `float16`, `float32`, `float64`
- A composite type: `any_number` (union of integer/float/decimal)

**Impact:** If downstream code imports from `extended_types`, it will get an empty module. If `extended_types` is supposed to define Python dataclasses for these types, they are missing.

**Action required:** Either regenerate `extended_types.py` with proper content, or add a comment explaining why it is intentionally empty (e.g., if these types are handled purely through the type system without needing Python classes).

---

## STRONG WARNINGS

### 3. `output.yaml` — stray file committed to repository root

**File:** `output.yaml` (repo root)

This file is a test/debug artifact containing a small "EquipmentSchema". It is NOT in `.gitignore`. This PR adds `metamodel_version: 1.7.0` to it, which appears to be a side effect of running a generator. This file should not be in the repository.

**Action required:** Remove `output.yaml` from the PR and add it to `.gitignore`.

### 4. `GITHUB_IO_PATH_FOR` — breaking API change (removed `version` parameter)

**File:** `linkml_files.py`

The old signature was:
```python
def GITHUB_IO_PATH_FOR(source: Source, fmt: Format, version="latest") -> str:
```

The new signature is:
```python
def GITHUB_IO_PATH_FOR(source: Source, fmt: Format) -> str:
```

Changes:
- The `version` parameter is removed — any caller passing `version=` will get a `TypeError`
- The URL structure changed: old produced `https://linkml.github.io/linkml-model/latest/linkml_model/rdf/meta.ttl`, new produces `https://linkml.github.io/linkml-model/rdf/meta.ttl`
- The `linkml_model` path segment is also removed

**Mitigating factors:** The test for this function is `@pytest.mark.skip("github paths largely unused")`. The only internal caller (`ModelFile.ModelLoc.github_loc()`) does not pass the `version` argument.

**Risk:** Medium. Any external code calling `GITHUB_IO_PATH_FOR(source, fmt, version="1.0.0")` will break. The URLs may also 404 depending on the GitHub Pages structure at `linkml.github.io/linkml-model/`.

### 5. `GITHUB_PATH_FOR` — `linkml_model` path segment removed from raw GitHub URLs

**File:** `linkml_files.py`

Old: `https://raw.githubusercontent.com/linkml/linkml-model/main/linkml_model/rdf/meta.ttl`
New: `https://raw.githubusercontent.com/linkml/linkml-model/main/rdf/meta.ttl`

This changes the expected repository layout from files being under `linkml_model/` to being at the repo root. Whether the new URLs are valid depends on the structure of the `linkml/linkml-model` repository.

**Risk:** Medium. The test is also skipped ("github rate limiting").

### 6. `default_prefix` str coercion removed from `SchemaDefinition.__post_init__`

**File:** `meta.py`, `SchemaDefinition.__post_init__`

The old code had:
```python
if self.default_prefix is not None and not isinstance(self.default_prefix, str):
    self.default_prefix = str(self.default_prefix)
```

This line was completely removed. If someone passes a non-string value as `default_prefix` (e.g., a `URIorCURIE`), it will no longer be coerced to `str`.

**Mitigating factors:** The field is typed as `Optional[str]`, users virtually always pass strings, and the `sfx()` fallback always produces a string. This is an edge case.

**Risk:** Low-Medium. Could cause subtle type errors in rare circumstances.

---

## SMALLER WARNINGS

### 7. `NATIVE_JSONLD` and `NATIVE_RDF` extensions changed (intentional bug fix)

**File:** `linkml_files.py`

Old: `NATIVE_JSONLD` = `"context.jsonld"` (same as `JSONLD`), `NATIVE_RDF` = `"ttl"` (same as `RDF`)
New: `NATIVE_JSONLD` = `"model.context.jsonld"`, `NATIVE_RDF` = `"model.ttl"`

This is a **bug fix** — the old values were duplicates of `JSONLD`/`RDF`, making `NATIVE_*` formats useless. The new values correctly point to the `model.*` file variants which exist in the directory.

**Risk:** Low. Any code relying on `NATIVE_JSONLD`/`NATIVE_RDF` producing the same paths as `JSONLD`/`RDF` was using the buggy behavior.

### 8. `LOCAL_BASE` type changed from `pathlib.Path` to `str`

**File:** `linkml_files.py`

`LOCAL_BASE = Path(__file__).parent.resolve()` → `LOCAL_BASE = os.path.abspath(os.path.dirname(__file__))`

**Risk:** Low. The test file correctly adapts. Any external code accessing `LOCAL_BASE` and calling `.iterdir()` or other `Path` methods will break, but `LOCAL_BASE` is not part of the documented public API.

### 9. `sqlddl/meta.sql` missing new elements

**File:** `packages/linkml_runtime/src/linkml_runtime/linkml_model/sqlddl/meta.sql`

Only description text changes; no new tables/columns for `ExtraSlotsExpression`, `extra_slots`, or `allowed`. The `sqlschema/meta.sql` DOES include them (6 occurrences).

**Risk:** Low. The DDL generator may intentionally not handle inlined complex types. Worth verifying.

### 10. `array.yaml` — `Array` renamed to `NDArray`, contributors commented out

**File:** `model/schema/array.yaml`

- Class `Array` renamed to `NDArray`
- One contributor (`sneakers-the-rat`) was silently removed from the commented-out list
- All contributors commented out with `#`

**Risk:** Low for the rename (abstract class, likely not directly instantiated). The contributor removal could be a social/attribution issue.

### 11. `__all__` removed from `__init__.py`

**File:** `linkml_model/__init__.py`

The explicit `__all__` list was replaced with the PEP 484 `X as X` re-export pattern. The set of 44 exported symbols is identical.

**Risk:** Very low. `from linkml_runtime.linkml_model import *` behavior is functionally identical. No code in the codebase uses wildcard imports from this module.

### 12. JSON Schema — massive nullable type changes

**File:** `jsonschema/meta.schema.json` (+6,429 / -1,543)

Beyond the new metamodel elements, this artifact has a massive change: all optional fields now use `"type": ["string", "null"]` instead of `"type": "string"`. This is a generator-level behavioral change making optional fields explicitly nullable.

**Risk:** Low-Medium. Downstream JSON Schema validators will now accept explicit `null` values where before they only accepted omission. This is more permissive but could affect strict validators.

---

## SAFE (Green Checkmarks)

### CI / Tests
- [x] CI test (Ubuntu 3.9) — PASS
- [x] CI test (Ubuntu 3.10) — PASS
- [x] CI test (Ubuntu 3.12) — PASS
- [x] CI test (Windows 3.9) — PASS
- [x] CI Slow Tests (Ubuntu 3.9, 3.13, Windows 3.9, 3.13) — all PASS
- [x] CI quality-checks — PASS
- [x] CI codecov/project and codecov/patch — PASS
- [x] CI check-external-links — PASS
- [x] CI test-rustgen — PASS
- [x] CI test_docs — PASS
- [x] CI Docker build — PASS
- [x] CI tests still running (Ubuntu 3.11, 3.13, Windows 3.13) — no failures so far

### meta.py — Core Python metamodel
- [x] `__post_init__` signature change (`*_: list[str]` → `*_: str`, `**kwargs: dict[str, Any]` → `**kwargs: Any`) — type hint correction only, zero runtime impact
- [x] `_normalize_inlined_as_dict` → `_normalize_inlined_as_list` for `structured_aliases` (keyed=False) — **bug fix**, prevents data loss when aliases share `literal_form`
- [x] Forward reference strings removed where class is already defined earlier — verified safe for all occurrences, remaining forward refs kept where needed
- [x] `@dataclass` → `@dataclass(repr=False)` across all generated Python files (annotations.py, extensions.py, datasets.py, units.py, validation.py, meta.py) — consistency fix, defers to `YAMLRoot.__repr__()` like all other classes
- [x] `SchemaDefinition.default_prefix` reordering — now set after `id` validation, actually safer than before
- [x] `ExtraSlotsExpression` class added correctly with `allowed` and `range_expression` fields
- [x] `ClassDefinition` gains `extra_slots` (ExtraSlotsExpression) and `alias` (str) — clean additions
- [x] `array` slot moved from `SlotDefinition`-only to `SlotExpression` mixin (now available on `AnonymousSlotExpression` too) — still in `SlotDefinition._inherited_slots`
- [x] New `slots.extra_slots` and `slots.allowed` Slot definitions added correctly
- [x] Generation date updated from 2024-06-27 to 2026-02-17 across all generated files

### linkml_files.py — Path infrastructure
- [x] `FormatPath` dataclass → `_PathInfo` namedtuple — no external usages of `FormatPath` or `model_path()` found
- [x] `PathParts` NamedTuple removed — no external usages found
- [x] `__class_getitem__` removed from `_Path` — no `_Path[...]` usages found in codebase
- [x] `_build_path` returns string instead of tuple — all callers updated correctly
- [x] `_build_loc` uses string concatenation instead of `urljoin` — semantically equivalent when base ends with `/`
- [x] `_Path.items()` filter tightened from `not k.startswith("_")` to `isinstance(v, _PathInfo)` — more precise, functionally equivalent
- [x] Import simplification: removed `dataclasses`, `pathlib`, `typing.NamedTuple`, `urllib.parse` — replaced with `os`, `collections.namedtuple`

### __init__.py — Public API surface
- [x] Same 44 symbols exported — no additions, no removals
- [x] `X as X` re-export pattern — PEP 484 compliant, ruff-friendly
- [x] No downstream code uses wildcard imports from this module

### Schema changes (meta.yaml)
- [x] New `extra_slots` slot with comprehensive documentation and examples
- [x] New `allowed` slot — generic boolean, context-dependent semantics
- [x] New `extra_slots_expression` class — properly mixes in `expression`, has `allowed` + `range_expression`
- [x] `array` moved from `slot_definition` slots to `slot_expression` slots — correct for enabling array in anonymous expressions
- [x] `alias` added to `class_definition` — enables class-level aliases
- [x] New comments: "to use a URI or CURIE as the range, create a class with the URI or curie as the class_uri"
- [x] Typo fixes: "re-used" → "reused", "GENERARE" → "GENERATE"
- [x] `ifabsent` description expanded with module reference

### extensions.yaml
- [x] `simple_dict_value: true` annotation removed from `extension_value` slot — reflected consistently in all 4 snapshot files (annotations.json, extensions.json, issue_167b.yaml, simple_uri_test.jsonld)

### Generated artifacts — consistency with metamodel
- [x] `meta.graphql` — `ExtraSlotsExpression` type, `extraSlots`/`alias` on ClassDefinition present
- [x] `meta.proto` — `ExtraSlotsExpression` message, `extraSlots`/`alias` fields added to ClassDefinition
- [x] `meta.shex` — `ExtraSlotsExpression` shape and `extra_slots` property on ClassDefinition
- [x] `meta.owl.ttl` — `ExtraSlotsExpression` class and restrictions (10 occurrences)
- [x] `meta.shacl.ttl` — `ExtraSlotsExpression` shape constraints (7 occurrences)
- [x] `meta.schema.json` — `ExtraSlotsExpression` definition with proper `allowed`/`range_expression` (4 occurrences)
- [x] `sqlschema/meta.sql` — `extra_slots` in SQL schema (6 occurrences)
- [x] `meta.context.jsonld` — new entries for `allowed`, `ExtraSlotsExpression`, explicit `@id` entries
- [x] `meta.model.ttl` and `meta.ttl` (RDF) — description updates only, minimal

### Test changes
- [x] `test_linkml_files.py` — defensive `Path(LOCAL_BASE)` conversion for the `str`→`Path` type change
- [x] Snapshot `annotations.json` and `extensions.json` — `simple_dict_value` annotation correctly removed
- [x] Snapshot `issue_167b.yaml` — same annotation removal, consistent
- [x] Snapshot `simple_uri_test.jsonld` — same, consistent
- [x] Snapshot `meta.synopsis` — count deltas match: +1 class, +2 slots, all derivative counts consistent
- [x] All snapshot `.ttl`, `.owl`, `.shex`, `.shexj`, `.sqla.py` regenerated — bulk changes from generator improvements

### Specification docs
- [x] `02instances.md`, `03schemas.md`, `04derived-schemas.md`, `05validation.md` — editorial updates only
- [x] `05validation.md` removed 108 lines of content — this is the validation spec being simplified

### New files
- [x] `model/schema/extended_types.yaml` — well-documented numeric type definitions (int8-int64, uint8-uint64, float16-float64)
- [x] Excel file `excel/meta.xlsx` — binary, assumed regenerated

---

## Summary

| Category | Count |
|----------|-------|
| Critical errors | 2 (meta.json not regenerated, empty extended_types.py) |
| Strong warnings | 4 (stray output.yaml, GITHUB_IO_PATH_FOR breaking change, GITHUB_PATH_FOR URL change, default_prefix coercion removed) |
| Smaller warnings | 6 |
| Safe items verified | 60+ |

**Recommendation:** Fix the two critical errors before merging. The `json/meta.json` regeneration gap is the most impactful issue — every other generated format was properly updated except this one. The empty `extended_types.py` needs clarification or regeneration. The `output.yaml` should be removed. The `linkml_files.py` API changes are unlikely to cause real-world breakage given the "largely unused" status, but should be documented in release notes.

packages/linkml_runtime/src/linkml_runtime/linkml_model/json/meta.json
```diff
@@ -2590,7 +2590,7 @@
     {
       "name": "ifabsent",
       "definition_uri": "https://w3id.org/linkml/ifabsent",
-      "description": "function that provides a default value for the slot.\n  * [Tt]rue -- boolean True\n  * [Ff]alse -- boolean False\n  * int(value) -- integer value\n  * str(value) -- string value\n  * default_range -- schema default range\n  * bnode -- blank node identifier\n  * slot_uri -- URI for the slot\n  * class_curie -- CURIE for the containing class\n  * class_uri -- URI for the containing class",
+      "description": "function that provides a default value for the slot.  Possible values for this slot are defined in\nlinkml_runtime.utils.ifabsent_functions.default_library:\n  * [Tt]rue -- boolean True\n  * [Ff]alse -- boolean False\n  * int(value) -- integer value\n  * str(value) -- string value\n  * default_range -- schema default range\n  * bnode -- blank node identifier\n  * slot_uri -- URI for the slot\n  * class_curie -- CURIE for the containing class\n  * class_uri -- URI for the containing class",
```
amc-corey-cox: This file was not regenerated — only the `ifabsent` description changed.

I used Claude to help verify this: the shipped file has **245 slots / 39 classes**, while the test snapshot (`tests/linkml/test_base/__snapshots__/meta.json`) has **271 slots / 46 classes**. The new `ExtraSlotsExpression` class, `extra_slots` slot, and `allowed` slot are all missing here, but present in every other generated format (meta.py, meta.schema.json, meta.graphql, meta.shex, meta.shacl.ttl, meta.owl.ttl, meta.proto, sqlschema/meta.sql).

I think this just needs regeneration before merge.

Here is some additional content I was able to glean, with Claude's help:

The file hasn't been regenerated since July 2022. It's not just this PR — the entire json/ directory has been stuck at 2022-era content.  The Python files (meta.py, etc.) were selectively regenerated in 2024 for the 1.8.0rc effort, but json/meta.json was left behind.

 There's no build infrastructure to regenerate it. The gen-project tool's GEN_MAP doesn't include a "json" entry — it only has "jsonld". These files were historically generated in the external linkml-model repo and vendored in via cp -pr. There's no Makefile target or CI  job that regenerates json/meta.json specifically.

 The shipped file and test snapshot use different serialization options. Both are JSON-LD from JSONLDGenerator, but:
  - The test snapshot uses context_kwargs={"model": True} (inline @context, ~1000 extra lines)
  - The shipped file uses external URL references only

output.yaml
```diff
@@ -7,3 +7,4 @@ prefixes:
   xsd: http://www.w3.org/2001/XMLSchema#
 default_prefix: equipment_schema
 default_range: string
+metamodel_version: 1.7.0
```
amc-corey-cox, line 10: This is a debug/test artifact (`EquipmentSchema`) that already existed on main. We should probably remove it from main altogether but it isn't blocking anything.

amc-corey-cox (changes requested):
@matentzn Thanks for this thorough review. The adversarial review was also quite helpful. I did skip over many of the files because their changes appear trivial (curie changes). However, there are three items that stand out to me. One significantly and two likely unimportant or trivial. The most important is that it appears meta.json was not actually regenerated, or if it was something is inconsistent with __snapshots__/meta.json. I think this does need to be resolved, see the in-line comment. There is also that extended_types.py appears to be empty. This appears to be intentional but is worth a second look. Finally, output.yaml was updated but that file should probably just be removed. The second two issues don't need to be corrected but the meta.json needs to be investigated. I have additional content in the in-line comment there.

packages/linkml_runtime/src/linkml_runtime/linkml_model/json/meta.json
matentzn: Would you say this file is conceptually redundant with

- packages/linkml_runtime/src/linkml_runtime/linkml_model/jsonld/meta.jsonld or
- packages/linkml_runtime/src/linkml_runtime/linkml_model/jsonld/meta.schema.json

and should therefore be `git rm`'ed?

output.yaml
matentzn: Agreed: https://github.com/linkml/linkml/issues/3173

packages/linkml_runtime/src/linkml_runtime/linkml_model/extended_types.py
matentzn, line 1: Should this empty file be commited? Is it correct that this file is empty? The yaml file is not empty (and also [part of this PR](https://github.com/linkml/linkml/pull/3163/changes#diff-04c3b63fa04ec39da981c3968eb616c611b4514c99b1ced608d9b13a2c1f46a8))

packages/linkml_runtime/src/linkml_runtime/linkml_model/json/meta.json
amc-corey-cox: Yes, that sounds correct to me @matentzn

packages/linkml_runtime/src/linkml_runtime/linkml_model/extended_types.py
amc-corey-cox: I dug a little more into this... It looks like pythongen crashes trying to create this file.
```
ValueError: type "any_number" must declare a type base or parent (typeof)
```
I think this file isn't supposed to be empty and that there is something wrong in the schema that bears further examination.  I think we should not commit the empty file and create an issue about schemaloader crashing on any_of in this case. I'll create that issue if you want.

amc-corey-cox: Oh, and nothing seems to need this file right now so we can go ahead with this PR without it.

amc-corey-cox:
Regarding the empty `extended_types.py`: `gen-python` crashes on `extended_types.yaml` because the `any_number` union type lacks `typeof`/`base`, which `SchemaLoader` currently requires. This is tracked in #3129.

Suggest dropping the empty `extended_types.py` from this PR — nothing imports it and it masks the generator failure. Once #3129 is fixed, it can be regenerated properly.

packages/linkml_runtime/src/linkml_runtime/linkml_model/json/meta.json
matentzn: This has now been done

packages/linkml/src/linkml/__init__.py
matentzn, line 1: @amc-corey-cox please check this is acceptable. This change is new.

packages/linkml_runtime/src/linkml_runtime/linkml_model/linkml_files.py
```diff
@@ -58,50 +56,40 @@ class Format(_AutoName):
...
-    SHEXC = FormatPath("shex", "shex")
-    SHEXJ = FormatPath("shex", "shexj")
-    SQLDDL = FormatPath("sqlddl", "sql")
-    SQLSCHEMA = FormatPath("sqlschema", "sql")
-    YAML = FormatPath((Path("model") / "schema").as_posix(), "yaml")
+    """LinkML Relative paths — maps each Format to its directory and file extension."""
+
+    EXCEL = _PathInfo("excel", "xlsx")
+    GRAPHQL = _PathInfo("graphql", "graphql")
+    JSONLD = _PathInfo("jsonld", "context.jsonld")
```
matentzn, line 67: Note this commit (you cant see it in the diff):

https://github.com/linkml/linkml/commit/295eba1ea1eb093696cc26e71ac035efc49e3745

JSON is entirely removed. @amc-corey-cox

packages/linkml_runtime/src/linkml_runtime/linkml_model/linkml_files.py
```diff
@@ -241,10 +221,6 @@ def __repr__(self):
     def graphql(self) -> ModelLoc:
         return ModelFile.ModelLoc(self._model, Format.GRAPHQL)
 
-    @property
-    def json(self) -> ModelLoc:
-        return ModelFile.ModelLoc(self._model, Format.JSON)
```
matentzn, line 246: https://github.com/linkml/linkml/commit/295eba1ea1eb093696cc26e71ac035efc49e3745

matentzn (commented):
@amc-corey-cox left some commends for you to resolve

amc-corey-cox:
Here is my thought on the final json/ removal. Yes, removing it from the package might break some things downstream for people. However, if it does they have been relying on a json model that was generated around 4 years ago and known issues. I think they will appreciate being moved to the correct modern jsonschema, at least I hope so anyway.

PR #3164, opened 2026-02-11
Skip duplicate key tracking in _normalize_inlined when keyed=False

### LINKED ISSUE #3161: `_normalize_inlined` crashes on unhashable key types
`_normalize_inlined` in `yamlutils.py` unconditionally adds the key to a `cooked_keys` set for duplicate detection. But the set is only checked when `keyed=True`. When the key field resolves to a complex object (e.g. `TypeObj`, a dict-like), `set.add()` raises `TypeError: unhashable type`.

### To reproduce

```python
# Schema with inlined list where key field range is a complex type
yaml_loader.loads(data_str, target_class=Container)
```

```
TypeError: unhashable type: 'TypeObj'
```

at `yamlutils.py:139`.

### Expected

When `keyed=False`, duplicate key detection is irrelevant — keys don't need to be tracked.

matentzn:
Fixes #3161

- Guard `cooked_keys.add(key)` with `if keyed:` — duplicate detection only matters for keyed slots

## Changed files
- `packages/linkml_runtime/src/linkml_runtime/utils/yamlutils.py` — `order_up()` inner function

## Test plan
- [x] `test_linkml_issue_463::test_inlined` passes — TypeObj keys no longer crash (this has to be be tested against #3163)
- [x] Full suite passes — keyed slots still get duplicate detection

### DIFF
--- a/packages/linkml_runtime/src/linkml_runtime/utils/yamlutils.py
+++ b/packages/linkml_runtime/src/linkml_runtime/utils/yamlutils.py
@@ -134,9 +134,10 @@ class YAMLRoot(JsonObj):
                     f"Slot: {loc(slot_name)} - attribute {loc(key_name)} "
                     f"value ({loc(cooked_entry[key_name])}) does not match key ({loc(key)})"
                 )
-            if keyed and key in cooked_keys:
-                raise ValueError(f"{loc(key)}: duplicate key")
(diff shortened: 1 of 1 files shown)


matentzn:
Now subsumed by @amc-corey-cox #3165

PR #3165, opened 2026-02-12
Fix pythongen skipping _normalize_inlined for inlined_as_list slots

### LINKED ISSUE #3160: pythongen skips `_normalize_inlined` for `inlined_as_list` slots without identifier

amc-corey-cox:
## Summary

Fixes #3160.

- **pythongen.py**: Remove the `not slot.inlined_as_list` guard that prevented the fallback required-field search from running for `inlined_as_list` slots. Also skip required fields whose range is a class (unhashable as dict keys), preserving the fix for #463.
- **yamlutils.py**: Only track keys in `cooked_keys` set when `keyed=True`, since the set is only used for duplicate detection in that path. Prevents `TypeError: unhashable type` for complex key types when `keyed=False`.

Without this fix, `inlined_as_list` slots with no identifier but a required simple-range field got a naive list comprehension instead of `_normalize_inlined_as_list()`, which meant dict-form YAML input was not handled.

## Test plan

- [x] `test_issue_3160.py` — 8 tests covering:
  - codegen emits `_normalize_inlined_as_list` for simple required field
  - list-form and dict-form YAML input both load correctly
  - duplicate key values allowed when `keyed=False`
  - no required fields → falls back to naive comprehension
  - class-range required field → falls back to naive comprehension (no unhashable error)
- [x] `test_linkml_issue_463.py` — regression test still passes
- [x] `test_inlined_compliance.py` — 752 passed
- [x] `test_pythongen.py` — 6 passed
- [x] `ruff check` — clean


matentzn:
Note I had fixed all the issues already, but doing it again is probably a good idea. Maybe compare to the solution I ended up with: https://github.com/linkml/linkml/pull/3162

amc-corey-cox:
@matentzn sorry I missed that. I'll check your solution too and see if there is anything I missed.

matentzn:
@amc-corey-cox I am very worried about shifting the mental load of this PR back to me.. I have spend many many hours to figure out the minimal fix needed for making re-integrateding the LinkML model possible. You have tried to fix the problem again, and by that created a bit of uncertainty - I knew my fix worked so I feel a bit worried about this change here. Look for example at this: 

https://github.com/linkml/linkml/pull/3164

This is a separate PR now again duplicated by your PR. 

Just to be clear - maybe your changes are MUCH better than what I had created. Its not only quite possible, its likely. But I cannot review this PR as I dont personally understand anything that is happening in it (or at least not much), so my hope was the reverse - I provide the fix, and you tell me whether the fix and the reasoning is good. 

I am not opposed to continue on this PR but I wanted to let you know this so you can make a call. If you as the senior developer tells me you are _certain_ that your changes address the issue without breaking anything else (as much as one can be certain, say 96%), we can close https://github.com/linkml/linkml/pull/3164 as well and have this PR here be the one subsuming both issues.

tests/linkml/test_issues/__snapshots__/issue_84.py
matentzn, line 1: same as what was in https://github.com/linkml/linkml/pull/3162/changes so ok!

tests/linkml/test_scripts/__snapshots__/genpython/meta.py
matentzn, line 1: same as what was in https://github.com/linkml/linkml/pull/3162/changes so ok!

tests/linkml/test_biolink_model/__snapshots__/biolink.py
matentzn, line 1: This test change for example was not needed for the _minimal_ fix made in https://github.com/linkml/linkml/pull/3162/changes

packages/linkml_runtime/src/linkml_runtime/utils/yamlutils.py
matentzn, line 1: Subsumes https://github.com/linkml/linkml/pull/3164! So if we merge this, close that.

tests/linkml/test_base/__snapshots__/meta.py
matentzn, line 1: You have caught MANY more cases here then I did in tests/linkml/test_scripts/__snapshots__/genpython/meta.py, do you understand why?

matentzn: (it looks right, just checking)

packages/linkml/src/linkml/generators/pythongen.py
```diff
@@ -940,19 +939,14 @@ def gen_postinit(self, cls: ClassDefinition, slot: SlotDefinition) -> Optional[s
...
-            # If we don't have an identifier, and we are expecting to be inlined first class elements
-            # (inlined_as_list is not True), we will use the first required field as the key.
-            #  Note that this may not always work, but the workaround is straight forward -- set inlined_as_list to
-            #  True
-            if not identifier and not slot.inlined_as_list:
+            if not identifier:
                 for range_slot_name in slot_range_cls.slots:
                     range_slot = self.schema.slots[range_slot_name]
-                    if range_slot.required:
+                    if range_slot.required and range_slot.range not in self.schema.classes:
```
matentzn, line 946: Why "`and range_slot.range not in self.schema.classes`" - I did not have that before. Maybe just add a note to the related issue so we have it documented

matentzn:
> @amc-corey-cox I am very worried about shifting the mental load of this PR back to me.. 

Scratch that - I think your PR looks pareto optimal over my initial suggestions:

1. You add two reasonable tests
2. You seem to "fix" some more cases in meta.py
3. In the end functionally there is only one bigger question, which you can see in the comment. 

Lets proceed with this!

amc-corey-cox:
@matentzn I apologize for that. I simply hadn't noticed the other PR when I started this one.  I'll make sure to look out for that next time. I would have been happy to switch back to your PR but I noticed it was failing CI, at least when I looked at it, and it was because of a few of these issue-based tests that I had already solved in making sure my CI wasn't failing. I thought it better to not have two authors on a PR so wanted to stick with this one.

With that said, I should have just given that review on your PR and I would have done so if I had noticed it in the first place. I'll look more closely on issues where I'm assigned with someone else to make sure I don't duplicate work and increase cognitive load like this.

packages/linkml/src/linkml/generators/pythongen.py
amc-corey-cox: This guard is needed to avoid a regression with issue #463. The test schema there has Contained with inlined_as_list=True where the  first required field (type) has range TypeObj — a class. Without this check, that field gets picked as key_name and runtime crashes with TypeError: unhashable type.

The yamlutils fix from #3164 (which we subsumed) prevents the crash, but it's better to just skip class-range fields here — you can't  meaningfully key a list by a complex object for dict-form normalization. Those slots correctly fall through to the naive list comprehension instead.

matentzn: Ok this _sounds_ good; I will let @kevinschaper have the last word on this PR. 

Thanks for the explanations and the swift implementation!

PR #3169, opened 2026-02-13
Update metamodel files

matentzn:


tests/linkml/test_base/__snapshots__/meta.py
```diff
@@ -620,8 +620,8 @@ def __post_init__(self, *_: str, **kwargs: Any):
             self.default_curi_maps = [self.default_curi_maps] if self.default_curi_maps is not None else []
         self.default_curi_maps = [v if isinstance(v, str) else str(v) for v in self.default_curi_maps]
 
-        if self.default_prefix is not None and not isinstance(self.default_prefix, str):
-            self.default_prefix = str(self.default_prefix)
+        if self.default_prefix is None:
+            self.default_prefix = sfx(str(self.id))
```
amc-corey-cox, line 624: Behavioral change: previously `default_prefix=None` was valid and preserved. Now it auto-defaults to `sfx(self.id)`. Any downstream code that checks `if schema.default_prefix is None` will never hit that branch anymore. This could be a little bigger deal so it's worth calling out in release notes.

tests/linkml/test_base/__snapshots__/meta.py
```diff
@@ -2043,6 +2043,7 @@ class SlotExpression(Expression):
     exactly_one_of: Optional[Union[Union[dict, "AnonymousSlotExpression"], list[Union[dict, "AnonymousSlotExpression"]]]] = empty_list()
     any_of: Optional[Union[Union[dict, "AnonymousSlotExpression"], list[Union[dict, "AnonymousSlotExpression"]]]] = empty_list()
     all_of: Optional[Union[Union[dict, "AnonymousSlotExpression"], list[Union[dict, "AnonymousSlotExpression"]]]] = empty_list()
+    array: Optional[Union[dict, "ArrayExpression"]] = None
```
amc-corey-cox, line 2046: `array` moved from `SlotDefinition`-only to `SlotExpression` (and thus inherited by `AnonymousSlotExpression` and `SlotDefinition`). Functionally correct — slot expressions are the right place for constraints — but code that introspects `SlotDefinition.__dataclass_fields__` expecting `array` to be declared directly there may need updating.

tests/linkml/test_base/__snapshots__/meta.py
```diff
@@ -2683,6 +2691,8 @@ class ClassDefinition(Definition):
     represents_relationship: Optional[Union[bool, Bool]] = None
     disjoint_with: Optional[Union[Union[str, ClassDefinitionName], list[Union[str, ClassDefinitionName]]]] = empty_list()
     children_are_mutually_disjoint: Optional[Union[bool, Bool]] = None
+    extra_slots: Optional[Union[dict, "ExtraSlotsExpression"]] = None
+    alias: Optional[str] = None
```
amc-corey-cox, line 2695: New: classes can now have aliases (previously only slots could). This adds a new field to `ClassDefinition` that downstream schema processors may not expect.

tests/linkml/test_base/__snapshots__/meta.json
```diff
@@ -5382,13 +5465,6 @@
...
       "name": "extension_value",
       "definition_uri": "https://w3id.org/linkml/extension_value",
-      "annotations": [
-        {
-          "tag": "simple_dict_value",
-          "value": true,
-          "@type": "Annotation"
-        }
-      ],
       "description": "the actual annotation",
```
amc-corey-cox, line 5468: The `simple_dict_value` annotation was removed from `extension_value`. Anything downstream that reads this annotation to decide how to serialize extensions will silently stop doing but I think that is something that should be fixed downstream if it exists.

output.yaml
```diff
@@ -7,3 +7,4 @@ prefixes:
   xsd: http://www.w3.org/2001/XMLSchema#
 default_prefix: equipment_schema
 default_range: string
+metamodel_version: 1.7.0
```
amc-corey-cox, line 10: This file looks like a stale test artifact (`EquipmentSchema`). Should it be in `.gitignore`?

PR #3172, opened 2026-02-17
Enable python 3.14

### LINKED ISSUE #3154: Add support for Python 3.14
Python 3.14 has been the official [current version for 4 months](https://devguide.python.org/versions/) by now. Therefore it should be added to the list of supported versions.

melonora:
closes #3154 

Pretty much as the title says. Changes to LinkML are pretty minimal other than changing toml files, adding 3.14 to CI and fixing some notebooks. Main updates are in dependencies for which I have taken over maintenance and fixed any issues when enabling 3.14 in those libraries.

Note:
Build time for --dev on 3.14 is currently long due to the morph-kgc dependency requiring duckdb 1.4.1 which takes a very long time to build. Officially, morph-kgc does not support 3.14 yet and the version they use of duckdb is outdated. I will check whether I can contribute by enabling 3.14 officially and updating dependencies to reduce build time (1.4.1 slow build is apparently a known issue).

Edit Note:
It was possible to just add duckdb as a dependency with a lowerbound. I set this to 1.5.2 which avoids the slow build in python 3.14 but also still works with python 3.10.


melonora:
All the tests that fail are due to an outdated library being used, namely `pyjsg`. In particular, prior to Python 3.14, the `is_union` function from this library would work (it is called from within the library itself). However, in Python 3.14, the following won't work:

```python
def is_union(etype) -> bool:
    """ Determine whether etype is a Union """
    return getattr(etype, '__origin__', None) is not None and \
           getattr(etype.__origin__, '_name', None) and\
           etype.__origin__._name == 'Union'
 ```
 This is because the private attribute is `__name__` and not `_name`. This library hasn't received any updates in 4 years. The latest Python version that the library supports is Python 3.9.

melonora:
@cmungall I can attempt getting `pyjsg` up to date, but I think if we are going to do that we would need to also bring the library under the linkml organization and we need a transfer of the project on pypi. Do you think that is possible?

cc. @matentzn

matentzn:
> @cmungall I can attempt getting pyjsg up to date, but I think if we are going to do that we would need to also bring the library under the linkml organization and we need a transfer of the project on pypi. Do you think that is possible?

I would prefer if we remove that dependency. Could you try and make a separate PR based on the current setup (w/o th 3.14 concerns) to propose what would need to happen to actually get rid of that lib as a dependency entirely?

melonora:
@matentzn, then this would probably be equivalent to deciding whether LinkML will continue to support ShEx. The reason is that `pyjsg` is not used directly in LinkML itself, but in `ShExJSG`, which in turn is used by LinkML for the ShEx generator.

matentzn:
I see now, https://github.com/hsolbrig/pyjsg is actually developed by @hsolbrig; I added this to the dev call agenda.

You also have not yet accepted my invite to the dev team?

melonora:
@matentzn I think I am already. I can also see the projects etc. Is it somewhere stated that I am not? I will be there for the developer call today

amc-corey-cox (approved):
There is some notebook weirdness we should clean up separately but this looks good to me.

PR #3174, opened 2026-02-18
Improvements and Bug Fixes for Golang Code Generator

michaeladler:
Hi all,

this is my first contribution to this project. I have a series of improvements and bug fixes to the Golang code generator which I would like to share with upstream. Each change is in a separate commit, accompanied by a corresponding test. I can squash the commits if preferred, just let me know. Looking forward to your feedback.

Below is a summary of the individual commits:

- Fixes mapping of the Decimal type to float64, aligning with the Rust code generator.
- Resolves missing time imports in generated code by addressing Jinja2 template scoping issues.
- Uses omitempty for optional fields in generated structs.
- Sorts classes for deterministic code generation output.
- Ensures fields inherited from mixins are not duplicated.
- Uses arrays for multivalued primitive slots.
- Replaces incorrect usage of time.Date with time.Time for XSDDate fields.
- Uses pointers for optional slots to correctly represent optionality in Go.
- Adds missing type aliases for class references.
- Fixes a syntax error in array declarations.

Issue #2943, opened 2025-10-15
Inconsistencies with inlining in SchemaView and SchemaLoader and between them


Silvanoc:
**Describe the bug**
Both SchemaLoader and SchemaView have inconsistencies in the way they handle inlining. See Sub-issues for more details.

**Version of LinkML you are using** v1.9.2

**Please provide a schema (and if applicable, a data file) that replicates the issue**
See Sub-issues for details

Issue #2451, opened 2024-12-13
[meta] "slot" attribute "required" must be "true" if it is an "identifier" or a "key"


Silvanoc:
**Describe the bug**

As of now the [meta](https://github.com/linkml/linkml-runtime/blob/ad74966db72be214a38de19f361df88489286e83/linkml_runtime/linkml_model/meta.py) module does not ensure that a `key` or `identifier` slot is `required`.

When [setting the `identifier`](https://github.com/linkml/linkml-runtime/blob/ad74966db72be214a38de19f361df88489286e83/linkml_runtime/linkml_model/meta.py#L2348) attribute, the `required` attribute should be also set.

Also when [setting the `key`](https://github.com/linkml/linkml-runtime/blob/ad74966db72be214a38de19f361df88489286e83/linkml_runtime/linkml_model/meta.py#L2345) attribute, the `required` attribute should be also set.

**Version of LinkML you are using** v1.8.5

Silvanoc:
This issue is related to #2448. Since `identifier`s and `key`s are not being identified as `required`, the JSON-Schema generator is considering that `identifier`s and `key`s can be `null`.

vincentkelleher:
From a developer's perspective it seems logical to enforce that identifiers/keys mustn't be empty 👍 
I tried thinking of a case where this would be possible and I cannot find one.

Then comes the question of should it be done at a `linkml-runtime` level or at a `linkml` level.
To me this would belong to `linkml-runtime` in the `SchemaView`.
As I was having a look at the `SchemaView` class I found [this line](https://github.com/linkml/linkml-runtime/blob/ad74966db72be214a38de19f361df88489286e83/linkml_runtime/utils/schemaview.py#L1412) which seems to make the identifier required (not the key though) 🤔 

I see you added a link to the Python generator in your last comment, is that the reason you want to fix this ?
If so, doesn't the Python generator use `SchemaView` ?

ticapix:
It's a bit funny because I have the exact opposite behavior as described in the first post and I wanted to find a way to make identifier `required: False` by default, meaning required is already at true today for identifier. (I haven't check with keys)

My need is to have a class that can be used as both an identified element and with a blank node, like described here https://github.com/orgs/linkml/discussions/2472

Silvanoc:
Added discussion #2487 to first clarify if my expectations are wrong (in which case I would close this issue) or right (in which case this issue remains as a bug).

[...]

PR #3175, opened 2026-02-18
feature(golanggen): Re-implementation of Golang Generator

speedfl:
Hello

Here is a complete rebuilt of Golang generator. I validated it on a complex model and it works. It provides almost all features than pythongen / pydanticgen offers.

This includes:

- moving golangen logic in its own directory (like pydantic)
- implementation of `is_a` using flag `json:",inline"` flag
- supports following types:
```
    "str": "string",
    "string": "string",
    "int": "int",
    "integer": "int",
    "Bool": "bool",
    "boolean": "bool",
    "float": "float64",
    "double": "float64",
    "XSDDate": "time.Time",
    "XSDDateTime": "time.Time",
    "date": "time.Time",
    "datetime": "time.Time",
    "time": "time.Time",
    "uri": "string",
    "uriorcurie": "string",
    "ncname": "string",
```
- add support for template dir override (that allows including extra template)
- add root_struct_names that allows to identify a linkml root class vs a class that is used as a slot

Don't hesitate if you have remarks.

I can also add a parameter --v2 so that it is backward compatible.

Geoffrey

### DIFF
--- a/tests/linkml/test_generators/test_golanggen.py
+++ b/tests/linkml/test_generators/test_golanggen.py
@@ -5,6 +5,8 @@ Tests the golanggen package implementation based on pydanticgen architecture.
 Schemas are defined inline to keep tests self-contained.
 """
 
+import re
+
 import pytest
 
 from linkml.generators.golanggen import GolangGenerator
@@ -1392,3 +1394,79 @@ def test_slot_is_a_description_on_type():
     assert "// A base signature slot\ntype Signature string" in code
     # Score has description "A base score slot"
     assert "// A base score slot\ntype Score int" in code
+
+
+def test_kitchen_sink(kitchen_sink_path):
+    """golang"""
+    code = GolangGenerator(kitchen_sink_path, mergeimports=True, alphabetical_sort=True).serialize()
+
+    def assert_in(s: str) -> None:
+        assert s.replace(" ", "") in code.replace(" ", "").replace("\t", "")
+
+    assert "package kitchen" in code
+    assert_in("type Person struct {")
+    assert_in("HasFamilialRelationships []FamilialRelationship")
+    assert_in("CodeSystems []CodeSystem")
+    assert_in("type ActivityId string")
+    assert_in("Id string")
+    assert_in("WasInformedBy *ActivityId")
+    assert_in("StartedAtTime *time.Time")
+    assert_in("Aliases []string")
+    assert_in("""
+type Place struct {
+    // Embedded types for inheritance
+    HasAliases `json:",inline"`
+    Id string `json:"id"`
+    Name *string `json:"name,omitempty"`
+}
+""")
+    assert_in("""
+import (
+    "time"
+)
+""")
+    assert_in("Altitude *float64")
+
+    # for reproducible codegen, we check that the generated structs are in sorted order
+    matches = re.findall(r"^type\s+(\w+)\s+struct", code, re.MULTILINE)
+    assert matches == sorted(matches)
+
+
+def test_multivalued_non_id(tmp_path):
+    schema = tmp_path / "multivalued_non_id.yaml"
+    schema.write_text(
+        """
+id: http://example.org/test_multivalued_non_id
+name: test_multivalued_non_id
+
+imports:
+  - https://w3id.org/linkml/types
+
+slots:
+  int_dict:
+    range: KeyedInt
+    multivalued: true
+
+  id:
+    range: string
+    identifier: true
+
+  value:
+    range: integer
+    required: true
+
+classes:
+  KeyedInt:
+    slots:
+      - id
+      - value
+  Test:
+    tree_root: true
+    slots:
+      - int_dict
+""",
+        encoding="utf-8",
+    )
+
+    code = GolangGenerator(schema, mergeimports=True).serialize()
+    assert 'IntDict []KeyedIntId `json:"int_dict,omitempty"`' in code


matentzn:
@speedfl thank you for your contribution.. I must be very confused by something but just a few minutes before your PR we got this PR:

https://github.com/linkml/linkml/pull/3174

Can you explain how these relate?

speedfl:
> @speedfl thank you for your contribution.. I must be very confused by something but just a few minutes before your PR we got this PR:
> 
> #3174
> 
> Can you explain how these relate?

Hello. This is totally unrelated. I was working as well on my side on getting the Golang Generator working for my case :smile:

matentzn:
> Hello. This is totally unrelated.

😱 This must be one of the biggest coincidences I have ever experienced in Open Data :D Two big refactorings of the same generator just 90 minutes apart! I will tell this to my kids :P 

Anyways, thanks. I think we will have to first merge the other one though because it came first, and it was already reviewed and approved by a community member. Just to set expectations - would you be willing to wait until that one is finalised and then rebase before proceeding?

Silvanoc:
@speedfl since you appear to be another user of the Golang generator and there are not many of them outing them, it would be good to have your opinion on PR #3174

speedfl:
> @speedfl since you appear to be another user of the Golang generator and there are not many of them outing them, it would be good to have your opinion on PR #3174

Let me check locally

michaeladler:
1. There is a problem with the `kitchen_sink.yaml` input, for which your generator produces invalid Go code:

```
// An organization.

This description
includes newlines

## Markdown headers

 * and
 * a
 * list
type Organization struct {
	// Embedded types for inheritance
	HasAliases `json:",inline"`
	Id string `json:"id"`
	Name string `json:"name,omitempty"`
}
```

2. Could you ensure that the structs are in sorted order? This would make it much easier to diff your generated code against mine (my generator generates the structs in sorted order which is also useful if you want to store the generated code in git and avoid having diffs when there are actually no changes in the model).

3. I have a test case `test_multivalued_non_id` which has the following model (based on some other test in LinkML):
```yaml
id: http://example.org/test_multivalued_non_id
name: test_multivalued_non_id

imports:

slots:
  int_dict:
    range: KeyedInt
    multivalued: true

  id:
    range: string
    identifier: true

  value:
    range: integer
    required: true

classes:
  KeyedInt:
    slots:
      - id
      - value
  Test:
    tree_root: true
    slots:
      - int_dict
```

My generator produces:

```go
type KeyedInt struct {
	Id *string `json:"id,omitempty"`
	Value int `json:"value"`
}

type KeyedIntId string


type Test struct {
	IntDict []KeyedIntId `json:"int_dict,omitempty"`
}
```

Yours:

```go
type KeyedInt struct {
	Id string `json:"id"`
	Value int `json:"value"`
}

type Test struct {
	IntDict []*KeyedInt `json:"int_dict,omitempty"`
}
```

My take on this (LinkML experts please correct me if I'm wrong, I'm not an expert):

- an identifier is not automatically required (see https://linkml.io/linkml-model/latest/docs/identifier/), so `*string` is correct
- You inlined `KeyedInt` but the `inlined` property is not set, so I'd argue my version is the correct one.

4. Most other changes are pointer vs. no pointer, we can check again once it's easier to diff the code.

Silvanoc:
> * an identifier is not automatically required (see https://linkml.io/linkml-model/latest/docs/identifier/), so `*string` is correct

@michaeladler you are right, that's the current behavior. Although [I disagree with that behavior](https://github.com/linkml/linkml/issues/2451) 😄 

> You inlined KeyedInt but the inlined property is not set, so I'd argue my version is the correct one.

In theory, the generator shouldn't care about it, since that's the task of [SchemaView's method `is_inlined`](https://linkml.io/linkml/developers/schemaview.html#linkml_runtime.utils.schemaview.SchemaView.is_inlined).

The fact that [SchemaLoader (deprecated) and SchemaView have a slightly different behavior](https://github.com/linkml/linkml/issues/2943) is not the concern of the generator, but of those core components.

speedfl:
Yop @michaeladler 

- I pushed a fix to allow alphabetical order using command line argument
- I pushed a fix to support primitives as nil (I like this idea :smile: , I just made it optional by using a command line argument
- I pushed a fix for the multiline comment
- I pushed a fix regarding inlined KeyedInt

concerning this point:

> an identifier is not automatically required (see https://linkml.io/linkml-model/latest/docs/identifier/), so *string is correct. The doc states:

According to the doc we can see 
> a key slot is automatically required. Identifiers cannot be optional

So I may be wrong but I would say it is required

michaeladler:
## Core

https://github.com/linkml/linkml/blob/main/tests/linkml/test_generators/input/core.yaml

My generator:
```go
type Activity struct {
	Id *string `json:"id,omitempty"`
	StartedAtTime *time.Time `json:"started_at_time,omitempty"`
	EndedAtTime *time.Time `json:"ended_at_time,omitempty"`
	WasInformedBy *ActivityId `json:"was_informed_by,omitempty"`
	WasAssociatedWith *AgentId `json:"was_associated_with,omitempty"`
	Used *string `json:"used,omitempty"`
	Description *string `json:"description,omitempty"`
}
```

Yours:
```go
type Activity struct {
	Id string `json:"id"`
	StartedAtTime time.Time `json:"started_at_time,omitempty"`
	EndedAtTime time.Time `json:"ended_at_time,omitempty"`
	WasInformedBy *ActivityId `json:"was_informed_by,omitempty"`
	WasAssociatedWith *AgentId `json:"was_associated_with,omitempty"`
	Used string `json:"used,omitempty"`
	Description string `json:"description,omitempty"`
}
```

My take: 

1. I think `time.Time` must be a pointer; `omitempty` doesn't work with `Time`.
2. I think it's better to have `*string` to distinguish between the string being empty or absent (this can sometimes make a difference in framing).
3. For the `Id` field I don't mind if you make it a string due to it being a key slot (it's more natural, I agree).

## Kitchen Sink

https://github.com/linkml/linkml/blob/main/tests/linkml/test_generators/input/kitchen_sink.yaml

Mine:

```go
type Person struct {
	/*
	 * parent types
	 */
	HasAliases
	Id string `json:"id"`
	Name *string `json:"name,omitempty"`
	HasEmploymentHistory []EmploymentEvent `json:"has_employment_history,omitempty"`
	HasFamilialRelationships []FamilialRelationship `json:"has_familial_relationships,omitempty"`
	HasMedicalHistory []MedicalEvent `json:"has_medical_history,omitempty"`
	/*
	 * number of years since birth
	 */
	AgeInYears *int `json:"age_in_years,omitempty"`
	Addresses []Address `json:"addresses,omitempty"`
	HasBirthEvent *BirthEvent `json:"has_birth_event,omitempty"`
	SpeciesName *string `json:"species_name,omitempty"`
	StomachCount *int `json:"stomach_count,omitempty"`
	IsLiving *string `json:"is_living,omitempty"`
}
```

Yours:

```go
type Person struct {
	// Embedded types for inheritance
	HasAliases `json:",inline"`
	Id string `json:"id"`
	Name string `json:"name,omitempty"`
	HasEmploymentHistory []EmploymentEvent `json:"has_employment_history,omitempty"`
	HasFamilialRelationships []FamilialRelationship `json:"has_familial_relationships,omitempty"`
	HasMedicalHistory []MedicalEvent `json:"has_medical_history,omitempty"`
	// number of years since birth
	AgeInYears int `json:"age_in_years,omitempty"`
	Addresses []Address `json:"addresses,omitempty"`
	HasBirthEvent *BirthEvent `json:"has_birth_event,omitempty"`
	SpeciesName string `json:"species_name,omitempty"`
	StomachCount int `json:"stomach_count,omitempty"`
	IsLiving LifeStatusEnum `json:"is_living,omitempty"`
}
```
Here I see a real issue with using `int` instead of `*int` for `StomachCount`. Being absent means we might not know or care, being `0` could mean we know it's zero. With your version, we cannot distinguish these cases. **So overall I'd recommend using pointers for optional strings (see previous example) and ints. I guess there is more affected, probably all primitives?**

speedfl:
> ## Core
> 
> https://github.com/linkml/linkml/blob/main/tests/linkml/test_generators/input/core.yaml
> 
> My generator:
> 
> ```go
> type Activity struct {
> 	Id *string `json:"id,omitempty"`
> 	StartedAtTime *time.Time `json:"started_at_time,omitempty"`
> 	EndedAtTime *time.Time `json:"ended_at_time,omitempty"`
> 	WasInformedBy *ActivityId `json:"was_informed_by,omitempty"`
> 	WasAssociatedWith *AgentId `json:"was_associated_with,omitempty"`
> 	Used *string `json:"used,omitempty"`
> 	Description *string `json:"description,omitempty"`
> }
> ```
> 
> Yours:
> 
> ```go
> type Activity struct {
> 	Id string `json:"id"`
> 	StartedAtTime time.Time `json:"started_at_time,omitempty"`
> 	EndedAtTime time.Time `json:"ended_at_time,omitempty"`
> 	WasInformedBy *ActivityId `json:"was_informed_by,omitempty"`
> 	WasAssociatedWith *AgentId `json:"was_associated_with,omitempty"`
> 	Used string `json:"used,omitempty"`
> 	Description string `json:"description,omitempty"`
> }
> ```
> 
> My take:
> 
>     1. I think `time.Time` must be a pointer; `omitempty` doesn't work with `Time`.
> 
>     2. I think it's better to have `*string` to distinguish between the string being empty or absent (this can sometimes make a difference in framing).
> 
>     3. For the `Id` field I don't mind if you make it a string due to it being a key slot (it's more natural, I agree).
> 
> 
> ## Kitchen Sink
> 
> https://github.com/linkml/linkml/blob/main/tests/linkml/test_generators/input/kitchen_sink.yaml
> 
> Mine:
> 
> ```go
> type Person struct {
> 	/*
> 	 * parent types
> 	 */
> 	HasAliases
> 	Id string `json:"id"`
> 	Name *string `json:"name,omitempty"`
> 	HasEmploymentHistory []EmploymentEvent `json:"has_employment_history,omitempty"`
> 	HasFamilialRelationships []FamilialRelationship `json:"has_familial_relationships,omitempty"`
> 	HasMedicalHistory []MedicalEvent `json:"has_medical_history,omitempty"`
> 	/*
> 	 * number of years since birth
> 	 */
> 	AgeInYears *int `json:"age_in_years,omitempty"`
> 	Addresses []Address `json:"addresses,omitempty"`
> 	HasBirthEvent *BirthEvent `json:"has_birth_event,omitempty"`
> 	SpeciesName *string `json:"species_name,omitempty"`
> 	StomachCount *int `json:"stomach_count,omitempty"`
> 	IsLiving *string `json:"is_living,omitempty"`
> }
> ```
> 
> Yours:
> 
> ```go
> type Person struct {
> 	// Embedded types for inheritance
> 	HasAliases `json:",inline"`
> 	Id string `json:"id"`
> 	Name string `json:"name,omitempty"`
> 	HasEmploymentHistory []EmploymentEvent `json:"has_employment_history,omitempty"`
> 	HasFamilialRelationships []FamilialRelationship `json:"has_familial_relationships,omitempty"`
> 	HasMedicalHistory []MedicalEvent `json:"has_medical_history,omitempty"`
> 	// number of years since birth
> 	AgeInYears int `json:"age_in_years,omitempty"`
> 	Addresses []Address `json:"addresses,omitempty"`
> 	HasBirthEvent *BirthEvent `json:"has_birth_event,omitempty"`
> 	SpeciesName string `json:"species_name,omitempty"`
> 	StomachCount int `json:"stomach_count,omitempty"`
> 	IsLiving LifeStatusEnum `json:"is_living,omitempty"`
> }
> ```
> 
> Here I see a real issue with using `int` instead of `*int` for `StomachCount`. Being absent means we might not know or care, being `0` could mean we know it's zero. With your version, we cannot distinguish these cases. **So overall I'd recommend using pointers for optional strings (see previous example) and ints. I guess there is more affected, probably all primitives?**

Hello for the nullable primitive you can use command line `--nullable-primitives`. We can set it to true by default if you want. I'll also add time.Time

michaeladler:
Thanks for the hint, I just discovered that option as well :) Yes, I think it would make sense to set it to `true` by default. However, I think `time.Time` must always be a pointer if is an optional slot because `omitempty` does not work on it:

```go
type Event struct {
	Name string    `json:"name,omitempty"`
	Time time.Time `json:"time,omitempty"`
}

func main() {
	var e Event
	data, _ := json.MarshalIndent(e, "", "  ")
	fmt.Println(string(data))
}
```

gives you

```json
{
  "time": "0001-01-01T00:00:00Z"
}
```
which is not what you'd expect.

speedfl:
@michaeladler  fixed. I also added a feature related to type safety

If you have a slot where something references it with `is_a` it will result in becoming a type
It is already the case for structs, but it could be good to have it for strings for example.

See my last commit (Add namedslot)

```yaml
id: https://example.org/slot_isa
name: slot_isa_test
default_range: string

prefixes:
imports:
  - linkml:types

classes:
  Document:
    slots:
      - signature
      - owner_signature
      - witness_signature
      - score
      - final_score

slots:
  signature:
    range: string
    description: A base signature slot
  owner_signature:
    is_a: signature
    range: string
    required: true
  witness_signature:
    is_a: signature
    range: string
  score:
    range: integer
    description: A base score slot
  final_score:
    is_a: score
    range: integer
"""
```

Here this will produce:

```golang
type Document struct {
	// A base signature slot
	Signature *Signature `json:"signature,omitempty"`
	OwnerSignature Signature `json:"owner_signature"`
	WitnessSignature *Signature `json:"witness_signature,omitempty"`
	// A base score slot
	Score *Score `json:"score,omitempty"`
	FinalScore *Score `json:"final_score,omitempty"`
}

// A base signature slot
type Signature string

// A base score slot
type Score int
```

The option is disable by default but could be used with `--named-slot-types`

michaeladler:
Looks good to me now. I think all my concerns have been addressed now :) Btw, I just realized you could use the newly introduced (since Go 1.24) `omitzero` annotation to prevent an "empty" `time.Time` from being serialized to JSON.

speedfl:
What other contributor think ? 
I am not against using omitzero if this is what your prefer. But I am wondering if we could simply be consistent with the primitives types + struct with: if optional => pointer

Silvanoc:
> What other contributor think ?

@speedfl I'm afraid that you are not going to find in this community many Go developers at the needed level to have an opinion at the depth that you've reached...

If you really see the need for another opinion, then you'll have to elaborate a bit more on the advantages and drawbacks of both approaches.

speedfl:
To be honest this hasn't a lot of impact on my side. @michaeladler would you prefer to stay consistent with other types or does the omitzero looks better for you ?
If it is better so we keep it only for time or do we add it as well for primitives ? (We could consider another cli Param)

michaeladler:
@speedfl I'm all for consistency :) I didn't mean to change the rule "optional => pointer"; instead, I meant to add `omitzero` for the other case where you don't use pointers for optional slots, with the intention being to avoid JSON serializations such as `{ "time": "0001-01-01T00:00:00Z" }` for an empty `time.Time`. (For my use-case it's not relevant though as I will use the pointers approach.)

speedfl:
That's perfect. I'll work on it this afternoon!

speedfl:
Everything is ok with the omitzero. I pushed a commit.

speedfl:
@michaeladler  @Silvanoc, as https://github.com/linkml/linkml/pull/3174 has been merged. 

What do you recommend ? Because my MR is supposed to override all the change. That's why I added everything that was requested.

michaeladler:
> Everything is ok with the omitzero. I pushed a commit.

Yep looks good, thanks!

> What do you recommend ? Because my MR is supposed to override all the change. That's why I added everything that was requested.

Just rebase your branch and throw away the existing generator while keeping the tests (which I think are already on your branch too). You have my +1 for merging your generator, it's definitely an improvement over the existing generator!

Silvanoc:
Having you both @speedfl and @michaeladler improving this contribution with your conversation is great! Thanks for your time and patience in the discussion to improve LinkML with high quality code!

@speedfl once you've fixed the conflicts and the tests are passing, I'll approve this PR and hand it over to a maintainer for merging.

PR #3187, opened 2026-02-23
fix(owlgen): implement children_are_mutually_disjoint

### LINKED ISSUE #3186: children_are_mutually_disjoint not implemented by OWL generator
## Description

The `children_are_mutually_disjoint` field on `ClassDefinition` is part of the LinkML metamodel but is silently ignored by `gen-owl` — no `owl:AllDisjointClasses` axiom is emitted when the flag is set.

## Reproducer

```yaml
classes:
  Animal:
    children_are_mutually_disjoint: true
  Dog:
    is_a: Animal
  Cat:
    is_a: Animal
```
Expected output (absent from gen-owl):

```
[] a owl:AllDisjointClasses ;
   owl:members ( :Cat :Dog ) .
```

## Fix

Add a method to OwlSchemaGenerator that reads children_are_mutually_disjoint and emits owl:AllDisjointClasses for all direct is_a children. The axiom should be suppressed when fewer than two children exist.

mgskjaeveland:
The `children_are_mutually_disjoint` flag on ClassDefinition was silently ignored by the OWL generator. When set, the immediate subclasses of a class should be stated as mutually disjoint in OWL.

Add `_add_disjoint_children()` which emits an `owl:AllDisjointClasses` axiom covering the direct `is_a` children of a class when the flag is True. The method is called for every class after the main schema elements have been processed. The axiom is suppressed when fewer than two qualifying children exist.

Fixes #3186.

### DIFF
--- a/packages/linkml/src/linkml/generators/jsonldgen.py
+++ b/packages/linkml/src/linkml/generators/jsonldgen.py
@@ -60,6 +60,9 @@ class JSONLDGenerator(Generator):
     context: str = None
     """Path to a JSONLD context file"""
 
+    metamodel_context: str = None
+    """Override for metamodel context URI/path. When None, uses METAMODEL_CONTEXT_URI."""
+
     def __post_init__(self) -> None:
         self.original_schema = deepcopy(self.schema)
         super().__post_init__()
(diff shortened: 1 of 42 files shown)


amc-corey-cox:
@mgskjaeveland sorry. There will be a bit of a delay on merging this while we resolve #3214. I'll try to get back to this quickly after we have that in.

amc-corey-cox (approved):
This looks good to me!

PR #3192, opened 2026-02-23
Fix/suppress vacuous axioms

### LINKED ISSUE #3190: Support skipping generating vacuous OWL axioms
Add support for skipping generating vacuous OWL axioms, such as: 
 - min 0 cardinality axioms
 - universally quantified restrictions that are qualified to range Thing, or to the range or the property/role

Some users may prefer these to be generated for information, so suggest to make translation customisable:

- "--skip-min-zero-cardinality-axioms"
- "--skip-vacuous-local-range-axioms"

### LINKED ISSUE #3191: Support for minimum and maximum cardinality with OWL generator
The OWL generator only supports required and multivalued for slots - and ignores maximum and minimum cardinality settings on slots.

mgskjaeveland:
Two additions to OWL Generator: Fixes #3190, Fixes #3191.

Suppress vacuous OWL axioms — two new opt-in flags:

- skip_min_zero_cardinality_axioms: drops owl:minCardinality 0 restrictions, which every individual trivially satisfies and which carry no information
- skip_vacuous_local_range_axioms: drops owl:allValuesFrom restrictions whose filler is owl:Thing or duplicates the property's global rdfs:range (already emitted by add_slot for top-level slots, making the local restriction fully entailed). Attribute slots, which have no global rdfs:range, are unaffected.
- OWL cardinality for minimum_cardinality / maximum_cardinality — slot fields are now translated into OWL cardinality restrictions; when both are equal a single owl:cardinality axiom is emitted instead of two. Cardinality node creation is centralised in a new helper _add_cardinality_restriction.

Snapshots were updated — the refactored cardinality logic changes axioms emitted for slots with explicit minimum_cardinality / maximum_cardinality.

### DIFF
--- a/.github/workflows/check-external-links.yaml
+++ b/.github/workflows/check-external-links.yaml
@@ -21,7 +21,7 @@ jobs:
           python-version: "3.12"
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v7.2.1
+        uses: astral-sh/setup-uv@v7.3.1
         with:
           version: ${{ env.UV_VERSION }}
 
--- a/.github/workflows/doc-pages.yaml
+++ b/.github/workflows/doc-pages.yaml
@@ -24,7 +24,7 @@ jobs:
           git fetch upstream --tags
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v7.2.1
+        uses: astral-sh/setup-uv@v7.3.1
         with:
           version: ${{ env.UV_VERSION }}
           enable-cache: true
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -54,7 +54,7 @@ jobs:
           password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
 
       - name: Build and push
-        uses: docker/build-push-action@v6.18.0
+        uses: docker/build-push-action@v6.19.2
         with:
           context: .
           platforms: linux/amd64,linux/arm64/v8
--- a/.github/workflows/docs-test.yaml
+++ b/.github/workflows/docs-test.yaml
@@ -29,7 +29,7 @@ jobs:
           git fetch upstream --tags
 
       - name: Install uv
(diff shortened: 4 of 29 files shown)


turbomam:
I up-voted the intent of this PR. I haven't reviewed the code.

tests/linkml/test_biolink_model/__snapshots__/biolink.owl.ttl
matentzn: These diffs are huge.. Its a little scary to merge them without some clarifications of what was changed.

mgskjaeveland:
> Question: should the long-term default for `skip_min_zero_cardinality_axioms` and `skip_vacuous_local_range_axioms` be `true`? 

My vote would be yes.

> It's hard to imagine a use case where someone intentionally wants `minCardinality 0` axioms or redundant `allValuesFrom` restrictions in their output.

One use case for "min 0" is to signal "may" or a "default" value; see https://ceur-ws.org/Vol-432/owled2008eu_submission_14.pdf. Personally, I dont like this, but others may. 

> If we agree these should eventually be the default, we could add deprecation warnings now (per our [deprecation policy](docs/maintainers/deprecation.md)) to signal that the defaults will flip in a future minor version, giving downstream users a heads-up.

Sounds good to me.

> Separately — the cardinality consolidation (emitting `owl:cardinality 1` instead of separate `minCardinality 1` + `maxCardinality 1` for required non-multivalued slots) changes generator output for existing schemas without a flag or opt-in. Our deprecation policy treats output changes as breaking changes. Should this also be behind a flag or go through a deprecation cycle, or is it better characterized as a bug fix?

Perhaps the safe option is to hide the change behind a flag that eventually can become the default?

Note that there is also a discussion in #3190. If this PR should really be in draft status, I apologise. Let me know if it should be converted to draft.

amc-corey-cox:
@mgskjaeveland  No, this PR is in the correct status. We only want PRs in draft if you are still actively working on them. Please add the flag and start the deprecation cycle and I think this will be ready to merge. Thanks for this great PR!

PR #3204, opened 2026-02-25
Add PR state labels system to make clearer in "whose court the ball" is

matentzn:
Added guidelines for PR state labels to ensure clarity on review responsibilities.

I am thinking of creating this as a strong signal of in "whose court the ball" is.


docs/maintainers/contributing.md
```diff
@@ -228,6 +228,7 @@ If the PyPI release failed, make fixes, [delete](https://docs.github.com/en/ente
...
+* PR state labels — All open, non-draft PRs must carry exactly one of the following labels at all times: `awaiting-review`, `awaiting-author`, `in-review`, or `on-hold`. These labels are the authoritative signal of who holds the ball. When a reviewer requests changes, they apply `awaiting-author`. When an author has addressed feedback, they switch it to `awaiting-review`. A PR with `awaiting-author` that receives no activity for 3 months it will be automatically marked stale; after a further 14 days it will be moved to draft and marked orphaned.
```
dalito, line 232: You could prefix these labels with "pr-" to make them easier to find/select.

docs/maintainers/contributing.md
```diff
@@ -228,6 +228,7 @@ If the PyPI release failed, make fixes, [delete](https://docs.github.com/en/ente
...
```
matentzn, line 231:
```suggestion
* PR state labels — All open, non-draft PRs must carry exactly one of the following labels at all times: `pr-awaiting-review`, `pr-awaiting-author`, `pr-in-review`, or `pr-on-hold`. These labels are the authoritative signal of who holds the ball. When a reviewer requests changes, they apply `pr-awaiting-author`. When an author has addressed feedback, they switch it to `pr-awaiting-review`. A PR with `pr-awaiting-author` that receives no activity for 3 months it will be automatically marked stale; after a further 14 days it will be moved to draft and marked orphaned.
```

Great idea from @dalito - thanks!

docs/maintainers/contributing.md
matentzn: https://github.com/linkml/linkml/pull/3204/changes#r2854990476

docs/maintainers/contributing.md
gouttegd, line 231: > When an author has addressed feedback, they switch it to `awaiting-review`.

This assumes that the author has the right to change labels on a PR, which requires the “triage” permission on the repository.

So if an external contributor (who does not have this triage access) has a open PR that has been flagged with `pr-awaiting-author`, once they have addressed the feedback the contributor will have no way of switching back the label to `pr-awaiting-review`. They will be dependent on someone with triage access to do it for them. And if nobody does it for them in 3 months, then their PR will be marked as stale without them being able to do anything about it.

amc-corey-cox: While I like this idea in general, I do think @gouttegd brings up an important point that we would have to rely on internal contributors to manually manage these labels. I'm also concerned that this adds additional friction for all contributors that touch a PR and it is easy to forget to change labels on a PR, especially as this isn't standard practice in most work-flows. I'm concerned we may end up in a state of relying on a failure-prone manual workflow.

amc-corey-cox:
@matentzn  Giving this a little more thought. I'm having a hard time understanding why we can't rely on the GitHub native workflow instead of inventing our own:
  - Review requested → reviewer's turn
  - Changes requested → author's turn
  - Draft → author's not ready
  - Approved → ready to merge

What am I missing here?

matentzn:
> What am I missing here?

FAIR point. The whole system assumes some kind of triage person. What we are trying to do is provide a clear SOP for what to do during phases of inactivity.

Maybe lets use your workflow as a scaffolding and model the cases I need on top:

1. (OPTIONAL) PR opened in draft mode → author's turn
2. PR set to ready-to-review  → triage person turn (assigning "official" reviewers)
3. Reviewer assigned → reviewer's turn
4. Review:
    a. Changes requested → author's turn
    b. Approved → triagers turn (**should be avoided**)
         - If author has merge - rights they CAN merge (but dont have to, default responsibility here is triager)
         - if author does not have merge - rights they are out of the loop -> triager has to merge
   c. Approved and merged (if the reviewer feels entitled to merge, they should) → done
5. Merge → done

So the only case I think that needs a tag is if a reviewer was provided by non-official reviewer (community member, etc) - I cant see, as triager, from the outside if the first reviewer approved but said something like "please lets also ask Corey for his opinion". So I will have to click to check (this is a false case of 4b - so cant merge). I guess this _is_ ok but it really hardens the need for a regular triager checkin. So at least here I would want a tag to not keep clicking on PRs to check why this is approved but not merged (`pr-needs-additional-review`).


So remains a solid SOP:

Whenever the ball is in the authors court, we can close after 3 months of inactivity.

Maybe that is better then what I propose here with all the tags. What do you think @amc-corey-cox?

gouttegd:
There is also another aspect that, unless I missed something, is not covered:

**Whose responsibility it is to merge a PR once it has been approved?** Is that particular ball in the court of the reviewer, of the PR’s author, or of anybody else?

If the answer is “it’s the author responsibility”, please keep in mind once again that the author of the PR may _not_ have the write permission required to do so.

matentzn:
> Whose responsibility it is to merge a PR once it has been approved? Is that particular ball in the court of the reviewer, of the PR’s author, or of anybody else?

I clarified how I see it in my comments above, good point.

matentzn:
I am closing the PR here in favour of an issue based discussion: #3248

Issue #2985, opened 2025-10-30
Warnings during tests are out of control


sneakers-the-rat:
i have [previously](https://github.com/linkml/linkml/pull/2288) done some [work](https://github.com/linkml/linkml/pull/1939) on reducing the quantity of warnings emitted during testing, and last i left it we were in a pretty good place where the warnings were actually informative and actionable, and the only ones left were from packages that were out of our control.

I am dismayed to see this sort of impressive quantity of warnings emitted during testing, i've never seen anything like it:

<img width="752" height="65" alt="Image" src="https://github.com/user-attachments/assets/65a110fe-981b-44d8-a7be-b6a608f026b7" />

i don't know about anyone else, but i personally can't cognize 178035 problems at once, and running the tests during local development involves scrolling up and down through 495 lines of warnings.

I guess most people must just never run the tests locally? but yeah, i would describe it as "significantly detrimental to developer experience"

to simulate the experience and motivate why we might want to not emit 178035 warnings, i have reproduced it below.

```
============================================================================== warnings summary ==============================================================================
linkml/generators/common/template.py:274: 12 warnings
tests/test_base/test_python.py: 56 warnings
tests/test_compliance/test_aliasing_compliance.py: 252 warnings
tests/test_compliance/test_annotation_compliance.py: 378 warnings
tests/test_compliance/test_array_compliance.py: 28 warnings
tests/test_compliance/test_boolean_slot_compliance.py: 1540 warnings
tests/test_compliance/test_core_compliance.py: 1358 warnings
tests/test_compliance/test_custom_type_compliance.py: 70 warnings
tests/test_compliance/test_designates_type_compliance.py: 98 warnings
tests/test_compliance/test_enum_compliance.py: 910 warnings
tests/test_compliance/test_ifabsent_compliance.py: 238 warnings
tests/test_compliance/test_import_compliance.py: 112 warnings
tests/test_compliance/test_inheritance_compliance.py: 658 warnings
tests/test_compliance/test_inlined_compliance.py: 490 warnings
tests/test_compliance/test_instantiates_compliance.py: 28 warnings
tests/test_compliance/test_jsonpointer_compliance.py: 14 warnings
tests/test_compliance/test_metadata_compliance.py: 70 warnings
tests/test_compliance/test_pattern_compliance.py: 196 warnings
tests/test_compliance/test_uniqueness_compliance.py: 420 warnings
tests/test_data/test_sqlite.py: 84 warnings
[...]

PR #3207, opened 2026-02-26
dont leave files open

sneakers-the-rat:
Chipping away at #2985 as part of a lineage of "make the tests behave nicely" PRs.

There are currently 10 "unclosed file" warnings in the test output (e.g. ctrl+f for "unclosed file" in https://github.com/linkml/linkml/actions/runs/22421739990/job/64920807241?pr=2940 ).

Some of these are just in the tests and probably not a big deal since these are temporary files, but there is one case in the delimited file loader where a file is left open because the `json_flattener` package leaves it open. the delimited file loader calls `json_flattener.flattener.unflatten_from_csv` which does this:

```python
if isinstance(source, str):
    instream = open(source)
else:
    instream = source
```

but then never closes `instream`. so to avoid that we mimick the check that the function does and manage the file opening/closing from within `delimited_file_loader`.

otherwise it's just making sure we close files in the tests.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,4 +1,8 @@
 {
+  "access_right": "open",
+  "communities": [
+    {"identifier": "linkml"}
+  ],
   "creators": [
     {
       "name": "Moxon, Sierra A.T.",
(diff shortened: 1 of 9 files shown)


amc-corey-cox:
@sneakers-the-rat Looks like we have some CI breakage. Can you look into it?

sneakers-the-rat:
Baffling to me, but will try and figure it out

dalito:
Mark sent a [message](https://obo-communitygroup.slack.com/archives/C06SMUB1NK1/p1772120621860629) on slack about these failures of CI jobs (also the tests on main fail now as he expected).

Issue #3216, opened 2026-02-26
Allow to generate Java code with annotations expected by the LinkML-Java runtime


gouttegd:
The [LinkML-Java runtime](https://github.com/gouttegd/linkml-java) (to be released soon) will offer helper methods to facilitate working with LinkML-defined data using LinkML-generated Java code.

To fully support LinkML features, the runtime will need some additional informations from the generated code, which it will expect to find on annotations on the classes and fields.

Those informations include, for any given class:

* the URI associated with the class (as specified in `class_uri`, in the defining schema);
* which slot (if any) is intended to hold the identifier or key for the objects of that class;
* which slot (if any) acts as the type designator slot;

and for each slot:

* the URI associated with the slot;
* the expected inlining mode (no inlining, inlined as dict, inlined as list);
* whether the slot is required, recommended, or optional.

Therefore, I would like to propose a new “template variant” (building on top of what was added in #3124) for the Java generator, which would generate code with the annotations expected by the LinkML-Java runtime.

gouttegd:
@matentzn Can you explicitly assign this to me please?

Issue #3213, opened 2026-02-26
`xsd:time` should not be mapped to Java’s `java.time.Instant`


gouttegd:
Currently, the Java generator maps the `xsd:time` base type to `java.time.Instant`.

I believe this is a wrong decision, maybe resulting from a misunderstanding of what `java.time.Instant` is supposed to represent compared to what `xsd:time` is supposed to represent.

From the Python implementation (which uses `datetime.time.fromisoformat()`) to parse a `xsd:time` type), it is clear the LinkML type is intended to store _only_ a time **without any date information**.

You cannot do that with `java.time.Instant`: that type is intended to store all the informations needed to represent a precise point in time, **including the day**. In effect, it is very similar to `java.time.ZonedDateTime`, which the Java code generator maps to `xsd:datetime`. It is _not_ suitable to represent a “date-less” time.

I believe the more correct Java type would be `java.time.LocalTime`, which **is** intended to represent a time only regardless of the day that time occurs.

gouttegd:
> I believe this is a wrong decision, maybe resulting from a misunderstanding of what java.time.Instant is supposed to represent compared to what xsd:time is supposed to represent.

Funnily enough, the person responsible for that misunderstanding is… [me](https://github.com/linkml/linkml/pull/1710). 🫣 🤦‍♂️

Issue #3177, opened 2026-02-18
Java rendering of `linkml:Any` cannot work


gouttegd:
From the [documentation](https://linkml.io/linkml/schemas/advanced.html#linkml-any-type):

> any class in the schema can take on [the role of `linkml:Any`] by being declared as `linkml:Any` using `class_uri`:
>
> ```yaml
> classes:
>   Any:
>     class_uri: linkml:Any
>
>  Person:
>    attributes:
>      id:
>      metadata:
>        range: Any
> ```

The _intention_ here is that the `metadata` slot in `Person` can hold literally _any kind of value_.

However this is rendered in Java as a `Any`-typed slot:

```java
private Any metadata;
```

where `Any` is a generated class like any other:

```java
public class Any {
}
```

This achieves the _complete opposite_ of what is intended: Java’s typing rules will enforce that only an instance of the `Any` class can ever be assigned to the `metadata` slot.

To really make `metadata` able to hold any kind of value, it should be rendered as a `Object`-typed field instead. (`Object` is the base class of Java’s entire object hierarchy; _everything_ ultimately inherits from `Object`, so a `Object`-typed variable/field can hold anything.)

gouttegd:
@kevinschaper I notice that @matentzn assigned you to this issue.

FYI, I am _already_ working on it, it will be part of a larger PR that I plan to submit soon to bring several improvements to the Java code generator.

(As a general rule, whenever I open a ticket about the Java code generator, unless I explicitly say otherwise you can assume that I plan to work on it.)

matentzn:
> (As a general rule, whenever I open a ticket about the Java code generator, unless I explicitly say otherwise you can assume that I plan to work on it.)

Got it! Reassigned!

Silvanoc:
> (As a general rule, whenever I open a ticket about the Java code generator, unless I explicitly say otherwise you can assume that I plan to work on it.)

@gouttegd instead of this rule that will be known (also remembered?) only for those reading your comment, I would recommend to directly assign you the issue on creation to explicit your intention of fixing it

gouttegd:
@Silvanoc And how should I do that without triage access to this repository, please?

Silvanoc:
> [@Silvanoc](https://github.com/Silvanoc) And how should I do that without triage access to this repository, please?

[...]

Issue #1449, opened 2023-05-12
Facilitate the generation of Visitor-type patterns in Java (and possibly other languages)


gouttegd:
**Is your feature request related to a problem? Please describe.**
I have a schema (the [KGCL schema](https://github.com/INCATools/kgcl/blob/main/src/kgcl_schema/schema/kgcl.yaml)) which contains, among other classes, a `ChangeLanguageElement` class. I would like to generate a Visitor-type design pattern in Java for that class and all its subclasses.

That is, I would like to generate Java code in such a way that:

1. all the generated classes for `ChangeLanguageElement` and derived classes (but not any other class that may be present in the schema) implement a _accept(IChangeVisitor)_ method;
2. in addition to the classes themselves, a `IChangeVisitor.java` interface is also generated, with an _visit()_ method for each subclass of `ChangeLanguageElement` (but again, not any other class).

**Describe the solution you'd like**
Unless I have missed something, `gen-java` does not currently allow to do what’s described above, even when using a custom template with `--template-file`.

For the (1), what’s missing is a way, from within the custom template, to determine whether the class being currently generated is a subclass (direct _or indirect_) of a given class. That is, I need a way to do the following:

```Jinja2
public class {{ cls.name }} {% if cls.is_a -%} extends {{ cls.is_a }} {%- endif %} {
{%- for f in cls.fields %}
    private {{ f.range }} {{ f.name }};
{%- endfor -%}

{% if cls.name == "ChangeLanguageElement" or HAS_ANCESTOR(cls, "ChangeLanguageElement") %}
    public void accept(IChangeVisitor v) {
        v.visit(this);
    }
{% endif -%}
}
```

where `HAS_ANCESTOR` is a hypothetical method that would need to be added.

For the (2), what’s missing is the possibility to specify an _extra_ template to generate a single supplementary file, _in addition_ to the files that are already generated for each class. In that template, it should be possible to iterate over all the subclasses (direct and indirect) of a given class, so that it would be possible to do the following:

```Jinja2
public interface IChangeVisitor {
    void visit(ChangeLanguageElement v);
{%- for descendant in GET_DESCENDANTS("ChangeLanguageElement") %}
    void visit({{ descendant }} v);
{%- endfor %}
}
```

**How important is this feature?** Select from the options below:
• Low - it's an enhancement but not crucial for work

[...]

PR #3217, opened 2026-02-26
Misc improvements to the Java code generator

gouttegd:
This PR groups various improvements to the Java code generator.

Specifically, it:

* fixes the use of an incorrect Java type to represent `xsd:time` (#3213);
* fixes the rendering of `linkml:Any` by translating it to Java’s `Object` class (#3177);
* adds a mechanism to easily generate a visitor interface for an arbitrary class (#1449); as part of that, it also adds a mechanism by which one can request the generation of an additional code file, that is not tied to any class or enum;
* ensures that the name of a class or a slot does not clash with a reserved Java keyword (#3254);
* adds a new template variant that produces code suitable for use with the [LinkML-Java runtime](https://github.com/gouttegd/linkml-java) (#3216).

Reviewers: Happy to break down the PR in several PRs that each addresses one issue if that is preferred. Note, however, that the PR is already broken down in issue-specific commits, which can be reviewed one at a time.

### DIFF
--- a/packages/linkml/src/linkml/generators/javagen.py
+++ b/packages/linkml/src/linkml/generators/javagen.py
@@ -24,7 +24,7 @@ TYPEMAP = {
     "xsd:dateTime": "ZonedDateTime",
     "xsd:date": "LocalDate",
     "xsd:time": "LocalTime",
-    "xsd:anyURI": "String",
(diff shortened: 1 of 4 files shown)


gouttegd:
Superseded by #3255

PR #3219, opened 2026-02-26
feat(owlgen): add covering axiom for abstract classes 

### LINKED ISSUE #3193: Support translation of abstract class to OWL
LinkML classes can be marked as abstract, which does not directly translate to OWL. However, a common(?) pattern is that the abstract class is expressed as a subclass of the union of its subclasses, hence forcing instances of the abstract class to also be instance of one of its subclasses. I suggest to support this with a flagged option, e.g., "--skip-abstract-class-as-unionof-subclasses".

Example:
```
Animal:
  abstract: true

Dog:
  is_a: Animal

Cat:
  is_a: Animal
```
translates to
```
Dog subClassOf Animal
Cat subClassOf Animal
Animal subClassOf (Dog unionOf Cat)
```

mgskjaeveland:
Fixes #3193

For every abstract class with at least one direct is_a child, emit:

    AbstractClass rdfs:subClassOf ( Child1 or Child2 or … )

This expresses the open-world covering constraint that every instance of the abstract class must also be an instance of one of its direct subclasses, a common OWL modelling pattern for abstract types.

The behaviour is on by default and can be suppressed with --skip-abstract-class-as-unionof-subclasses.


matentzn:
@mgskjaeveland can you resolve conflicts? I will review. Thanks for the contribution!

mgskjaeveland:
> @mgskjaeveland can you resolve conflicts?

Done, @matentzn.

mgskjaeveland:
Sorry, force-push reverting a spillover of my local developments to upstream.

matentzn:
@amc-corey-cox I have reviewed the principles behind this PR and I am happy - but since there is more than 3 lines of code here, I would prefer if you did the technical (code) review. Let me know if thats not ok, then I will do my best here.

amc-corey-cox:
Thanks for this — the implementation is clean and the test coverage is solid.

Two things I'd like to discuss:

**Default behavior:** Having this on by default changes the output of `gen-owl`, which our [deprecation guidelines](https://linkml.io/linkml/maintainers/deprecation.html) would normally flag as a breaking change ("Changes to defaults"). @matentzn — would OWL users generally expect this axiom to be present for abstract classes? If this is more like fixing a gap in the OWL translation than changing behavior, I think we can skip the deprecation cycle. Curious what you think.

**Flag naming:** Do you think there's a shorter flag name that still captures the intent clearly enough?

matentzn:
> Default behavior: Having this on by default changes the output of gen-owl, which our [deprecation guidelines](https://linkml.io/linkml/maintainers/deprecation.html) would normally flag as a breaking change ("Changes to defaults"). @matentzn — would OWL users generally expect this axiom to be present for abstract classes? If this is more like fixing a gap in the OWL translation than changing behavior, I think we can skip the deprecation cycle. Curious what you think.

No one would _expect_ these axioms to be there (as few people think about these kinds of things), but they do reflect the semantics of the LinkML model better than the default. So my inclination is that this should be default for that reason. If this was merely a change for "preference" I would opt for not default - but it does reflect the original intention!

> Flag naming: Do you think there's a shorter flag name that still captures the intent clearly enough?

I personally think the long form of an option can be long and descriptive; I personally am fine with it, as there were few scenarios where you would type these out on the CLI.

Issue #3208, opened 2026-02-26
DotGen `test_cli_with_directory` occupies 10% of total testing time


sneakers-the-rat:
---
name: Feature request
about: Suggest an idea for this project
title: 'ahhh um what do i put here can this test be shorter'
labels: feature
assignees: 'me!'

---

**What is your feature request?**

I guess this isn't exactly a feature request or a bug per se, but a point of improvement.

Running the tests locally with pytest-profiling takes 19 minutes (1150 seconds). Of that, 171 seconds is spent in a single test: `test_dotgen.py:test_cli_with_directory`

It just seems like the overhead from the shell call to graphviz is very expensive.

The test doesn't test much, aside from the cli call not failing and generating files: 

https://github.com/linkml/linkml/blob/15570439b45b0c90a1af836ca41d7c9ad82a7fef/tests/linkml/test_generators/test_dotgen.py#L342-L351

If instead of using kitchen sink we used a simpler schema, we could make the tests 10% faster for free.

**How important is this feature to you?** Select from the options below:
• Low - it's an enhancement but not crucial for work

**Additional context**

just lookin out for DX!

PR #3220, opened 2026-02-26
cheaper docgen test and use threadpoolexecutor to generate classes

sneakers-the-rat:
Fix: #3208 

It just takes a real long time to render graphviz diagrams, particularly when we are rendering one for every class in a schema. So in addition to swapping out the schema in the test, I also did v simple concurrency with a ThreadPoolExecutor - rather than rendering them as soon as the class is finished, save them up until the end and render them all then. since the python graphviz already calls out to a subprocess we should get actual concurrency not fake GIL concurrency, the thread is basically just waiting for the subprocess to complete.

When running the full `test_dotgen.py` module locally:

- `main`: sequential/kitchen sink: 227.8s
- sequential/organization: 64.1s
- ThreadPoolExecutor/kitchen sink: 78.5s
- ThreadPoolExecutor/organization: 34.6s


sneakers-the-rat:
Well, the coverage report is certainly wrong.

edit: nevermind, now it's correct. maybe the tests failing was evaluated as not covering the lines?

packages/linkml/src/linkml/generators/dotgen.py
```diff
@@ -96,13 +110,8 @@ def end_class(self, cls: ClassDefinition) -> None:
...
         if self.classdot:
-            self.classdot.render(
-                underscore(cls.name),
-                self.dirname,
-                view=False,
-                cleanup=True,
-                format=self.format,
-            )
+            self.classdots[underscore(cls.name)] = self.classdot
+            self.classdot = None
```
sneakers-the-rat, line 114: This feels a little janky, and not exactly sure if it is necessary, but elsewhere i believe there are checks for truthiness of classdot, and trying to avoid the implicit state of one class graph leaking into other parts of the class

amc-corey-cox: @sneakers-the-rat would you like me to hold off on this while you look for something better or do you think this is okay to merge?

sneakers-the-rat: i think it's ok, my feelings are just code smell feelings

PR #3222, opened 2026-02-26
keep environments in a classvar for panderagen

### LINKED ISSUE #3209: [perf] panderagen loads the jinja environment on every `serialize` call

Profiling the tests, the `dataframe_helper.py:check_data_pandera` occupies 187 seconds of 1310 seconds (14%).

Of this, the jinja rendering takes 116s. Just the template loading phase takes some 20 seconds just at the top level for the dataframe generator, and jinja's `get_template` takes 130s (across all uses) which include all the imports that happen within the templates themselves (which would only happen once as well if the environment was preserved)

Jinja environments are expensive to create, but they usually only need to be created once. However the panderagen re-creates the environment, and thus re-loads the templates, on every `serialize` call through its `load_template` method:

https://github.com/linkml/linkml/blob/15570439b45b0c90a1af836ca41d7c9ad82a7fef/packages/linkml/src/linkml/generators/panderagen/dataframe_generator.py#L102-L107

I haven't read the code enough to know when that `template_path` actually changes, but it looks like it can really only take two values - `"panderagen_class_based"` and `"panderagen_polars_schema"`. So rather than the 2598 calls to `load_template`, there could only be 2.

e.g. the pydantic generator loads it as a classvar - https://github.com/linkml/linkml/blob/15570439b45b0c90a1af836ca41d7c9ad82a7fef/packages/linkml/src/linkml/generators/pydanticgen/template.py#L80-L82 which is not necessarily the most efficient choice since it happens at import time, but it is an example of reusing the environment rather than re-loading. in this case we might keep a class attribute that's a dict mapping the `template_path` to loaded environments, if they haven't been loaded already -> do load, otherwise return already-loaded environment.

**Additional context**
Perf good!

sneakers-the-rat:
fix: #3209

super simple. instead of creating a new environment every time, we store environments in a classvar, which lets them be reused and not have to load the template and parse them and compile them and so on every time we invoke them. since this is invoked thousands of times e.g. in the compliance tests, should speed things up a bit with no loss of generality (since those template directories don't change, and if a custom template path is supplied, it should work the same as it would otherwise).

### DIFF
--- a/packages/linkml/src/linkml/generators/panderagen/dataframe_generator.py
+++ b/packages/linkml/src/linkml/generators/panderagen/dataframe_generator.py
@@ -7,6 +7,7 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from pathlib import PurePosixPath
 from types import ModuleType
+from typing import ClassVar
 
 from jinja2 import Environment, PackageLoader
 
@@ -40,6 +41,7 @@ class DataframeGenerator(OOCodeGenerator, ABC):
     file_extension = "py"
     java_style = False
     TYPE_MAP: dict = None
+    environments: ClassVar[dict[str, Environment]] = {}
 
     # ObjectVars
     template_file: str = None
@@ -106,8 +108,11 @@ class DataframeGenerator(OOCodeGenerator, ABC):
         """Load the template for code generation."""
         if template_path is None:
             template_path = self.default_template_path()
-        jinja_env = Environment(loader=PackageLoader("linkml.generators.panderagen", template_path))
-        return jinja_env.get_template(template_filename)
+        if template_path not in self.environments:
+            self.environments[template_path] = Environment(
+                loader=PackageLoader("linkml.generators.panderagen", template_path)
+            )
+        return self.environments[template_path].get_template(template_filename)
 
     def serialize(self, directory: str = None, rendered_module: OODocument | None = None) -> str:
         """


sneakers-the-rat:
doctor keeps telling me about this one weird trick that they hate, and it turns out the one weird trick is just profiling the code.

not earth shattering, but hey, it's free

main:
Screenshot 2026-02-26 at 2 49 00 PM

this pr (failures unrelated, same as the rest):

Screenshot 2026-02-26 at 2 49 40 PM

matentzn:
@tfliss are you happy to review this PR as it is in an area I think you are very invested in?

packages/linkml/src/linkml/generators/panderagen/dataframe_generator.py
```diff
@@ -40,6 +41,7 @@ class DataframeGenerator(OOCodeGenerator, ABC):
     file_extension = "py"
     java_style = False
     TYPE_MAP: dict = None
+    environments: ClassVar[dict[str, Environment]] = {}
```
noelmcloughlin, line 44: Reading https://typing.python.org/en/latest/spec/class-compat.html 
 This is "an explicit type, using the syntax ClassVar[<type>]" so LGTM

packages/linkml/src/linkml/generators/panderagen/dataframe_generator.py
```diff
@@ -106,8 +108,11 @@ def load_template(self, template_filename, template_path=None):
         """Load the template for code generation."""
         if template_path is None:
             template_path = self.default_template_path()
-        jinja_env = Environment(loader=PackageLoader("linkml.generators.panderagen", template_path))
-        return jinja_env.get_template(template_filename)
+        if template_path not in self.environments:
+            self.environments[template_path] = Environment(
```
noelmcloughlin, line 112: `[template_path]` may not be enough

packages/linkml/src/linkml/generators/panderagen/dataframe_generator.py
```diff
@@ -106,8 +108,11 @@ def load_template(self, template_filename, template_path=None):
...
+                loader=PackageLoader("linkml.generators.panderagen", template_path)
```
noelmcloughlin, line 113: because the package literal is hard coded here

But if base class is extended to load from different package without method override, cache would miss
```
class ArrowDataframeGenerator(DataframeGenerator):
    PACKAGE_NAME: ClassVar[str] = "linkml.generators.arrowgen"  # different package!
```

**Suggestion**

Make an explicit top level variable - so its obvious the package name is "variable"
```
    PACKAGE_NAME: ClassVar[str] = "linkml.generators.panderagen"
```
And then adjust the cache_key.
```
        cache_key = (self.PACKAGE_NAME, template_path)
        if cache_key not in self.environments:
            self.environments[cache_key] = Environment(
                loader=PackageLoader(self.PACKAGE_NAME, template_path)
            )
        return self.environments[cache_key].get_template(template_filename)
```

noelmcloughlin: A corner case - but worth considering - feel free to push back as its corner case, but needs tests

noelmcloughlin (changes requested):

packages/linkml/src/linkml/generators/panderagen/dataframe_generator.py
sneakers-the-rat: What do you mean? Enough for what

packages/linkml/src/linkml/generators/panderagen/dataframe_generator.py
noelmcloughlin: Test coverage could be expanded:

- expand tests to assert cache hit occurrs;
- thread-safety of cache;
- Jinja template edit/auto_reload regression;
- load_template/defualt_template_path

sneakers-the-rat: Its not even really a corner case, its unsupported functionality. The existing implementation also hard coded the package, this PR is a function-neutral caching of the environment. 

Moving the package name to another classvar wouldn't be the right way to provide an extension point for customizing templates, unless we are expecting subclasses of Panderagen, which we shouldn't. Making the key a tuple of a classvar and a variable passed into the method doesn't really make sense, because the classvar should not mutate, that's sort of the point of it being a classvar. So the tuple key just reduces to the method var but with extra hashing.

Pydanticgen supports customizing templates by passing a directory of overrides: https://linkml.io/linkml/generators/pydantic.html#linkml.generators.pydanticgen.PydanticGenerator.template_dir

That would be the way to support overriding the templates if that was desired. Otherwise, someone could just extend the classvar with an environment before invoking the generator. 

This PR is a focused, minimal change to improve the performance of the generator by caching its environment, not trying to add the ability to customize template loading behavior.

packages/linkml/src/linkml/generators/panderagen/dataframe_generator.py
sneakers-the-rat: Yes and this has a specific meaning for dataclasses - https://docs.python.org/3/library/dataclasses.html#class-variables

noelmcloughlin (approved):

noelmcloughlin:
Ok, thanks

PR #3224, opened 2026-02-27
Use vendored JSON-LD context in generators

### LINKED ISSUE #3223: Tests should use vendored JSON-LD context to prevent CI breakage from upstream metamodel changes
### Problem

When the upstream linkml-model changes (e.g., the `xsd:anyURI` context change from linkml-model #236/#241), CI breaks on **all open PRs** — even ones touching completely unrelated code. This happened most recently with #3214, where 6 tests failed across every branch because the live `meta.context.jsonld` served at `w3id.org` changed.

The root cause is an inconsistency in how our tests resolve metamodel artifacts:

**YAML imports** (`linkml:types`) resolve **locally** — hardcoded in `SchemaView.load_import()`:
```python
# packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py:337
default_import_map = {"linkml:": str(SCHEMA_DIRECTORY)}
```

**JSON-LD context** resolves from the **live network** — generators default to `METAMODEL_CONTEXT_URI`:
```python
# packages/linkml/src/linkml/generators/jsonldgen.py (when context is None):
context = [METAMODEL_CONTEXT_URI, add_prefixes_json["@context"]]
#           ^ "https://w3id.org/linkml/meta.context.jsonld"

# packages/linkml/src/linkml/generators/rdfgen.py (CLI default):
@click.option("--context", default=[METAMODEL_CONTEXT_URI], ...)
```

The vendored local copy **already exists** and is exported but never used by generators:
```
packages/linkml_runtime/src/linkml_runtime/linkml_model/jsonld/meta.context.jsonld
```
```python
# packages/linkml/src/linkml/__init__.py
LOCAL_METAMODEL_LDCONTEXT_FILE = linkml_files.LOCAL_PATH_FOR(Source.META, Format.JSONLD)
```

### Proposed Fix

1. **Tests should use the vendored context by default** — Tests that call `RDFGenerator` or `JSONLDGenerator` should pass `LOCAL_METAMODEL_LDCONTEXT_FILE` as the context instead of relying on the generators' default of `METAMODEL_CONTEXT_URI`. This makes non-network tests truly network-independent and deterministic.

2. **Convert network tests to upstream-drift detection** — Instead of "generate RDF with live context and compare to snapshots," the `@pytest.mark.network` tests that hit `w3id.org/linkml/` for context should verify the live upstream matches the vendored copy. When this fails, the message is clear: "upstream context has drifted from vendored copy — update vendored files."

3. **Attribute breakage correctly** — With this change, when the vendored metamodel is updated (via a dedicated PR), any real downstream breakage (like the `anyURI` → `URIRef` fix in #3221) appears on *that* PR — not on random unrelated PRs that happened to trigger CI.

**Question for discussion:** Should generators also default to the vendored context at runtime (not just in tests)? This would make generator output deterministic regardless of network availability, matching how YAML imports already resolve locally. The `--context` CLI flag would still allow users to specify any URL for explicit upstream testing.

### Context

- #3214 — the immediate breakage this would prevent
- #3221 — real fix for `anyURI` serialization, correctly identified via the upstream change but surfaced on the wrong PRs
- #3215 — snapshot-update approach (superseded by #3221)
- #2213 — broader URI serialization discussion
- @sneakers-the-rat's [observation](https://github.com/linkml/linkml/issues/3214#issuecomment-2688098781): "within linkml we try not to resolve the metamodel using the network since we have a copy of it locally"

### Impact

- CI stops breaking on unrelated PRs when upstream metamodel changes
- Real breakage is still caught — just on the right PR (the one updating vendored files)
- Upstream verification preserved via targeted network tests
- Existing `--context` CLI flag still allows explicit upstream testing

amc-corey-cox:
Fixes #3223

## Summary

- RDFGenerator now resolves `meta.context.jsonld` from the vendored copy in linkml-runtime instead of fetching from `w3id.org` at runtime. This prevents CI breakage on open PRs when upstream linkml-model changes.
- Added `metamodel_context` attribute to JSONLDGenerator so RDFGenerator can override the context path without losing model-specific prefix generation.
- Added drift detection test (`test_vendored_files_match_upstream`) that compares vendored YAML and JSONLD files against upstream `w3id.org` — runs with `--with-network` in CI to catch silent divergence.
- Removed `@pytest.mark.network` from tests that no longer need network access.

## Approach: "Load locally, reference URL"

The vendored file content is used for resolution at runtime, but `METAMODEL_CONTEXT_URI` (the `w3id.org` URL) is still emitted in JSON-LD `@context` output. No output change for users.

## Files changed

| File | Change |
|------|--------|
| `generators/jsonldgen.py` | Add `metamodel_context` attribute, use it in `end_schema()` |
| `generators/rdfgen.py` | Pass local context to JSONLDGenerator, update CLI default |
| `tests/.../test_rdfgen.py` | Remove `@pytest.mark.network` from `test_annotation_extensions` |
| `tests/.../test_rdf.py` | Remove `@pytest.mark.network` from `test_model_rdf` |
| `tests/.../test_linkml_files.py` | Add `test_vendored_files_match_upstream` drift detection |

## Test plan

- [x] `test_model_rdf` passes without network (4/4 parametrized cases)
- [x] `test_annotation_extensions` runs without network (pre-existing failure, unrelated to this change)
- [x] All `test_generators/` rdf/jsonld tests pass (46 passed)
- [x] All script tests pass (`test_gen_rdf.py`, `test_gen_jsonld.py`)
- [x] `gen-jsonld` CLI output still references `https://w3id.org/linkml/meta.context.jsonld` in `@context`
- [x] Drift detection test catches real drift (meta.yaml, types.yaml are currently out of sync)

### DIFF
--- a/tests/linkml/test_utils/__snapshots__/meta.synopsis
+++ b/tests/linkml/test_utils/__snapshots__/meta.synopsis
@@ -1,6 +1,6 @@
 
 Classes: 46
-	Referenced by: 29 classes, 235 slots, 0 types, 0 enums 
+	Referenced by: 29 classes, 236 slots, 0 types, 0 enums 
 	Root: 32
 	Leaf: 40
 	Standalone: 45
@@ -9,8 +9,8 @@ Classes: 46
 		enum_expression
 	Abstract: 5
 
-Slots: 271
-	Referenced by: 42 classes, 63 slots, 0 types, 0 enums 
+Slots: 272
+	Referenced by: 42 classes, 64 slots, 0 types, 0 enums 
 	Root: 208
 	Leaf: 243
 	Standalone: 275
@@ -53,7 +53,7 @@ Slots: 271
 			annotation: 1
 			anonymous_class_expression: 9
 			anonymous_enum_expression: 2
-			anonymous_slot_expression: 6
+			anonymous_slot_expression: 7
 			anonymous_type_expression: 4
 			array_expression: 1
 			class_definition: 8


amc-corey-cox:
## Expected CI failures

The new `test_vendored_files_match_upstream` drift detection test will fail for two cases:

- `Source.META-Format.YAML` — `meta.yaml` differs from upstream
- `Source.TYPES-Format.YAML` — `types.yaml` differs from upstream

This is **pre-existing drift**, not caused by this PR. The vendored YAML files in `linkml-runtime` are out of sync with what's currently served at `w3id.org`. The drift detection test is doing exactly what it's supposed to — surfacing this.

All JSON-LD context files (the focus of this PR) match upstream and pass.

### Options to unblock CI

1. **Sync the vendored YAML files** to match upstream in this PR (or a prerequisite PR)
2. **Scope the drift test to JSONLD only** — only test the files whose resolution behavior this PR changes
3. **`xfail` the known-drifted cases** — document them as known drift while still testing the rest

Happy to go whichever direction you prefer.

sierra-moxon (approved):
- move the JSONld generator and tests to use local copies of linkml-model components.

PR #3225, opened 2026-02-27
feat(owlgen): add --inline-mixins flag

### LINKED ISSUE #3226: Alternative translation of mixins: inline/flatten into using classes
The current OWL generator translates mixin classes either as 
 - the default: ordinary OWL classes and emits `rdfs:subClassOf` axioms, or, 
 - with the flag `--mixins-as-expressions`, as existential restrictions connected to the users of the mixin with the property `mixin`
(See https://linkml.io/linkml/generators/owl.html#tips)

In both cases, mixin LinkML classes are translated to OWL classes. 

In my use case, I am using mixins on the LinkML side as a code sharing construct, they dont have any proper semantic meaning. One of my main motivations for using LinkML is to maintain a natural ontology *and* a data model with a single (LinkML) model. Hence, I propose an alternative translation of mixins to OWL, where the mixin is *inlined* (or flattened/de-normalised). The proposal is motivated by generating a "clean" ontology, but it could also be useful for other formats. 

## Proposed feature

Add an `--inline-mixins` flag that produces a *mixin-free* OWL representation:

1. **Mixin classes are not emitted as OWL classes.** The mixins' slot restrictions are copied directly onto every class that uses the mixin, as if the slots had been defined locally.

2. **References to a mixin class (e.g. as a slot range) are replaced by a union of all concrete classes that use that mixin.** This preserves polymorphism: a slot that previously ranged over the mixin now ranges over the union of its users, so any concrete filler remains valid.

### Example

```yaml
id: https://example.org/inlining-mixins-mwe-1
name: inlining-mixins-mwe-1

classes:

  HasAliases:
    mixin: true
    attributes:
      aliases:
        multivalued: true

  Person:
    mixins:
      - HasAliases

  Organization:
    mixins:
      - HasAliases

slots:
  test_slot:
    range: HasAliases
````

Current output includes a HasAliases OWL class and Person/Organization rdfs:subClassOf HasAliases.

With mixins inlined:
- No HasAliases class is emitted.
- Both Person and Organization get the aliases slot restriction directly.
- The range of test_slot: becomes (Person or Organization).

Examples of default, `--mixins-as-expressions`, and test implementation (#3225) attached:

- [mwe-1-default.owl.ttl.txt](https://github.com/user-attachments/files/25610414/mwe-1-default.owl.ttl.txt)
- [mwe-1-expressions.owl.ttl.txt](https://github.com/user-attachments/files/25610413/mwe-1-expressions.owl.ttl.txt)
- [mwe-1-inline-mixin.owl.ttl.txt](https://github.com/user-attachments/files/25610412/mwe-1-inline-mixin.owl.ttl.txt)

This currently seems to work as intended for my use case, but I claim no complete coverage of other cases. Please test. 

## Pros and cons
- Pro: Produces ontologies without unnatural mixins
- Pro: Preserves polymorphic slot ranges without introducing artificial superclasses.
- Con: Bloats the ontology by replicating the mixin content for each of its users
- Con: The polymorphic slot ranges (union of all mixin user classes) may end up hiding the intended semantics of the LinkML model ("increased cognitive complexity"). 

## Implementation notes
- When only one class uses a mixin, the mixin naturally collapses into its single user.
- Mixin classes that have no users produce no output.
- Mixins-of-mixins (mixin inheritance) should be flattened transitively before inlining.

mgskjaeveland:
This is a draft to make code available for testing by others. A better explanation will be given if/once the feature has matured.
Please see fixes #3226 for a description, and please use #3226 for discussions.

### DIFF
--- a/packages/linkml/src/linkml/generators/owlgen.py
+++ b/packages/linkml/src/linkml/generators/owlgen.py
@@ -174,7 +174,7 @@ class OwlSchemaGenerator(Generator):
     enum_iri_separator: str = "#"
     """Separator for enum IRI. Can be overridden for example if your namespace IRI already contains a #"""
 
-    inline_mixins: bool = False
+    mixins_inlined: bool = False
     """If True, mixin classes are not emitted as OWL classes. Instead, each mixin's slot
     restrictions are inlined as ``rdfs:subClassOf`` axioms on every class that uses the
     mixin.  Any slot or property whose range is a mixin class is replaced by a union of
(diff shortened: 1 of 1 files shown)


matentzn:
thanks @mgskjaeveland you let us know if you want any feedback!

PR #3227, opened 2026-02-27
downgrade null/empty enum errors to warnings for optional slots

madanucd:
## Summary

Adds `--allow-null-for-optional-enums` flag to `linkml-validate` CLI that downgrades enum validation errors to warnings when the value is null/empty and the slot is not required. Addresses [dm-bip#233](https://github.com/linkml/dm-bip/issues/233) where schemas generated from TSV/JSON data with nullable enum columns produced spurious `None is not one of [...]` errors.

## Changes

- `validator/cli.py`: Added `--allow-null-for-optional-enums` flag and `_is_null_enum_error()` helper that extracts the slot name from the validation message and checks `required: false` + enum range via `SchemaView`
- `test_validator/test_cli.py`: 3 new tests using existing `personinfo.yaml` fixture covering flag-off (backward compat), flag-on (ERROR → WARN), and valid values unaffected

## Behavior

Without flag (default, unchanged): [ERROR] '' is not one of ['cisgender man', ...] in /gender
With --allow-null-for-optional-enums: [WARN] '' is not one of ['cisgender man', ...] in /gender

Only optional (required: false) enum slots are affected. Required enum slots still produce errors.

### DIFF
--- a/docs/ecosystem.md
+++ b/docs/ecosystem.md
@@ -51,6 +51,7 @@ from the LinkML community meeting.
 ### Data Rendering
 
  - [linkml-renderer](https://github.com/linkml/linkml-renderer) allows linkml data to be rendered as markdowm or HTML.
+ - [plugin-link](https://github.com/plugin-healthcare/plugin-link) allows linkml to be viewed interactively using Svelte Flow.
 
 ### Semantic DSL
 


amc-corey-cox:
@madanucd can you check out why this is failing CI?

packages/linkml/src/linkml/validator/cli.py
```diff
@@ -61,6 +62,46 @@ def _resolve_loaders(loader_config: Iterable[str | dict[str, dict[str, str]]]) -
     return loaders
 
 
+def _is_null_enum_error(result, sv, target_class: str | None) -> bool:
```
amc-corey-cox: This function reconstructs schema knowledge the validation plugin already had at validation time. This logic belongs in the plugin or validator, not the CLI.

packages/linkml/src/linkml/validator/cli.py
```diff
@@ -61,6 +62,46 @@ def _resolve_loaders(loader_config: Iterable[str | dict[str, dict[str, str]]]) -
...
+    This addresses issues where schema-automator generated schemas from TSV/JSON
+    data with null values in enum columns, causing spurious validation errors.
+    """
+    msg = result.message
+
+    # Only handle null/empty value enum violations
+    # Message format from JsonschemaValidationPlugin:
+    #   "'' is not one of [...] in /slot_name"
+    #   "None is not one of [...] in /slot_name"
+    is_null_value = msg.startswith("None is not") or msg.startswith("'' is not")
```
amc-corey-cox: These message strings come from jsonschema internals and can change between library versions, silently breaking this feature.

packages/linkml/src/linkml/validator/cli.py
```diff
@@ -61,6 +62,46 @@ def _resolve_loaders(loader_config: Iterable[str | dict[str, dict[str, str]]]) -
...
+    if not is_null_value:
+        return False
+
+    # Extract slot name from end of message e.g. "'' is not one of ['M', 'F'] in /sex" → "sex"
+    match = re.search(r"in /(\w+)$", msg)
```
amc-corey-cox: This only matches root-level slots — nested paths like `/persons/0/gender` won't be captured.

packages/linkml/src/linkml/validator/cli.py
```diff
@@ -61,6 +62,46 @@ def _resolve_loaders(loader_config: Iterable[str | dict[str, dict[str, str]]]) -
...
+        return False
+    slot_name = match.group(1)
+
+    try:
+        induced = sv.induced_slot(slot_name, target_class)
+        # Downgrade only if slot is not required AND has an enum range
+        if not induced.required:
+            if induced.range and induced.range in sv.all_enums():
+                return True
+    except Exception:
```
amc-corey-cox: This silently swallows errors from `induced_slot()`, masking bugs.

packages/linkml/src/linkml/validator/cli.py
```diff
@@ -125,8 +176,21 @@ def cli(
     loaders = _resolve_loaders(config.data_sources)
     validator = Validator(config.schema_path, validation_plugins=plugins, strict=exit_on_first_failure)
     severity_counter = Counter()
+
+    # Only load SchemaView if the flag is set — avoids overhead otherwise
+    if allow_null_for_optional_enums:
+        from linkml_runtime import SchemaView
```
amc-corey-cox: This should be a top-level import.

packages/linkml/src/linkml/validator/cli.py
```diff
@@ -125,8 +176,21 @@ def cli(
...
+
+        sv = SchemaView(str(config.schema_path))
```
amc-corey-cox: The `SchemaView` construction here triggers deferred schema resolution logging that writes to the CliRunner's closed stream — this is why CI breaks. Consider constructing it outside the CLI function or earlier in its lifecycle.

packages/linkml/src/linkml/validator/cli.py
```diff
@@ -125,8 +176,21 @@ def cli(
...
+    else:
+        sv = None
+
     for loader in loaders:
         for result in validator.iter_results_from_source(loader, config.target_class):
+            # If flag is set, downgrade null/empty enum errors on optional slots to warnings
+            if allow_null_for_optional_enums and result.severity == Severity.ERROR:
+                if _is_null_enum_error(result, sv, config.target_class):
+                    result.severity = Severity.WARN
```
amc-corey-cox: Mutating a Pydantic model owned by the validation framework from the CLI layer is a surprising side effect — if this logic lived in the plugin it would emit the correct severity from the start.

tests/linkml/test_validator/test_cli.py
```diff
@@ -148,3 +148,69 @@ def test_custom_loader_config(tmp_path, cli_runner, csv_data_file, json_data_fil
...
+
+
+def test_allow_null_for_optional_enums_with_flag(cli_runner, json_data_file):
+    """
+    With the flag, empty string in an optional enum slot is downgraded to WARN.
+    Uses personinfo.yaml: gender slot is optional with GenderType enum range.
+    """
+    data = [
+        {"id": "P:001", "name": "Alice", "gender": "cisgender woman"},
+        {"id": "P:002", "name": "Bob", "gender": ""},  # empty — WARN with flag
```
amc-corey-cox: The code handles both `''` and `None` message prefixes but only empty strings are tested.

madanucd:
@amc-corey-cox Thank you for the detailed review. I have tried to address all your points:

- **Logic moved to `JsonschemaValidationPlugin`** — no logic in CLI, just a thin pass-through
- **No regex parsing** — uses `best_error.instance` and `best_error.absolute_path` directly
- **Nested paths handled** — `absolute_path[-1]` works for `/persons/0/gender` too
- **Exceptions logged** — `logger.debug()` instead of silent swallow
- **No deferred import / no `SchemaView` in CLI** — uses `context.schema_view` already in the plugin, CI logging issue resolved
- **No mutation from CLI** — severity set at result creation time in `process()`
- **Both `''` and `None` tested** — `None`/JSON null passes cleanly with no error raised, so no downgrade needed for that case

Issue #3235, opened 2026-03-02
Summary generator prefixes enums in ranges with "Unknown_"


sveinugu:
---
name: Bug report
about: Create a report to help us improve
title: 'Summary generator prefixes enums in ranges with "Unknown_"'
labels: bug
assignees: ''

---

**Describe the bug**
A clear and concise description of what the bug is (including the error message that is printed, if any)

**Version of LinkML you are using**

**Please provide a schema (and if applicable, a data file) that replicates the issue**

Consider a simple schema `MySchema.yaml` with a slot defined with an enum as range:

```yaml
id: https://w3id.org/MySchema
name: MySchema
enums:
  MyEnum:
    permissible_values:
      'Yes':
        description: 'Yes'
      'No':
        description: 'No'
classes:
  MyClass:
    slots:
    - my_slot
slots:
  my_slot:
    range: MyEnum
```

Running `linkml generate summary MySchema.yaml` gives:

```tsv
Class Name	Parent Class	YAML Class Name	Description	Flags	Slot Name	YAML Slot Name	Range	Card	Slot Description	URI
MyClass		MyClass								
					my_slot		Unknown_MyEnum	0..1		https://w3id.org/MySchema/my_slot
```

Here Range is `Unknown_MyEnum` instead of `MyEnum`.

A PR fixing the issue will be provided

matentzn:
@sveinugu thank you for reporting and addressing this!

sveinugu:
> [@sveinugu](https://github.com/sveinugu) thank you for reporting and addressing this!

@matentzn Wouldn't it make more sense to assign someone with with merge permissions as reviewer instead of/in addition to myself?

matentzn:
@sveinugu this is just the current SOP! Your core team liaison is @kevinschaper and he is on your case (currently at position 6 on his list but he is prolific)

PR #3236, opened 2026-03-02
Fixes issue #3235: Summary generator prefixes enums in ranges with "Unknown_"

sveinugu:
Fixes issue #3235.


kevinschaper (approved):
thanks @sveinugu, and sorry for taking so long to get to this PR. 

I think we could handle this by supporting enums in class_or_type_name(), but bringing this PR in now doesn't prevent us from doing that in the future. 

Much appreciated!

PR #3238, opened 2026-03-02
Fix pydantic generator ifabsent=bnode producing invalid code

### LINKED ISSUE #3237: pydanticgen generates invalid code for ifabsent=bnode
## Description

The pydantic generator produces invalid Python code when a slot has `ifabsent: bnode`. The generated code contains `default=bnode()` but `bnode` is not imported in the generated module, resulting in a `NameError` at import time.

```yaml
slots:
  bnode_slot:
    ifabsent: bnode
```

Generates:
```python
bnode_slot: Optional[str] = Field(default=bnode(), ...)
```

## Root Cause

`PydanticIfAbsentProcessor` inherits from `PythonIfAbsentProcessor`, which returns the string `"bnode()"` for the `bnode` ifabsent case. The Python dataclass generator has `bnode` available via `linkml_runtime.utils.metamodelcore`, but the pydantic generator does not import it.

## Proposed Fix

Following the approach from @cmungall in #2166: special-case `bnode` in `predefined_slot_values` to use `uuid.uuid4().hex` (stdlib, no runtime dependency on linkml_runtime), and add `uuid` to pydanticgen's default imports.

## Related

- #1334 — original report of pydanticgen generating invalid code for `ifabsent=class_curie` (fixed by #2366, but `bnode` case remains)
- #2165 — "add ifabsent handling to pydanticgen" (prematurely closed as done)
- #2166 — @cmungall's PR that included this fix alongside other changes

amc-corey-cox:
## Summary

- Fixes the pydantic generator producing invalid `Field(default=bnode())` for slots with `ifabsent: bnode`. The `bnode()` function is a dataclasses helper from `linkml_runtime.utils.metamodelcore` that returns a `dataclasses.field()` — not valid in pydantic context.
- Now generates `Field(default_factory=lambda: "_:" + uuid.uuid4().hex)` which produces unique blank node IDs (e.g., `_:8e2ed3f0cde74461accfa7e222de2687`).
- Adds `default_factory` support to the pydantic template system (`PydanticAttribute.default_factory` field + conditional rendering in `attribute.py.jinja`).

Salvages the valuable piece from #2166 (by @cmungall) — the pydantic bnode fix — in a clean, minimal form.

Fixes #3237
Supersedes #2166

## Test plan

- [x] New `test_ifabsent_bnode` verifies generated code contains `default_factory`, compiles, and produces unique bnode IDs at runtime
- [x] All 350 pydantic generator tests pass
- [x] All 171 ifabsent compliance tests pass


packages/linkml/src/linkml/generators/pydanticgen/pydanticgen.py
```diff
@@ -584,6 +585,8 @@ def generate_slot(self, slot: SlotDefinition, cls: ClassDefinition) -> SlotResul
         predef = self.predefined_slot_values.get(camelcase(cls.name), {}).get(slot.name, None)
         if predef is not None:
             slot_args["predefined"] = str(predef)
+        elif slot.ifabsent == "bnode":
+            slot_args["default_factory"] = 'lambda: "_:" + uuid.uuid4().hex'
```
sneakers-the-rat: Hang on shouldnt this go in the ifabsent processor?

sneakers-the-rat: Have been trying to get all the string literal mutation out of the generator.

amc-corey-cox: Yes, it should. My bad. I'll fix that.

amc-corey-cox: Fixed. Moved bnode factory string into `PydanticIfAbsentProcessor.map_custom_default_values` and added a `FACTORY_IFABSENT_VALUES` frozenset so the generator only checks set membership to route to `default_factory` vs `predefined`. The `predefined_slot_values` property now lets all ifabsent values flow through the processor without special-casing.

amc-corey-cox:
I wanted to extract what was still left over in #2166 but I'm not sure I've done this right here. I'm going to think about it some more and try to make sure I'm fitting into @sneakers-the-rat new framework properly instead of just tacking something on. I'm converting this to draft for now while I work that out. Suggestions welcome.

packages/linkml/src/linkml/generators/pydanticgen/templates/attribute.py.jinja
```diff
@@ -1,4 +1,4 @@
-{{name}}: {{ range }} = Field(default={{ field }}
+{{name}}: {{ range }} = Field({% if default_factory %}default_factory={{ default_factory }}{% else %}default={{ field }}{% endif %}
```
sneakers-the-rat: to mimick style of surrounding lines and keep template readable, could use the whitespace stripping chars and put on newlines

packages/linkml/src/linkml/generators/pydanticgen/pydanticgen.py
```diff
@@ -583,7 +584,10 @@ def generate_slot(self, slot: SlotDefinition, cls: ClassDefinition) -> SlotResul
         slot_args["description"] = slot.description.replace('"', '\\"') if slot.description is not None else None
         predef = self.predefined_slot_values.get(camelcase(cls.name), {}).get(slot.name, None)
         if predef is not None:
-            slot_args["predefined"] = str(predef)
+            if isinstance(predef, FactoryDefault):
+                slot_args["default_factory"] = str(predef)
+            else:
+                slot_args["predefined"] = str(predef)
```
sneakers-the-rat: the name "predefined" is a bit awkward (not the fault of this PR obviously) - it is mostly for historical reasons - but we now end up with four names in slightly overlapping scopes: `ifabsent`, `predefined`, `default` and `default_factory` which is a bit awkward. `default_factory` is a name from the target (pydantic) domain, `ifabsent` is from the source (linkml) domain, `predefined` is from... neither, and so i think that's why this is reading odd to me. If we were to add `default_factory` as a separate field (as opposed to having some other attr that switches whether `predefined` to be a factory or not) then it might be worth it to rename `predefined` to be `default` at the same time? not a strong opinion, i never really liked the use of "predefined" and this is heightening that feeling.

packages/linkml/src/linkml/generators/pydanticgen/pydanticgen.py
```diff
@@ -672,7 +676,8 @@ def predefined_slot_values(self) -> dict[str, dict[str, str]]:
                         ]
                     elif slot.ifabsent is not None:
                         value = ifabsent_processor.process_slot(slot, class_def)
-                        slot_values[camelcase(class_def.name)][slot.name] = value
+                        if value is not None:
+                            slot_values[camelcase(class_def.name)][slot.name] = value
```
sneakers-the-rat: should end up being the same, i remember writing this this way to minimize levels of nesting and if/else checks in the generator and delegating interpreting the input values to the models. more of an informational comment than a request for change.

packages/linkml/src/linkml/generators/pydanticgen/template.py
```diff
@@ -197,6 +197,8 @@ class PydanticAttribute(PydanticTemplateModel):
     key: bool = False
     predefined: str | None = None
     """Fixed string to use in body of field"""
+    default_factory: str | None = None
+    """Factory callable string to use as default_factory in Field"""
```
sneakers-the-rat: if we add additional field, we should add an `after`-style model validator that asserts that only one of them is non-None.

sneakers-the-rat (commented):
i think the structure of this is fine, using a special string subclass in the ifabsent processor is also fine to me, although it should probably be in the parent `generators/python/python_ifabsent_processor.py` since this would also be a mechanism that the dataclass generator should have access to, since dataclass fields have the same default/default_factory situation.

I guess i'm not entirely sure if we should be doing this though - like a bnode is given a name (skolemized?) in the context of its graph, and the whole challenge with giving bnodes unique identifiers is that in order to be stable/useful, they have to be generated using the context of their connected edges. I am not sure when we would want to fill in a random bnode-looking-string in a pydantic version of a schema because i'm not exactly sure how the pydantic models are used alongside rdf generation - is there some example of the problem this change is supposed to address? I see the linked issue is just that the code that's currently generated is invalid (creates some undefined `bnode()` call), so that obviously should be fixed, but it's not clear to me that the fix should be generating a random bnode-like-string rather than the default being `None`

packages/linkml/src/linkml/generators/pydanticgen/templates/attribute.py.jinja
amc-corey-cox: Moot — reverted the template change entirely. Based on your overall review, we simplified the fix to just treat bnode as unimplemented (returns None), same as class_curie, class_uri, etc. No more default_factory/template machinery.

packages/linkml/src/linkml/generators/pydanticgen/pydanticgen.py
amc-corey-cox: Moot — removed the default_factory field entirely. The PR no longer adds any new fields to PydanticAttribute. Agreed on the predefined naming awkwardness though — worth a separate cleanup.

packages/linkml/src/linkml/generators/pydanticgen/template.py
amc-corey-cox: Moot — removed the default_factory field entirely, so no mutual exclusion needed.

packages/linkml/src/linkml/generators/pydanticgen/pydanticgen.py
amc-corey-cox: Yep, your way is better.

sneakers-the-rat:
sorry if i threw a spanner in the works! i just wasn't sure when one would encounter a bnode in a pydantic generated artifact. my comment wasn't intended as an authoritative "bnondes shouldn't default to random" but "i'm not sure why that would be the case" from what i know about naming/canonicalizing bnodes

amc-corey-cox:
Thanks for pushing back on this @sneakers-the-rat . I think you're right  that we should just return none. I also changed this for the python ifabsent. I don't see any value in generating random bnodes with no context. This is much simpler.

tests/linkml/test_generators/test_pydanticgen.py
sneakers-the-rat, line 1: lovely tidying!

tests/linkml/test_generators/python/test_python_ifabsent_processor.py
sneakers-the-rat, line 1: i assume this would have also failed in the python dataclases, right? like i don't see any `bnode` callable imported or defined in generated python dataclasses. not sure how this slipped past like this

tests/linkml/test_generators/test_pydanticgen.py
amc-corey-cox: Thanks!

tests/linkml/test_generators/python/test_python_ifabsent_processor.py
amc-corey-cox: Yeah, I think so. Even if not, it didn't feel right to have them do different things.

PR #3247, opened 2026-03-04
Fix: use kwargs in _normalize_inlined to support inherited classes

### LINKED ISSUE #3244: _normalize_inlined passes positional args to slot_type, breaks when key field is not first  __init__ parameter
_normalize_inlined in yamlutils.py instantiates slot_type using positional arguments in three places. This assumes the key/identifier field is the first parameter of __init__. With dataclass  inheritance, parent fields come before child fields in the MRO, so a key field declared on a subclass is not the first parameter. The positional call assigns the value to the wrong field, leading to silent data corruption or ValueError.

Found in https://github.com/nfdi-de/dcat-ap-plus/pull/64 after upgrading to linkml 1.10.0, which started generating _normalize_inlined_as_list for the other_identifier slot with key_name="notation". The Identifier class has is_a: SupportiveEntity (contributing title, description), making notation the third __init__ parameter.

dalito:
Changed all three sites to use keyword arguments. For the two SimpleDict paths, uses `dataclasses.fields()` to find the value field by name instead of by position.

Also fixed the `lek == key_name` guard to allow list values (multivalued fields) instead of falling through to `form_1` which crashes on lists.

Closes #3244

### DIFF
--- a/packages/linkml_runtime/src/linkml_runtime/utils/yamlutils.py
+++ b/packages/linkml_runtime/src/linkml_runtime/utils/yamlutils.py
@@ -202,8 +202,12 @@ class YAMLRoot(JsonObj):
                         cooked_obj = slot_type(**as_dict(list_entry))
                         order_up(cooked_obj[key_name], cooked_obj)
                 elif isinstance(list_entry, list):
-                    # *args
-                    cooked_obj = slot_type(*list_entry)
+                    # First element is the key; remaining map to non-key fields in order
+                    non_key_fields = [f.name for f in dataclasses.fields(slot_type) if f.name != key_name]
+                    kwargs = {key_name: list_entry[0]}
+                    for fname, val in zip(non_key_fields, list_entry[1:]):
+                        kwargs[fname] = val
+                    cooked_obj = slot_type(**kwargs)
                     order_up(cooked_obj[key_name], cooked_obj)
                 else:
                     # lone key [key1: , key2: ... }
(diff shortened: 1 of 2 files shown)


dalito:
The action failures are network errors unrelated to the changes made here.

dalito:
Updated test strategy to make the problem more obvious:

The `_normalize_inlined` tests in `test_yaml_utils.py` now use fixture classes derived from PythonGenerator output rather than hand-written dataclasses. Each fixture section documents the complete source schema so the generated code can be reproduced.

All 5 tests confirmed RED on `main` and GREEN with the fix.

packages/linkml_runtime/src/linkml_runtime/utils/yamlutils.py
```diff
@@ -177,13 +185,16 @@ def form_1(entries: dict[Any, dict | JsonObj | None]) -> None:
...
                                 # key_name:value
-                                order_up(list_entry[lek], slot_type(list_entry))
+                                order_up(lev, slot_type(**{key_name: lev}))
                                 break  # Not strictly necessary, but
                             elif not isinstance(lev, list | dict | JsonObj):
-                                # key: value --> slot_type(key, value)
-                                order_up(lek, slot_type(lek, lev))
+                                # key: value — SimpleDict shorthand
+                                kwargs = {key_name: lek}
+                                if _value_field is not None:
```
amc-corey-cox, line 195: This `slot_type(*list_entry)` call has the same positional-arg assumption that the rest of the PR is fixing. If the key field isn't the first `__init__` parameter (inheritance case), this will mis-assign values too. Was this intentionally left, or just overlooked?

dalito: Thanks for noticing. I committed a fix so that the list branch now uses kwargs. The first element maps to key_name, remaining elements map to non-key fields in declaration order. Also added a test for this case that fails without the fix and passes with it.

Issue #3254, opened 2026-03-05
Java code generator should detect and avoid Java keywords


gouttegd:
The Java code generator should detect when an element within a schema (e.g. a class or a slot) has a name that happens to be a valid Java keyword, and have a mechanism to automatically alter the element’s name so as to avoid generating invalid code.

Name clashes between the elements of a schema and the keywords of the language may be unlikely to happen for most schemas, but the risk is much higher for some schemas – and first and foremost for LinkML’s own meta schema, which by nature (since it describes an object model) is bound to contain terms similar to those used in object oriented programming languages.

For example, with LinkML’s meta schema the code generator shamelessly produces code such as the following:

```java
private List<String> implements;
private Boolean abstract;
```

which will upset any Java compiler, since `implements` and `abstract` are Java keywords which cannot be used as field names.

Suggestions welcome as to how resolve a name clash. For now, I envision simply appending a dummy suffix to the element name (for example, turning `implements` to `implementsA`).

@matentzn You may assign to me.

dalito:
Maybe a trailing underscore? (that is convention in Python)

But I am not even sure if linkml follows it in the python/pydantic generators. --> it seems they do: https://github.com/linkml/linkml/pull/2862/changes

PR #3255, opened 2026-03-06
Misc improvements to the Java code generator

gouttegd:
(This PR supersedes #3217 with a couple of added fixes and a cleaner history.)

This PR groups various improvements to the Java code generator.

Specifically, it:

* fixes the use of an incorrect Java type to represent `xsd:time` (#3213) and makes sure the the Java classes representing dates and times are imported in the templates;
* similarly, replaces `String` by `java.net.URI` to represent `xsd:anyURI`;
* fixes the rendering of linkml:Any by translating it to Java’s Object class (#3177);
*  adds a mechanism to easily generate a visitor interface for an arbitrary class (#1449); as part of that, it also adds a mechanism by which one can request the generation of an additional code file, that is not tied to any class or enum;
* ensures that the name of a class or a slot does not clash with a reserved Java keyword (#3254);
* adds a new template variant that produces code suitable for use with the LinkML-Java runtime (#3216).

Reviewers: Happy to break down the PR in several PRs that each addresses one issue if that is preferred. Note, however, that the PR is already broken down in issue-specific commits, which can be reviewed one at a time.


packages/linkml/src/linkml/generators/javagen/class-org.incenp.linkml.jinja2
```diff
@@ -0,0 +1,193 @@
...
+{#- No need to call the parent's method if we are directly derived from Object. #}
+{%- if cls.is_a %}
+        if ( !super.equals(o) ) return false;
+{% endif %}
+{#- Now compare every slots. #}
+{%- for f in cls.fields %}
+        final Object this${{ f.name }} = this.{{ gen.get_read_accessor_name(f.source_slot) }}();
+        final Object other${{ f.name }} = other.{{ gen.get_read_accessor_name(f.source_slot) }}();
+{%-   if f.range in ["int", "float", "double", "boolean"] %}
+        if ( this${{ f.name }} != $other{{ f.name }} ) return false;
```
kevinschaper:
```suggestion
        if ( this${{ f.name }} != other${{ f.name }} ) return false;
```

The $ was in the wrong place on this one

tests/linkml/test_generators/test_javagen.py
```diff
@@ -119,3 +125,38 @@ def test_create_documents_includes_enums(kitchen_sink_path):
...
+
+
+def test_visitor_generation(kitchen_sink_path, tmp_path):
+    """Check that we can generator a visitor interface."""
+    gen = JavaGenerator(kitchen_sink_path, package=PACKAGE)
+    gen.serialize(directory=str(tmp_path), visitors=["Concept"])
+    assert_file_contains(tmp_path / "IConceptVisitor.java", "public void visit(DiagnosisConcept visited);")
+    assert_file_contains(tmp_path / "ProcedureConcept.java", "public void accept(IConceptVisitor visitor)")
+
+
```
kevinschaper, line 155:
```suggestion
def test_org_incenp_linkml_primitive_equals(input_path, tmp_path):
    """Primitive fields in org.incenp.linkml template must generate valid equals()."""
    gen = JavaGenerator(input_path("primitive_types.yaml"))
    gen.serialize(directory=str(tmp_path), template_variant="org.incenp.linkml")
    assert_file_contains(tmp_path / "SimpleClass.java", "this$requiredBoolean \!= other$requiredBoolean")

```

This adds another test path that catches the misplaced `$`

packages/linkml/src/linkml/generators/javagen/class-org.incenp.linkml.jinja2
gouttegd: Good catch, thanks!

tests/linkml/test_generators/test_javagen.py
gouttegd: No objection to adding the test, but I believe the backslash in this line (`\!=`) has nothing to do here:

> ```
> assert_file_contains(tmp_path / "SimpleClass.java", "this$requiredBoolean \!= other$requiredBoolean")
> ```

kevinschaper: you're right! the backslash definitely doesn't belong in there.

kevinschaper (approved):

PR #3257, opened 2026-03-08
Add plugin-link for interactive LinkML viewing

dkapitan:
Thanks for maintaining LinkML, we have started using it at PLUGIN and created this interactive ERD viewer with Svelte Flow.

### DIFF
--- a/docs/ecosystem.md
+++ b/docs/ecosystem.md
@@ -51,6 +51,7 @@ from the LinkML community meeting.
 ### Data Rendering
 
  - [linkml-renderer](https://github.com/linkml/linkml-renderer) allows linkml data to be rendered as markdowm or HTML.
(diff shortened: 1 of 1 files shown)


dkapitan:
@matentzn thanks for the enthusiastic response. It is early days yet, but we at [PLUGIN](https://github.com/plugin-healthcare) are serious about this. It is not just an one-off experiment but part of a larger ongoing programme to implement a nationwide federated processing infrastructure for hospitals.

Ultimately, I want to use LinkML as a [Rosetta stone](https://github.com/plugin-healthcare/plugin-rosetta) to create artefacts that hospitals can use in their ETL pipelines. Our end-goal is to have a practical integration of openEHR, FHIR and OMOP:

matentzn:
@dkapitan would you be able to present this work at an upcoming LinkML community call?

dkapitan:
@matentzn happy to join a community call and present it! Generally, if I have at least three weeks leadtime I can fit it in my calendar.

PR #3258, opened 2026-03-09
terminusdb: updates to the modern schema format

philocalyst:
Plugs into the latest client API and generates the new format, tested against an actual terminus instance.


matentzn:
Thank you @philocalyst! 

@dwinston @sneakers-the-rat can I ask you both to take a look at this?

packages/linkml/src/linkml/generators/terminusdbgen.py
```diff
@@ -39,106 +32,138 @@
...
+        elif not slot.required:
+            prop_value = {"@type": "Optional", "@class": rng}
+        else:
+            prop_value = rng
+
+        self.current_class_doc[prop_name] = prop_value
+
+        # Add property documentation
+        if slot.description and "@documentation" in self.current_class_doc:
+            self.current_class_doc["@documentation"]["@properties"][prop_name] = slot.description
```
noelmcloughlin: Maybe extend checks to ensure "@properties" dict exists unless there is contract saying it will.

packages/linkml/src/linkml/generators/terminusdbgen.py
```diff
@@ -39,106 +32,138 @@
...
+                doc.setdefault("@inherits", []).append("Document")
+        self.current_class_doc = doc
         return True
 
     def end_class(self, cls: ClassDefinition) -> None:
-        self.classes.append(self.clswq)
+        self.documents.append(self.current_class_doc)
 
     def visit_class_slot(self, cls: ClassDefinition, aliased_slot_name: str, slot: SlotDefinition) -> None:
+        rng = self._resolve_range(slot)
```
noelmcloughlin, line 120: Noting that `_resolve_range(slot)` may return `xsd:string` which is not in `XSD_TRANSLATE` mapper, presumably because no mapping is needed.

tests/linkml/test_generators/test_terminusdbgen.py
```diff
@@ -0,0 +1,217 @@
...
+def test_slot_description_in_documentation():
+    docs = _generate(SIMPLE_SCHEMA)
+    person = next(d for d in docs if d.get("@id") == "Person")
+    assert person["@documentation"]["@properties"]["name"] == "The name of the entity"
+
+
+@pytest.mark.parametrize(
+    "xsd_uri,expected",
+    [
+        ("xsd:int", "xsd:integer"),
```
noelmcloughlin, line 141: Noting this is not full coverage of `XSD_TRANSLATE` values, but may be sufficient coverage?

noelmcloughlin (commented):
LGTM, I have three comments which you could address/reply to, but otherwise looks fine. The unit tests are located under correct location `test_generators`.

matentzn:
@noelmcloughlin Thanks a ton for your review!

@philocalyst can you check out Noels review? Once everything is addressed, please tag @amc-corey-cox  in a PR comment and he will finalise everything!

packages/linkml/src/linkml/generators/terminusdbgen.py
```diff
@@ -39,106 +32,138 @@
...
 
+# Map LinkML XSD types that don't exist in TerminusDB to supported equivalents.
+XSD_TRANSLATE = {
+    "xsd:int": "xsd:integer",
+    "xsd:float": "xsd:double",
+    "xsd:language": "xsd:string",
+    "xsd:date": "xsd:dateTime",
+    "xsd:time": "xsd:dateTime",
+    "xsd:nonNegativeInteger": "xsd:integer",
+    "xsd:positiveInteger": "xsd:integer",
```
amc-corey-cox: `xsd:nonNegativeInteger` and `xsd:positiveInteger` are both already in `XSD_OK` (lines 27-28), meaning TerminusDB supports them natively. But because `XSD_TRANSLATE.get()` runs unconditionally before the `XSD_OK` check in `_resolve_range`, these will always get downgraded to `xsd:integer`, silently losing precision.

Same issue with `xsd:float` on line 38 — it's in `XSD_OK` (line 22) but gets translated to `xsd:double`.

I think the fix is to remove these three entries from `XSD_TRANSLATE` (lines 38, 42, 43), since TerminusDB handles them natively. Alternatively, you could check `XSD_OK` *before* `XSD_TRANSLATE` in `_resolve_range`, but removing the dead translations is simpler.

packages/linkml/src/linkml/generators/terminusdbgen.py
```diff
@@ -39,106 +32,138 @@
...
+                prop_value = {"@type": "Set", "@class": rng}
+        elif not slot.required:
+            prop_value = {"@type": "Optional", "@class": rng}
+        else:
+            prop_value = rng
+
+        self.current_class_doc[prop_name] = prop_value
+
+        # Add property documentation
+        if slot.description and "@documentation" in self.current_class_doc:
```
amc-corey-cox: Noel also flagged this: `@documentation` is only added to the class doc when `cls.description` is truthy (line 104-105). If a class has no description but its slots *do* have descriptions, those slot descriptions will be silently dropped because the `"@documentation" in self.current_class_doc` check fails.

Consider lazily initializing `@documentation` here:
```python
if slot.description:
    doc = self.current_class_doc.setdefault("@documentation", {"@comment": "", "@properties": {}})
    doc.setdefault("@properties", {})[prop_name] = slot.description
```

packages/linkml/src/linkml/generators/terminusdbgen.py
```diff
@@ -39,106 +32,138 @@
     ]
 }
 
+# Map LinkML XSD types that don't exist in TerminusDB to supported equivalents.
```
amc-corey-cox: Nit: the comment says these are types that "don't exist in TerminusDB", but three of the entries (`xsd:float`, `xsd:nonNegativeInteger`, `xsd:positiveInteger`) *do* exist in `XSD_OK` above. Once the overlap is resolved (see other comment), updating this comment to stay accurate would be good.

amc-corey-cox (approved):

philocalyst:
Good, I believe :)

amc-corey-cox:
Pushed a one-line fix to get CI green: switched `tests/linkml/test_generators/test_terminusdbgen.py` to import `SchemaBuilder` from `linkml_runtime.utils.schema_builder` instead of the deprecated `linkml.utils.schema_builder` shim.

PR #3265, opened 2026-03-13
Refactor: cleanup dead code

Silvanoc:
Remove modules clearly identified as useless in [#2371][1].

[1]: https://github.com/linkml/linkml/issues/2371


sujaypatil96:
Thanks for removing the comparefiles utility from the library @Silvanoc. Probably also a good idea to tack on the removal of the utility entrypoint from the pyproject.toml file: https://github.com/linkml/linkml/blob/main/packages/linkml_runtime/pyproject.toml#L60?

Silvanoc:
> Thanks for removing the comparefiles utility from the library @Silvanoc. Probably also a good idea to tack on the removal of the utility entrypoint from the pyproject.toml file: https://github.com/linkml/linkml/blob/main/packages/linkml_runtime/pyproject.toml#L60?

Wait a sec! I checked that the module is not used by any other module in LinkML. But I wasn't aware that it's being offered as a CLI tool! Probably installing it globally as `comparefiles` was never a good idea... nevertheless, I would be careful removing user-facing parts.

sujaypatil96:
@Silvanoc agreed about being careful removing user-facing parts/CLI tools but I think this comparefiles utility is trivial enough that it can be accomplished with file `diff` or any number of standard file comparison utilities. I don't know if there's too many people using this tool, so I think it can go!

If you want to be conservative we can keep the entrypoint and have it emit a deprecation warning for one release cycle and then get rid of it, but i'd say it's okay to just flat out remove it even now.

amc-corey-cox:
@Silvanoc  I agree with @sujaypatil96 on this one. It's unlikely this will break anyone in a way that 15 minutes with understanding diff wouldn't leave them better off. Could you remove it from the CLI, though?

Silvanoc:
@amc-corey-cox @sujaypatil96 removal from scripts added to the commit.

PR #3273, opened 2026-03-15
fix(gen-json): include mixins/abstracts for union(anyof) support

noelmcloughlin:
This PR is to resovle issue with dangling references in JSON schema.

- Classes with abstract=true or mixin:true may be referenced by `anyOf` (or `mixin` classes). But the generator skips them so they are missing under $defs.
- The pydantic model includes the class definitions missing in JSON schema.


noelmcloughlin:
re-basing into https://github.com/linkml/linkml/pull/3274

PR #3274, opened 2026-03-15
fix(gen-json): include mixins/abstracts for union(anyof) support

noelmcloughlin:
This PR is to fix issue with dangling references in JSON schema.

- Classes with abstract=true or mixin:true may be referenced by `anyOf` (or `mixin` classes). 
- However the generator skips them so they are missing under `$defs` in output schema file. 
- The pydantic model contains the missing class definitions.

We should generate definitions for abstract/mixin classes anyway since they can be referenced by other classes.  

Tested patch locally and issue resolved.

### DIFF
--- a/docs/maintainers/generator-governance.md
+++ b/docs/maintainers/generator-governance.md
@@ -1,0 +1,782 @@
+# Generator and Validator Governance
+
+## Why generator-level governance matters
+
+LinkML supports a large and growing number of output formats, each maintained by
+different contributors with varying levels of activity. Without clear ownership
+and decision-making processes at the generator level, several problems arise:
+
+- **Unreviewed changes** can break downstream consumers who depend on a specific
+  output format. A change to the JSON Schema generator, for example, can silently
+  invalidate validation pipelines across many projects.
+- **Stalled maintenance** becomes invisible. When the sole contributor to a
+  generator moves on, there is no clear path for someone else to pick up
+  responsibility, and issues accumulate without triage.
+- **Inconsistent quality** across generators. Some have extensive test coverage
+  and active reviewers; others have little of either. Users have no way to gauge
+  the maturity or reliability of a given output format.
+- **Decision deadlocks** around breaking changes. Without a designated maintainer
+  who understands the design intent and user base, even straightforward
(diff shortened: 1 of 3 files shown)


matentzn:
@noelmcloughlin thank your for your contribution! Before we get to this, can you fix the QC errors please? Much obliged!

noelmcloughlin:
Hi @amc-corey-cox @matentzn 
Thanks for the feedback.  I now have tests working locally except for unrelated "bug" on main branch which fails test.
```
FAILED tests/linkml/test_scripts/test_gen_python.py::test_head_deprecated - Failed: DID NOT WARN. No warnings of type (<class 'Warning'>,) were emitted.
!!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!
```

The issue is `depreciated_in` which starts at 1.9.6 so LinkML v1.1.13 is considered before the deprecation
https://github.com/linkml/linkml/blob/de7fae4c5f56bd398569bc9f8545074b03bbc7ad/packages/linkml/src/linkml/utils/deprecation.py#L227

Should I submit a separate PR for that, and if merged before this PR full tests could pass?
`        deprecated_in=SemVer.from_str("1.1.0"),            # <---- proposed change
`

matentzn:
> Thanks for the feedback. I now have tests working locally except for unrelated "bug" on main branch which fails test.

See also #3301 | dont worry about these locally failing tests we are handling them in  a different PR. Just make sure CI/CD passes here on the branch.

amc-corey-cox:
After digging into this more, I'm satisfied with this approach. The previous guard was generating broken JSON Schema — abstract classes were omitted from `$defs`, leaving dangling `$ref` pointers to definitions that didn't exist. The "enforcement" of abstract classes was an accidental side effect of that omission, not an intentional feature.

JSON Schema has no concept of abstract types — there's no way to say "this `$defs` entry exists for `$ref` resolution but can't be instantiated directly." Marking the compliance test as `INCOMPLETE` is the honest acknowledgment of that format limitation. The generated schema is now valid and correct.

Issue #2970, opened 2025-10-23
JSON-LD Context Generator Does Not Support Type Overriding


alexbrdn:
**Describe the bug**
When a class overrides the type of an attribute, which is also declared in slots, the generated JSON-LD context ignores the override and contains only one term definition for the attribute.

**Version of LinkML you are using**

1.9.4

**Details**

Here's the linkml schema. Note that `value` in `Attribute` has type `integer`, while `Measurement` overrides that by declaring an attribute `value` of type `double`.

<details>
<summary>type-conflict.linkml.yaml</summary>

```
id: example_schema
name: ExampleSchema
prefixes:
  ex: http://example.org/
  linkml: https://w3id.org/linkml/
default_prefix: ex
default_range: string
imports:
  - linkml:types

classes:
  Attribute:
    slots:
      - name
      - value

  Measurement:
    slots:
      - unit
    attributes:
      value:
        required: true
        range: double

slots:
  name:
    required: true
    range: string

  value:
    required: true
    range: integer

  unit:
    required: true
    range: string
```
</details>

When I generate a JSON-LD context from this schema, it contains a single definition for `value`:

<details>
<summary>generated-jsonld-context.json</summary>

```
{
   "comments": {
      "description": "Auto generated by LinkML jsonld context generator",
      "generation_date": "2025-10-23T13:03:19",
      "source": "type-conflict.yaml"
   },
   "@context": {
      "xsd": "http://www.w3.org/2001/XMLSchema#",
      "ex": "http://example.org/",
      "linkml": "https://w3id.org/linkml/",
      "@vocab": "http://example.org/",
      "value": {
         "@type": "xsd:integer",
         "@id": "value"
      },
      "name": {
         "@id": "name"
      },
      "unit": {
         "@id": "unit"
      },
      "Attribute": {
         "@id": "Attribute"
      },
      "Measurement": {
         "@id": "Measurement"
      }
   }
}
```
</details>

Compare this to the JSON schema generated from the same model:

<details>
<summary>generated-json-schema.json</summary>

```
{
    "$defs": {
        "Attribute": {
            "additionalProperties": false,
            "description": "",
            "properties": {
                "name": {
                    "type": "string"
                },
                "value": {
                    "type": "integer"
                }
            },
            "required": [
                "name",
                "value"
            ],
            "title": "Attribute",
            "type": "object"
        },
        "Measurement": {
            "additionalProperties": false,
            "description": "",
            "properties": {
                "unit": {
[...]

PR #3276, opened 2026-03-16
fix(generators): prefer literal coercion in mixed any_of unions

jdsika:
## Summary

When a LinkML slot uses `any_of` with both literal types and class ranges, the JSON-LD context generator previously collapsed the term definition to `"@type": "@id"`. That is wrong for scalar literals such as integers or decimals because JSON-LD then interprets plain scalar values as identifiers instead of typed literals.

This PR changes the JSON-LD contract for these mixed unions:
- if all literal branches collapse to one unambiguous coercion, emit that coercion
- if literal branches disagree, omit coercion instead of guessing
- if all branches are class-like, keep `"@type": "@id"`

## Why this is correct

JSON-LD term definitions can only carry a single `@type` coercion ([JSON-LD 1.1, section 4.2.2 -- Type Coercion](https://www.w3.org/TR/json-ld11/#type-coercion)). In a mixed union such as:

```yaml
any_of:
  - range: integer
  - range: QuantitativeValue
```

`"@type": "@id"` is not a correct default because it misrepresents the scalar branch. By contrast, `"@type": "xsd:integer"` correctly handles plain scalar values, while object values still serialize as node objects.

This also matches the direction of LinkML issue `#1483`: a range expression should not be silently dominated by unrelated default behavior.

## Alternatives considered

1. **Always omit coercion for mixed unions** -- simpler, but discards useful type information when all literal branches agree (e.g. `integer + SomeClass` can safely use `xsd:integer`).
2. **Emit an array of coercions** -- JSON-LD 1.1 does not support multiple `@type` values in a term definition, so this is not spec-compliant.
3. **Prefer `@id` (status quo)** -- incorrect for literal branches as explained above.

The chosen approach (prefer literal when unambiguous, omit when conflicting) is the least-surprising behavior that stays within JSON-LD 1.1 constraints.

## Implementation

Added `_literal_coercion_for_ranges()` to `jsonldcontextgen.py`:
1. collect literal coercions from `any_of` branches
2. if they all agree, use that one coercion
3. if they conflict, omit coercion
4. if there are no literal branches, fall through to existing class behavior

## Test coverage

Generator-level coverage:
- `decimal + class -> xsd:decimal`
- `integer + class -> xsd:integer`
- `string + class -> no coercion`
- `integer + string + class -> no coercion`
- existing class-only behavior remains `@id`

Compliance-level coverage:
- updated `test_slot_any_of(..., framework=jsonld_context)` to reflect the new JSON-LD contract
- covers the same mixed `any_of` behavior with and without `default_range`
- covers the same behavior with and without an explicit `linkml:Any` range

## Breaking change assessment

This is an intentional behavior change for `jsonld_context` generation in one narrow case: mixed literal/class unions.

The previous output was convenient for object references but incorrect for scalar literal expansion. The new behavior is more correct for JSON-LD processors and still allows object-valued branches.

## References

- Refs #1483
- Related #2970

### DIFF
--- a/packages/linkml/src/linkml/generators/pythongen.py
+++ b/packages/linkml/src/linkml/generators/pythongen.py
@@ -2,9 +2,11 @@ import keyword
 import logging
 import os
 import re
+from collections import defaultdict
 from collections.abc import Callable, Iterator
 from copy import copy
 from dataclasses import dataclass
+from graphlib import TopologicalSorter
 from pathlib import Path
 from types import ModuleType
 
@@ -405,7 +407,8 @@ version = {'"' + self.schema.version + '"' if self.schema.version else None}
 
     def gen_references(self) -> str:
         """Generate python type declarations for all identifiers (primary keys)"""
-        rval = []
+        rval = dict()
+        graph = defaultdict(set)
         for cls in self._sort_classes(self.schema.classes.values()):
             if not cls.imported_from:
                 pkeys = self.primary_keys_for(cls)
@@ -425,9 +428,10 @@ version = {'"' + self.schema.version + '"' if self.schema.version else None}
                         parent_cls = (
                             "extended_" + parents[-1] if parents[-1] in ["str", "float", "int"] else parents[-1]
                         )
-                        rval.append(f"class {classname}({parent_cls}):\n\tpass")
+                        rval[classname] = f"class {classname}({parent_cls}):\n\tpass"
+                        graph[classname].add(parent_cls)
                         break  # We only do the first primary key
-        return "\n\n\n".join(rval)
+        return "\n\n\n".join(rval[name] for name in TopologicalSorter(graph).static_order() if name in rval)
 
     def gen_typedefs(self) -> str:
         """Generate python type declarations for all defined types"""
--- a/packages/linkml_runtime/src/linkml_runtime/dumpers/rdflib_dumper.py
+++ b/packages/linkml_runtime/src/linkml_runtime/dumpers/rdflib_dumper.py
@@ -116,7 +116,14 @@ class RDFLibDumper(Dumper):
             return self._as_uri(element, id_slot, schemaview)
             # return URIRef(schemaview.expand_curie(str(element)))
         element_type = type(element)
-        cn = element_type.class_name
+        if hasattr(element_type, "class_name"):
+            cn = element_type.class_name
+        else:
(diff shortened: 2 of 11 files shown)


packages/linkml/src/linkml/generators/jsonldcontextgen.py
```diff
@@ -192,14 +192,64 @@ def visit_class(self, cls: ClassDefinition) -> bool:
...
+                    rng in self.schema.classes for rng in any_of_ranges
+                )
+                has_literal_range = any(rng in self.schema.types for rng in any_of_ranges)
+
+                if has_class_range and has_literal_range:
+                    # Mixed any_of: both literal and class branches exist.
+                    # Prefer the literal type for JSON-LD coercion so plain
+                    # values are typed correctly. Object branches still work
+                    # via explicit {"@type": "ClassName"} in instance data.
+                    #
```
amc-corey-cox: Nit: this comment block could be trimmed — the method docstring on `_literal_coercion_for_ranges` already covers the conflict case.

tests/linkml/test_compliance/test_boolean_slot_compliance.py
```diff
@@ -114,7 +114,12 @@ def test_slot_any_of(framework, data_name, value, is_valid, use_any_type, use_de
                     "_mappings": {
                         PYDANTIC: f"{SLOT_S1}: Optional[Union[D, int]]",
                         JSON_SCHEMA: expected_json_schema,
-                        JSONLD_CONTEXT: {SLOT_S1: {"@id": SLOT_S1, "@type": "@id"}},
+                        # JSON-LD term definitions can only carry one coercion.
+                        # For mixed any_of branches, LinkML now prefers the
+                        # unambiguous literal coercion so scalar values expand
+                        # correctly, while object values still serialize as
```
amc-corey-cox: Nit: this comment re-explains the feature — a one-liner like `# mixed any_of: prefer literal coercion` would be enough.

packages/linkml/src/linkml/generators/jsonldcontextgen.py
jdsika: resolved

tests/linkml/test_compliance/test_boolean_slot_compliance.py
jdsika: resolved

PR #3277, opened 2026-03-16
fix(generators): pass importmap and base_dir through ShaclGenerator

### LINKED ISSUE #2913: No easy way to map with SchemaView Element <-> "import name"
**Describe the issue**

Schemas can be imported with the `import` SchemaDefinition attribute. If I want to find out with **SchemaView** if and how an **element** has been imported, there is no easy way. The `SchemaLoader` returns for "resolved" schemas the [attribute `imported_from`](https://linkml.io/linkml-model/latest/docs/imported_from/) for each imported element (see [here](https://github.com/linkml/linkml/blob/ef5c24c32688d069b9a285ef6a7def8e68760950/tests/test_scripts/__snapshots__/genjsonld/simple_uri_test.jsonld#L48) how it results in some generated JSON-LD).

The [`SchemaView`](https://linkml.io/linkml/developers/schemaview.html) of a schema importing other schemas provides the method [`in_schema`](https://linkml.io/linkml/developers/schemaview.html#linkml_runtime.utils.schemaview.SchemaView.in_schema) to find out in which schema the **element** can be found. It returns  a `SchemaDefinitionName` without any reference to the URI or CURIE used to import that schema (if imported, of course).

Also [`imports_closure`](https://linkml.io/linkml/developers/schemaview.html#linkml_runtime.utils.schemaview.SchemaView.imports_closure) to get a list of all the imports (including for whatever reason the importing schema itself) as a list of `SchemaDefinitionName`s. Without any reference to the URIs or CURIEs used to import those schemas.
 
The only workaround that I've found to do that mapping is calling the `SchemaView` method [`load_import`](https://linkml.io/linkml/developers/schemaview.html#linkml_runtime.utils.schemaview.SchemaView.load_import) for each "import UriOrCurie", comparing the name of the imported schema with the `SchemaDefinitionName` that `in_schema` returns.

IMO a method called `import_uri` providing this functionality is missing. As of now the JSON-LD generator (which still uses `SchemaLoader`) is providing the attribute `imported_from` in its output. Having such a method would simplify the adaption of the JSON-LD generator to use `SchemaView` instead.

**Version of LinkML you are using** v1.9.4

jdsika:
## Summary

`ShaclGenerator.__post_init__` constructs its own `SchemaView` but previously hardcoded `SchemaView(self.schema)` without forwarding the `importmap` or `base_dir` keyword arguments. This causes `FileNotFoundError` when generating SHACL for schemas that use cross-directory imports resolved via an importmap.

## Problem

The `Generator` base class (line 211 of `generator.py`) correctly passes both parameters:
```python
self.schemaview = SchemaView(schema, importmap=self.importmap, base_dir=self.base_dir)
```

But `ShaclGenerator` overrides `__post_init__` with:
```python
self.schemaview = SchemaView(self.schema)  # missing importmap and base_dir!
```

This means any schema that relies on an importmap to resolve imports (e.g. mapping a schema ID to a local file path in a different directory) fails with a `FileNotFoundError` when using the SHACL generator, even though other generators (OWL, JSON-LD context) work correctly.

## Fix

Pass `importmap` and `base_dir` to the `SchemaView` constructor in `ShaclGenerator.__post_init__`:
```python
self.schemaview = SchemaView(
    self.schema,
    importmap=self.importmap,
    base_dir=self.base_dir,
)
```

## Test Coverage

Added `test_shacl_generator_respects_importmap` — creates a schema that imports a library from a subdirectory via an importmap, then verifies SHACL generation succeeds and produces the expected shapes.

## References

- Fixes #2913

### DIFF
--- a/packages/linkml/src/linkml/generators/shaclgen.py
+++ b/packages/linkml/src/linkml/generators/shaclgen.py
@@ -45,7 +45,11 @@ class ShaclGenerator(Generator):
     uses_schemaloader = True
 
     def __post_init__(self) -> None:
-        self.schemaview = SchemaView(self.schema)
+        self.schemaview = SchemaView(
+            self.schema,
+            importmap=self.importmap,
+            base_dir=self.base_dir,
+        )
         super().__post_init__()
         self.generate_header()
 
--- a/tests/linkml/test_generators/test_shaclgen.py
+++ b/tests/linkml/test_generators/test_shaclgen.py
@@ -1005,3 +1005,80 @@ classes:
         URIRef("https://example.org/related_to"),
     ]
     assert sorted(in_values, key=str) == expected_uris
+
+
+def test_shacl_generator_respects_importmap(tmp_path):
+    """ShaclGenerator must pass importmap and base_dir to SchemaView.
+
+    When a schema uses imports resolved via an importmap (e.g. remapping
+    a schema ID to a local file path), the SHACL generator must propagate
+    these parameters to its internal SchemaView. Previously, __post_init__
+    hardcoded SchemaView(self.schema) without forwarding importmap/base_dir,
+    causing FileNotFoundError for cross-directory imports.
+
+    See: https://github.com/linkml/linkml/issues/2913
+    """
+    # Create a "library" schema in a subdirectory
+    lib_dir = tmp_path / "lib"
+    lib_dir.mkdir()
+    (lib_dir / "base_types.yaml").write_text(
+        """
+id: https://example.org/base-types
+name: base_types
+default_prefix: bt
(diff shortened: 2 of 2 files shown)


jdsika:
Closing in favour of #3293, which fixes the root cause by migrating ShaclGenerator from SchemaLoader to SchemaView entirely.

This PR patched the symptom (forwarding importmap/base_dir to SchemaView), but #3293 removes the code this PR modified -- setting `uses_schemaloader = False` lets the parent `Generator.__post_init__` create the SchemaView with correct parameters, aligning with the documented migration plan (#923) and the pattern used by 13 other generators.

PR #3279, opened 2026-03-16
feat(generators): add --exclude-external-imports flag for JSON-LD context generator

jdsika:
## Summary

When a LinkML schema imports terms from an external vocabulary (e.g. W3C Verifiable Credentials v2), the generated JSON-LD context can redefine terms that are `@protected` in the external context, causing JSON-LD 1.1 processing errors per [section 4.1.11](https://www.w3.org/TR/json-ld11/#protected-term-definitions).

This PR adds a new `--exclude-external-imports` flag to the JSON-LD context generator that excludes elements from URL-based external vocabulary imports while keeping local file imports and linkml standard imports.

## Changes

| File | Change |
|------|--------|
| `jsonldcontextgen.py` | Add `exclude_external_imports` field, `_collect_external_elements()` helper using SchemaView `schema_map`, filtering in `visit_class`/`visit_slot`, and CLI flag |
| `test_jsonldcontextgen.py` | Add parametrized test over `mergeimports={True,False}` proving the flag works independently |

### Dedicated flag instead of overloading `--no-mergeimports`

The original approach piggybacked filtering onto `--no-mergeimports`, which changes its established semantics. Per reviewer feedback from @kevinschaper, a dedicated `--exclude-external-imports` flag is cleaner because:

- It's explicit -- users opt in to the behavior
- It doesn't change existing `--no-mergeimports` behavior
- It works independently of `mergeimports` (tested both `True` and `False`)
- It follows the principle of least surprise

### SchemaView `schema_map` instead of `imported_from` attribute

The original approach relied on the `imported_from` attribute which is only populated when `mergeimports=False`. The new approach uses SchemaView's `schema_map` (populated by `imports_closure()`) to identify URL-based imports by checking the map keys. This makes the flag work regardless of `mergeimports`.

### URL heuristic

Schemas whose import key starts with `http://` or `https://` are treated as external vocabularies. This covers the motivating cases (W3C/VC-style external contexts). Projects modularizing schemas across canonical URLs may want a future policy knob -- documented as a known limitation.

## How to use

```bash
# Exclude external vocabulary terms from generated context
gen-jsonld-context schema.yaml --exclude-external-imports

# Combine with other flags
gen-jsonld-context schema.yaml --exclude-external-imports --no-mergeimports
```

## Test coverage

Parametrized regression test verifying with both `mergeimports=True` and `mergeimports=False`:
- Local terms remain present
- URL-based external vocabulary terms are excluded
- linkml standard imports are unaffected

## References

- JSON-LD 1.1 section 4.1.11 -- [Protected Term Definitions](https://www.w3.org/TR/json-ld11/#protected-term-definitions)
- Builds on `--exclude-imports` from PR #3294

### DIFF
--- a/docs/generators/dashboard.md
+++ b/docs/generators/dashboard.md
@@ -25,7 +25,7 @@ Each cell shows the aggregate result across all tests in that category. Scroll d
 | <a href="#arrays">Arrays</a> | ✅ | ❓ | ⚠️ | ⚠️ | ❓ | ⚠️ | ❓ | ⚠️ | ❓ | ⚠️ | ⚠️ | ⚠️ |
 | <a href="#boolean-expressions">Boolean Expressions</a> | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ |
 | <a href="#cardinality-presence">Cardinality & Presence</a> | ⚠️ | ⚠️ | ✅ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ |
-| <a href="#core-structure">Core Structure</a> | ⚠️ | ⚠️ | ✅ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ |
+| <a href="#core-structure">Core Structure</a> | ⚠️ | ⚠️ | ⚠️ | ❓ | ⚠️ | ❓ | ⚠️ | ✅ | ⚠️ | ❓ | ⚠️ | ✅ |
 | <a href="#defaults-computed">Defaults & Computed</a> | ✅ | ✅ | ⚠️ | ❓ | ✅ | ❓ | ❓ | ✅ | ⚠️ | ❓ | ✅ | ✅ |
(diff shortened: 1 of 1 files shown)


kevinschaper:
I'm a little concerned about this PR changing the existing behavior of `--no-mergeimports`, what do you think about adding a new `--exclude-external-imports` flag for this path?

jdsika:
ok, I still need to test a bit more but does this go into the right direction?

packages/linkml/src/linkml/generators/jsonldcontextgen.py
```diff
@@ -56,8 +56,22 @@ class ContextGenerator(Generator):
...
+    """If True, elements from URL-based external vocabulary imports are excluded.
+
+    Local file imports and linkml standard imports are kept.  This is useful
+    when extending an external ontology (e.g. W3C Verifiable Credentials)
+    whose terms are ``@protected`` in their own JSON-LD context — redefining
+    them locally would violate JSON-LD 1.1 §4.1.11.
+
+    Note: this flag has no effect when ``mergeimports=False`` because
+    non-local elements are already absent from the visitor iteration
+    in that mode.
```
amc-corey-cox: This note says the flag has no effect when `mergeimports=False`, but the second commit message says it "corrects stale docstring claim that it has no effect in that mode" — and the tests (`test_exclude_external_imports_works_with_mergeimports_false`) explicitly verify it *does* work with `mergeimports=False`. Looks like the docstring fix didn't make it into the code — the test docstrings are correct but this note is still stale.

jdsika: AH, good catch! I try to. I need to go to bed now and will be away over the easter weekend with the family. Thank you for the review and I am happy to help!

PR #3286, opened 2026-03-17
Create office-hours.md

sagehrke:
Adding a new office-hours.md that has all the info for our new monthly LinkML office hours. It should live under the "Get Involved" section of the documentation.


sagehrke:
> I wonder if the slack invite will really work beyond 30 days, but else looks good!

In theory it should! This invite has no expiration date.

Issue #3285, opened 2026-03-16
Bug: `_DelimitedFileLoader` coerces numeric-looking strings before consulting schema


Sigfried:
### Summary

The TSV/CSV loader's `_parse_numeric` function unconditionally converts all numeric-looking string values to Python `int` or `float` before the schema is consulted. This breaks validation and mapping for slots with `range: string` or `range: <enum>` when the data values look numeric (e.g., zip codes like `"90210"`, or enum permissible values like `"1"`, `"2"`).

### Reproduction

Given this schema:

```yaml
classes:
  Survey:
    attributes:
      id:
        range: integer
        identifier: true
      zipcode:
        range: string
      satisfaction:
        range: satisfaction_enum

enums:
  satisfaction_enum:
    permissible_values:
      '1':
        description: Very Unsatisfied
      '2':
        description: Unsatisfied
      '3':
        description: Neutral
      '4':
        description: Satisfied
      '5':
        description: Very Satisfied
```

And this TSV:

```
id	zipcode	satisfaction
1	90210	4
2	10001	2
```

Running `linkml validate --schema schema.yaml --target-class Survey data.tsv` rejects both rows because `_parse_numeric` converts `"90210"` → `90210` (int) and `"4"` → `4` (int) before validation sees them.

### Root cause

`_parse_numeric` in `delimited_file_loader.py` is applied unconditionally to every cell value:

```python
yield {k: _parse_numeric(v) for k, v in row.items() if k is not None and v != ""}
```

The loader has no schema context, so it cannot distinguish between a `"90210"` destined for `range: string` (should stay a string) and a `"42"` destined for `range: integer` (should be coerced).

### Downstream impact

This also breaks linkml-map enum derivations. `ObjectTransformer.transform_enum()` compares `source_value == pv_deriv.populated_from` — when `source_value` is `4` (int) and `populated_from` is `"4"` (str), the match fails silently.

### Proposed fix

Make `_DelimitedFileLoader` schema-aware by accepting optional `schema` and `target_class` parameters. When provided, use `SchemaView` to determine which columns have `range: string` or `range: <enum>` and skip `_parse_numeric` for those columns. Without these parameters, behavior is unchanged.

Callers to update:

- `validate_file()` in `linkml/validator/__init__.py` — already has `schema` + `target_class`
- `_resolve_loaders()` in `linkml/validator/cli.py` — `schema` + `target_class` available in `cli()` scope
[...]

PR #3289, opened 2026-03-17
Make delimited file loader schema-aware to preserve string/enum columns

Sigfried:
## Summary

- Addresses #3285 
- The TSV/CSV loader's `_parse_numeric` converts all numeric-looking values to Python `int`/`float` before the schema is consulted, breaking validation and mapping for `range: string` slots (e.g. zipcode `"90210"` → `90210`) and enum permissible values (e.g. satisfaction `"4"` → `4`).
- Add optional `schema_path` and `target_class` params to `_DelimitedFileLoader`. When provided, `SchemaView` identifies columns whose range is `string` or an enum, and `_parse_numeric` is skipped for those. Without the params, behavior is unchanged.
- Update callers (`validate_file`, `default_loader_for_file`, CLI `_resolve_loaders`) to pass schema info through when available.

## Motivation

When a LinkML schema defines a slot with `range: string` or `range: <some_enum>`, and the TSV data contains numeric-looking values like `"90210"` or `"4"`, the loader converts them to integers before validation or mapping sees them. This causes:
- **Validation failures**: `90210 is not of type 'string'` and `4 is not one of ['1', '2', '3', '4', '5']`
- **Mapping failures**: downstream tools like linkml-map receive `int` instead of `str`

The fix is minimal and backward-compatible — the new params are optional and default to `None`, preserving existing behavior for all current callers.

## Test plan

- [x] 6 new parametrized tests covering schema-aware loading for both CSV and TSV
- [x] Existing test suite passes (101 passed, 3 skipped)
- [x] Backward compatibility: tests verify that without schema params, numeric coercion still happens

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -86,9 +86,7 @@ jobs:
         uses: ts-graphviz/setup-graphviz@v2.0.2
       - name: Test linkml package
         run: |
-          uv run coverage run -m pytest tests/linkml/ --with-network -m "not kroki"
-          uv run coverage xml -o coverage-linkml.xml
-          uv run coverage report -m
+          uv run pytest tests/linkml/ --with-network -m "not kroki" -n auto --dist loadgroup --cov --cov-report xml:coverage-linkml.xml --cov-report term-missing
         shell: bash
       - name: Test linkml-runtime package
         run: |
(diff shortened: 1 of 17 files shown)


Sigfried:
> thanks @Sigfried!

Thank you!!!

Issue #2495, opened 2025-01-13
Make it possible to map a URI to an RDF Node


roblinksdata:
**What is your feature request?**

According to [Direct Translation of instance graphs to RDF](https://linkml.io/linkml-model/latest/docs/specification/06mapping/#direct-translation-of-instance-graphs-to-rdf) it does not seem that it is possible to design a model in which the use specifies a URI and that URI is then referenced as a **graph node** when the data is translated to RDF.

It appears that it is only possible to express URIs as a **literal** as in the following example:

```ttl
@prefix schema: <http://schema.org/> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

<https://example.com/some-identifier> a schema:Dataset ;
    schema:isBasedOn "http://example.com/some-previous-dataset"^^xsd:anyURI.
```

This information could better be represented in the following way, where the URI points at a node in the graph:

```ttl
@prefix schema: <http://schema.org/> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

<https://example.com/some-identifier> a schema:Dataset ;
    schema:isBasedOn <http://example.com/some-previous-dataset>.
```

Given that RDF is so deeply integrated into the linkml metamodel, this seems like a key feature that is missing. It's clear that the ability exists when `identifier` is set to `true` on a given slot, however there are plenty of predicates which don't identify an instance of a class, but nonetheless provide deep links into the web of linked data.

*How important is this feature to you?** 
• Important - it's a blocker and can't do work without it

roblinksdata:
It seems it's possible to trick linkml in to serialising a URI as a reference to a node in the graph:

```yaml
...

prefixes: 
  schema: http://schema.org/
...

classes:

  UriasNode:
    description: Use this only in the `range` field of a slot to make a URI point at a node in an RDF graph.
    attributes:
      id: 
        description: You probably don't want to actually populate this field of the class, it's just here to support a reference.
        identifier: True
        required: True
        range: uriorcurie
...

slots:
  source_dataset_identifier:
    required: False
    multivalued: True
    slot_uri: schema:isBasedOn
    range: UriasNode
```

This can be used to generate RDF translations of the form

```ttl
@prefix schema: <http://schema.org/> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

<https://example.com/some-identifier> a schema:Dataset ;
    schema:isBasedOn <http://example.com/some-previous-dataset>.
```

[...]

PR #3292, opened 2026-03-17
feat(generators): add --xsd-anyuri-as-iri flag for cross-generator IRI consistency

jdsika:
## Summary

Add an opt-in `--xsd-anyuri-as-iri` flag to both the **JSON-LD context generator** and the **OWL generator**, aligning their treatment of `range: uri` / `range: uriorcurie` slots with the SHACL generator's built-in behavior.

## Problem

Without this flag, the three generators disagree on URI-typed properties:

| Generator | Output | Semantics |
|-----------|--------|-----------|
| **OWL** | `owl:DatatypeProperty` + `rdfs:range xsd:anyURI` | Literal string |
| **SHACL** | `sh:nodeKind sh:IRI` | IRI node (built-in, no flag) |
| **JSON-LD Context** | `@type: xsd:anyURI` | Typed literal string |

This creates a semantic inconsistency: SHACL says the value is an IRI node, but OWL and JSON-LD say it's a literal.

## Solution

With `--xsd-anyuri-as-iri` enabled:

| Generator | Output | Semantics |
|-----------|--------|-----------|
| **OWL** | `owl:ObjectProperty` (no `rdfs:range`) | IRI node ✅ |
| **SHACL** | `sh:nodeKind sh:IRI` (unchanged) | IRI node ✅ |
| **JSON-LD Context** | `@type: @id` | IRI node ✅ |

Per [OWL 2 §5.3–5.4](https://www.w3.org/TR/owl2-syntax/#Object_Properties), `ObjectProperty` connects individuals to individuals (IRI nodes), while `DatatypeProperty` connects individuals to literals.

### Context generator (`jsonldcontextgen.py`)
- New `xsd_anyuri_as_iri` field (default `False`)
- `URI_RANGES_WITH_XSD` constant extends `URI_RANGES` with `XSD.anyURI`
- `visit_slot()` and `_literal_coercion_for_ranges()` use the flag
- CLI: `--xsd-anyuri-as-iri / --no-xsd-anyuri-as-iri`

### OWL generator (`owlgen.py`)
- New `xsd_anyuri_as_iri` field (default `False`)
- `slot_owl_type()`: promotes URI-range slots to `ObjectProperty`
- `transform_class_slot_expression()`: skips literal classification for URI ranges
- `_range_is_datatype()`: returns `False` for URI ranges when flag set
- CLI: `--xsd-anyuri-as-iri / --no-xsd-anyuri-as-iri`

### Shared utility (`common/subproperty.py`)
- `is_uri_range()` used by both generators to detect URI-like ranges

## Testing

- All 80 existing OWL generator tests pass
- All 23 deterministic output tests pass
- Verified with synthetic schema: `range: uri` → `ObjectProperty`, `range: string` → `DatatypeProperty` (unchanged)
- Flag is opt-in, default `False` — no impact on existing users

## Motivation

Gaia-X Trust Framework properties like `url`, `endpointURL`, `attestationServiceURI` are semantically IRI references (links to dereferenceable resources). The SHACL generator already treats them correctly (`sh:IRI`). This flag makes OWL and JSON-LD context consistent with SHACL.

### DIFF
--- a/packages/linkml/src/linkml/generators/common/subproperty.py
+++ b/packages/linkml/src/linkml/generators/common/subproperty.py
@@ -15,6 +15,10 @@ from linkml_runtime.utils.schemaview import SchemaView
 CURIE_TYPES: frozenset[str] = frozenset({"uriorcurie", "curie"})
 URI_TYPES: frozenset[str] = frozenset({"uri"})
 
+# Types whose XSD mapping is xsd:anyURI (not xsd:string).
+# ``curie`` maps to xsd:string and is deliberately excluded.
+_ANYURI_TYPES: frozenset[str] = frozenset({"uri", "uriorcurie"})
+
 
 def is_uri_range(sv: SchemaView, range_type: str | None) -> bool:
     """
@@ -58,6 +62,35 @@ def is_curie_range(sv: SchemaView, range_type: str | None) -> bool:
     if range_type in sv.all_types():
         type_ancestors = set(sv.type_ancestors(range_type))
         if type_ancestors & CURIE_TYPES:
+            return True
+
+    return False
+
+
+def is_xsd_anyuri_range(sv: SchemaView, range_type: str | None) -> bool:
+    """Check if range type resolves to ``xsd:anyURI``.
+
+    Returns True for ``uri``, ``uriorcurie``, and types that inherit from them.
+    Returns False for ``curie`` (which maps to ``xsd:string``).
+
+    This is the correct predicate for the ``--xsd-anyuri-as-iri`` flag: only
+    types whose XSD representation is ``xsd:anyURI`` should be promoted from
+    literal to IRI semantics.  ``curie`` is a compact string representation
+    that resolves to ``xsd:string`` and must not be affected.
+
+    :param sv: SchemaView for type ancestry lookup
+    :param range_type: The range type to check
+    :return: True if range type maps to xsd:anyURI
+    """
+    if range_type is None:
+        return False
+
+    if range_type in _ANYURI_TYPES:
+        return True
+
+    if range_type in sv.all_types():
+        type_ancestors = set(sv.type_ancestors(range_type))
+        if type_ancestors & _ANYURI_TYPES:
             return True
 
     return False
--- a/packages/linkml/src/linkml/generators/jsonldcontextgen.py
+++ b/packages/linkml/src/linkml/generators/jsonldcontextgen.py
@@ -23,6 +23,10 @@ from linkml_runtime.utils.schemaview import SchemaView
 
 URI_RANGES = (SHEX.nonliteral, SHEX.bnode, SHEX.iri)
 
+# Extended URI_RANGES that also treats xsd:anyURI as an IRI reference (@id)
+# rather than a typed literal. Opt-in via --xsd-anyuri-as-iri flag.
+URI_RANGES_WITH_XSD = (*URI_RANGES, XSD.anyURI)
+
 ENUM_CONTEXT = {
     "text": "skos:notation",
     "description": "skos:prefLabel",
@@ -72,6 +76,12 @@ class ContextGenerator(Generator):
     _local_slots: set | None = field(default=None, repr=False)
     _external_classes: set | None = field(default=None, repr=False)
     _external_slots: set | None = field(default=None, repr=False)
+    xsd_anyuri_as_iri: bool = False
+    """Map xsd:anyURI-typed ranges (uri, uriorcurie) to ``@type: @id`` instead of ``@type: xsd:anyURI``.
+
+    This aligns the JSON-LD context with the SHACL generator, which emits
+    ``sh:nodeKind sh:IRI`` for the same types.
+    """
 
     # Framing (opt-in via CLI flag)
     emit_frame: bool = False
@@ -263,6 +273,7 @@ class ContextGenerator(Generator):
         and "could not resolve safely because the branches disagree".
         """
         coercions: set[str | None] = set()
+        uri_ranges = URI_RANGES_WITH_XSD if self.xsd_anyuri_as_iri else URI_RANGES
         for range_name in ranges:
             if range_name not in self.schema.types:
                 continue
@@ -271,7 +282,7 @@ class ContextGenerator(Generator):
             range_uri = self.namespaces.uri_for(range_type.uri)
             if range_uri == XSD.string:
                 coercions.add(None)
-            elif range_uri in URI_RANGES:
+            elif range_uri in uri_ranges:
                 coercions.add("@id")
             else:
                 coercions.add(range_type.uri)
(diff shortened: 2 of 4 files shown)


matentzn:
Thank you @jdsika for this contribution, we will try to find a suitable reviewer today!

jdsika:
Still needs some testing, but the general idea can already be reviewed

jdsika:
@amc-corey-cox I am getting my examples green in validation and it "seems to make sense for me" like that. Can you critically review it for me, please?

jdsika:
`-> Disclaimer: Conversation with myself summarized by agent <-`

## Alternative Design Consideration: Dedicated `iri` Type in the LinkML Metamodel

This PR is a solid, backward-compatible fix for the cross-generator inconsistency on `range: uri` slots. However, I want to raise a longer-term alternative for discussion: **introducing a dedicated `iri` type in `linkml-model`** that would make the flag unnecessary.

### The Root Cause

The current [`uri` type definition](https://github.com/linkml/linkml-model/blob/main/linkml_model/model/schema/types.yaml) maps to `xsd:anyURI`, with a comment that acknowledges the ambiguity:

> *"in RDF serializations a slot with range of uri is treated as a literal or type xsd:anyURI **unless** it is an identifier or a reference to an identifier, in which case it is translated directly to a node"*

This "it depends on context" definition is the root cause: the same `range: uri` declaration must mean different things in different generators, and a flag is needed to reconcile them.

### What the W3C Standards Say

The standards draw a **hard line** between IRI nodes and `xsd:anyURI` literals — they are not interchangeable:

| Standard | Key distinction | Reference |
|----------|----------------|-----------|
| **RDF 1.1 Concepts** | §3.2: IRIs are **graph nodes** that identify resources. §3.3: Literals are **typed values** — `"http://example.org"^^xsd:anyURI` is a literal, not an IRI node. | [§3.2 IRIs](https://www.w3.org/TR/rdf11-concepts/#section-IRIs), [§3.3 Literals](https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal) |
| **OWL 2 Syntax** | §5.3: **Object Properties** connect individuals to individuals (IRI nodes). §5.4: **Data Properties** connect individuals to literals. `xsd:anyURI` is a datatype (§4.6) — its values are literals. In OWL 2 DL a property cannot be both (§5.8.1). | [§5.3 Object Properties](https://www.w3.org/TR/owl2-syntax/#Object_Properties), [§5.4 Data Properties](https://www.w3.org/TR/owl2-syntax/#Data_Properties) |
| **XML Schema** | `xsd:anyURI` is a **datatype** whose value space is character sequences conforming to URI/IRI syntax — i.e. strings, not node references. | [XSD §3.3.17 anyURI](https://www.w3.org/TR/xmlschema11-2/#anyURI) |
| **SHACL** | `sh:nodeKind sh:IRI` constrains the value to be an **IRI node**, not a literal. This is what the SHACL generator already emits for `range: uri`. | [SHACL §4.8.1](https://www.w3.org/TR/shacl/#NodeKindConstraintComponent) |
| **JSON-LD 1.1** | `@type: @id` = IRI node reference; `@type: xsd:anyURI` = typed literal string. Fundamentally different JSON-LD processing semantics. | [JSON-LD 1.1 §4.2.2](https://www.w3.org/TR/json-ld11/#type-coercion) |

**Summary:** `xsd:anyURI` is normatively a datatype for literals. Treating it as an IRI node (as SHACL already does, and as this flag enables in OWL/JSON-LD) is semantically correct **for the intent**, but contradicts the type's own definition. The clean solution is a type that maps to an IRI-node construct from the start.

### LinkML Already Has the Building Blocks

LinkML already defines types that unambiguously mean "IRI node":

| Type | `uri:` field | Semantics |
|------|-------------|-----------|
| `objectidentifier` | `shex:iri` | IRI node ([ShEx vocabulary](http://www.w3.org/ns/shex#iri)) |
| `nodeidentifier` | `shex:nonLiteral` | IRI or blank node |

These already produce `sh:nodeKind sh:IRI` and `@type: @id` **without any flag**, because `shex:iri` and `shex:nonLiteral` are in the context generator's `URI_RANGES` tuple.

However, `objectidentifier` carries metamodel-level identity semantics ("represents an object in the model") and uses `ElementIdentifier` as its Python base, making it unsuitable as a drop-in replacement for `range: uri`.

### Proposed: A New `iri` Type

A dedicated type in `linkml-model/types.yaml`:

```yaml
iri:
  uri: shex:iri
  base: URI
  repr: str
  description: >-
    An IRI reference that identifies a resource node in the RDF graph.
    Unlike `uri` (xsd:anyURI, a literal datatype), `iri` always
    produces an IRI node across all generators: owl:ObjectProperty in
    OWL, sh:nodeKind sh:IRI in SHACL, @type: @id in JSON-LD context.
  notes: >-
    Use `range: iri` when the slot value should reference another
    resource (e.g., a homepage URL, a related entity). Use `range: uri`
    when the value is truly a URI-as-string literal.
  conforms_to: https://www.w3.org/TR/rdf11-concepts/#section-IRIs
```

### Comparison: Flag vs. New Type

| Criterion | `--xsd-anyuri-as-iri` flag (this PR) | New `iri` type |
|-----------|--------------------------------------|----------------|
| Schema self-documentation | ❌ `range: uri` is ambiguous — intent depends on a build flag | ✅ `range: iri` explicitly declares IRI-node intent |
| Generator coupling | ❌ Every generator must check the flag | ✅ Generators dispatch on the `uri:` field naturally |
| Per-slot granularity | ❌ All-or-nothing (flag applies globally) | ✅ Each slot independently chooses `uri` (literal) vs `iri` (node) |
| Backward compatibility | ✅ Opt-in flag, default off | ✅ Old schemas keep `range: uri` unchanged |
| Standards alignment | ⚠️ Reinterprets `xsd:anyURI` as IRI node | ✅ Uses `shex:iri` — correct by construction |
| Upstream dependency | ✅ No metamodel change needed | ❌ Requires `linkml-model` change + release |

### Recommendation

These approaches are **not mutually exclusive**. This PR is a pragmatic short-term bridge. But I'd suggest:

1. **Merge this PR** as-is (the flag is well-implemented and backward-compatible)
2. **Open an issue/RFC on `linkml/linkml-model`** proposing the `iri` type
3. Once the type ships, **deprecate the flag** and migrate schemas from `range: uri` + flag to `range: iri`

jdsika:
@amc-corey-cox the flag works for my usecase which is the reson I am keeping it in my develop branch but I am actually in favor of adding an additional type as in the summary above. What do you think about it?

I converted this into a draft as we then keep it for the discussion and I can pivot maybe to the other implementation.

amc-corey-cox (approved):
@jdsika Sorry this took so long for me to really wrap my head around. I think this is a good interim solution. I agree with your analysis that we need a better solution for IRI in the data model. I'm not inclined to think a dedicated IRI slot in the model is that solution but I think we need to have a broader discussion on this. 
I did a bit of an analysis on this that I'll add to the context of #2495 which I'm not going to close with this PR. I think you have provided a good solution for the immediate problem but there is a deeper issue there we still need to address. 

When you feel like this is ready please take it off of draft and I'll get it merged in.

jdsika:
@amc-corey-cox I resolved the merge conflict. It is great to see that you linked the PR in the issue that exactly discusses the case! Maybe this flag should be removed after a real modelling solution is found

jdsika:
@kevinschaper does the merge queue not work for PRs from a fork? The status checks apparently did not return a resul. I rebased the branch again now.

PR #3300, opened 2026-03-19
first working version openapi generator

### LINKED ISSUE #3299: Add OpenAPI generator

OpenAPI specification files with complex resources are very though to write if detailed schemas are desired. It the resources are interrelated, then having a data model is meaningful and LinkML appears to be a very good option for this purpose.

**Additional context**

LinkML can only help to generate the resource schemas (found under `./components/schemas`). All headers, endpoints, security schemes,... need to be provided externally.

Resource schemas are created in a JSON-Schema dialect, therefore the JSON-Schema generator can help on this task.

In my use-case I also need MarkDown documentation of the data model and JSON-Schemas to validate the data independently from the OpenAPI specification. Therefore the polyglot capabilities of LinkML make it a good fit for the task.

Silvanoc:
Initial version of the OpenAPI generator. It requires two arguments:
1. An incomplete OpenAPI YAML specification file that provides everything needed except for the resource schemas (`./components/schemas`).
2. A LinkML model of the API resources.

It uses the JSON-Schema generator to create the resource schemas. Unfortunatelly OpenAPI resource schemas are a dialect of JSON-Schema and therefore some small changes are needed. The JSON-Schema is also converted into YAML.

Finally the generated resource schemas are appended to the provided incomplete OpenAPI YAML specification file.

Parts of the generator feel like a better fit for the JSON-Schema, but I didn't want to "touch" such an important generator.

In theory Pydantic can generate OpenAPI-conformant schemas, but I've failed to get it running with that approach.

Fixes: #3299


Silvanoc:
Apparently Pydantic can generate [OpenAPI v3.1.0 comform resource schemas](https://github.com/pydantic/pydantic/blob/main/docs/concepts/json_schema.md). I've tried and failed. I don't know if there's an issue in the Pydantic generator or somewhere else.

My implementation generates Open v3.0.3 conform resource schemas (what I need). @sneakers-the-rat as the main contributor to the Pydantic generator, do you think you could have a look at it to generate Open v3.1.0 conform resource schemas out of Pydantic generated code?

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -0,0 +1,131 @@
...
+            for class_name in class_schema_names:
+                if class_name not in referenced_classes:
+                    del class_schemas[class_name]
+            referenced_classes = endpoint_referenced_classes.copy()
+            for class_schema in class_schemas.values():
+                self._find_references(class_schema, referenced_classes)
+        return class_schemas
+
+    def serialize(self, template_file: str = "", **kwargs) -> str:
+        self._template = yaml.safe_load(open(template_file))
```
amc-corey-cox, line 100: The `template_file` parameter on `serialize()` is unusual — every other generator's `serialize()` takes no required positional args. Would it work to make this a field on the generator dataclass instead, so it follows the standard pattern? Open to other approaches if that doesn't fit — it might be nice to keep the generator contract consistent.

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -0,0 +1,131 @@
...
+        class_schemas = {}
+        for class_name in referenced_classes:
+            json_schema = JsonSchemaGenerator(self.schema, include_null=False, top_class=class_name).generate()
+            json_schema_classes = json.loads(json_schema.to_json())["$defs"]
+            class_schemas = class_schemas | json_schema_classes
+        class_schemas = self._sanitize_schemas(class_schemas, referenced_classes)
+        class_schemas = self._fix_openapi_spec(
+            class_schemas,
+            ["description", "title"],
+        )
```
amc-corey-cox, line 112: Curious about stripping `description` and `title` — OpenAPI schemas support both and they're useful in generated API docs (Swagger UI, Redoc). Is there a reason these need to be removed, or could they be kept?

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -0,0 +1,131 @@
...
+        return class_schemas
+
+    def serialize(self, template_file: str = "", **kwargs) -> str:
+        self._template = yaml.safe_load(open(template_file))
+        referenced_classes = self._find_referenced_classes()
+        self._template["components"]["schemas"] = {}
+        class_schemas = {}
+        for class_name in referenced_classes:
+            json_schema = JsonSchemaGenerator(self.schema, include_null=False, top_class=class_name).generate()
+            json_schema_classes = json.loads(json_schema.to_json())["$defs"]
```
amc-corey-cox, line 106: This runs `JsonSchemaGenerator` once per referenced class. Would a single generation covering all classes with filtering afterward be more efficient?

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -0,0 +1,131 @@
...
+import yaml
+
+from linkml._version import __version__
+from linkml.generators.common.lifecycle import LifecycleMixin
+from linkml.generators.jsonschemagen import JsonSchemaGenerator
+from linkml.utils.generator import Generator, shared_arguments
+
+
+@dataclass
+class OpenApiGenerator(Generator, LifecycleMixin):
```
amc-corey-cox, line 17: Nit: project convention is to document all classes — a brief class docstring would be helpful. I recognize we haven't done a great job of holding to that convention.

amc-corey-cox (changes requested):
@Silvanoc thanks for the effort on implementing this feature. I have a couple of questions and a couple of nit-picks. None of this is blocking. I'm leaving this request changes just so we don't merge it before you have a chance to respond. If you don't feel these changes are needed just say so I'll change my review and merge. Nice work!

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: Originally I implemented it that way. But then I realized that the template is only relevant at the time of the serialization.

With this implementation, it would be possible to generate multiple different API specification files sharing the same data model. So an `OpenApiGenerator` instance could be created once and multiple `serialize` calls with different templates (with different endpoints) could be then done.

I have to admit though, that I don't know how realistic such as scenario can be.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: I suppose that it depends on the use scenario. In my case this approach is more efficient, because the data model is pretty big, but only a small part of it is relevant for the OpenAPI specification.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: Is a good point. I'll add it as long as the contribution is considered appropriate to be merged.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: I probably need to document the reason. Even for myself.

I have to admit that I've forgotten the reason for stripping them. But there for a reason, for sure, otherwise I wouldn't have added it. I'll find it out and document it. Probably that description and title are valid at attribute level, but not at class level.

amc-corey-cox (approved):
This looks good to me now. 

Do you want to fix the template_file dataclass field and description/title before I merge this in? If I don't hear from you by mid next week I'll just merge.

sneakers-the-rat:
> @sneakers-the-rat as the main contributor to the Pydantic generator, do you think you could have a look at it to generate Open v3.1.0 conform resource schemas out of Pydantic generated code?

sure, what's da problem?

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: Should be fixed by #3511.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: Should be fixed by #3511.

Silvanoc:
> > @sneakers-the-rat as the main contributor to the Pydantic generator, do you think you could have a look at it to generate Open v3.1.0 conform resource schemas out of Pydantic generated code?
> 
> sure, what's da problem?

@sneakers-the-rat mainly that I had no idea how to do it, not even where to start 😄 But I've asked an AI and got a big picture. So I might provide a PR to add it and will set you as reviewer.

Issue #3301, opened 2026-03-19
OWL compliance tests silently skip ROBOT coherency checks in CI


matentzn:
The OWL compliance tests include a ROBOT/HermiT coherency check step, but ROBOT is not installed in CI. This means:

- **In CI**: `robot_check_coherency()` returns `None`, the check is skipped, and the test is marked `UNTESTED` for that data point. The test "passes" based only on whether schema generation and RDF conversion succeed.: 
   ```python
                coherent = robot_check_coherency(ttl_path, output_path, str(ttl_path) + ".reasoned.owl")
                if coherent is not None:
                    if valid:
                        assert coherent, f"Coherency check failed for {ttl_path}"
                    else:
                        assert not coherent, f"Coherency check succeeded for {ttl_path}"
                else:
                    expected_behavior = ValidationBehavior.UNTESTED
   ```
- **Locally** (if ROBOT is on PATH): the coherency check actually runs, and at least 2 tests fail against ROBOT 1.9.10 with HermiT.

The result is that OWL compliance tests have never been validated against an actual OWL reasoner in CI.

## The 2 failing tests (local only)

```
FAILED test_class_boolean_with_expressions[owl-any_of_streq_MV-...-False]
  AssertionError: Coherency check succeeded for ...
  (expected incoherent, but HermiT found it coherent)

FAILED test_min_max_values[owl-float-float-1.5-10.5-1.6-True]
  AssertionError: Coherency check failed for ...
  (expected coherent, but HermiT found it incoherent)
```

Both pass when run in isolation — the failures only manifest when running the full compliance suite, suggesting test-ordering pollution (likely through the `cached_generator_output` dict in `helper.py`).

## Why ROBOT is in the compliance suite

The OWL branch in `check_data()` (`helper.py:866-886`) does:

1. Convert test data to RDF via `_convert_data_to_rdf()`
2. Call `robot_check_coherency()` which merges the data with the generated OWL ontology and runs `robot reason -r hermit`
3. Assert coherent if `is_valid=True`, incoherent if `is_valid=False`

This is the only way to actually validate OWL output — unlike JSON Schema or Pydantic where you can directly validate data against the schema, OWL validation requires a reasoner. Without ROBOT, the OWL tests only verify that the generator produces syntactically valid output, not that the output correctly captures the intended constraints.

## Why this is a problem

1. **Silent degradation**: There is no warning, skip marker, or CI log indicating that OWL coherency checks are being skipped. The tests appear green.
[...]

Issue #2333, opened 2024-09-30
Generator Support Matrix for LinkML Language Constructs


stanleyj-edsn:
Currently, LinkML's language definition includes constructs that aren't fully supported by all generators.

Identifying these gaps often requires trial-and-error (ie finding out the hard way), creating individual bug reports and digging through source code. This proposal suggests creating a centralized "Generator Support Matrix" to provide a clear, at-a-glance overview of language construct support across all LinkML generators.

The matrix may be structured as follows:

| Language Construct | Generator A | Generator B | Generator C | ... |
|---|---|---|---|---|
| Construct 1 | ✅ | ⚠️ Partial (see Issue #xxx) | ❌ Not Supported | ... |
| Construct 2 | ⚪ NA | ✅ | ✅ | ... |
| ... | ... | ... | ... | ... |

**Legend:**

* **✅ Full:** The generator fully supports this language construct.
* **⚠️ Partial:** The generator has partial support, potentially with limitations or known bugs. A linked issue number can provide further context.
* **❌ Not Supported:** The generator does not support this language construct. A linked issue number can track feature requests.
* **⚪ Not Applicable:** The language construct is not relevant to this specific generator.

**Advantages:**

* **Improved Transparency:** Users can quickly identify which generators support their desired language constructs.
* **Streamlined Development:**  Developers gain a clear roadmap for prioritizing generator improvements and addressing compatibility issues.
* **Reduced Bug Reports:**  By clearly outlining supported features, the matrix can preemptively address common questions and prevent unnecessary bug reports.
* **Enhanced Collaboration:**  The matrix can serve as a collaborative resource, allowing community members to contribute by identifying gaps and proposing solutions.

If this idea is accepted, I can contribute to this effort by helping to populate the initial matrix.

cmungall:
We are almost there, help most welcome!

@sneakers-the-rat made a PR in March (!!):

- #2021

I was a bit concerned about including in the main repo as the full output of the compliance suite can be large (we had envisioned each cell in your table as being linked to examples and proofs). I like the idea of deploying on another site that doesn't have the limitations of sphinx so it can be more interactive. But I think it's better to make some incremental progress on this.

I think ultimately it's also important to separate these categories:

- not implemented (linkml limitation)
- not directly supported in target framework (e.g multivalued in pure SQL; is_a in many frameworks)
[...]

PR #3303, opened 2026-03-19
Generate compliance dashboard

matentzn:
Depends on #3302 

- Adds a feature dashboard (`docs/generators/dashboard.md`) showing which metamodel features each generator supports, directly addressing #2333
- The compliance test suite is the **sole source of truth** — there is no manually maintained matrix
- Each compliance test is decorated with `@feature_category("Category", "Display Name")` which flows into a `summary.yaml` (gitignored) and is used to group features in the dashboard (#3302)
- A dedicated GitHub Action (`update-feature-dashboard.yaml`) rebuilds the dashboard **only** when generators, compliance tests, or the dashboard script change on `main` — it does not run on every CI build

## How it works

1. Compliance tests run with `--with-output` → produces `summary.yaml` with per-framework `ValidationBehavior` results
2. `scripts/generate_dashboard.py` reads `summary.yaml` → writes `dashboard.md`
3. The GitHub Action commits the updated dashboard back to `main` if it changed
4. The existing Sphinx workflow picks it up on the next docs deploy

Locally: `make build-generator-dashboard` runs both steps.

## Other changes

- `helper.py`: `report()` now uses `model_dump(mode="json")` with `SafeDumper` for portable YAML output (fixes Python-tagged enum serialization)
- `helper.py`: OWL/ROBOT coherency checks disabled with a warning — ROBOT was never installed in CI, so these checks silently skipped in CI but could fail locally (#3301)
- `Feature` model gains `category` and `display_name` fields populated from the `@feature_category` decorator
- `conftest.py`: no changes to report generation guards

## Test plan

- [x] `make build-generator-dashboard` produces readable `dashboard.md`
- [x] Existing compliance tests pass unchanged
- [ ] Verify dashboard renders correctly in Sphinx docs build


scripts/generate_dashboard.py
```diff
@@ -0,0 +1,320 @@
...
+
+def generate_dashboard(features: list[dict]) -> str:
+    """Generate the complete dashboard Markdown."""
+    frameworks = discover_frameworks(features)
+    fw_names = [GENERATOR_DISPLAY_NAMES.get(f, f) for f in frameworks]
+
+    # Group features by category (from @feature_category decorator)
+    categories: dict[str, list[dict]] = defaultdict(list)
+    for feat in features:
+        cat = feat.get("category", "Uncategorized")
```
kevinschaper, line 165: The `Feature` model in `tests/linkml/test_compliance/helper.py` only has `name`, `description`, `implementations`, and `num_tests` — there are no `category` or `display_name` fields. The `@feature_category` decorator referenced in the docstring and footer doesn't exist either.

As a result, running this script against real compliance output will put **all features under "Uncategorized"** and display names will fall back to raw test function names (e.g. `test_inlined` instead of "Inlined objects").

The `Feature` model needs `category` and `display_name` fields, and the `@feature_category` decorator (or equivalent mechanism to populate them) needs to be implemented for the auto-generation to work as described.

scripts/generate_dashboard.py
```diff
@@ -0,0 +1,320 @@
...
+    "python_dataclasses": "Python DC",
+    "jsonschema": "JSON Schema",
+    "java": "Java",
+    "shacl": "SHACL",
+    "shex": "ShEx",
+    "jsonld_context": "JSON-LD Ctx",
+    "jsonld": "JSON-LD",
+    "sql_ddl_sqlite": "SQLite DDL",
+    "sql_ddl_postgres": "Postgres DDL",
+    "sql_alchemy_imperative": "SQLAlchemy Imp",
```
kevinschaper: Framework ID mismatch: the compliance helper defines `SQL_ALCHEMY_IMPERATIVE = "sqlalchemy_imperative"` and `SQL_ALCHEMY_DECLARATIVE = "sqlalchemy_declarative"` (no underscore between sql and alchemy), but the keys here use `sql_alchemy_imperative` and `sql_alchemy_declarative`.

These won't match at runtime, so the display name lookup will fall through to showing the raw framework key.

kevinschaper (commented):
Review comments on the dashboard generation script.

scripts/generate_dashboard.py
kevinschaper: actually, this is a tiny fix, I'll just push up the change

scripts/generate_dashboard.py
kevinschaper: I'll just push up a quick fix on this

matentzn:
@kevinschaper just be a little cognisant of the dependency: https://github.com/linkml/linkml/pull/3302

matentzn:
Thanks for the review will work on these!

matentzn:
@amc-corey-cox @kevinschaper thanks for the feedback. I will address most of it but:

> The CI workflow can't push to main. Branch protection requires status checks (quality-checks, test on 3 platforms) with non_admins enforcement. github-actions[bot] isn't an admin, so the direct push will be rejected.

Good point! We could use a PR mechanism instead, read on.

> Committing dashboard.md to the repo will cause recurring merge conflicts. Any two PRs that touch compliance tests will conflict on the dashboard file. Since it's fully regenerated each time, the resolution is always "regenerate," but it's friction on every merge.

No the summary.yaml is not checked in, and dashboard.md is only created when deliberatly invoked. So individual PRs should not ever get merge conflicts!

> 15 minutes of CI per regeneration is expensive for what's essentially a docs update, and there's no caching or incremental approach.

Exactly, hence my solution of completely separating it from everything - including docs building! The action just rebuilds the dashboard from time to time when updates have occured.

> I think the right fix for all three is to move dashboard generation into the docs deploy pipeline instead of a separate workflow that pushes to main. 

I dont (yet) think so, read on.

> The compliance tests already produce summary.yaml during CI — the docs build just needs to run generate_dashboard.py before Sphinx builds. 

This file is NOT checked in. So the docs would have to run the full pipeline!

> That eliminates the branch protection issue, the merge conflicts (nothing committed), and the standalone CI cost.

Branch protection can be eliminated by PR mechanism, and the merge conflict does not really exist. If standalone CI cost is a real concern, we can just rebuild the dashboard ONCE every week - but I would say we try.

> One more thing: generate_dashboard.py is 320 lines with real logic (status aggregation, YAML parsing, framework discovery/ordering) but no tests. If the summary.yaml format or ValidationBehavior values change, the script will silently produce a dashboard full of ❓ icons with no error. A few focused tests on aggregate_status, status_icon, and the YAML loading would go a long way.

Good point - I will add a schema check!

matentzn:
@kevinschaper and @amc-corey-cox I think this is ready to be pushed over the finish line - I have addressed most of the feedback you gave (everything I agree with). Let me know what you think!

amc-corey-cox:
@matentzn This is all reasonable pushback and your solution is better than what I proposed. I'll give this another quick review. I suspect this is ready to go but I want to go through my whole review process again since it isn't very expensive.

.github/workflows/update-feature-dashboard.yaml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v6.2.0
+        with:
+          python-version: "3.12"
+
+      - name: Install dependencies
+        run: uv sync --all-groups
+
+      - name: Run compliance tests
+        run: uv run pytest tests/linkml/test_compliance/ --with-output -q
```
amc-corey-cox, line 50: If a compliance test fails, this step fails and the dashboard never gets generated. But the dashboard's purpose is to capture *current* state — including partial/failing features. Consider adding `continue-on-error: true` to this step so the dashboard always reflects reality.

(The `summary.yaml` should still be written even when tests fail, since it records per-framework `ValidationBehavior` results rather than pass/fail.)

scripts/generate_dashboard.py
```diff
@@ -0,0 +1,319 @@
...
+    # Coverage scores
+    lines.append("## Coverage Scores")
+    lines.append("")
+    lines.append("Percentage of tests where the generator fully implements the feature (excluding not-applicable).")
+    lines.append("")
+
+    lines.append("| Generator | Implements | Partial | Ignores | N/A | Total | Score |")
+    lines.append("|-----------|:----------:|:-------:|:-------:|:---:|:-----:|:-----:|")
+
+    for fw, name in zip(frameworks, fw_names):
```
amc-corey-cox: nit: `zip(frameworks, fw_names, strict=True)` would catch a length mismatch instead of silently truncating. Both lists are derived from `discover_frameworks` so they *should* always match, but `strict=True` is cheap insurance (Python 3.10+).

scripts/generate_dashboard.py
```diff
@@ -0,0 +1,319 @@
...
+        "This dashboard shows which LinkML metamodel features each generator supports. "
+        "**The compliance test suite "
+        "([tests/linkml/test_compliance/](https://github.com/linkml/linkml/tree/main/tests/linkml/test_compliance)) "
+        "is the source of truth** — every cell below is derived from actual test results."
+    )
+    lines.append("")
+    lines.append(
+        "*Regenerate with: "
+        "`uv run pytest tests/linkml/test_compliance/ --with-output && "
+        "python scripts/generate_dashboard.py`*"
```
amc-corey-cox: nit: The second command is missing the `uv run` prefix — should be `uv run python scripts/generate_dashboard.py` per project conventions. Same in the footer text (line ~287) and `dashboard.md` header.

.github/workflows/update-feature-dashboard.yaml
amc-corey-cox: Now, generally, we don't allow merges to main when tests fail - but it can happen.

melonora: Agree here, also when potentially certain concepts get added, but are not yet supported in each framework (e.g. perhaps array work).

docs/generators/dashboard.md
```diff
@@ -0,0 +1,200 @@
...
+| <a href="#identity-keys">Identity & Keys</a> | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ❌ | ⚠️ |
+| <a href="#inheritance-refinement">Inheritance & Refinement</a> | ✅ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ❌ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ |
+| <a href="#inlining-references">Inlining & References</a> | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ |
+| <a href="#metadata">Metadata</a> | ⚠️ | ⚠️ | ⚠️ | ❌ | ⚠️ | ❌ | ❌ | ⚠️ | ⚠️ | ❌ | ⚠️ | ⚠️ |
+| <a href="#other">Other</a> | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ❌ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ |
+| <a href="#rules-classification">Rules & Classification</a> | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ❌ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ |
+| <a href="#schema-level">Schema-Level</a> | ✅ | ✅ | ✅ | ❌ | ⚠️ | ❌ | ❌ | ✅ | ✅ | ❌ | ⚠️ | ✅ |
+| <a href="#slot-typing-ranges">Slot Typing & Ranges</a> | ⚠️ | ⚠️ | ⚠️ | ❌ | ⚠️ | ⚠️ | ⚠️ | ✅ | ⚠️ | ❌ | ⚠️ | ⚠️ |
+| <a href="#value-constraints">Value Constraints</a> | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ |
+
```
melonora, line 40: just simply amazing resource:)

.github/workflows/update-feature-dashboard.yaml
```diff
@@ -0,0 +1,64 @@
...
+jobs:
+  update-dashboard:
+    runs-on: ubuntu-latest
+    if: github.repository == 'linkml/linkml' && github.actor != 'github-actions[bot]'
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
```
dalito: Use the long/full version specifier to get patch and minor updates via dependabot. The current version of the checkout action is "v6.0.2".

.github/workflows/update-feature-dashboard.yaml
```diff
@@ -0,0 +1,64 @@
...
+        run: uv sync --all-groups
+
+      - name: Run compliance tests
+        run: uv run pytest tests/linkml/test_compliance/ --with-output -q
+
+      - name: Generate dashboard
+        run: uv run python scripts/generate_dashboard.py
+
+      - name: Create PR with updated dashboard
+        uses: peter-evans/create-pull-request@v8
```
dalito: Same as above: Here the current version is "v8.1.0".

matentzn:
@kevinschaper @amc-corey-cox @dalito @melonora all feedback should be addressed.

scripts/generate_dashboard.py
```diff
@@ -0,0 +1,319 @@
...
+    """Compute an aggregate icon for a list of per-feature statuses."""
+    if not statuses:
+        return "❓"
+    applicable = [s for s in statuses if s != "not_applicable"]
+    if not applicable:
+        return "⚪"
+    if all(s in ("implements", "accepts") for s in applicable):
+        return "✅"
+    if all(s in ("ignores", "untested") for s in applicable):
+        return "❌"
```
melonora, line 145: small question here, do we actually want untested to show as an error? Why not a question mark here?

melonora: for untested that is.

sneakers-the-rat: or just blank, for visual noise's sake?

matentzn: I played with both ? and space and ended up with `?` @melonora - for the simple purpose to have something recognisable in my "legend" at the top: https://github.com/linkml/linkml/blob/6249219bb0bcf521a065509006ec0a69eb8c7c37/docs/generators/dashboard.md 

Thanks botH!

Issue #3305, opened 2026-03-20
Reduce snapshot file count to improve PR reviewability


amc-corey-cox:
Generator changes often produce PRs with dozens to hundreds of changed snapshot files, making review impractical. The test suite currently has ~500 snapshot files across 8 `__snapshots__/` directories.

The worst offender is the biolink GOLR snapshots — 236 individual YAML config files (one per biolink class), which account for nearly half of all snapshots. These could be concatenated into a single file in a predictable order, turning a 236-file diff into a 1-file diff while preserving the same coverage.

The issue test snapshots (69 files, mostly generated Python) could be rolled into their appropriate generator test files, reducing both file count and the blast radius of generator changes.

SproutSeeds:
I took a local first pass on this from current `main`.

Exact results:
- local branch: `cody/issue-3305-snapshot-reduction-first-pass`
- commit: `2f03271f4`
- branch: `https://github.com/SproutSeeds/linkml/tree/cody/issue-3305-snapshot-reduction-first-pass`
- targeted verification:
  - `uv run --python 3.13 pytest tests/linkml/test_issues/test_issue_80.py tests/linkml/test_issues/test_issue_121.py tests/linkml/test_issues/test_curie_prefix_matching.py tests/linkml/test_issues/test_linkml_issue_384.py tests/linkml/test_issues/test_linkml_issue_388.py --with-network -q`
  - result: `7 passed`

One useful correction first: the issue body is directionally right, but some of the concrete counts are stale on current `main`.

- repo-wide snapshot count here is `241`, not `~500`
- `test_biolink_golr(...)` is already collapsed to one concatenated snapshot
- the biggest remaining hotspot is now `tests/linkml/test_issues/__snapshots__` at `69` files

So this first pass stays inside that narrower live target:

- consolidates five multi-output issue-test modules into bundled snapshots
- reduces `tests/linkml/test_issues/__snapshots__` from `69 -> 55`
- reduces repo-wide `tests/**/__snapshots__` from `241 -> 227`
- keeps the change out of the already-fixed Biolink GOLR path

One implementation detail that seemed necessary: once several outputs share one snapshot file, the bundled helper needs to normalize JSON sections and canonicalize RDF/OWL sections, otherwise harmless serialization-order changes become flaky.

If this direction looks useful, I can turn the branch into a small PR next.

matentzn:
@amc-corey-cox can you liaison with @SproutSeeds on this?

matentzn:
[...]

PR #3307, opened 2026-03-20
Consolidate GOLR biolink snapshots into single file

amc-corey-cox:
Partial fix for #3305

## Summary

GOLR snapshot directories stored one YAML file per class, causing generator or model changes to produce massive PR diffs. This consolidates them into single concatenated files with `# --- filename.yaml ---` separators.

| Location | Before | After |
|----------|--------|-------|
| biolink model test | 236 files | 1 file |
| gen-golr script test | 22 files | 1 file |
| **Total snapshot files in test suite** | **500** | **241** |

Same coverage, same content, reviewable diffs.

## Test plan

- [x] `test_biolink_golr` passes against generated snapshot
- [x] `test_metamodel_valid_call` passes against generated snapshot
- [x] Snapshot regeneration produces identical output


kevinschaper (approved):
nice optimization!

PR #3309, opened 2026-03-20
fix(owlgen): warn on covering axiom edge cases for abstract classes

jdsika:
## Summary

Emit a warning when an abstract class has only one direct `is_a` child, since the covering axiom degenerates to an equivalence (`Parent ≡ Child`). The warning recommends `--skip-abstract-class-as-unionof-subclasses` to suppress covering axioms if the equivalence is undesired.

## Motivation

The covering axiom feature (linkml/linkml#3219) correctly expresses that every instance of an abstract class must belong to one of its subclasses. With a single child, this is semantically equivalent to `Parent ≡ Child` per OWL 2 Primer §4.2. While OWL-correct, this may surprise users building extensible ontologies where more children are added downstream — the equivalence causes new siblings to inherit constraints from the first child via RDFS transitivity.

Rather than silently skipping the axiom (which would change the OWL semantics), a warning alerts users to the implication and points them to the existing escape hatch.

## Changes

- **`owlgen.py`**: Add `logger.warning()` when an abstract class has exactly 1 child before the covering axiom is emitted. Update field docstring and CLI help text to document the warning.
- **`test_owlgen.py`**: Add `test_abstract_class_with_single_child_emits_warning` (validates warning is logged with flag recommendation) and `test_single_child_warning_suppressed_by_skip_flag` (validates no warning when skip flag is set).

## Design Decision

Per review feedback from @mgskjaeveland (linkml/linkml#3309): the single-child equivalence is **not a generator error** — it is the expected OWL semantics. The proper response is a warning, not a silent skip. Users who need to avoid the equivalence should use `--skip-abstract-class-as-unionof-subclasses`.

## References

- [W3C OWL 2 Primer §4.2](https://www.w3.org/TR/owl2-primer/#Class_Hierarchies) — bidirectional `rdfs:subClassOf` = `owl:equivalentClass`
- [W3C OWL 2 Syntax §9.1.4](https://www.w3.org/TR/owl2-syntax/#Disjoint_Union_of_Class_Expressions) — `DisjointUnion` as proper covering construct
- linkml/linkml#3219 — Original covering axiom feature
- linkml/linkml#3309 — Upstream discussion


jdsika:
ok, this might be ready for a proper review @amc-corey-cox

matentzn:
@mgskjaeveland Can you perhaps take a look as well?

mgskjaeveland:
The behavior you describe is a logical consequence of how LinkML currently translates abstract classes into OWL covering axioms. The generator asserts that instances of an abstract class must belong to at least one of its subclasses:

Parent ⊑ Child1 ⊔ Child2 ⊔ ...

If there is only one subclass, this reduces to Parent ⊑ Child. Combined with the existing Child ⊑ Parent, the two classes become equivalent. This is expected from the semantics of OWL and not a generator error.

The issue arises when additional subclasses are later introduced directly in the generated OWL ontology without updating the covering axiom accordingly. In such cases the ontology must also update the union expression, or alternatively disable the abstract-class translation in the generator. Note that this is not only limited to the case when there is a single subclass. 

That said, it may be worth considering whether the generator should emit a warning when an abstract class has fewer than two subclasses, since the covering axiom then collapses the hierarchy.

jdsika:
Alright, so I take the suggestion of reworking it into a warning if this case occurs and tell the user about the existing flag from #3219 - using this flag would resolve my case, as I understand it?

jdsika:
The flag resolves my validation issue but I must admit that I am not fully aware of the side effects that come with the flag?
I also need to increase the tests in my framework to see if all my intended constraints are actually met and detected by the validation suite. But that is another story and I hope the implementation proposal for the warning is clear and helps users to navigate this situation.

mgskjaeveland:
Thanks!

> The flag resolves my validation issue but I must admit that I am not fully aware of the side effects that come with the flag?

My guess is that you can safely use the flag. Bare in mind that "abstract" does not fully accurately translate to OWL. You might miss out on some corner reasoning cases where the reasoner would a able to determine that a instance of an abstract class also is a member of one of its subclasses. 

A comment to your latest commit: as we are including warnings, for completeness, I suggest to also include a warning when an abstract class has no children. In this case the covering axiom has nothing to cover and warning could be e.g., "Abstract class '%s' has no children. No covering axiom will be generated."

jdsika:
Like this now?

mgskjaeveland:
> Like this now?

Yes, I do. Thanks! 

I have not checked against any code standards, though, but leave that for the LinkML team.

packages/linkml/src/linkml/generators/owlgen.py
```diff
@@ -471,6 +475,26 @@ def condition_to_bnode(expr: AnonymousClassExpression) -> BNode | None:
         # must be an instance of at least one of its direct subclasses.
         if cls.abstract and not self.skip_abstract_class_as_unionof_subclasses:
             children = sorted(sv.class_children(cls.name, imports=self.mergeimports, mixins=False, is_a=True))
+            if not children:
```
matentzn, line 478: The warning for 0 children is arguably noise for legitimate use. Abstract classes with no children in a *base* schema that's meant to be extended downstream is a perfectly normal pattern. Every abstract class starts with 0 children before you add subclasses. This could be quite chatty for schemas under development. Consider whether this should be `logger.info()` instead of `logger.warning()`, or whether there should be a way to suppress just the 0-children warning without also suppressing covering axioms for classes that *do* have children. I know @mgskjaeveland explicitly asked for this, but I am just wondering whether we have carefully considered this scenario?

matentzn (commented):
LGTM in general, but maybe smaller issue worth thinking about. I will approve after I get both of your responses to that.

mgskjaeveland:
- I agree with @matentzn and would add that it is a bit weird to return warnings then when there is nothing wrong, which one can argue is the case here: covering axioms are added when applicable, also for cases when there are 0 and 1 subclasses.
- I also see @jdsika 's point that getting an equivalence axiom then expecting a subclass axioms is surprising

Perhaps making abstract classes into covering axioms should not be the default? That will avoid future confusion. If it is not the default, then I suggest that both these messages could be turned into info messages.

matentzn:
>  Perhaps making abstract classes into covering axioms should not be the default? That will avoid future confusion. If it is not the default, then I suggest that both these messages could be turned into info messages.

What do you think @jdsika? I am a little inclined but acknowledge this is a 180 on my previous position.

jdsika:
Hey guys,
i am in the road and will think about it in Monday. I am sure we will find a good solution!

jdsika:
Hi all, sorry for the wait — I took some time to think this through properly and here's what I'd conclude:

---

**0-children: downgrade to `logger.info()`**

@matentzn you're right — abstract classes with no children are a completely normal pattern in base schemas designed for downstream extension. Every abstract class starts with 0 children before subclasses are added. A `warning` implies something is wrong when it isn't. `info` provides sufficient observability without the noise.

**1-child: keep as `warning`**

The equivalence consequence (Parent ≡ Child) is a genuine semantic trap that's hard to debug after the fact. When users later add sibling subclasses, those siblings can unexpectedly inherit constraints from the first child via RDFS transitivity. A `warning` is proportionate here — it flags a real, non-obvious implication and points directly to `--skip-abstract-class-as-unionof-subclasses` as the escape hatch.

**Keep covering axioms as default (opt-out)**

@mgskjaeveland I see the appeal of making it opt-in, but that would be a breaking change for users who already rely on the covering axiom behavior — which is the correct OWL translation of `abstract`. With the differentiated log levels (info for 0-children, warning for 1-child), users get enough signal to make an informed choice without us changing defaults.

I'll push the update shortly.

jdsika:
@matentzn I think this is ready for a final review

jdsika:
@amc-corey-cox this is IMO rdy to be merged

packages/linkml/src/linkml/generators/owlgen.py
```diff
@@ -505,6 +509,26 @@ def condition_to_bnode(expr: AnonymousClassExpression) -> OWL_EXPRESSION | None:
...
+            elif len(children) == 1:
+                # Warn: with one child C, the covering axiom degenerates to
+                # Parent ⊑ C which, combined with C ⊑ Parent (from is_a),
+                # creates Parent ≡ C (equivalence).  This is semantically
+                # correct per OWL 2 but may be surprising for extensible
+                # ontologies where more children are added later.
+                logger.warning(
+                    "Abstract class '%s' has only 1 direct child ('%s'). "
+                    "The covering axiom makes them equivalent (%s ≡ %s). "
+                    "Use --skip-abstract-class-as-unionof-subclasses to suppress.",
```
matentzn, line 588: NIT: This messages is framed for CLI users - folks that use the generator programmatically might pause over this framing.

matentzn (approved):
@mgskjaeveland's broader question about whether covering axioms should be opt-in by default is a fair one and I went back and forth on it, but I agree with @jdsika that flipping the default is a behaviour change for existing users and not something we should sneak into this PR. If anyone wants to push on that, it's worth its own issue and a discussion across more generators.

matentzn:
@amc-corey-cox this is fine by me - just leaving the more tech side of the PR to you.

amc-corey-cox (approved):
Yep, this looks good to me.

Issue #3297, opened 2026-03-18
doc-gen fails when rendering out rule without precondition


DnlRKorn:
---
name: Doc-Gen fails when rendering rule without preconditions
about: 
title: Doc-Gen fails when rendering rule without preconditions
labels: bug
assignees: Nico

---

**Describe the bug**
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[4], line 1
----> 1 out_str = template.render(gen=docgen, element=c_19, schemaview=sv_19, **template_vars_19)

File ~/ontology-access-kit/.venv/lib/python3.12/site-packages/jinja2/environment.py:1295, in Template.render(self, *args, **kwargs)
   1293     return self.environment.concat(self.root_render_func(ctx))  # type: ignore
   1294 except Exception:
-> 1295     self.environment.handle_exception()

File ~/ontology-access-kit/.venv/lib/python3.12/site-packages/jinja2/environment.py:942, in Environment.handle_exception(self, source)
    937 """Exception handling helper.  This is used internally to either raise
    938 rewritten exceptions or return a rendered traceback for the template.
    939 """
    940 from .debug import rewrite_traceback_stack
--> 942 raise rewrite_traceback_stack(source=source)

File ~/ontology-access-kit/.venv/lib/python3.12/site-packages/linkml/generators/docgen/class.md.jinja2:227, in top-level template code()
    225 | Rule Applied | Preconditions | Postconditions | Elseconditions |
    226 |--------------|---------------|----------------|----------------|
--> 227 {% for key in rule.preconditions -%}
    228 | {{ key }} |
    229 {%- if rule.preconditions[key] is defined -%}

TypeError: 'NoneType' object is not iterable

```
More details are provided at https://github.com/INCATools/ontology-access-kit/issues/850

**Version of LinkML you are using**
1.10.0 (this functions passes normally on release 1.9.1 since this md change was introduced in commit 3728203)

**Please provide a schema (and if applicable, a data file) that replicates the issue**
If you clone https://github.com/INCATools/ontology-access-kit/ the following command will replicate.
uv run gen-doc --dialect myst src/oaklib/datamodels/ontology_metadata.yaml -d  docs/datamodels/ontology-metadata/

**Proposed Fix**
[...]

PR #3310, opened 2026-03-21
fix: handle None preconditions in doc-gen class template

saschabuehrle:
## Bug
[#3297](https://github.com/linkml/linkml/issues/3297) — `gen-doc` crashes with `TypeError: 'NoneType' object is not iterable` when rendering a class rule that has no preconditions.

## Fix
Added Jinja2 `|default({})` filter to the preconditions loop in `class.md.jinja2`. When `rule.preconditions` is `None`, the loop now safely iterates over an empty dict instead of raising a `TypeError`.

This matches the existing guard pattern already used for `rule.postconditions` and `rule.elseconditions` in the same template.

## Testing
Verified that the template renders correctly for rules with preconditions (unchanged behavior) and rules without preconditions (no longer crashes, table section is simply empty).

Regression introduced in commit 3728203 (v1.10.0); v1.9.1 was unaffected since the markdown table was not present.

Happy to address any feedback.

Greetings, saschabuehrle


matentzn:
@saschabuehrle wonderful, thanks. I assigned @sujaypatil96 as your reviewer.

@sujaypatil96 if you dont have time dont worry, just reassign it to me!

sujaypatil96:
@matentzn i'll leave a review on this PR soon.

PR #3311, opened 2026-03-21
javagen: Render `uriorcurie`-typed slots as String fields.

gouttegd:
This PR is a follow-up to the PR that introduced the `org.incenp.linkml` template variant for the Java code generator (#3217).

Among other things, that PR changed the Java type used to represent the `xsd:anyURI` type from `String` to `java.net.URI`.

While I still believe this is the correct thing to do, I failed to realise that it applied to `uriorcurie`-typed slots as well (instead of only `uri`-typed slots), because that type, like `uri`, is associated to the `xsd:anyURI` XSD type. And `java.net.URI` is _not_ an appropriate type for `uriorcurie`-typed slots, at least when the code is intended to be used with the LinkML-Java runtime – which assumes those slots are String-typed instead.

So this commit amends the `org.incenp.linkml` template to ensure that `uriorcurie`-typed slots are rendered as String-typed fields as the runtime expects them to be.

Apologies for not catching that until _after_ #3217 was merged.


gouttegd:
Marking as draft as there is yet another issue with the template. Working on it right now.

kevinschaper (approved):
looks good!

PR #3317, opened 2026-03-23
fix(pep8): append underscore to reserved word named slots

### LINKED ISSUE #3316: Handling slot names which are python reserved words
**Describe the bug**
Python has about a dozen [reserved words](https://docs.python.org/3/reference/lexical_analysis.html#keywords).  When a LinkML slot name is a python reserved words, there is problems. 

For example, a slot named `class` results in `class: Optional[str] = None`, and SyntaxError on import.

**Cause**
The `pythongen.py` generator has no automatic mangling for Python reserved keywords as suggested by [PEP8](https://peps.python.org/pep-0008/):
```
_If your public attribute name collides with a reserved keyword, append a single trailing underscore to 
your attribute name. This is preferable to an abbreviation or corrupted spelling. (However, notwithstanding
this rule, ‘cls’ is the preferred spelling for any variable or argument which is known to be a class, especially
_the_ first argument to a class method.)_
```

**Workaround**
Renamed the slot from `class` to `_class`.  Aliasing does not work (causes aliased_slot_name() to return class, reproducing the problem). Also `class_` had some problem also.

**Example input data**
```
  groups:
    - id: ac
      class: family                     <------------- reserved word
      title: Access Control
      props:
        - name: label
          value: AC
      controls:
        - id: ac-1
          class: SP800-53                    <------------- reserved word
          title: Policy and Procedures
```

```
$  uv run linkml --version
linkml, version 1.1.13.post3557.dev0+e36df57a
```

noelmcloughlin:
## Summary

Fixes #3316 

## How was this tested?

- Updated `unmasked_python_keywords_example.yaml` to include all keywords and new unit test
- Fully test suite locally

## Checklist


packages/linkml/src/linkml/generators/pythongen.py
```diff
@@ -1085,6 +1085,10 @@ def gen_slotdefs(self) -> str:
 
     def gen_slot(self, slot: SlotDefinition) -> str:
         python_slot_name = underscore(slot.name)
+        # If slot name is a reserved keyword, follow PEP8 conventions to append an underscore.
+        # The transformed slot will be accessible as `slots.class_`, not `slots.class`.
+        if keyword.iskeyword(python_slot_name):
+            python_slot_name = python_slot_name + "_"
```
kevinschaper, line 1150: Rather than mangling inline here, override `slot_name()` on `PythonGenerator` so the trailing-underscore rule applies uniformly to `gen_slot`, `gen_class_variable`, `gen_postinit`, and `gen_constructor`. This mirrors the existing extension point used in `panderagen/dataframe_generator.py:363`:

```suggestion
    def slot_name(self, name: str) -> str:
        """Python-safe slot identifier. Appends a trailing underscore (PEP 8)
        if the underscored name collides with a Python reserved keyword, so
        that emissions like ``self.class`` become ``self.class_``. The
        original slot name is preserved on the runtime ``Slot(name=...)``
        argument, so schema lookups and URI resolution are unaffected.
        """
        pyname = super().slot_name(name)
        if keyword.iskeyword(pyname):
            pyname += "_"
        return pyname

    def gen_slot(self, slot: SlotDefinition) -> str:
        python_slot_name = underscore(slot.name)
        if keyword.iskeyword(python_slot_name):
            python_slot_name += "_"
```

(`gen_slot` keeps its own check because it operates on `slot.name` directly — for inline class attributes that have a qualified name like `thing__class` — rather than going through `slot_name()` / `aliased_slot_name()`.)

Separately, see the review body for a one-line change to `python_uri_for()` at line 1074, which is outside this hunk but emits `EX.class` (also a `SyntaxError`) for the same family of inputs.

tests/linkml/test_generators/test_pythongen.py
```diff
@@ -177,6 +177,37 @@ def __post_init__(self, *args, **kwargs):
...
+"""
+    output = PythonGenerator(yaml, mergeimports=False).serialize()
+
+    # The slots block must use the mangled name (keyword + "_")
+    assert f"slots.{keyword_slot}_" in output, (
+        f"Expected 'slots.{keyword_slot}_' in generated output but got:\n{output}"
+    )
+    # The reserved word must NOT appear as a bare attribute assignment in the class.
+    assert f"slots.{keyword_slot} " not in output, f"Found bare 'slots.{keyword_slot} ' (unmangled) in generated output"
+
```
kevinschaper: The current test only registers a top-level slot with no owning class, so `gen_class_variable` / `gen_postinit` / `python_uri_for` are never exercised. Attach the slot to a class and compile the generated module — that catches all the emission sites at once:

```suggestion
@pytest.mark.parametrize("keyword_slot", ["class", "import", "return", "yield", "global"])
def test_keyword_slot_names(keyword_slot: str):
    """
    Slots whose LinkML name is a Python reserved keyword must be emitted as
    ``<name>_`` (PEP 8 trailing-underscore convention) everywhere they appear
    as a Python identifier: in the ``slots`` registry, in dataclass fields,
    and in ``__post_init__`` self-references. The generated module must also
    be valid Python.
    """
    yaml = f"""id: http://example.org/test
name: test
prefixes:
  ex: http://example.org/
  xsd: http://www.w3.org/2001/XMLSchema#
default_prefix: ex
default_range: string
types:
  string:
    uri: xsd:string
    base: str
classes:
  Thing:
    attributes:
      {keyword_slot}:
        range: string
slots:
  {keyword_slot}:
    range: string
"""
    output = PythonGenerator(yaml, mergeimports=False).serialize()

    # The slots registry must use the mangled name
    assert f"slots.{keyword_slot}_" in output, (
        f"Expected 'slots.{keyword_slot}_' in generated output but got:\n{output}"
    )
    # The reserved word must NOT appear as a bare attribute assignment
    assert f"slots.{keyword_slot} " not in output, (
        f"Found bare 'slots.{keyword_slot} ' (unmangled) in generated output"
    )
    # The generated module must be valid Python (catches dataclass fields,
    # __post_init__ self-refs, and CurieNamespace attribute access).
    compile(output, "<generated>", "exec")
```

kevinschaper (changes requested):
Thanks for the fix — the direction is right, but I tested it locally against a schema that puts a reserved-word slot on a class (e.g. `Thing.attributes.class`), and the generated module still fails to compile. The current patch only handles the top-level `slots.<name> =` registry, but three other sites in `pythongen.py` also emit the bare keyword and produce `SyntaxError`:

1. `gen_class_variable()` (pythongen.py:623) → `class: Optional[str] = None`
2. `gen_postinit()` / related (pythongen.py:883, 898, 904–985) → `self.class = ...`
3. `python_uri_for()` (pythongen.py:1074) → `EX.class` (because `'class'.isidentifier()` is `True`, so the code takes the attribute-access branch instead of the subscript branch)

I verified this by generating a schema with `classes.Thing.attributes.class` and running `compile()` on the output — it raises `SyntaxError` in all three places. After applying the suggestions below, an extended parametrized test (class-owned slot + `compile()` assertion) passes across five keywords, and all 13 existing `test_pythongen.py` tests still pass.

### Summary of suggested changes

- Replace the inline `iskeyword` check in `gen_slot()` with an override of `slot_name()` local to `PythonGenerator`. This is the existing extension point (see `panderagen/dataframe_generator.py:363`) and routes the mangling through every callsite that emits a Python-form slot name — `gen_class_variable`, `gen_postinit`, `gen_constructor`. Reserved words are irrelevant to JSON Schema / SHACL / OWL / SQL generators, so `Generator.slot_name()` itself should stay unchanged.
- Patch `python_uri_for()` (line 1074, outside this PR's hunks but a latent bug of the same family). The one-line change is:
  ```python
  ns.upper() + (f".{ln}" if ln.isidentifier() and not keyword.iskeyword(ln) else f"['{ln}']"),
  ```
  so a keyword local-name falls back to subscript form (`EX['class']`).
- Extend the test to attach the reserved-word slot to a class and assert the output compiles.

packages/linkml/src/linkml/generators/pythongen.py
noelmcloughlin: Applied this solution.

tests/linkml/test_generators/test_pythongen.py
noelmcloughlin: Applied this solution with expanded schema in `pre-existing `unmasked_python_keywords_example.yaml`.

noelmcloughlin:
Thanks very much @kevinschaper for your review and I have updated the PR accordingly.  I also extended the test to cover all keywords.

kevinschaper (approved):

PR #3322, opened 2026-03-24
fix(induced_slot): resolve spaced/underscored slots to generators names

### LINKED ISSUE #3321: Handling induced_slot resolution of underscored hypenated slots
**Describe the bug**
LinkML Runtime (`utils/schemaview.py -> SchemaView.induced_slot`) does not resolve underscored variants of hyphenated slot names.  We get error even though the JSON generator emits `how_many` as a property (applying underscore() to the slot name how-many). Related to #3318

https://github.com/linkml/linkml/blob/5f3601b99c17a7efe474d816f804b7fcaf26e996/packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py#L1653-L1656

**Cause**
The code looks up slot_name literally in all_slots() and class.attributes with no fallback to slot_name_mappings() (underscore(name) → slot) causing mismatch between JSON Schema (underscored) and the slot names induced_slot understands.

**Workaround**
Same as #3320 and #3318 I used `--no-use-type-designators` to bypass induced_slot lookup entirely.

**Input data**
```
            - id: ac-01_odp.03
              props:
                - name: alt-identifier
                  value: ac-1_prm_2
                - name: label
                  value: AC-01_ODP[03]
                  class: sp800-53a
              select:
                how-many: one-or-more                     <----------- Hyphenated key
                choice:
                  - organization-level
                  - mission/business process-level
                  - system-level
```
`linkml, version 1.1.13.post3593.dev0+5f3601b9`

noelmcloughlin:
This PR is to address #3321 

Let me know of any feedback or concerns.


kevinschaper (approved):
looks good!

PR #3325, opened 2026-03-25
Extend sqlvalidation generator to process rules from schema

### LINKED ISSUE #3262: Extend sqlvalidation to process rules from schema
---
name: Extend sqlvalidation to process rules from schema 
title: 'Extend sqlvalidation to process rules from schema'
labels: feature
assignees: '@FlorianK13'
---

The sqlvalidation from #3123 is a MVP and needs to be extended. I now want to add the [rules](https://linkml.io/linkml/schemas/advanced.html#rules) from linkml schemas. These rules translate to two `WHERE` statements that are concatenated with `AND`. The precondition is checked as is, whereas for the postcondition, the opposite is checked (to find wrong entries):

```yaml
classes:
  LivingThings:
    rules:
      - preconditions:
          slot_conditions:
            type:
              equals_string: Human
        postconditions:
          slot_conditions:
            age:
              maximum_value: 150
```

translates to

```sql
SELECT * FROM LivingThings WHERE type = 'Human' AND age > 150;  
```

**Additional context**
I'll implement this soon.

FlorianK13:
Closes #3262 

In this PR I extend the sqlvalidation generator to process rules. I also add some bug fixes (not best practice, I know, sorry for that)

# What has changed

<details>
<summary>Adding `rules` to sqlvalidationgenerator</summary>

- extends the sqlvalidation generator to handle `rules` as described in #3262
  - rules evaluate to a `WHERE precondition AND NOT postcondition;`
  - types of condition that are supported are: `equals_string`, `equals_number`, `equals_string_in`, `minimum_value`, `maximum_value`, `pattern`, `required` - these by far not all conditions stated [here](https://linkml.io/linkml-model/latest/docs/SlotDefinition/) but i struggled to choose the ones that are relevant (see general remark)
  - it does not yet:
    - handle "any_of", "all_of", "none_of", "exactly_one_of" to combine several conditions 
- refactors code, so that existing functionality can be reused:
  - A new function `_build_violation_query` builds consistent final SELECT statements for every violation type
  - `_required_condition`, `_range_condition`, `_pattern_condition` build WHERE components that can be used at several locations at the code
  - apparently, rules are not in the `RelationalModelTransformer(SchemaView)` so I need to iterate over the original `SchemaView` to extract `rules`
- Added new tests
- Updated the `personinfo.sql` with the latest generated query

</details>

<details>
<summary>Bug Fixes</summary>

- Removed usage of `literal_column` from sqlalchemy and instead using `literal(..., type=TYPE)`
- Removed usage of handcrafted dialect handling for patterns (postgres uses ~ vs. sqlite uses REGEXP) and instead using the builtin sqlalchemy function for pattern
- Removed the `NOT NULL` Check for patterns (Only check string values with REGEX, if the attribute is not allowed to be null it should be stated via the `required` field
- Added a warning if users specify a dialect which is not implemented yet 
- Added missing tests to improve code coverage

</details>

# General Remark
When I started building this, I really liked the capabilities of `rules` for data validation. However during implementation, I realized that the `rules` is not a central part of linkml (yet?). The [documentation](https://linkml.io/linkml/schemas/advanced.html#rules) of rules is not really large and I had difficulties understanding all cases that need to be taken into account. Users of the generator will have the same problem eventually when they try to formulate `rules`.

* Is it even a good idea to include rules to the sqlvalidation at this point? I am also fine with closing this PR and adding a new PR only for the bug fixes. Rules could then be added later as well.


FlorianK13:
Looking at the [Generator Governance](https://linkml.io/linkml/maintainers/generator-governance.html), @cmungall has the closest relation to these changes based on the work on the sql / sqlalchemy generators

tests/linkml/test_generators/test_sqlvalidationgen.py
```diff
@@ -501,3 +561,407 @@ def test_validation_interop_with_invalid_data(input_path, tmp_path):
     )
 
     conn.close()
+
+
+def _schema_with_rules(rules, slots=None, class_name="LivingThings"):
```
cmungall: add type hints

cmungall (changes requested):
Thank you!

haven't had time to fully review, but it looks like this switches the rule generator to use the source schema rather than the relationalized version? If so, this will result in errors if the source schema doesn't fit in a pure relational profile

FlorianK13:
> Thank you!
> 
> haven't had time to fully review, but it looks like this switches the rule generator to use the source schema rather than the relationalized version? If so, this will result in errors if the source schema doesn't fit in a pure relational profile

Thanks, I did not have this in mind. I'll work on it and convert this PR back to draft until then.

tests/linkml/test_generators/test_sqlvalidationgen.py
FlorianK13: Done

FlorianK13:
I now use the relational schema view for everything except rules. There is a second loop over the non transformed schema view to extract rules. I also added the **General Remark** to the first message in this issue about the general relevance of this PR.

packages/linkml/src/linkml/generators/sqlvalidationgen.py
```diff
@@ -223,6 +270,92 @@ def _generate_header(self) -> str:
...
+        Uses SQLAlchemy's ``regexp_match`` which compiles to the dialect-specific
+        syntax (PostgreSQL: ``~``, SQLite: ``REGEXP``).
+
+        :param col: SQLAlchemy column expression
+        :param pattern: Regular expression pattern string
+        :param negate: If True, return the violation condition; if False, the conformance condition
+        :return: SQLAlchemy condition
+        """
+        match = col.regexp_match(literal(pattern, type_=Text()))
+        return ~match if negate else match
```
kevinschaper: The old `_generate_pattern_violations` wrapped this in `and_(col.isnot(None), pattern_check)`. That NULL guard is gone here. On PostgreSQL `~` propagates NULL fine, but on SQLite `REGEXP` compiles to a call into a user-registered scalar function (e.g. `re.search(pattern, value) is not None`), and SQLite passes `NULL` straight through rather than short-circuiting — so a NULL value in a non-required, pattern-constrained column raises `TypeError` inside the UDF, surfaced as `sqlite3.OperationalError`, crashing the whole `UNION ALL` query. This is untested: the interop tests explicitly disable pattern checks for SQLite. Suggest restoring the `col.isnot(None)` guard here regardless of dialect.

packages/linkml/src/linkml/generators/sqlvalidationgen.py
```diff
@@ -460,24 +563,146 @@ def _generate_unique_key_violations(
...
+        if slot_condition.equals_string is not None:
+            lit = literal(slot_condition.equals_string, type_=Text())
+            conditions.append(col != lit if negate else col == lit)
+
+        if slot_condition.equals_number is not None:
+            lit_num = _literal_num(slot_condition.equals_number)
+            conditions.append(col != lit_num if negate else col == lit_num)
+
+        if slot_condition.equals_string_in:
+            lit_vals = [literal(v, type_=Text()) for v in slot_condition.equals_string_in]
```
kevinschaper, line 711: These negated conditions (`col != lit`, `col.notin_(lit_vals)`) are used as the violation/negated branch in rule postcondition checks. In SQL three-valued logic `NULL != 'x'` and `NULL NOT IN (...)` both evaluate to NULL, not TRUE — so a row where this column is actually NULL (which should clearly count as a rule violation, since NULL != any required value) is silently excluded from the violation results. Suggest a NULL-safe form for the negated branch, e.g. `or_(col != lit, col.is_(None))`. None of the new tests exercise a NULL value in an equals-type postcondition column.

packages/linkml/src/linkml/generators/sqlvalidationgen.py
```diff
@@ -460,24 +563,146 @@ def _generate_unique_key_violations(
...
+        """
+        Convert an AnonymousClassExpression to a composite WHERE clause.
+
+        :param tbl: SQLAlchemy table object
+        :param expression: AnonymousClassExpression with slot_conditions
+        :param negate: If True, negate the expression (for postcondition violation detection). Note that negated
+        conditions are concatenated with OR -> De Morgan's law.
+        :return: SQLAlchemy condition or None
+        """
+        if not expression or not expression.slot_conditions:
```
kevinschaper: This early return fires before the `any_of`/`all_of`/`none_of`/`exactly_one_of` warning loop below ever runs. So a rule `precondition` expressed purely via `any_of` (no top-level `slot_conditions`) returns `None` here with **no warning logged**, and in `_generate_rule_violations` (`if pre is not None: where_parts.append(pre)`) the precondition is simply dropped — the postcondition check then applies to every row in the table instead of just the intended subset, producing false-positive violations. The one `any_of` test in this PR puts `any_of` on the *postcondition* (which still has `slot_conditions`), so this branch isn't covered. Consider moving the unsupported-combinator warning above this early return, and/or having the caller distinguish "no precondition" from "precondition present but untranslatable."

packages/linkml/src/linkml/generators/sqlvalidationgen.py
```diff
@@ -460,24 +563,146 @@ def _generate_unique_key_violations(
...
+        # Collect all referenced column names
+        col_names = {identifier_slot_name}
+        if rule.preconditions and rule.preconditions.slot_conditions:
+            col_names.update(rule.preconditions.slot_conditions.keys())
+        col_names.update(rule.postconditions.slot_conditions.keys())
+
+        postcondition_slot_names = list(rule.postconditions.slot_conditions.keys())
+        column_name_label = ",".join(postcondition_slot_names)
+
+        tbl = table(class_name, *[column(c) for c in col_names])
```
kevinschaper: `col_names` here comes straight from `rule.preconditions/postconditions.slot_conditions.keys()` — raw schema slot names, never passed through `underscore(slot.alias or slot.name)` like every other query path in this file (see the main loop building `induced_slots`). A rule referencing a slot with an `alias`, or a name `underscore()` would rewrite, will generate `column(...)` referencing a name that doesn't match the real/underscored SQL column used everywhere else for that class. Worth resolving these names through the same alias map the main loop already builds before constructing `tbl`/`col_names`. Untested — all new rule tests use plain, alias-free, already-underscored slot names.

packages/linkml/src/linkml/generators/sqlvalidationgen.py
```diff
@@ -384,66 +474,79 @@ def _generate_enum_violations(
...
         :param identifier_slot_name: Name of the identifier slot
         :return: SQLAlchemy select object or None
         """
-        # Get column names from unique key slots
+        # Get column names from unique key slots (underscored for SQL)
         columns = []
         for slot_name in uk.unique_key_slots:
-            if slot_name in class_def.attributes:
-                columns.append(slot_name)
+            sql_name = underscore(slot_name)
```
kevinschaper: `slot_names` (built by the caller a few lines up) is `{underscore(slot.alias or slot.name) for slot in induced_slots}` — alias-resolved. But here `sql_name = underscore(slot_name)` underscores the raw slot *name* from `uk.unique_key_slots` (which is typed as `SlotDefinitionName`, not an alias). For a slot with `alias != name` referenced in a unique key, these won't match, so the column is silently dropped from `columns` (and the whole query is skipped via `if not columns: return None` if it was the only column) — no warning. Suggest resolving `uk.unique_key_slots` entries through the same alias map used to build `slot_names`. Untested — no unique-key test in this PR uses an aliased slot.

packages/linkml/src/linkml/generators/sqlvalidationgen.py
```diff
@@ -182,8 +172,32 @@ def generate_validation_queries(self, **kwargs: dict[str, Any]) -> str:
...
-                    query = self._generate_unique_key_violations(class_name, class_def, uk, identifier_slot_name)
+                    query = self._generate_unique_key_violations(class_name, slot_names, uk, identifier_slot_name)
+                    if query is not None:
+                        query_objects.append(query)
+
+        # Check rules (precondition/postcondition constraints)
+        if self.check_rules:
+            # We need to iterate over the untransformed schema, since the
+            # RelationalModelTransformer removes rules
+            untransformed_sv = SchemaView(self.schema)
```
kevinschaper: Minor efficiency note: `RelationalModelTransformer.transform()` builds a brand-new `SchemaDefinition` rather than mutating the `SchemaView` it's given in place (it just omits `rules` when copying class attrs) — so the `SchemaView(schema)` constructed a few lines up (and handed to the transformer, with imports already merged) still has `class_def.rules` intact. Building a second full `SchemaView` here just to read rules seems avoidable — could this reuse the original `sv` (or `sqltr.schemaview`) instead?

packages/linkml/src/linkml/generators/sqlvalidationgen.py
```diff
@@ -1,25 +1,33 @@
...
 from linkml.utils.generator import Generator, shared_arguments
-from linkml_runtime.linkml_model.meta import ClassDefinition, SlotDefinition
+from linkml_runtime.linkml_model.meta import SlotDefinition
+from linkml_runtime.utils.formatutils import underscore
 from linkml_runtime.utils.schemaview import SchemaView
 
 logger = logging.getLogger(__name__)
 
 
+def _literal_num(val):
```
kevinschaper, line 32: Per repo convention ("always use type hints"), this and a few other new helpers below (`_slot_condition_to_sqlalchemy`, `_class_expression_to_sqlalchemy`, `_generate_rule_violations`) have docstrings but no parameter/return type hints, unlike their sibling methods added in the same diff (e.g. `_pattern_condition(col, pattern: str, negate: bool = False)`).

packages/linkml/src/linkml/generators/sqlvalidationgen.py
```diff
@@ -460,24 +563,146 @@ def _generate_unique_key_violations(
...
+            column_name=uk.unique_key_name,
+            constraint_type="unique_key",
+            identifier_slot_name=identifier_slot_name,
+            invalid_value=concat_expr,
+            tbl=tbl,
+            where_condition=where_clause,
         )
 
-        return query
+    def _slot_condition_to_sqlalchemy(self, tbl, slot_name, slot_condition, negate=False):
```
kevinschaper: Missing type hints on `tbl`, `slot_name`, `slot_condition`, `negate`, and return type — see note on `_literal_num` above.

packages/linkml/src/linkml/generators/sqlvalidationgen.py
```diff
@@ -460,24 +563,146 @@ def _generate_unique_key_violations(
...
+
+        if slot_condition.pattern is not None:
+            conditions.append(self._pattern_condition(col, slot_condition.pattern, negate=negate))
+
+        if slot_condition.required:
+            conditions.append(self._required_condition(col, negate=negate))
+
+        return conditions
+
+    def _class_expression_to_sqlalchemy(self, tbl, expression, negate=False):
```
kevinschaper: Missing type hints on `tbl`, `expression`, `negate`, and return type — see note on `_literal_num` above.

packages/linkml/src/linkml/generators/sqlvalidationgen.py
```diff
@@ -460,24 +563,146 @@ def _generate_unique_key_violations(
...
+        if not all_conditions:
+            return None
+
+        if negate:
+            # De Morgan's law: negating AND → OR
+            return or_(*all_conditions)
+        else:
+            return and_(*all_conditions)
+
+    def _generate_rule_violations(self, class_name, rule, identifier_slot_name):
```
kevinschaper: Missing type hints on `class_name`, `rule`, `identifier_slot_name`, and return type — inconsistent with the fully-typed `_generate_required_violations(self, class_name: str, slot: SlotDefinition, identifier_slot_name: str)` it parallels.

tests/linkml/test_generators/test_sqlvalidationgen.py
```diff
@@ -501,3 +562,411 @@ def test_validation_interop_with_invalid_data(input_path, tmp_path):
...
+    for s in slots:
+        b.add_slot(s)
+    b.add_class(class_name, slots=[s.name for s in slots])
+    b.add_defaults()
+    # Attach rules directly
+    b.schema.classes[class_name].rules = rules
+    return b.schema
+
+
+def test_simple_rule_equals_string_and_maximum_value():
```
kevinschaper, line 676: This and the next several tests (`test_rule_equals_string_in`, `test_rule_minimum_value_postcondition`, `test_rule_postcondition_only`, `test_rule_multiple_postcondition_slots`, `test_rule_equals_number`, `test_rule_required_slot_condition`) share identical structure and differ only in the slot-condition kwarg and expected SQL fragment. Per repo convention ("use modern pytest idioms, including `@pytest.mark.parametrize` to test for combinations of inputs"), these look like a good fit for a single parametrized test.

kevinschaper (changes requested):
Hi @FlorianK13,

Thanks for the PR, and for building this generator, at some point I'm hoping to add duckdb support and use it for a project of mine. 

I attached some claude-generated inline review comments, and I'm marking as requesting review based on a few that look like real bugs or important tidying. 

I think rules are less supported and shouldn't be used in place of better supported constraints or good data modeling practices, but supporting them here in the SQL validation generator is definitely a good thing.

FlorianK13:
Hi @kevinschaper - duckdb support would be awesome, also to directly use it with .csv files. Right now I am super busy with other stuff, so this PR will again take some more time.

packages/linkml/src/linkml/generators/sqlvalidationgen.py
FlorianK13: As suggested, I restored the NOT NULL part of the pattern match for both dialects in 2bffe52ab838101f672f569683ceaa673c5ac884

packages/linkml/src/linkml/generators/sqlvalidationgen.py
FlorianK13: Done in 450f7b2e11a690d12d55a45ddf90c5af1dfd8695

packages/linkml/src/linkml/generators/sqlvalidationgen.py
FlorianK13: Done in 76ce8ce80290846ea08f64359a8955c9fe7bc8fc - it is made sure now that if no precondition is generated even though it is given in the schema then postconditions are skipped as well.

packages/linkml/src/linkml/generators/sqlvalidationgen.py
FlorianK13: Done in cb6a765f39a281854d6beeef18b2a466dc6e1c7b - the alias map was created and in the next commit is also used for correct mapping for validation logic outside of rules.

packages/linkml/src/linkml/generators/sqlvalidationgen.py
FlorianK13: Created one static `_column_name_map` function in 383cb9fcdfaf65e362febc333b90a7ec4537b1b9 that is used throughout the generator.

packages/linkml/src/linkml/generators/sqlvalidationgen.py
FlorianK13: Added type hints in 48fb83c3eccd38fe26c00f67382a14a82e324830

packages/linkml/src/linkml/generators/sqlvalidationgen.py
FlorianK13: Added type hints in 48fb83c3eccd38fe26c00f67382a14a82e324830

packages/linkml/src/linkml/generators/sqlvalidationgen.py
FlorianK13: Added type hints in 48fb83c3eccd38fe26c00f67382a14a82e324830

tests/linkml/test_generators/test_sqlvalidationgen.py
FlorianK13: Using one parametrized test in 1ce3de9bac120803236d9798e08b366538289885

packages/linkml/src/linkml/generators/sqlvalidationgen.py
FlorianK13: Added type hints in 48fb83c3eccd38fe26c00f67382a14a82e324830

packages/linkml/src/linkml/generators/sqlvalidationgen.py
FlorianK13: Done, untransformed SchemaView is created once and then the corresponding schema object  is used throughout the generator.

kevinschaper (approved):
beautiful! thanks @FlorianK13!

PR #3326, opened 2026-03-25
oocodegen/javagen: Add option to honour slot aliases when deriving field names

### LINKED ISSUE #3323: `oocodegen` does not honour the `alias` attribute on slots
When a slot is defined as having an alias, as in the following example (from the `personinfo.yaml` schema in the test suite):

```yaml
slots:
  age_in_years:
    range: integer
    alias: age

classes:
  Person:
    slots:
      - age_in_years
```

then the Python code generator will generate a `Person` class that has a `age` attribute, not a `age_in_years` attribute. This I believe is the **expected** behaviour, as per the documentation of the `alias` slot in LinkML’s meta-schema:

> the name used for a slot in the context of its owning class. If present, this is used instead of the actual slot name.

But the `oocodegen` generator (on which the Java code generator is based) does not follow this rule, and systematically use the “actual slot name” (so in the example above, it would create a `Person` class with a `ageInYears` field).

I believe `oocodegen` should be made, at least optionally, to follow both the behaviour of the Python code generator and LinkML’s own documentation, and use aliases when they are present.

gouttegd:
The current behaviour of `OOCodeGenerator` is to always use the _name_ of a slot to derive the name of the corresponding field in the object-oriented representation of a schema.

This is different from the behaviour of the Python code generator, which uses the slot alias if it exists – in agreement with the documentation of the `alias` slot in LinkML meta-model, which explicitly says that, when present, the value of the `alias` slot should be used wherever the slot name is expected.

This PR adds a `use_aliases` option to `OOCodeGenerator` to follow the behaviour of the Python code generator (and of the spec). For backwards compatibility, that option is set to False by default, so the current behaviour is preserved (aliases are ignored) unless the new behaviour is explicitly requested.

A new `--use-aliases` option is added to the command-line interface of the Java generator.

(I leave it to LinkML devs to decide whether to set option to `True` by default; for what it’s worth, I do believe the new behaviour should be enabled by default, since (1) that’s what the Python code generator does and (2) that’s the _correct_ behaviour as per the documentation of the `alias` slot.)

The PR also updates the `org.incenp.linkml` template variant to make sure the slot _alias_ (if present), not the slot _name_, is given to the Java runtime if needed.

closes #3323


kevinschaper (approved):
looks good!

PR #3328, opened 2026-03-26
normalize: Correctly recognise linkml:Any classes.

### LINKED ISSUE #3246: `linkml-normalize` is seemingly unaware of the meaning of `linkml:Any`
**Describe the bug**
Given the schema and sample YAML file below, the `linkml-normalize` command reports a _ClosedClassConstraint_ violation:

```sh
$ uv run linkml-normalize -C Foo -s schema.yaml sample.yaml
# Errors:
type: ClosedClassConstraint
subject: '{''description'': ''I am an instance of Anything'', ''but_can_i_really_contain_anything'':
  {''according_to_linkml_validate'': ''yes!'', ''according_to_linkml_normalize'':
  ''no :(''}}'
instantiates: Anything
predicate: description
type: ClosedClassConstraint
subject: '{''description'': ''I am an instance of Anything'', ''but_can_i_really_contain_anything'':
  {''according_to_linkml_validate'': ''yes!'', ''according_to_linkml_normalize'':
  ''no :(''}}'
instantiates: Anything
predicate: but_can_i_really_contain_anything
```

whereas the `linkml-validate` command reports (as expected) that everything is fine:

```sh
$ uv run linkml validate -C Foo -s schemas.yaml sample.yaml
No issues found
```

Tip of the main branch as of March 3rd (commit 5ed3c0ba7b74daa2b2cad9da6fded616c6c2da35).

* `schema.yaml`:

```yaml
id: https://example.org/anything
name: anything
default_range: string
imports:
  - linkml:types
prefixes:

classes:
  Anything:
    class_uri: linkml:Any

  Foo:
    attributes:
      name:
      extra:
        range: Anything
```

* `sample.yaml`:

```yaml
name: test
extra:
  description: I am an instance of Anything
  but_can_i_really_contain_anything:
    according_to_linkml_validate: yes!
    according_to_linkml_normalize: no :(
```

According to the [documentation of linkml:Any](https://linkml.io/linkml/schemas/advanced.html#linkml-any-type), the example above should be valid: the `extra` slot has for range a class that “takes the role of `linkml:Any`” (by being marked with `class_uri: linkml:Any`), so that slot can legitimately contain any object.

`linkml-validate` agrees with that, but it seems that `linkml-normalize` treats the `Anything` class in that example as a completely “normal” class, ignoring its `linkml:Any` role.

gouttegd:
A slot whose range is set to a class that is marked with the special class URI `linkml:Any` can contain any object. Therefore, the normalizer should _not_ flag the presence of unrecognized slots as a ClosedClassConstraint violation -- no constraints are being violated since it is precisely the purpose of such a class to contain arbitrary, unspecified slots.

This commit simply makes the `normalize_object` method of the ReferenceValidator to check early on whether the class we are validating against is a `linkml:Any` class, and if so skip the entire validation.

closes #3246

### DIFF
--- a/tests/linkml/test_metamodel_compat/input/metamodel/datasets.yaml
+++ b/tests/linkml/test_metamodel_compat/input/metamodel/datasets.yaml
@@ -88,7 +88,7 @@ classes:
       - https://specs.frictionlessdata.io/data-package
     slots:
       - resources
-    
+
   DataResource:
     is_a: Information
     description: An individual file or table
@@ -118,7 +118,7 @@ classes:
       double_quote:
       header:
       quote_char:
-      
+
 # ==================================
 # Slots                            #
 # ==================================
@@ -140,18 +140,18 @@ slots:
 
   language:
     description: language in which the information is expressed
-    
+
   publisher:
     slot_uri: dcterms:publisher
     range: uriorcurie
-    
+
   issued:
     slot_uri: dcterms:issued
     range: datetime
 
   page:
     slot_uri: dcat:landingPage
-    
+
   dialect:
     slot_uri: csvw:dialect
 
@@ -185,7 +185,7 @@ slots:
     is_a: hash
   md5:
     is_a: hash
-    
+
   media_type:
     # range: MediaTypeEnum
     range: string
@@ -209,7 +209,7 @@ slots:
   conforms_to_class:
     is_a: conforms_to
     description: class in schema which the data object instantiates
-      
+
   profile:
     range: uriorcurie
     exact_mappings:
@@ -222,13 +222,13 @@ slots:
     slot_uri: dcat:keyword
     exact_mappings:
       - schema:keywords
-    
+
   themes:
     singular_name: theme
     multivalued: true
     range: uriorcurie
     slot_uri: dcat:theme
-    
+
   resources:
     multivalued: true
     range: DataResource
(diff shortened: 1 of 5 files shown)


matentzn:
@gouttegd let me know if you are happy if you would like to adopt @sneakers-the-rat personal preference here - I personally have no problem with issue based test design, so I gave my approval.

gouttegd:
@matentzn I have no personal preferences here, but I note that [LinkML’s contributor documentation](https://github.com/linkml/linkml/blob/main/docs/maintainers/contributing.md#linkml-testing-framework) does recommend “issue based test design”:

> Outside of generators, users will want to create issue specific unit tests.

And incidentally re-reading that page made me realize that I should have named the test file `test_linkml_issue_3246.py`, _not_ `test_issue_3246.py`:

> While creating a test for a bug, you should make a file called `test_linkml_issue_NNN.py`. […] You will see a number of issues which are named `test_issue_NNN.py`. The numbers and convention for those issues are with reference to the old [biolinkml](https://github.com/biolink/biolinkml) issue numbering convention.

My bad.

sneakers-the-rat:
> I personally have no problem with issue based test design, so I gave my approval.

> While creating a test for a bug, you should make a file called 

especially seeing how doing it this way is recommended in the contributor docs, i retract my comment there. (i shouldn't have said anything lol i was just being nitpicky, completely good with me and absolutely do not want to slow things down for that style pref.)

PR #3332, opened 2026-03-28
feat: add C++ header generator (gen-cpp-header)

### LINKED ISSUE #3333: C++ header generator (gen-cpp-header)
LinkML has code generators for Python, TypeScript, Java, Go, Rust but nothing for C or C++.

I needed C++ header generation for a project where a LinkML schema defines the type system (game engine ontology with entity types, material enums, constraint structs). The schema is the source of truth and headers get generated at build time.

Put together a gen-cpp-header generator following the Go generator's architecture (extends OOCodeGenerator, uses SchemaView, Jinja2 templates). Handles classes to structs with public inheritance, enums to enum class with string conversion helpers, std::optional for optional fields, std::vector for multivalued.

Implementation in #3332, 26 tests, ruff-clean.

Happy to iterate on feedback.

Kieleth:
Closes #3333

Adds gen-cpp-header, a C++17 header generator following the same architecture as the Go generator (OOCodeGenerator + SchemaView + Jinja2 templates).

Given a LinkML schema it produces:

- `enum class` from enums with inline `to_string()`/`from_string()` helpers
- `struct` from classes with `public` inheritance for `is_a`
- `std::optional<T>` for optional fields, `std::vector<T>` for multivalued
- Topological ordering so base structs come before derived ones
- Automatic `#include` tracking based on types used
- Doxygen comments from descriptions
- Namespace wrapping

```bash
gen-cpp-header schema.yaml --namespace my::project
```

Example output for a schema with an entity hierarchy and material enum:

```cpp
#pragma once

#include <cstdint>
#include <optional>
#include <string>

namespace my_schema {

enum class MaterialType {
    WOOD,
    STONE,
    IRON
};

inline const char* to_string(MaterialType value) { /* ... */ }
inline bool from_string(const char* str, MaterialType& out) { /* ... */ }

struct WorldEntity {
    std::string name;
    float position_x;
};

struct LivingEntity : public WorldEntity {
    float health;
    std::optional<MaterialType> material = std::nullopt;
};

} // namespace my_schema
```

26 tests covering struct generation, inheritance, type mapping, enums, multivalued fields, optional fields, includes, namespaces, multiline descriptions, CLI options. Ruff clean.


Kieleth:
Hey @kevinschaper and @matentzn let me know if any questions in the Review. I've been using this branch for a project of mine, and I can say with confidence that it is working as expected. 
BTW: amazing work with LinkML, I love it and use it in many projects. It's an awesome idea.

packages/linkml/src/linkml/generators/cppgen/templates/module.h.jinja
```diff
@@ -0,0 +1,20 @@
+// Code generated by linkml gen-cpp-header. DO NOT EDIT.
+#pragma once
+{% if includes and includes.strip() %}
```
noelmcloughlin: The `cppgen/template.py` file has no `strip()` method so `includes.strip()` would fail since the object is not string. some aternative approaches are:
```
{% if includes and includes.includes %}  # simplest solution ..  or
{% if includes|length > 0 %}   # jinja2 length filter .. or
{% if includes %}   # see note
```
Note: 
assume jinja2 treats as truthy (if non-null), 
 or add __bool__() method to `cppgen/template.py -> Includes` to return false when empty.

packages/linkml/src/linkml/generators/cppgen/template.py
```diff
@@ -0,0 +1,175 @@
...
+    template: ClassVar[str]
+    _environment: ClassVar[Environment] = Environment(
+        loader=PackageLoader("linkml.generators.cppgen", "templates"),
+        trim_blocks=True,
+        lstrip_blocks=True,
+    )
+    _environment.filters["cpp_comment"] = lambda text, indent="": "\n".join(
+        f"{indent}/// {line}" if line.strip() else f"{indent}///" for line in text.splitlines()
+    )
+    _environment.filters["upper_snake"] = lambda text: (
```
noelmcloughlin: The `upper_snake` filter is not used so could be dropped.

packages/linkml/src/linkml/generators/cppgen/template.py
```diff
@@ -0,0 +1,175 @@
...
+        "".join(f"_{c}" if c.isupper() and i > 0 else c for i, c in enumerate(text)).upper().lstrip("_")
```
noelmcloughlin: The `upper_snake` implementation looks incorrect (given "INPUT" it produces "I_N_P_U_T") which is not snake case. So should  be dropped and use `cppgen.py -> _to_upper_snake()` instead when needed.

packages/linkml/src/linkml/generators/cppgen/cppgen.py
```diff
@@ -0,0 +1,545 @@
...
+Pass a directory containing templates with the same name as any of the default
+CppTemplateModel templates to override them.
+
+\b
+Available templates to override:
+"""
+    + "\n".join(["- " + name for name in _TEMPLATE_NAMES]),
+)
+@click.version_option(__version__, "-V", "--version")
+@click.command(name="cpp-header")
```
noelmcloughlin: pyproject.toml has `gen-cpp-header` not `cpp-header.

noelmcloughlin (changes requested):
Thanks for interesting PR @Kieleth 
I left a review and LGTM if addressed.

packages/linkml/src/linkml/generators/cppgen/templates/module.h.jinja
Kieleth: nice catch, ty

packages/linkml/src/linkml/generators/cppgen/template.py
Kieleth: yup, good one

packages/linkml/src/linkml/generators/cppgen/template.py
Kieleth: yup

packages/linkml/src/linkml/generators/cppgen/cppgen.py
Kieleth: nice, fixed

Kieleth:
@noelmcloughlin Nice review, thanks for catching those, new push with the fixes. Plz take a look.

packages/linkml/src/linkml/generators/cppgen/templates/module.h.jinja
```diff
@@ -0,0 +1,20 @@
+// Code generated by linkml gen-cpp-header. DO NOT EDIT.
+#pragma once
+{% if includes and includes|trim %}
```
noelmcloughlin, line 3: LGTM, `trim` is a classic Jinja2 filter.

noelmcloughlin (approved):
Nice work @Kieleth
@matentzn  this could be merged, new feature, good code test coverage.

Kieleth:
hey @noelmcloughlin , @matentzn just checking if we're waiting for a release cycle or... some other reason not to merge? curious.

kevinschaper (approved):
Looks great! thanks @Kieleth, and thanks @noelmcloughlin for the review!

> hey @noelmcloughlin , @matentzn just checking if we're waiting for a release cycle or... some other reason not to merge? curious.

Nope, I was just slow to get to the PR, I'll be spending lots of time this week catching up - sorry for the long wait!

PR #3338, opened 2026-03-31
YARRRML generator bug fixes and E2E tests update

lapkinvladimir:
This PR fixes 7 bugs in the YARRRML generator to match the official YARRRML specification. It also upgrades tests to include strict JSONSchema validation

### What it does

1. **Inlined objects without IDs:** Generates predictable IRIs (like `ex:Child_$(parent_id)`) so data stays connected in the graph. Throws a clear error if it's a list without an ID
2. **XSD datatypes:** Properly outputs specific datatypes (like `datatype: xsd:integer`) per YARRRML [Spec Section 9.5](https://rml.io/yarrrml/spec/#datatypes)
3. **Cleaner YAML:** Formats data sources in a correct way, single-line arrays (e.g., `[data.json~jsonpath, '$[*]']`) so external YAML parsers don't break
4. **Smart iterators:** Automatically uses `$` as the JSONPath iterator if the schema has a `tree_root`, otherwise defaults to `$[*]`
5. **Correct `rdf:type`s:** Stops accidentally assigning the target's class to the subject when linking objects
6. **Mixin fixes:** Merges mixin classes directly into the main class's `rdf:type` instead of creating broken, separate mappings
7. **Identifier triples:** Ensures identifier slots actually generate a triple if they have an explicit `slot_uri` (like `schema:identifier`)

### Tests & JSONSchema validation

The E2E test suite now checks the full pipeline: **JSON -> JSONSchema -> YARRRML -> morph-kgc -> SHACL**

To test different data shapes fairly, I use two validation strategies:
* **For flat data (CSV/TSV):** Since multiple classes can mix in one row, I point the validator to the class using `$ref` and allow extra sibling properties. This checks required fields without crashing on a neighbor's fields
* **For hierarchical data (`tree_root: true`):** Uses JSON validation. Any extra or unexpected fields will instantly fail the test

Added 7 new E2E tests for these specific fixes


docs/generators/yarrrml.rst
```diff
@@ -122,8 +121,10 @@ Overview
   ``rdf:type`` and predicate IRIs (including full IRIs if present)
 - subject from identifier slot (else key; else safe fallback)
 - ``po`` for all class attributes (slot aliases respected)
-- emits ``rdf:type`` as CURIEs or IRIs (depending on availability)
-- JSON by default: ``sources: [[data.json~jsonpath, $.items[*]]]``
+- emits ``a`` (rdf:type) as CURIEs or IRIs (depending on availability), and automatically aggregates mixin classes into this array
+- emits predicate-object mappings for identifier slots if they have an explicit ``slot_uri``
+- JSON by default: ``sources: [[data.json~jsonpath, '$[*]']]`` (or '$' automatically if tree_root is used)
```
niegrzybkowski: Could you describe the tree_root / no tree_root behavior separately?

packages/linkml/src/linkml/generators/yarrrmlgen.py
```diff
@@ -133,24 +207,50 @@ def _prefixes_with_defaults(self) -> dict[str, str]:
...
         key_slot = sv.get_key_slot(c.name)
         if key_slot:
-            return f"{default_prefix}:$({key_slot.name})"
-        return f"{default_prefix}:{c.name}/$(subject_id)"
+            return f"{prefix}:$({key_slot.name})"
 
-    def _po_list_for_class(self, c: ClassDefinition) -> list[dict[str, Any]]:
+        # YARRRML parsers fail to execute RML Joins (`condition: equal`) on generated Blank Nodes.
+        # For inlined objects lacking an ID, we synthesize a deterministic IRI using the parent's ID
+        # to guarantee graph connectivity and avoid broken/orphaned triples during the RDF roundtrip.
```
niegrzybkowski: Please add this reasoning to the docs as well.

packages/linkml/src/linkml/generators/yarrrmlgen.py
```diff
@@ -54,54 +74,108 @@ def as_dict(self) -> dict[str, Any]:
...
+                            is_multi = s.multivalued
+                            break
+
+                    if not has_own_id:
+                        # Iterate over the parent's array, we will access child properties via dot notation
+                        owner_cls = sv.get_class(owner_name)
+                        iterator = self._iterator_for_class(owner_cls)
+                    else:
+                        # Iterate directly over the nested child array
+                        iterator = f"$..{slot_var}[*]" if is_multi else f"$..{slot_var}"
```
niegrzybkowski, line 165: Please add the limitations that `$..{slot}` iterator causes to the docs as well.

packages/linkml/src/linkml/generators/yarrrmlgen.py
```diff
@@ -54,54 +74,108 @@ def as_dict(self) -> dict[str, Any]:
...
+                has_id = False
+                if range_cls:
+                    has_id = (
+                        sv.get_identifier_slot(range_cls.name) is not None
+                        or sv.get_key_slot(range_cls.name) is not None
+                    )
 
                 if inlined is None:
-                    inlined = False
+                    inlined = not has_id
```
niegrzybkowski: Please throw an error if it's not inlined and doesn't have an id. Describe it in docs.
I wouldn't want adding an identifier to an inner class to change the inlining behavior of an outer class.

niegrzybkowski (changes requested):
Please add some more docs about the things happening in the generator, keep track of the limitations.

lapkinvladimir:
@niegrzybkowski requested changes are done

docs/generators/yarrrml.rst
```diff
@@ -168,8 +170,11 @@ Limitations
 - Object slots:
   - ``inlined: false`` → emitted as IRIs
   - ``inlined: true`` → emitted using YARRRML ``mapping`` + ``condition`` (join) pattern
-- Inlined objects without an identifier are assigned a synthetic IRI (e.g. ex:Child_$(parent_id)). However, multivalued inlined objects (lists) still require an identifier.
+- Inlined objects without an identifier are assigned a synthetic IRI using the parent's ID (e.g. ``ex:Child_$(parent_id)``). This ensures graph connectivity and avoids broken/orphaned triples during RDF roundtrips, as many YARRRML parsers fail to execute Joins (``condition: equal``) on Blank Nodes.
```
niegrzybkowski:
```suggestion
- Inlined objects without an identifier are assigned a synthetic IRI using the parent's ID (e.g. ``ex:Child_$(parent_id)``). This ensures graph connectivity and avoids broken/orphaned triples during lifting, as some YARRRML implementations fail to execute joins (``condition: equal``) on Blank Nodes.
```

packages/linkml/src/linkml/generators/yarrrmlgen.py
```diff
@@ -133,24 +214,50 @@ def _prefixes_with_defaults(self) -> dict[str, str]:
...
         key_slot = sv.get_key_slot(c.name)
         if key_slot:
-            return f"{default_prefix}:$({key_slot.name})"
-        return f"{default_prefix}:{c.name}/$(subject_id)"
+            return f"{prefix}:$({key_slot.name})"
 
-    def _po_list_for_class(self, c: ClassDefinition) -> list[dict[str, Any]]:
+        # YARRRML parsers fail to execute RML Joins (`condition: equal`) on generated Blank Nodes.
+        # For inlined objects lacking an ID, we synthesize a deterministic IRI using the parent's ID
+        # to guarantee graph connectivity and avoid broken/orphaned triples during the RDF roundtrip.
```
niegrzybkowski:
```suggestion
        # to guarantee graph connectivity and avoid broken/orphaned triples during lifting.
```

niegrzybkowski (commented):
Two small suggestions, looks good otherwise.

niegrzybkowski (approved):

Ostrzyciel (approved):
Looks good to me as well! Excellent work :)

Ostrzyciel:
@kevinschaper @sierra-moxon could we get this merged? I'm not sure if the per-generator governance model is live already, but I think we meet the criteria for a merge :) https://github.com/linkml/linkml/blob/0fa6f931e9edfc10a6410885e88eed496f6b9249/docs/maintainers/generator-governance.md?plain=1#L719

tests/linkml/test_generators/test_yarrrmlgen.py
```diff
@@ -925,17 +958,343 @@ def test_yarrrml_e2e_multi_owner_inline_raises(tmp_path: Path):
...
     schema_path.write_text(schema, encoding="utf-8")
     data_path.write_text(json.dumps(data), encoding="utf-8")
 
-    with pytest.raises(ValueError, match="used in multiple owners"):
+    with pytest.raises(ValueError, match="without an identifier is used by multiple owners"):
         YarrrmlGenerator(str(schema_path), source=f"{data_path.resolve()}~jsonpath").serialize()
+
+
+@pytest.mark.yarrrml
+def test_fix1_inlined_slots_without_id_synthetic_iri(tmp_path: Path):
```
kevinschaper, line 977: The `test_fix1` … `test_fix7` naming ties these tests to an ordered bug list in the PR description rather than to what they verify. A few concerns:

- When one fails in CI, the name alone doesn't tell a maintainer what broke.
- The numbering invites churn if fixes are reordered or split.
- It's inconsistent with the rest of the file, which uses descriptive names like `test_yarrrml_e2e_inline_object_join_mapping`.

Could you rename these to describe the behavior under test? e.g.:

- `test_fix1_inlined_slots_without_id_synthetic_iri` → `test_inlined_without_id_gets_synthetic_iri`
- `test_fix2_preserve_xsd_datatypes` → `test_xsd_datatype_preserved_on_typed_slot`
- `test_fix3_readable_yaml_flow_lists` → `test_sources_emitted_as_flow_sequence`
- `test_fix4_unusable_defaults_and_tree_root` → `test_tree_root_uses_root_iterator_and_skips_non_root_classes`
- `test_fix5_incorrect_classes_for_non_inlined_slots` → `test_non_inlined_link_does_not_stamp_target_type_on_subject`
- `test_fix6_incorrect_subjects_for_mixins` → `test_mixins_contribute_additional_rdf_types`
- `test_fix7_identifier_missing_triple_mapping` → `test_identifier_slot_with_slot_uri_emits_triple`

tests/linkml/test_generators/test_yarrrmlgen.py
```diff
@@ -363,8 +362,19 @@ def test_yarrrml_e2e_basic_json_morph_shacl(tmp_path: Path):
...
+    schema_person = JsonSchemaGenerator(str(schema_path), top_class="Person").generate()
+    schema_person["$ref"] = "#/$defs/Person"
+    schema_person["$defs"]["Person"]["additionalProperties"] = True
+    schema_org = JsonSchemaGenerator(str(schema_path), top_class="Organization").generate()
+    schema_org["$ref"] = "#/$defs/Organization"
+    schema_org["$defs"]["Organization"]["additionalProperties"] = True
+    for item in DATA_BASIC:
+        if "person_id" in item:
+            jsonschema.validate(instance=item, schema=schema_person)
+        elif "org_id" in item:
```
kevinschaper, line 375: The `elif "org_id" in item` branch seems to be dead code in this test. Every element of `DATA_BASIC` has *both* `person_id` and `org_id` (the data is denormalized — each row carries a person and their employer's org fields), so the `if` branch always wins and `schema_org` is never exercised.

Two ways to resolve:

1. If the intent is to validate both shapes, split `DATA_BASIC` into separate person rows and org rows, and use two separate `if` checks (not `elif`).
2. If only `Person` is meaningful here, drop `schema_org` and the `elif` entirely.

packages/linkml/src/linkml/generators/yarrrmlgen.py
```diff
@@ -172,50 +279,80 @@ def _po_list_for_class(self, c: ClassDefinition) -> list[dict[str, Any]]:
...
-                                ],
-                            },
+                if range_id:
+                    if multivalued:
+                        left = f"$({var}[*].{range_id.name})"
+                    else:
+                        left = f"$({var}.{range_id.name})"
+                    right = f"$({range_id.name})"
+                elif parent_id:
+                    left = f"$({parent_id.name})"
```
kevinschaper, line 318: This fallback branch — used when the child class has no identifier/key but the parent does — generates a degenerate self-join: `$(parent_id) == $(parent_id)`. In practice this only fires for id-less inlined children, which the earlier validation at yarrrmlgen.py:144-159 and the multi-owner check at :166-173 are supposed to route through the synthetic-IRI path rather than a join.

A couple of concerns:

- If this branch is genuinely reachable, a Cartesian self-join is almost certainly not what we want and deserves its own E2E test to pin down the emitted graph.
- If it's *not* reachable (i.e. the earlier validation covers every case), this is silent dead code and should either be removed or replaced with an `AssertionError` / `raise ValueError("unreachable: …")` so future regressions surface immediately.

Could you either add a targeted test for this path, or assert unreachable with a comment explaining why the earlier validation makes it so?

kevinschaper (dismissed):
Thanks for the thorough spec-compliance pass and the much stronger E2E tests, and apologies for the slow review again.

I added a few (claude generated, lightly edited) inline comments. I don't think any of these are critical to address, so I'll approve and hold off for a day before merging, if you want to address any of them.

tests/linkml/test_generators/test_yarrrmlgen.py
kevinschaper, line 977: The `test_fix1` … `test_fix7` naming ties these tests to an ordered bug list in the PR description rather than to what they verify. A few concerns:

- When one fails in CI, the name alone doesn't tell a maintainer what broke.
- The numbering invites churn if fixes are reordered or split.
- It's inconsistent with the rest of the file, which uses descriptive names like `test_yarrrml_e2e_inline_object_join_mapping`.

Could you rename these to describe the behavior under test? e.g.:

- `test_fix1_inlined_slots_without_id_synthetic_iri` → `test_inlined_without_id_gets_synthetic_iri`
- `test_fix2_preserve_xsd_datatypes` → `test_xsd_datatype_preserved_on_typed_slot`
- `test_fix3_readable_yaml_flow_lists` → `test_sources_emitted_as_flow_sequence`
- `test_fix4_unusable_defaults_and_tree_root` → `test_tree_root_uses_root_iterator_and_skips_non_root_classes`
- `test_fix5_incorrect_classes_for_non_inlined_slots` → `test_non_inlined_link_does_not_stamp_target_type_on_subject`
- `test_fix6_incorrect_subjects_for_mixins` → `test_mixins_contribute_additional_rdf_types`
- `test_fix7_identifier_missing_triple_mapping` → `test_identifier_slot_with_slot_uri_emits_triple`

tests/linkml/test_generators/test_yarrrmlgen.py
kevinschaper, line 375: The `elif "org_id" in item` branch is dead code in this test. Every element of `DATA_BASIC` has *both* `person_id` and `org_id` (the data is denormalized — each row carries a person and their employer's org fields), so the `if` branch always wins and `schema_org` is never exercised.

Two ways to resolve:

1. If the intent is to validate both shapes, split `DATA_BASIC` into separate person rows and org rows, and use two separate `if` checks (not `elif`).
2. If only `Person` is meaningful here, drop `schema_org` and the `elif` entirely.

As written, the test silently gives less coverage than it appears to.

packages/linkml/src/linkml/generators/yarrrmlgen.py
kevinschaper, line 318: This fallback branch — used when the child class has no identifier/key but the parent does — generates a degenerate self-join: `$(parent_id) == $(parent_id)`. In practice this only fires for id-less inlined children, which the earlier validation at yarrrmlgen.py:144-159 and the multi-owner check at :166-173 are supposed to route through the synthetic-IRI path rather than a join.

A couple of concerns:

- If this branch is genuinely reachable, a Cartesian self-join is almost certainly not what we want and deserves its own E2E test to pin down the emitted graph.
- If it's *not* reachable (i.e. the earlier validation covers every case), this is silent dead code and should either be removed or replaced with an `AssertionError` / `raise ValueError("unreachable: …")` so future regressions surface immediately.

Could you either add a targeted test for this path, or assert unreachable with a comment explaining why the earlier validation makes it so?

kevinschaper (approved):
Thanks for the thorough spec-compliance pass and the much stronger E2E tests, and apologies for the slow review again.

I added a few (claude generated, lightly edited) inline comments. I don't think any of these are critical to address, so I'll approve and hold off for a day before merging, if you want to address any of them.

Issue #3196, opened 2026-02-24
Pydantic converter doesn't exclude abstract classes from emitted designates_type slots' typing unions


colinrsmall:
---
name: Bug report
about: Create a report to help us improve
title: ''
labels: bug
assignees: ''

---

**Describe the bug**

When using the Pydantic converter to generate Pydantic code for a given schema, abstract classes with a `designates_type` type slot are listed in the typing union for slots of the type of the abstract class. This does not seem like intended behavior, as I would expect the abstract class to be explicitly excluded from the typing union. 

This is more straightforward to see in the example below.

**Version of LinkML you are using** 

v1.9.6

**Please provide a schema (and if applicable, a data file) that replicates the issue**

```yaml
id: https://w3id.org/test
name: Test

imports:
  - linkml:types

classes:
  Foo:
    abstract: true
    attributes:
      foo_field: 
        range: string
        required: true
      type:
        range: string
        designates_type: true

  Bar:
    is_a: Foo
    attributes:
      bar_field: 
        range: string
        required: true
  Baz:
    is_a: Foo
    attributes:
      baz_field: 
        range: string
        required: true

  Qux:
    attributes:
      foo_object: 
        range: Foo
        required: true
```

Running the Pydantic generator emits the following Pydantic code:

```python
class Foo(ConfiguredBaseModel):
    foo_field: str = Field(default=...)
    type: Literal["Foo"] = Field(default="Foo")


class Bar(Foo):
    bar_field: str = Field(default=...)
    foo_field: str = Field(default=...)
    type: Literal["Bar"] = Field(default="Bar")


class Baz(Foo):
    baz_field: str = Field(default=...)
    foo_field: str = Field(default=...)
    type: Literal["Baz"] = Field(default="Baz")


class Qux(ConfiguredBaseModel):
    foo_object: Union[Foo,Bar,Baz] = Field(default=...)
```

Since `Foo` is marked as abstract, I would expect the type hint for the `foo_object` field of the `Qux` class to be `Union[Bar,Baz]`. At current, Pydantic would validate and accept a `Foo` for the `foo_object` slot, which should otherwise be disallowed since `Foo` is abstract.

colinrsmall:
[...]

PR #3344, opened 2026-04-01
Update get_class_slot_range() in pydanticgen to exclude abstract classes when generating inlined type unions

colinrsmall:
## Summary

Fixes #

* Addresses https://github.com/linkml/linkml/issues/3196
* Stops Pydantic generator from incorrectly including abstract classes when generating inlined type unions

## How was this tested?

* The example schema from https://github.com/linkml/linkml/issues/3196 was used to generate a Pydantic module in which an object `Qux` has a slot `foo_object` of type `Foo`. `Foo` is abstract and there exists two non-abstract child classes of `Foo`, `Bar` and `Baz`.
* The type union of that Pydantic module was inspected as `annotation = mod.Qux.model_fields["foo_object"].annotation`
* We assert that `Foo` (the abstract class in the example schema) is not included in the type union of the `foo_object` attribute.

## Areas of uncertainty

* I'm not sure what expected behavior is for the Pydantic generator if a user attempts to generate a Pydantic module with a slot of type some abstract class when no non-abstract child classes of that class were defined (e.g. if we remove `Bar` and `Baz` from the test schema). Is this even valid LinkML?

## Checklist


packages/linkml/src/linkml/generators/pydanticgen/pydanticgen.py
```diff
@@ -769,7 +769,11 @@ def get_class_slot_range(self, slot_range: str, inlined: bool, inlined_as_list:
                 len([x for x in sv.class_induced_slots(slot_range) if x.designates_type]) > 0
                 and len(sv.class_descendants(slot_range)) > 1
             ):
-                descendants = [self._get_class_python_name(c) for c in sv.class_descendants(slot_range)]
+                descendants = [
+                    self._get_class_python_name(c)
+                    for c in sv.class_descendants(slot_range)
+                    if not sv.get_class(c).abstract
+                ]
```
kevinschaper, line 776: The `len(sv.class_descendants(slot_range)) > 1` gate on line 770 counts abstract classes, so a hierarchy where every descendant is abstract (abstract root + abstract child) still enters this branch. With the new filter, `descendants` becomes `[]` and line 777 emits `Union[]`, which fails later in `compile_module` with `SyntaxError: invalid syntax`. Reproducer: abstract `Foo` with `designates_type` and an abstract `is_a` child, referenced by a concrete slot.

An abstract range with no concrete descendants is a schema bug — no value could ever satisfy it. Raising makes that explicit rather than silently producing code typed on an unusable class.

```suggestion
                descendants = [
                    self._get_class_python_name(c)
                    for c in sv.class_descendants(slot_range)
                    if not sv.get_class(c).abstract
                ]
                if not descendants:
                    raise ValueError(
                        f"Slot range '{slot_range}' is abstract and has no concrete descendants; "
                        f"cannot generate a valid Pydantic type."
                    )
```

tests/linkml/test_issues/test_linkml_issue_3196.py
```diff
@@ -0,0 +1,49 @@
...
+        required: true
+"""
+
+
+def test_pydantic_abstract_class_not_in_range_union():
+    mod = PydanticGenerator(test_schema).compile_module()
+    annotation = mod.Qux.model_fields["foo_object"].annotation
+
+    # Assert that our abstract class is not in the type union for the Qux.foo_object slot
+    assert mod.Foo not in typing.get_args(annotation)
```
kevinschaper, line 50: Could we flesh out the coverage a bit? A few cases I'd like pinned down:

1. **Positive assertion** that `Bar` and `Baz` *are* in the union (the current test only checks `Foo` isn't).
2. **Nested abstract intermediate** (`Foo` abstract → `Mid` abstract → concrete leaves) — confirms transitive descendants are filtered correctly.
3. **Mixin-only descendant regression guard** — mixin descendants are included in the current behavior (`MixinChild` with `mixins: [Foo]` and no `is_a` appears in the union). Worth locking that in.
4. **All-descendants-abstract** — should raise `ValueError` per the generator suggestion above.

I ran all of these locally against this branch (with the generator suggestion applied for #4).

```suggestion
    # Assert that our abstract class is not in the type union for the Qux.foo_object slot
    assert mod.Foo not in typing.get_args(annotation)


def test_concrete_descendants_are_in_range_union():
    mod = PydanticGenerator(test_schema).compile_module()
    annotation = mod.Qux.model_fields["foo_object"].annotation
    args = typing.get_args(annotation)
    assert mod.Bar in args
    assert mod.Baz in args


nested_abstract_schema = """
id: https://w3id.org/test
name: TestNested
imports:
  - linkml:types
classes:
  Foo:
    abstract: true
    attributes:
      type: {range: string, designates_type: true}
  Mid:
    is_a: Foo
    abstract: true
  Leaf1:
    is_a: Mid
    attributes: {f1: {range: string}}
  Leaf2:
    is_a: Foo
    attributes: {f2: {range: string}}
  Qux:
    attributes:
      foo_object: {range: Foo, required: true}
"""


def test_nested_abstract_intermediates_excluded():
    mod = PydanticGenerator(nested_abstract_schema).compile_module()
    args = typing.get_args(mod.Qux.model_fields["foo_object"].annotation)
    assert mod.Foo not in args
    assert mod.Mid not in args
    assert mod.Leaf1 in args
    assert mod.Leaf2 in args


mixin_schema = """
id: https://w3id.org/test
name: TestMixin
imports:
  - linkml:types
classes:
  Foo:
    abstract: true
    attributes:
      type: {range: string, designates_type: true}
  Bar:
    is_a: Foo
    attributes: {bar_field: {range: string}}
  MixinChild:
    mixins: [Foo]
    attributes: {mc_field: {range: string}}
  Qux:
    attributes:
      foo_object: {range: Foo, required: true}
"""


def test_mixin_descendant_included_in_union():
    mod = PydanticGenerator(mixin_schema).compile_module()
    args = typing.get_args(mod.Qux.model_fields["foo_object"].annotation)
    assert mod.Foo not in args
    assert mod.Bar in args
    assert mod.MixinChild in args


all_abstract_schema = """
id: https://w3id.org/test
name: TestAllAbstract
imports:
  - linkml:types
classes:
  Foo:
    abstract: true
    attributes:
      type: {range: string, designates_type: true}
  MidAbstract:
    is_a: Foo
    abstract: true
  Qux:
    attributes:
      foo_object: {range: Foo, required: true}
"""


def test_all_descendants_abstract_raises():
    """Abstract range with no concrete descendants is a schema error."""
    import pytest
    with pytest.raises(ValueError, match="no concrete descendants"):
        PydanticGenerator(all_abstract_schema).compile_module()
```

kevinschaper (changes requested):
Thanks for this fix! I played with this locally, trying out a handful of related schema shapes on this branch turned up one latent crash the filter exposes, plus a couple of behaviors worth pinning with tests.

**Verified on this branch:**
| Scenario | Result |
| --- | --- |
| Abstract root + 2 concrete `is_a` children | `Union[Bar, Baz]` ✅ |
| Abstract root + 1 concrete child | collapses to bare `Bar` (Python flattens `Union[X]` → `X`) ✅ |
| Abstract root → abstract intermediate → concrete leaves | `Union[Leaf1, Leaf2]`, intermediates excluded ✅ |
| Abstract root + pure-mixin descendant (`mixins: [Foo]`, no `is_a`) | `Union[Bar, MixinChild]` ✅ — mixin descendants ARE picked up |
| **All descendants abstract (abstract root + abstract child only)** | ❌ `SyntaxError: Union[]` at `compile_python` |

On the "Areas of uncertainty" question: a slot whose range is an abstract class with no concrete descendants is a schema error — whatever we emit, pydantic would never be able to validate an instance for that slot. Better to fail loudly at generation time than silently emit code typed on an unusable class. Two inline suggestions below: the generator raises a clear `ValueError`, and the test file gets extra coverage (positive assertions, nested-abstract, mixin regression guard, and a test for the new error).

Note that linkml-lint has no rule for this today — I'll open a follow-up issue to add one so schema authors catch it before they hit the generator.

tests/linkml/test_issues/test_linkml_issue_3196.py
colinrsmall: Great suggestions! Missing the positive assertion that `Bar` and `Baz` are included in the union was definitely an oversight on my part. Otherwise the other cases you suggest look great.

colinrsmall:
@kevinschaper Would love another look at this when you have a chance. We're distributing an internal version of LinkML with this patch but would love to switch back to using the main published version once this patch is live. If you think this is blocked on https://github.com/linkml/linkml/issues/3443, I'm happy to work on a PR for it.

kevinschaper (approved):
Thanks @colinrsmall! sorry for taking a while to pick this back up.

PR #3345, opened 2026-04-02
Update validation docs with details on rules

### LINKED ISSUE #3337: Documentation for rules is insufficient
We wanted to use LinkML rules to say that if one field exists in SSSOM, then another has to exist: https://github.com/mapping-commons/sssom/pull/514#discussion_r3014675699

I tried reading https://linkml.io/linkml-model/latest/docs/specification/05validation/#rules to understand how to use the rules slot, but I couldn't figure it out

It would be nice for this to be more detailed, improve the typography, and have examples.

matentzn:
## Summary

Fixes #3337

Improves the user facing docs for how rules should work

## How was this tested?

NA

## Areas of uncertainty

Just sanity check this. I found it weird that 

```
required: true
```

was used to denote the presence of a slot in this rule head:

```
- preconditions:
          slot_conditions:
            confidence:
              required: true
```

## Checklist

- [ ] ~~Existing tests pass locally with my changes~~

## AI Assistance


packages/linkml_runtime/src/linkml_runtime/linkml_model/model/docs/specification/05validation.md
```diff
@@ -188,6 +188,106 @@ may choose to assign the most specific value allowed to the slot.
...
+      object_id:
+        required: true
+      confidence:
+        range: float
+      confidence_source:
+    rules:
+      - preconditions:
+          slot_conditions:
+            confidence:
+              required: true
```
amc-corey-cox, line 227: Using `required: true` inside a precondition to mean "slot has a value" is semantically confusing — `required` normally means "must be provided," so reading this as "if confidence is required" rather than "if confidence is present" is a natural misreading. This is the exact confusion that motivated #3337.

I think documentation alone can't fully resolve this. Would it be worth proposing a dedicated metaslot like `present: true` for use in slot_conditions? That would make preconditions read naturally:

```yaml
preconditions:
  slot_conditions:
    confidence:
      present: true
```

This might increase the scope of this PR (or spawn a separate issue), but it would address the root cause rather than papering over it with docs.

packages/linkml_runtime/src/linkml_runtime/linkml_model/model/docs/specification/05validation.md
```diff
@@ -188,6 +188,106 @@ may choose to assign the most specific value allowed to the slot.
...
 
+Rules allow conditional constraints on classes: if certain **preconditions** hold for an
+instance, then **postconditions** must also hold. Rules are declared under the `rules` slot
+of a class definition.
+
+Each rule is a `ClassRule` with the following structure:
+
+- **preconditions** — an anonymous class expression that acts as the trigger ("if")
+- **postconditions** — an anonymous class expression that must hold when preconditions are met ("then")
+- **elseconditions** — an anonymous class expression that must hold when preconditions are *not* met ("else")
```
amc-corey-cox, line 199: `elseconditions` is introduced here and referenced in the evaluation pseudocode below, but none of the three examples demonstrate it. Was this intentional, or would a short example be worth adding for completeness?

amc-corey-cox (approved):
@matentzn I have a couple things here that don't need to block this. My AI agent also noticed: L315 says none_of for length 2 = "XOR" — should say "NOR", which isn't in this PR but might be a nice fix if you patch up anything else. This can go in as-is but I do think the 'present: true' versus 'required: true' in the preconditions is worth examining.

PR #3348, opened 2026-04-02
Fix identifier/key slots not marked required in induced_slot

### LINKED ISSUE #2448: Generated JSON-Schema allows "null" on identifiers

### LINKED ISSUE #2451: [meta] "slot" attribute "required" must be "true" if it is an "identifier" or a "key"

amc-corey-cox:
## Summary
- `SchemaView.induced_slot()` was setting `required=True` and `inlined=True` on the **original** slot object instead of the **induced copy**, so the inferred properties were silently lost
- Two-character fix: change `slot.` to `induced_slot.` on lines 1704-1707
- Updated one snapshot where an identifier's cardinality correctly changed from `0..1` to `1..1`

Closes #2448
Closes #2451


noelmcloughlin:
Related to #3360 and #3357

noelmcloughlin:
Hi @amc-corey-cox 
Good PR, I discovered same issue.
For the owl failures you could add this commit: https://github.com/linkml/linkml/pull/3357/changes/380172c63d96d6a70be828100487cf33bee314a0 which should resolve the OWL failures.

My `induced_slot` commit is slightly different: https://github.com/linkml/linkml/pull/3357/changes/d511e5ed776b3aa73ad27d435269d7d232a51dd6.  I think the OWL generator and JSON schema generator have dependency on mutation of `slot` (original) so I set the properties on BOTH (I did not investigate if both owl and json generators had bug and should be using induced_slot.required and induced_slot.inlined).

noelmcloughlin (commented):
LGTM, once tests are passing, can be merged.

packages/linkml/src/linkml/generators/shaclgen.py
```diff
@@ -173,7 +173,7 @@ def prop_pv_literal(p, v):
                     prop_pv_literal(SH.minCount, s.minimum_cardinality)
                 elif s.exact_cardinality:
                     prop_pv_literal(SH.minCount, s.exact_cardinality)
-                elif s.required:
+                elif s.required and not s.identifier:
```
matentzn, line 179: **shaclgen change** (`s.required and not s.identifier`): After your schemaview fix, induced slots with `identifier=True` will correctly have `required=True`. But the SHACL change explicitly *excludes* identifiers from getting `sh:minCount 1`. What's the reasoning here? If identifiers are required, shouldn't SHACL reflect that? Is this about identifiers mapping to the node IRI in RDF (so no property to constrain)? If so, a brief comment in the code would help.

packages/linkml/src/linkml/generators/owlgen.py
```diff
@@ -602,7 +602,9 @@ def transform_class_expression(
                 min_card = slot.minimum_cardinality
             elif top_slot.minimum_cardinality is not None:
                 min_card = top_slot.minimum_cardinality
-            elif slot.required or top_slot.required:
+            elif (
+                slot.required or top_slot.required or slot.identifier or slot.key or top_slot.identifier or top_slot.key
```
matentzn, line 606: Long-term owlgen should probably use induced slots too.

matentzn (approved):
LGTM root-cause fix for the `slot.` vs `induced_slot.` mutation bug; I assume you and @noelmcloughlin are both happy with the mild overlap between your PRs.

packages/linkml/src/linkml/generators/owlgen.py
noelmcloughlin: > LGTM root-cause fix for the `slot.` vs `induced_slot.` mutation bug; I assume you and @noelmcloughlin are both happy with the mild overlap between your PRs.

I will remove overlapping commits - the only concern I have is removing two `slot.xxx = yyy` lines from `schemaview.py` as noted in my comment above (some tests failed).

packages/linkml/src/linkml/generators/shaclgen.py
amc-corey-cox: Yes, that's the reasoning — an identifier slot becomes the subject IRI in RDF, so there's no `ex:id` triple to count. Requiring `sh:minCount 1` on the identifier path would fail on every valid instance. Added a comment at line 175 to make that explicit.

packages/linkml/src/linkml/generators/owlgen.py
amc-corey-cox: Agreed. Out of scope for this PR but noted — I'll track it as a follow-up so owlgen can migrate to induced slots once the rest of the generator landscape is aligned on that pattern. Want me to file the follow-up issue, or leave that to you?

amc-corey-cox: Filed as #3405 to track the migration. Resolving this thread.

Silvanoc:
@amc-corey-cox you haven't overseen that the xfail I introduced was meant to be removed once the bug has been fixed! Good eye! 💯

noelmcloughlin:
Fixes #3360 also :+1:

PR #3357, opened 2026-04-04
fix(linkml): relmodel performance #3353,  schemaview cachefix #3360

### LINKED ISSUE #3353: Relmodel_transformer invalidates lru_cache destroying performance
**Describe the bug**
The Relational model transformer performs expensive cache re-computations because `get_reference_map` + `merge_imports` both invalidation valid @lru_cache which destroys performance of`transform()` method:
https://github.com/linkml/linkml/blob/ed89a88d8e9766b7d3300c29315453db2ffad8ea/packages/linkml/src/linkml/transformers/relmodel_transformer.py#L153

Running gen-project on schema (4700 classes) runs for hours, with --log_level DEBUG unhelpful.

**Quick Analysis**
See (A), (B), and (C) problems:
```python
links = self.get_reference_map()   # (A) calls class_induced_slots for every class => populates lru_cache
source_sv.merge_imports()          # (B) calls set_modified() -> busts every `@lru_cache(None)` using `self` as key
   ... etc ...
for cn, c in source_sv.all_classes().items():
    for slot in source_sv.class_induced_slots(cn):  # (C) entire lru_cache recomputed from scratch
```

** Detailed analysis**
A. Populate Cache
The `get_reference_map()` method loops over every class and calls Schemaview's `class_induced_slots()` which in itself is not the problem as long as lru_cache is used.
```
        sv = self.schemaview
        for cn, c in sv.all_classes().items():
            for slot in sv.class_induced_slots(cn):                <==== every class
```
- And yes, Schemaview's `class_induced_slots()` method does populate cache:
https://github.com/linkml/linkml/blob/ed89a88d8e9766b7d3300c29315453db2ffad8ea/packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py#L1752-L1753

B. Destroy Cache
Next Schemaview's `merge_imports()`  calls `set_modified()`.
```
    def merge_imports(self) -> None:
        .... etc ...
        self.set_modified()         <=== here
```
And `set_modified()` sets the sv `SchemaView.modification` property.
```
    def set_modified(self) -> None:
        """Increase the number of schema modifications by 1."""
        self._hash = None
        self.modifications += 1         <== here
```
But scheaview's `__hash__` uses 'self.modifications' as a key:
```
def __key(self):
    return self.schema.id, self.uuid, self.modifications  <==== see here!!

def __hash__(self):
    if self._hash is None:
        self._hash = hash(self.__key())         <=== here inherited !!
    return self._hash

def set_modified(self):
    self._hash = None           # invalidate cached hash
    self.modifications += 1    # change the value that feeds into __key
```
lru_cache uses hash as as part of the cache key.
Every cached method on SchemaView is a bound method where `self` is the first argument.
So we have a valid hash for current loop iteration (in relmodel_transformer.py `get_reference_map()`)
.. until next iteration ..

C. Recompute Cache from scratch
The next call to sv.class_ancestors("Bob") computes a new key (new_hash(sv), "Bob") - never seen before - so lru_cache gets a cache miss and recomputes from scratch, and old entries (stored under previous hash) become orphaned memeory.

**Root Cause**
When `relmodel_transformer.py` calls `sv.set_modified()` inside its per-class loop shown above, **its very expensive, because we get new hash, new cache recompute, new memory orphans.***

**Tested solution**
In `relmodel_transformer.py`:
- swap method call order:  `source_sv.merge_imports()` executes before `self.get_reference_map()` so the cache populated during `get_reference_map` is no longer blown away before the class iteration loop uses it.
- remove `sv.set_modified()` from `add_primary_key` method. There is existing `target_sv.set_modified()` after the PK injection loop which does the single required invalidation. This eliminates 4700 individual cache flushes and reduces the PK phase from O(n²) to O(n).

`1.10.01`

### LINKED ISSUE #3360: Schemaview induced_slot stale/caching bug impacts
NOTE: Wait for https://github.com/linkml/linkml/pull/3348

**Describe the bug**
There is a stale/caching bug in `schemaview.py` revealed by testing PR #3357, which introduced newer call ordering (for example through `get_reference_map()`), meaning the stale cached result got used earlier and more consistently. Identifier slots are correctly materialized as required on first computation, the improvement revealed "hidden" issues:

- `induced_slot()` does not handle induced slot.
```
if slot.inlined_as_list:
            slot.inlined = True                 <= what about induced_slot.inlined ?
        if slot.identifier or slot.key:
            slot.required = True                 <= what about induced_slot.required ?
```
- jsonschema test data is wrong:
```
"id": {
                    "type": [
                        "string",
                        "null"                  <== incorrect, identifiers are required/non-nullable
                    ]
```
- SHACL MinCount uniqueness test failures caused `shaclgen.py` is too strict on identifiers (line 178). Identifiers may be represented as the subject IRI (no explicit id triple), so requiring sh:minCount 1 on the identifier path causes false violations.
```
                   elif s.required:
                    prop_pv_literal(SH.minCount, 1)      <-==  identifiers not excluded
```
- The `recommend_slots_plugin.py` is checking range class for the list itself, not what is wanted. 
```
                    if slot_def.multivalued:
                         if slot_def.inlined and isinstance(slot_value, dict):     <==  "and not slot_def.inlined_as_list" needed
                             for k, v in slot_value.items():
```
- Owl `snapshots` using stale value after identifier-required semantics fix (see also https://github.com/linkml/linkml/pull/3348)
```
             owl:minCardinality 0 ;        <== should be owl:allValuesFrom ex:C 
             owl:onProperty ex:has_c ],
         [ a owl:Restriction ;
             owl:maxCardinality 1 ;       <== should be owl:minCardinality 0 ;
```
- Markdown `snapshots` incorrectly showing identifiers as non-required (see https://github.com/linkml/linkml/pull/3348)
```
| **[id](#id)** | <sub>0..1</sub> | string |  |                <== should be <sub>1..1</sub>
```

**Causes**
- In `induced_slot()`, the code set `slot.inlined = True` and `slot.required = True` on the original slot object.
- But `induced_slot` variable had already been created as a copy before those mutations.
- So the returned copied object could still have stale values (`inlined`/`required` not updated).
- Because `class_induced_slots()` is cached (see `lru_cache`), that stale first result could be cached and reused.
- Some snapshots and unit tests had wrong test data (passes stale/cached values instead of schema-compliant data).

**Tested Solution**
I fixed using granular commits on #3357 but https://github.com/linkml/linkml/pull/3348 already spotted some issues so this issue should be revised for smaller scope once that PR is merged.

Fix code:
-  `schemaview.py`: also mutate the returned `induced_slot` object
- `shaclgen.py`: do not enforce  `sh:minCount 1` on the identifier path
- `recommended_slots_plugin.py`: update conditional with "only if the slot is not inlined as a list" clause.

Fix downstream impacts on snapshots
-  Fixes jsonschema `snapshots` so id is required/non-nullable
- Fixes markdown `snapshots` so id is required/non-nullable
- Fixes owl `snapshots` so owl:minCardinality 1 moves to the end of the restriction list for ex:id because serialized order differs.
- Fixes owl `snapshots` so owl:minCardinality 1 is now present for ex:id. The fixed `schemaview.py -> induced_slot`  means the OWL generator now sees s.required == True for the id slot and emits owl:minCardinality 1.
- 

`1.10.0`

noelmcloughlin:
## Summary

Fixes #3353 which explains the performance issue being addressed, add debug logs, testing.

## Reviewer 
It may be easier to review commit by commit.

## How was this tested?
Local testing

## Checklist


tests/linkml/test_generators/test_jsonschemagen.py
```diff
@@ -316,7 +316,6 @@ def test_slot_title_from_title_slot(subtests, input_path):
     external_file_test(subtests, input_path("jsonschema_slot_title_from_title.yaml"), {"title_from": "title"})
 
 
-@pytest.mark.xfail(reason="identifier slots incorrectly allow null (#2448)", strict=True)
```
noelmcloughlin, line 319: Included in https://github.com/linkml/linkml/pull/3348

tests/linkml/test_generators/__snapshots__/markdowndatadict_kitchen_sink.md
```diff
@@ -362,7 +362,7 @@ Person ||--}o MedicalEvent : "has medical history"
 
 | Name | Cardinality: | Type | Description |
 | --- | --- | --- | --- |
-| **[id](#id)** | <sub>0..1</sub> | string |  |
+| **[id](#id)** | <sub>1..1</sub> | string |  |
```
noelmcloughlin, line 365: Included in https://github.com/linkml/linkml/pull/3348

packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
```diff
@@ -1703,8 +1703,10 @@ def induced_slot(
                 setattr(induced_slot, metaslot_name, v)
         if slot.inlined_as_list:
             slot.inlined = True
+            induced_slot.inlined = True
```
noelmcloughlin, line 1706: Included in https://github.com/linkml/linkml/pull/3348

packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
```diff
@@ -1703,8 +1703,10 @@ def induced_slot(
...
         if slot.identifier or slot.key:
             slot.required = True
+            induced_slot.required = True
```
noelmcloughlin, line 1709: Included in https://github.com/linkml/linkml/pull/3348

amc-corey-cox:
Hi @noelmcloughlin — #3348 just merged, which overlaps with several commits here. Could you rebase on current `main` and re-scope this PR down to the pieces that are still unique and valuable:

- **`relmodel_transformer.py`** perf fix (the `merge_imports()` / `get_reference_map()` reordering and removal of per-class `set_modified()` in `add_primary_key`)
- **`recommended_slots_plugin.py`** inlined_as_list fix
- **The relmodel test suite**

Thanks!

noelmcloughlin:
Hey @amc-corey-cox  thanks for letting me know.  

Replaced by #3425 for cleaner commit history.

PR #3356, opened 2026-04-04
fix(sqltablegen): performance optimization #3355

noelmcloughlin:
## Summary

Fixes #3355 which explains the issue in detail.

## How was this tested?

1. Added unit tests
2. Tested replicator repo.
```
git clone https://github.com/lmodel/d3fend.git
cd d3fend/
just clean
```
(a) Before
```
just setup  # hangs
uv run gen-python src/d3fend/schema/d3fend/yaml > /tmp/defend.py   # hangs
uv run gen-pydantic src/d3fend/schema/d3fend.yaml > /tmp/d3fendic.py  # hangs
uv run gen-sqltables src/d3fend/schema/d3fend.yaml > /tmp/d3fend.sql  # hangs
```
(2) Apply corrections from #3364, #3363 and #3356 
```
## py3.14

curl -L -o .venv/lib/python3.14/site-packages/linkml_runtime/utils/schemaview.py https://raw.githubusercontent.com/noelmcloughlin/linkml/0791284b2d12bbf24ad36494be8022fc9f1d4be9/packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py 
curl -L -o .venv/lib/python3.14/site-packages/linkml/generators/pythongen.py https://raw.githubusercontent.com/noelmcloughlin/linkml/4033cbeae8b29cf43e65724c87eb8fd6b03c89da/packages/linkml/src/linkml/generators/pythongen.py
curl -L -o .venv/lib/python3.14/site-packages/linkml/generators/sqltablegen.py https://raw.githubusercontent.com/noelmcloughlin/linkml/331bcef98d3522c3001201874a65de39725a4ac5/packages/linkml/src/linkml/generators/sqltablegen.py
```
(3) After
```
uv run gen-python src/d3fend/schema/d3fend/yaml > /tmp/defend.py   # 6 minutes 6 seconds
uv run gen-pydantic src/d3fend/schema/d3fend.yaml > /tmp/defendic.py  # 8 minutes 14 seconds
uv run gen-sqltables src/d3fend/schema/d3fend.yaml > /tmp/d3fend.sql  # 6 minutes 16 seconds
```

## Checklist

- [x] My code follows the [contributor guidelines](https://linkml.io/linkml/maintainers/contributing.html)
- [x] I have added tests that prove my fix/feature works
- [x] Existing tests pass locally with my changes

Issue #3351, opened 2026-04-03
Linkml: python generator sort_classes() algorithm sub optimal


noelmcloughlin:
**Describe the bug**
This Python Generator `sort_classes` logic is incrementally non performant as schema size grows and can exhaust memory.
https://github.com/linkml/linkml/blob/ed89a88d8e9766b7d3300c29315453db2ffad8ea/packages/linkml/src/linkml/generators/pythongen.py#L776-L786

Running `gen-project` on schema (4700 classes)  runs for hours, with `--log_level DEBUG` unhelpful, exhausting memory.

**Troubleshooting with `py-spy`**
Stack traces blamed:  `_sort_classes` -> `gen_references` ->  `gen_schema` ->  `serialize` flow
```python
while len(clist) > 0:                                              # Object (n) passes
    for i in range(len(clist)):                                   # Object (n) per pass
        if candidate.is_a in [p.name for p in slist]:   # O(current slist size) - list scan!
```
The following math highlights the problem for my schema (4700 classes):
```
O(4700) means membership test is O(k) where k grows to n. 
For n=4700 -> up to ~52 billion comparisons in O(n³) worst case.
```

**Cause**
`slist` is defined as simple [list](https://docs.python.org/3/library/stdtypes.html#typesseq-list) so membership scanning is necessary using for loop. 

**Workaround**
None

**Tested solution**
Make `slist` a [set type](https://docs.python.org/3.10/library/stdtypes.html#set-types-set-frozenset) then use [slist.add](https://docs.python.org/3.10/library/stdtypes.html#frozenset.add) and worst case is ~22m (n² + 1). 

There is no functional impact because its just a algorithm optimization.

**Version of LinkML you are using**
`1.10.01`

noelmcloughlin:
PR #3363

noelmcloughlin:
Fixed by #3363 merge

PR #3363, opened 2026-04-05
fix(pythongen): improved sort_classes performance

noelmcloughlin:
## Summary

This PR addresses #3351 

Added commit https://github.com/linkml/linkml/pull/3363/changes/0568d17c4fece6422b59d46e7d102ff3e7475cb6 to fix previously hidden bug #3365

Added commit to ensure gen_references() produces ordered output (and refreshed snapshots).

## How was this tested?

1. Added unit tests
2. Tested replicator repo.
```
git clone https://github.com/lmodel/d3fend.git
cd d3fend/
just clean
```
(a) Before
```
just setup  # hangs
uv run gen-python src/d3fend/schema/d3fend/yaml > /tmp/defend.py   # hangs
uv run gen-pydantic src/d3fend/schema/d3fend.yaml > /tmp/d3fendantic.py  # hangs
uv run gen-sqltables src/d3fend/schema/d3fend.yaml > /tmp/d3fend.sql  # hangs
```
(2) Apply corrections from #3364, #3363 and #3356 
```
## py3.14

curl -L -o .venv/lib/python3.14/site-packages/linkml_runtime/utils/schemaview.py https://raw.githubusercontent.com/noelmcloughlin/linkml/0791284b2d12bbf24ad36494be8022fc9f1d4be9/packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py 
curl -L -o .venv/lib/python3.14/site-packages/linkml/generators/pythongen.py https://raw.githubusercontent.com/noelmcloughlin/linkml/4033cbeae8b29cf43e65724c87eb8fd6b03c89da/packages/linkml/src/linkml/generators/pythongen.py
curl -L -o .venv/lib/python3.14/site-packages/linkml/generators/sqltablegen.py https://raw.githubusercontent.com/noelmcloughlin/linkml/331bcef98d3522c3001201874a65de39725a4ac5/packages/linkml/src/linkml/generators/sqltablegen.py```
(3) After
```
uv run gen-python src/d3fend/schema/d3fend/yaml > /tmp/defend.py   # 6 minutes 6 seconds
uv run gen-pydantic src/d3fend/schema/d3fend.yaml > /tmp/pydantic.py  # 8 minutes 14 seconds
uv run gen-sqltables src/d3fend/schema/d3fend.yaml > /tmp/d3fend.sql  # 6 minutes 16 seconds
```

## Checklist

### CODE REVIEW

**Comment by noelmcloughlin** [comment]
Added replicator and testing results to description above ^^^

File: packages/linkml/src/linkml/generators/pythongen.py [hunk 1]
```diff
@@ -406,32 +404,77 @@ def gen_namespaces(self) -> str:
...
+        # Step 2: write each wrapper only after its parent has been written.
+        emitted: set[str] = set()
+        rval: list[str] = []
+
+        def _emit(name: str) -> None:
+            if name in emitted or name not in ref_info:
+                return  # Already written, or imported type — nothing to do.
+            parent_cls, line = ref_info[name]
+            _emit(parent_cls)  # Write the parent first.
+            if name not in emitted:  # Defensive check against malformed schemas.
```
amc-corey-cox, line 470: Minor concern: the old `TopologicalSorter` raised `CycleError` on a cyclic input, but I think this recursive `_emit` would infinite-recurse instead — the top-level `if name in emitted` check doesn't help because `name` is only added *after* the recursive call returns. The "defensive check against malformed schemas" comment oversells what the post-recurse check actually does.

Realistically class hierarchies don't cycle in valid schemas, so this is low risk. But if you're open to it, tracking an `in_progress: set[str]` alongside `emitted` would restore a clear cycle signal and let the new `_sort_classes` test suite include a cycle case (it currently covers empty / linear / diamond / large but not cyclic). Something like:

```python
in_progress: set[str] = set()

def _emit(name: str) -> None:
    if name in emitted or name not in ref_info:
        return
    if name in in_progress:
        raise ValueError(f"Cyclic wrapper inheritance at {name}")
    in_progress.add(name)
    parent_cls, line = ref_info[name]
    _emit(parent_cls)
    in_progress.discard(name)
    emitted.add(name)
    rval.append(line)
```

Happy to defer if you don't think it's worth the churn.

amc-corey-cox (changes requested):
One small concern inline about cycle handling in the new _emit; otherwise looks good to me.

packages/linkml/src/linkml/generators/pythongen.py
noelmcloughlin: Good catch @amc-corey-cox  I have updated the PR with your suggestion and unit tests

amc-corey-cox (approved):
This looks good to me.

PR #3368, opened 2026-04-08
Make gen_doc output ordering deterministic

sveinugu:
## Summary

Several places in the autogenerated docs, the output changes randomly between `gen-doc` runs, causing spurious diffs
in version-controlled output.

This includes:
- The list of types shown under *"Type and Range → Range"* on slot
documentation pages (and in the *"Cardinality and Range"* column of slot tables
on class pages)
- The *"Applicable Classes"* table on slot documentation pages lists

In both places, the root cause is the Python builtin `set`, which does not maintain insertion order. Unfortunately, there is no "OrderedSet" in the std libraries, so the fixes are more of workarounds, adding a list alongside the set. Adding a dependency just for this is too high cost for little gain.

## How was this tested?

No new functionality, so no need of new tests. The modified methods (`slot_range_as_union()` and `get_classes_by_slot()` of the class `SchemaView`) already have tests in `test_schemaview.py`. All relevant tests are running.

## Checklist

- [ x ] My code follows the [contributor guidelines](https://linkml.io/linkml/maintainers/contributing.html)
- [ x ] Existing tests pass locally with my changes

## AI Assistance


sujaypatil96 (changes requested):
Thanks for noting the issue with certain sections of the generated documentation pages not preserving ordering. I see you're suggesting changes to the SchemaView module in your PR, but I would refrain from making any modifications to the helper methods in there b/c a lot of generators besides docgen (documentation generator) are using the methods from there.

Instead we can achieve the behavior that you desire with the changes in this PR by making changes to the [slot jinja template](https://github.com/linkml/linkml/blob/main/packages/linkml/src/linkml/generators/docgen/slot.md.jinja2) and the [class jinja template](https://github.com/linkml/linkml/blob/main/packages/linkml/src/linkml/generators/docgen/class.md.jinja2) by adding a `|sort` filter ([sort()](https://jinja.palletsprojects.com/en/stable/templates/#jinja-filters.sort) in jinja) to these 4 places in the templates:

1.  [slot.md.jinja2](https://github.com/linkml/linkml/blob/main/packages/linkml/src/linkml/generators/docgen/slot.md.jinja2) line 18 — `compute_range` macro: `schemaview.slot_range_as_union(slot)|sort`
2.  [slot.md.jinja2](https://github.com/linkml/linkml/blob/main/packages/linkml/src/linkml/generators/docgen/slot.md.jinja2) line 70 — Applicable Classes loop: `classes_by_slot|sort`
3.  [slot.md.jinja2](https://github.com/linkml/linkml/blob/main/packages/linkml/src/linkml/generators/docgen/slot.md.jinja2) line 82 — Mixin Usage inline: `...|sort|join(', ')`
4.  [class.md.jinja2](https://github.com/linkml/linkml/blob/main/packages/linkml/src/linkml/generators/docgen/class.md.jinja2) line 18 — same `compute_range` macro: `schemaview.slot_range_as_union(slot)|sort`

Does that make sense?

These are good changes for the base jinja templates in the linkml library so we appreciate the contribution! for your changes to propagate up to linkml and have it be available in the linkml library you will need to wait for your PR to be merged and be made available in the following linkml release/cycle. If you would prefer not to wait, you can always duplicate the existing jinja templates we have (copy the ones you want to modify into a directory) and point your `gen-doc` command to the folder with custom jinja (changes including `|sort`) using the [--template-directory](https://linkml.io/linkml/generators/docgen.html#cmdoption-gen-doc-template-directory) option we have for the command in linkml.

sveinugu:
> Thanks for noting the issue with certain sections of the generated documentation pages not preserving ordering. I see you're suggesting changes to the SchemaView module in your PR, but I would refrain from making any modifications to the helper methods in there b/c a lot of generators besides docgen (documentation generator) are using the methods from there.
> 
> Instead we can achieve the behavior that you desire with the changes in this PR by making changes to the [slot jinja template](https://github.com/linkml/linkml/blob/main/packages/linkml/src/linkml/generators/docgen/slot.md.jinja2) and the [class jinja template](https://github.com/linkml/linkml/blob/main/packages/linkml/src/linkml/generators/docgen/class.md.jinja2) by adding a `|sort` filter ([sort()](https://jinja.palletsprojects.com/en/stable/templates/#jinja-filters.sort) in jinja) to these 4 places in the templates:
> 
> 1. [slot.md.jinja2](https://github.com/linkml/linkml/blob/main/packages/linkml/src/linkml/generators/docgen/slot.md.jinja2) line 18 — `compute_range` macro: `schemaview.slot_range_as_union(slot)|sort`
> 2. [slot.md.jinja2](https://github.com/linkml/linkml/blob/main/packages/linkml/src/linkml/generators/docgen/slot.md.jinja2) line 70 — Applicable Classes loop: `classes_by_slot|sort`
> 3. [slot.md.jinja2](https://github.com/linkml/linkml/blob/main/packages/linkml/src/linkml/generators/docgen/slot.md.jinja2) line 82 — Mixin Usage inline: `...|sort|join(', ')`
> 4. [class.md.jinja2](https://github.com/linkml/linkml/blob/main/packages/linkml/src/linkml/generators/docgen/class.md.jinja2) line 18 — same `compute_range` macro: `schemaview.slot_range_as_union(slot)|sort`
> 
> Does that make sense?
> 
> These are good changes for the base jinja templates in the linkml library so we appreciate the contribution! for your changes to propagate up to linkml and have it be available in the linkml library you will need to wait for your PR to be merged and be made available in the following linkml release/cycle. If you would prefer not to wait, you can always duplicate the existing jinja templates we have (copy the ones you want to modify into a directory) and point your `gen-doc` command to the folder with custom jinja (changes including `|sort`) using the [--template-directory](https://linkml.io/linkml/generators/docgen.html#cmdoption-gen-doc-template-directory) option we have for the command in linkml.

Thanks for the quick feedback!

Being new to linkml, I do not have the overview of the code and so everything I say should of course be taken with a grain of salt. That said, I cannot fathom how my suggested changes could be an issue for any downstream usage at all, other than a negligible slight reduction in performance and increased memory use. Rather, it is much more likely that other generators might get value from the deterministic order. I cannot see any use case where a non-deterministic listing is a required feature.

Also, as a new user of linkml, one of the initial annoyances with the software was actually the opposite, that most of the outputs ARE sorted and do not retain the order from the schema yaml files. Since, in many cases, maintaining the order often simplifies understanding of a model - then I would rather in the future add this as an option and not have to depend on sorting for deterministic output. So contributing towards more sorting was really not my intention with the PR!

sveinugu:
A less intrusive implementation could be to use a `dict` with `None` as values, and then return only the keys, as `dict` in contrast to `set` maintains insertion order.

sveinugu:
I have now updated the PR to use `dict` with `None` as value to mimic an ordered set, which should actually improve performance somewhat (but with slightly larger memory usage which should not matter in reality).

sujaypatil96 (approved):
Great points @sveinugu! Sorry about my reluctance to update methods in the SchemaView module, my initial thought was since this is a `gen-doc` issue, let's try to solve it with changes such that we're not touching any shared code from SchemaView since they often tend to be breaking changes.

But yes you're absolutely right, in that the changes you're introducing to the `slot_range_as_union()` and `get_classes_by_slot()` methods to have them use a "mimicked ordered set" instead of a regular old set looks like a fix that wouldn't have any side effects on downstream pieces that use these methods. In fact, like you said, generators like rustgen and pydanticgen (among other places where these methods are used) would benefit from the determinism.

Issue #3367, opened 2026-04-08
`_normalize_inlined` fails on `inlined_as_list` values containing commas


matentzn:
When a class is used as the range of an `inlined_as_list` slot, `linkml_runtime`'s `_normalize_inlined` uses the first required string slot as a dict key. If the value contains a comma (e.g. `48,XXYY syndrome`), the key parsing splits on the comma, causing a `ValueError` mismatch between the parsed key and the actual attribute value.

This affects both `linkml-owl` (via `load_structured_file`) and any other code path that instantiates dataclasses from YAML through `linkml_runtime`.

### Minimal reproducer

**Schema** (`schema.yaml`):

```yaml
id: https://example.org/comma-bug
name: comma_bug
prefixes:
  linkml: https://w3id.org/linkml/
  ex: https://example.org/
imports:
  - linkml:types
default_prefix: ex
default_range: string

classes:
  Container:
    tree_root: true
    slots:
      - items

  Item:
    slots:
      - item_id
      - synonyms

  Synonym:
    slots:
      - synonym_text
      - synonym_type

slots:
  items:
    range: Item
    multivalued: true
    inlined_as_list: true

  item_id:
    identifier: true
    range: string
    required: true

  synonyms:
    range: Synonym
    multivalued: true
    inlined_as_list: true

  synonym_text:
    required: true

  synonym_type:
    range: string
```

**Data that works** (`data_ok.yaml`):

```yaml
items:
- item_id: ex:001
  synonyms:
  - synonym_text: simple synonym
  - synonym_text: another synonym
```

**Data that fails** (`data_fail.yaml`):

```yaml
items:
- item_id: ex:001
  synonyms:
  - synonym_text: '48,XXYY syndrome'
  - synonym_text: simple synonym
```

**Validation passes for both:**

```bash
python -m linkml.validator.cli --schema schema.yaml --target-class Container data_fail.yaml
# No issues found
```

**linkml-owl fails on `data_fail.yaml`:**

```bash
python -m linkml_owl.dumpers.owl_dumper --schema schema.yaml -o out.owl data_fail.yaml
```

```
ValueError: Slot: synonyms - attribute synonym_text value (JsonObj(synonym_text='48,XXYY syndrome')) does not match key (File "<file>", line 5, col 19: 48,XXYY syndrome)
```

**linkml-owl succeeds on `data_ok.yaml`:**

```bash
[...]

PR #3371, opened 2026-04-08
Add regression tests for _normalize_inlined with comma-containing values

matentzn:
Adds regression tests for already fixed problem #3367 

The bug was caused by slot_type(list_entry) passing the dict as a positional arg instead of slot_type(**{key_name: lev}). Already fixed by prior kwargs refactor; these tests prevent regression.

## Areas of uncertainty

This is probably not needed - but because it was a real problem I stumbled over I thought I add a quick regression test. @amc-corey-cox its likely this is redundant. If you nose tells you to avoid adding another test because you fear test suite bloat more then this issue re-occurring, feel free to close the PR without concerns.

## Checklist

- [X] I have added tests that prove my fix/feature works
- [X] Existing tests pass locally with my changes

## AI Assistance


amc-corey-cox (approved):
I can't think of any reason this isn't worth having. I'll bring it in.

PR #3382, opened 2026-04-14
feat: add TypeDB generator (gen-typedb)

GullyBurns:
## Summary

- Adds \`gen-typedb\`, a new generator that converts a LinkML schema to a TypeQL 3.x \`define\` block for loading into a [TypeDB](https://typedb.com/) graph database
- Generator lives at \`linkml/generators/typedbgen.py\` and extends the base \`Generator\` class
- Registered as \`gen-typedb\` console script entry point in \`pyproject.toml\`
- TypeDB driver is an optional dependency (\`uv sync --group typedb\`)

## Mapping

| LinkML concept | TypeDB/TypeQL output |
|---|---|
| class | \`entity\` type |
| class with \`represents_relationship: true\` | \`relation\` type (with \`relates\` roles) |
| scalar slot | \`attribute\` type + \`owns\` |
| object-ranged slot on entity class | \`relation\` type + \`plays\` |
| object-ranged slot on relationship class | \`relates <slot-name>\` role |
| \`is_a\` | \`sub\` keyword |
| \`abstract: true\` | \`@abstract\` |
| \`identifier: true\` | \`@key\` |
| \`required: true\` (singular) | \`@card(1..1)\` |
| \`multivalued: true\` | \`@card(0..)\` |
| enum | \`string\` attribute with permitted values in comment |

## Changes since initial submission (addressing review feedback)

- **Removed** \`typedb-driver\` from root workspace \`[dependency-groups]\` (was incorrectly adding it to everyone's dev env)
- **Added** \`file_extension = "tql"\` ClassVar
- **Fixed** \`serialize(**kwargs)\` signature to match base class (removed \`# type: ignore\`)
- **Fixed** \`_resolve_typedb_value_type\` type-hierarchy walk: was using \`from_schema\` (schema URI) instead of \`typeof\` (parent type), and had a spurious \`break\` preventing iteration
- **Added** \`represents_relationship: true\` → TypeDB \`relation\` type mapping (requested by kevinschaper and flyingsilverfin)
- **Added** CI workflow \`.github/workflows/typedb-integration.yaml\` — path-conditional, only runs when the generator or its tests change
- **Refactored** \`_collect_entity_defs\` for DRY: extracted \`_ancestor_slots\` helper, \`_build_owns_stmt\` helper, cached ancestor slot lookups, fixed set→list comprehensions for deterministic output

## Tests

- Unit tests in \`tests/linkml/test_generators/test_typedbgen.py\` (28 tests, no external deps)
  - Added tests for reserved keyword handling, name collision, and \`represents_relationship\` → relation mapping
- Integration tests in \`tests/linkml/test_generators/test_typedbgen_integration.py\` (marked \`@pytest.mark.integration\`, requires TypeDB 3.x at \`localhost:1729\`)
- New CI workflow runs integration tests automatically when generator files change

## Documentation

- \`docs/generators/typedb.rst\` with type mapping table, usage examples, and limitations
- Added to \`Database\` section of \`docs/generators/index.rst\`
- Added to \`docs/maintainers/generator-governance.md\`

## Test plan

- [x] \`uv run pytest tests/linkml/test_generators/test_typedbgen.py -v\` — all 28 tests pass
- [x] \`uv tool run ruff check\` — no lint errors
- [ ] Integration tests pass against TypeDB 3.x (requires local server)


GullyBurns:
Note that I run tests on a locally running typedb database in a docker container: 

docker-compose.yaml
```
# Docker Compose for Skillful-Alhazen
# Usage: docker compose up -d
#
# Services:
# - Dashboard: http://localhost:3001
# - TypeDB: localhost:1729
#
# To run only TypeDB: docker compose up -d typedb
# To run everything: docker compose up -d

services:
  # ============================================
  # TypeDB - Knowledge Graph Database
  # ============================================
  typedb:
    image: typedb/typedb:3.8.0
    container_name: alhazen-typedb
    ports:
      - "127.0.0.1:1729:1729"
      - "127.0.0.1:8000:8000"
    volumes:
      - typedb-data:/var/lib/typedb/data
      - ./local_resources/typedb:/schema:ro
    healthcheck:
      test: ["CMD-SHELL", "timeout 3 bash -c 'echo > /dev/tcp/localhost/1729' 2>/dev/null || exit 1"]
      interval: 10s
      timeout: 5s
      retries: 10
      start_period: 30s
    restart: unless-stopped
    networks:
      - alhazen-network

  # ============================================
  # Schema Initialization (runs once)
  # ============================================
  typedb-init:
    image: python:3.11-slim
    container_name: alhazen-typedb-init
    depends_on:
      typedb:
        condition: service_healthy
    volumes:
      - ./local_resources/typedb:/schema:ro
      - ./local_skills:/local_skills:ro
      - ./scripts:/scripts:ro
    networks:
      - alhazen-network
    environment:
      - TYPEDB_HOST=typedb
      - TYPEDB_PORT=1729
      - TYPEDB_DATABASE=alhazen_notebook
    entrypoint: ["/bin/bash", "-c"]
    command:
      - |
        pip install typedb-driver -q
        python /scripts/db_init.py \
          /schema/alhazen_notebook.tql \
          /schema/namespaces/*.tql \
          /local_skills/*/schema.tql
    restart: "no"

volumes:
  typedb-data:
    driver: local

networks:
  alhazen-network:
    driver: bridge

```

with the db_init.py script:
```python
#!/usr/bin/env python3
"""
TypeDB 3.x database initialization script.

Creates the database (if it doesn't exist) and loads all schemas using
the TypeDB Python driver. This replaces the console-based approach.

Usage:
    uv run python scripts/db_init.py [--wait-only]
    uv run python scripts/db_init.py --host localhost --port 1729 --database alhazen_notebook

Environment:
    TYPEDB_HOST       TypeDB server host (default: localhost)
    TYPEDB_PORT       TypeDB server port (default: 1729)
    TYPEDB_DATABASE   Database name (default: alhazen_notebook)
    TYPEDB_USERNAME   TypeDB username (default: admin)
    TYPEDB_PASSWORD   TypeDB password (default: password)
"""

import argparse
import glob
import os
import sys
import time
from pathlib import Path

def wait_for_typedb(host, port, username, password, timeout=60):
    """Wait until TypeDB is accepting connections."""
    from typedb.driver import Credentials, DriverOptions, TypeDB
    deadline = time.time() + timeout
    while time.time() < deadline:
        try:
            driver = TypeDB.driver(
                f"{host}:{port}",
                Credentials(username, password),
                DriverOptions(is_tls_enabled=False),
            )
            driver.close()
            print(f"TypeDB is ready at {host}:{port}", flush=True)
            return True
        except Exception:
            print(".", end="", flush=True)
            time.sleep(2)
    print(f"\nTypeDB failed to start within {timeout} seconds", flush=True)
    return False


def init_database(host, port, username, password, database, schema_files):
    """Create database and load schemas."""
    from typedb.driver import Credentials, DriverOptions, TransactionType, TypeDB

    driver = TypeDB.driver(
        f"{host}:{port}",
        Credentials(username, password),
        DriverOptions(is_tls_enabled=False),
    )

    try:
        # Create database if it doesn't exist
        if not driver.databases.contains(database):
            driver.databases.create(database)
            print(f"Created database: {database}", flush=True)
        else:
            print(f"Database already exists: {database}", flush=True)

        # Load each schema file
        for schema_path in schema_files:
            schema_path = Path(schema_path)
            if not schema_path.exists():
                print(f"Warning: schema not found: {schema_path}", flush=True)
                continue

            print(f"Loading schema: {schema_path.name} ...", end=" ", flush=True)
            with open(schema_path) as f:
                schema = f.read()

            try:
                with driver.transaction(database, TransactionType.SCHEMA) as tx:
                    tx.query(schema).resolve()
                    tx.commit()
                print("OK", flush=True)
            except Exception as e:
                print(f"FAILED: {e}", flush=True)
                raise

    finally:
        driver.close()


def main():
    parser = argparse.ArgumentParser(description="Initialize TypeDB 3.x database")
    parser.add_argument("--host", default=os.getenv("TYPEDB_HOST", "localhost"))
    parser.add_argument("--port", type=int, default=int(os.getenv("TYPEDB_PORT", "1729")))
    parser.add_argument("--database", default=os.getenv("TYPEDB_DATABASE", "alhazen_notebook"))
    parser.add_argument("--username", default=os.getenv("TYPEDB_USERNAME", "admin"))
    parser.add_argument("--password", default=os.getenv("TYPEDB_PASSWORD", "password"))
    parser.add_argument("--timeout", type=int, default=60,
                        help="Seconds to wait for TypeDB to be ready")
    parser.add_argument("--wait-only", action="store_true",
                        help="Only wait for TypeDB readiness, don't load schemas")
    parser.add_argument("schemas", nargs="*",
                        help="Schema .tql files to load (in order)")
    args = parser.parse_args()

    # Wait for TypeDB to be ready
    print(f"Waiting for TypeDB at {args.host}:{args.port} ...", flush=True)
    if not wait_for_typedb(args.host, args.port, args.username, args.password, args.timeout):
        sys.exit(1)

    if args.wait_only:
        return

    # Load schemas
    if not args.schemas:
        print("No schema files specified. Use: scripts/db_init.py schema1.tql schema2.tql ...", flush=True)
        sys.exit(1)

    init_database(args.host, args.port, args.username, args.password, args.database, args.schemas)
    print(f"\nDatabase '{args.database}' initialized successfully.", flush=True)


if __name__ == "__main__":
    main()

```

GullyBurns:
Note - the typedb 3.x libraries have an issue with python 3.14.

cmungall:
Fantastic, thanks!

This may be better handled in a separate PR but it would be good to have a path to support classes that instantiate edge patterns in linkml, see:

https://linkml.io/linkml/howtos/model-property-graphs.html

cmungall (approved):

GullyBurns:
One of my hopes from doing this would be to get technical members from the TypeDB team directly contributing to this thread (since they know their own system inside out and seem amenable to connecting with user communities).

packages/linkml/src/linkml/generators/typedbgen.py
```diff
@@ -0,0 +1,470 @@
...
+    "decimal": "double",
+    # boolean
+    "boolean": "boolean",
+    "bool": "boolean",
+    # datetime-like
+    "date": "datetime",
+    "datetime": "datetime",
+    "datetimestamp": "datetime",
+    "time": "datetime",
+    "duration": "string",  # no native duration in TypeDB 3
```
krishnangovindraj: It does actually

flyingsilverfin:
Hi all! CTO of TypeDB here :) 

First of all - absolutely love LinkML, I just read the quickstart and thought about the concepts for a while. Incredible spec, very atomic and compact - wonderful!

Second of all - damn... how can we get TypeDB to be the default database backend for LinkML usage?? TypeDB's data model is almost a direct translation of your model! More direct than LPG for sure, and certainly none of the reification weirdness that RDF has!

- classes are either Entities or Relations
  - entities if they don't contain any slots that range over other classes
  - relations if thy contain slots that do range over classes
- primitive slots (eg. numerical, string, etc.) become Attribute in TypeDB
- class slots (eg. ranging over classes) become Roles in TypeDB
- `is_a` becomes `sub` in TypeDB's data model
  - this works for classes (eg. Entities and relations)
  - this _also_ works for reusable primitive slots (eg. Attributes in TypeDB can `sub`type other ones!)
  - it may not quite work for reusable complex slots, though we do support subtyping roles within a relation (they're not not global types, but belong to a part of a relation). Perhaps reusable complex slots can themselves become entities or relations?
 - all the inheritance and extension stuff works out of the box
 - monotonicity is exactly how our schemas work too
- I believe mixins are just composition of reusable schema definitions
 
- primitive enums become `@values`
- identifier becomes `@key`
- required becomes `@card(1..1)`
- multivalue becomes `@card(0..)` -- our default is @card(0..1) -- works for both relation's roles and attribute ownership (!)
- min/max value becomes `@range(min..max)` (or leave one open if not set)
- min/max/exact cardinality can be `@card(min, max)`

It's actually a really close mapping...  some things don't work so easily, not sure how important they are for actual usages:
- boolean combination of slot types
- arrays == ordered slot contents (coming this year though!)
- subsets (didn't read too much on these)
- default values (we dont have these yet)

Am I mad to think that TypeDB should be a better data store for link-ml model data than other databases??

kevinschaper (approved):
Thanks so much for this PR @GullyBurns, and thank you @flyingsilverfin for dropping in with an expert meta-model comparison!  

Even though long test times have been a challenge on this repo, I'm inclined to say that we should add an action so that the TypeDB integration tests are actually getting exercised - but maybe mitigated with a conditional path check? 

Something like:
<details>

```yaml
# .github/workflows/typedb-integration.yaml
name: TypeDB Integration Tests
on:
  pull_request:
    paths:
      - 'packages/linkml/src/linkml/generators/typedbgen.py'
      - 'tests/linkml/test_generators/test_typedbgen*.py'
  workflow_dispatch:

jobs:
  test_typedb:
    runs-on: ubuntu-latest
    services:
      typedb:
        image: typedb/typedb:3.8.0
        ports:
          - 1729:1729
        options: >-
          --health-cmd "timeout 3 bash -c 'echo > /dev/tcp/localhost/1729'"
          --health-interval 10s
          --health-timeout 5s
          --health-retries 10
          --health-start-period 30s
    steps:
      - uses: actions/checkout@v6
      - uses: astral-sh/setup-uv@v7.3.1
      - uses: actions/setup-python@v6.2.0
        with:
          python-version: "3.12"
      - run: uv sync --all-groups --group typedb
      - run: uv run pytest tests/linkml/test_generators/test_typedbgen_integration.py -m integration -v
```

</details> 

The relational modeling questions are maybe a bit more important. Do want to do some revision to chase after more explicit modeling of relationships with the `represents_relationship` slot at the class level in this PR, or leave that for later?  I was initially thinking that it might be a breaking change, but I guess it makes more sense to think of that as enhanced support for the metamodel. 

Even the extra testing isn't essential for this PR though, so I'll approve, but hold off on merging to see if you want to change anything now. Again, thanks for the PR!


(Also, here's my generated claude review markdown, if you want to point your agent at it to address some of the minor issues feel free, but nothing jumped out at me as 100% essential. )

<details>
# PR #3382 Review: feat: add TypeDB generator (gen-typedb)

**Author:** GullyBurns (+ Claude Sonnet co-authored)
**Status:** All CI checks pass, already approved by cmungall
**Tested:** 25 unit tests + 5 integration tests pass locally against TypeDB 3.8.0 in Docker

## Overall Assessment

Strong first generator for TypeDB. The mapping is sound, the code is clean, and it actually works end-to-end — I verified by loading both the `organization.yaml` and `kitchen_sink.yaml` schemas into a live TypeDB 3.8.0 instance, including data insertion and querying. The engagement from TypeDB's CTO (flyingsilverfin) in the comments is a great sign for future development.

**Recommendation:** Approve with minor suggestions (none blocking).

---

## What Works Well

1. **Correct architecture** — Uses `SchemaView` (`uses_schemaloader = False`), which is the modern/preferred approach
2. **Name collision handling** — The `_build_name_maps()` function that resolves entity/attribute/relation namespace collisions is well thought out
3. **TypeDB inheritance compatibility** — Correctly filters inherited slots to avoid TypeDB's `[SVL42]` redeclaration error
4. **Reserved keyword handling** — Comprehensive list of TypeDB 3.x reserved keywords with automatic `-attr` suffix
5. **Integration tests** — Properly gated behind `@pytest.mark.integration` and socket-based skip when no server available
6. **Idempotency test** — Verifying that `define` is idempotent is a smart test
7. **Good docstrings and doctests** on the helper functions

---

## Issues & Suggestions

### 1. Consider CI coverage for integration tests

The 5 integration tests currently self-skip in CI via socket check since no TypeDB server is available. This is consistent with how other generators are tested — most use in-process validation (e.g., `compile_python()`, `Graph().parse()`, `create_engine("sqlite://")`). TypeQL is the odd one out: there's no standalone Python parser for it, so the only way to validate the output is against a live server.

Since this would be the first (and likely only) generator requiring a sidecar container in CI, it's worth considering whether the cost is justified. A path-conditional workflow would keep it lightweight — only spinning up TypeDB when the generator or its tests actually change:

```yaml
# .github/workflows/typedb-integration.yaml
name: TypeDB Integration Tests
on:
  pull_request:
    paths:
      - 'packages/linkml/src/linkml/generators/typedbgen.py'
      - 'tests/linkml/test_generators/test_typedbgen*.py'
  workflow_dispatch:

jobs:
  test_typedb:
    runs-on: ubuntu-latest
    services:
      typedb:
        image: typedb/typedb:3.8.0
        ports:
          - 1729:1729
        options: >-
          --health-cmd "timeout 3 bash -c 'echo > /dev/tcp/localhost/1729'"
          --health-interval 10s
          --health-timeout 5s
          --health-retries 10
          --health-start-period 30s
    steps:
      - uses: actions/checkout@v6
      - uses: astral-sh/setup-uv@v7.3.1
      - uses: actions/setup-python@v6.2.0
        with:
          python-version: "3.12"
      - run: uv sync --all-groups --group typedb
      - run: uv run pytest tests/linkml/test_generators/test_typedbgen_integration.py -m integration -v
```

As a separate workflow with `paths` triggers, this has zero cost on unrelated PRs — no container startup, no job queued at all. The `workflow_dispatch` trigger allows manual runs when needed. This avoids setting a precedent of sidecar containers in the main CI matrix while still keeping the integration tests honest.

### 2. Missing `file_extension` ClassVar

Every other generator defines `file_extension` (e.g., `"sql"`, `"py"`, `"owl"`). This should be `"tql"`:
```python
file_extension = "tql"
```

### 3. Missing pytest marker

Other generator test files use `pytestmark = pytest.mark.typedbgen` (or similar). The unit test file doesn't have one. This matters for running subsets of tests:
```python
pytestmark = pytest.mark.typedbgen
```
And register it in `pyproject.toml` alongside the existing markers.

### 4. `_resolve_typedb_value_type` — broken type hierarchy walk

Lines 192-194 have a suspicious pattern:
```python
type_def = sv.get_type(type_def.from_schema) if type_def.from_schema else None
break  # avoid infinite loop; fall through to name-based lookup
```
This `break` immediately after the assignment means the `while` loop body only ever executes once. The field `from_schema` is the schema URI the type was defined in — it's **not** the parent type. The correct field for walking up the type hierarchy is `typeof`. This works in practice because the first-level type usually has the XSD URI or base name, but it's technically a bug for deeply nested custom types.

### 5. `serialize` signature mismatch

The `# type: ignore[override]` on `serialize(self)` exists because the base class defines `serialize(self, **kwargs) -> str`. Should accept `**kwargs` for consistency:
```python
```

### 6. Workspace-level `[dependency-groups]` in root `pyproject.toml`

```toml
[dependency-groups]
dev = [
    "typedb-driver>=3.8.1",
]
```
This adds `typedb-driver` to the default dev dependencies for **everyone** who runs `uv sync`. It should probably only be in the package-level optional group (`packages/linkml/pyproject.toml` already has the `typedb` group correctly). The root-level entry should be removed.

### 7. `_typedb_name` always appends `-attr` for reserved words

This applies uniformly regardless of whether the name is being used as an entity, attribute, or relation. For a class named `type`, it becomes `type-attr` even though `type-entity` would be more semantically correct. Minor, but worth noting in the docstring.

### 8. Duplicated ancestor-slot filtering logic

The ancestor slot filtering appears twice in `_collect_entity_defs` — once for `plays_map` computation (line 357-362) and again for the main entity loop (line 389-393). Could be extracted to a helper method.

---

## Design Observations (For Future Work)

### Mixins support

The docs say "partially supported" but the implementation only looks at `sv.get_class(class_name).slots` (direct slots) for relations. Mixin slots (contributed via `mixins:` rather than `is_a:`) may not produce the expected `owns`/`plays` declarations. Worth a dedicated test.

### Enum values not enforced

Enum values are only recorded as comments. As flyingsilverfin noted, TypeDB 3.x has `@values` which could enforce enum constraints at the database level. This would make the mapping significantly more useful.

### Use `represents_relationship` / `relational_role` for Entity vs Relation mapping

Currently the generator maps **every** class to a TypeDB entity, and derives relations purely from object-ranged slots. But LinkML already has explicit metadata for exactly the entity-vs-relation distinction TypeDB needs:

- **`represents_relationship: true`** (class-level) — explicitly marks a class as modeling a relationship rather than an entity. This is exactly a TypeDB `relation` type.
- **`relational_role`** (slot-level, on slots of relationship classes) — marks slots as `SUBJECT`, `OBJECT`, `PREDICATE`, `NODE`, or `OTHER_ROLE`. These map directly to TypeDB's `relates` role declarations.

For example, a LinkML schema like:
```yaml
classes:
  FamilialRelationship:
    represents_relationship: true
    slots:
      - subject
      - object
      - type
slots:
  subject:
    range: Person
    relational_role: SUBJECT
  object:
    range: Person
    relational_role: OBJECT
```

...should generate a TypeDB `relation` (not entity) with proper role names, rather than the current approach of creating an entity plus synthetic relations for each object-ranged slot.

This connects to both flyingsilverfin's comment about classes becoming entities or relations based on their slots, and cmungall's pointer to the [property graph modeling howto](https://linkml.io/linkml/howtos/model-property-graphs.html). The metadata fields are status `testing` but they exist in the metamodel and are the canonical way to express this. Using them would make the TypeDB mapping significantly more natural and would avoid the current pattern of creating one relation-per-slot which can be awkward for classes that represent edges.

---

## Test Quality

Tests are functional style (good), use real schemas (good), and cover the key mapping features. A few gaps:

- **No snapshot/golden file test** — Other generators use snapshot testing for regression detection. Would be valuable here.
- **No test for name collisions** — The `_build_name_maps` collision resolution logic (appending `-attr`/`-rel`) doesn't have a dedicated test with a schema that triggers collisions.
- **No test for reserved keyword handling** — No test schema that has a class or slot named `type`, `entity`, etc.
- **No mixin test** — As noted above.

---

## Docker Testing Notes

For reviewers who want to try the integration tests locally:

```bash
# Start TypeDB
docker run -d --name linkml-typedb-test -p 127.0.0.1:1729:1729 typedb/typedb:3.8.0

# Wait for it (about 5-10 seconds)
sleep 10

# Run integration tests
uv run pytest tests/linkml/test_generators/test_typedbgen_integration.py -m integration -v

# Cleanup
docker rm -f linkml-typedb-test
```

---

## Files Changed Summary

| File | Assessment |
|---|---|
| `typedbgen.py` (470 lines) | Core generator — well structured, minor issues noted above |
| `test_typedbgen.py` (234 lines) | Good coverage, could use collision/reserved/mixin tests |
| `test_typedbgen_integration.py` (142 lines) | Well gated, clean fixture management |
| `typedb.rst` (120 lines) | Clear docs with mapping table |
| `pyproject.toml` changes | Root-level dev dep should be removed |
| Other changes | Clean: `.gitignore`, `__init__.py`, `generator-governance.md`, `index.rst` |
<details>

GullyBurns:
You're so welcome @kevinschaper! 

I got claude to implement the suggested changes from you and @flyingsilverfin - I don't think I have permission to merge the branch. Please take a look at the changes and merge when you feel satisfied. 

I'm excited to dive into using LinkML with my TypeDB projects going forward!

Thanks Again!

G

PR #3390, opened 2026-04-15
Temporary update_model workflow

matentzn:
## Summary

Until we have figured out how to properly version the linkml-model one, this is the workflow I am using locally to update the linkml-model. This PR also updates in the validation model which has not happened in a while.

## Areas of uncertainty

This is temporary.

## Checklist

- [ ] My code follows the [contributor guidelines](https://linkml.io/linkml/maintainers/contributing.html)

## AI Assistance


packages/linkml_runtime/src/linkml_runtime/processing/validation_datamodel.py
```diff
@@ -57,18 +52,18 @@ class TypeSeverityKeyValueType(URIorCURIE):
...
 
 
-@dataclass
+@dataclass(repr=False)
 class ConstraintCheck(YAMLRoot):
     _inherited_slots: ClassVar[list[str]] = []
 
-    class_class_uri: ClassVar[URIRef] = VM.ConstraintCheck
-    class_class_curie: ClassVar[str] = "vm:ConstraintCheck"
+    class_class_uri: ClassVar[URIRef] = LINKML["ClassDefinition"]
```
amc-corey-cox: The regenerated file has all class URIs collapsed to `LINKML["ClassDefinition"]` and all slot URIs collapsed to `LINKML.slot_uri` instead of their own distinct URIs (e.g. `VM["ConstraintCheck"]`, `VM.instantiates`). This pattern repeats across every class and slot in the file.

I ran `gen-python` locally on the same YAML and it produces correct output — each class gets its own URI, each slot gets its own URI. So this isn't a generator bug; it looks like the Makefile invoked a different `gen-python` than the one in this project (which would also explain the `0.0.1` version in the header).

The fix should be straightforward: use `uv run gen-python` in the Makefile target so it picks up the project's own installation, and regenerate.

packages/linkml_runtime/src/linkml_runtime/processing/validation_datamodel.py
```diff
@@ -1,5 +1,5 @@
-# Auto generated from validation_datamodel.yaml by pythongen.py version: 0.9.0
-# Generation date: 2023-01-27T10:37:33
+# Auto generated from validation_datamodel.yaml by pythongen.py version: 0.0.1
```
amc-corey-cox, line 1: Generator version shows `0.0.1` vs the old `0.9.0` — this supports the theory that a different `gen-python` was picked up. Our current generator produces correct per-class/per-slot URIs.

amc-corey-cox (approved):
The Makefile improvements are good — the workflow is now self-contained instead of depending on a local ../linkml-model/ checkout. But the regenerated validation_datamodel.py has incorrect URIs because it was generated by a different gen-python than this project's. Using uv run gen-python in the Makefile and regenerating should fix it.

packages/linkml_runtime/src/linkml_runtime/processing/validation_datamodel.py
amc-corey-cox: This is likely due to the issue fixed in https://github.com/linkml/linkml/pull/3414. We need to regenerate this file with the tool-chain after this model fix.

packages/linkml_runtime/src/linkml_runtime/processing/validation_datamodel.py
amc-corey-cox: This is actually a non-issue that can now be ignored.

amc-corey-cox (dismissed):
See my previous review with new comment. This needs to be regenerated with the current linkml generators and the correct model.

packages/linkml_runtime/Makefile
```diff
@@ -1,18 +1,30 @@
 # Currently the main purposing of this makefile is to synchronize
 # latest releases of linkml-model into linkml_runtime/linkml_model
-#
-# See https://github.com/linkml/linkml-runtime/issues/10
 
-# TODO: make this mechanism more robust
-MODEL_DIR = ../linkml-model/linkml_model/
+BRANCH = main
```
sierra-moxon: This is `main`?  Not a release?

amc-corey-cox:
Hi @matentzn — I pushed a small commit to your branch applying pre-commit auto-fixes (end-of-file + ruff) so `quality-checks` should pass now.

Also, could you respond to @sierra-moxon's question about `BRANCH = main` in the Makefile (https://github.com/linkml/linkml/pull/3390#discussion_r3204706241)? She's asking whether we should pin to a release tag instead of tracking `main` of linkml-model.

matentzn:
@amc-corey-cox @sierra-moxon thanks! Yes, release tag is more correct. I will address this next week. Thanks for the reviews and fixes!

packages/linkml_runtime/Makefile
matentzn: Updated - right call!

matentzn:
@amc-corey-cox I updated to use a release tag instead of main, which is technically correct (thanks @sierra-moxon for feedback). I set to auto-merge, if you could approve.. :) Thanks!

amc-corey-cox (approved):
Yep, this looks good to me. I bumped the repo pin from v1.11.0 and v1.11.0-rc2. The both point at the same commit so this is just an aesthetic/hygiene bump with no impact.

matentzn:
Sweet thanks @amc-corey-cox

PR #3394, opened 2026-04-15
doc: do point to Zenodo "transient" DOI

yarikoptic:
Used DOI which would always redirect to the most recent and that is why did not ember any particular instance of rendered version since should change across releases

> ## AI Assistance
> 
> If you used AI tools while preparing this PR, you are still the author and responsible for understanding, verifying, and defending your submission. Please engage with reviewers personally rather than through your agent during feedback and revisions. See our [AI Covenant](AI_COVENANT.md) for details.

*well -- this is one of the few fully HI ones ;) -- enjoy while they last*

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -23,7 +23,7 @@ from linkml.utils.deprecation import DEPRECATIONS
 # -- Project information -----------------------------------------------------
 
 project = 'linkml'
-copyright = '2021-2024, LinkML Authors'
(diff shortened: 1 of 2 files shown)


dalito:
The Gigascience paper is also out and should be added, too. Would be great if you could include it here.

yarikoptic:
done

dalito (approved):
Thank you!

matentzn (approved):

PR #3388, opened 2026-04-15
Update linkml-model-1.11.0rc1

matentzn:
## Summary

Updates the linkml-model to 1.11.0rc1.

## Areas of uncertainty

- I would like @sneakers-the-rat or @gouttegd to confirm that this is enough for: https://github.com/linkml/linkml/pull/2940

## Checklist

- [X] My code follows the [contributor guidelines](https://linkml.io/linkml/maintainers/contributing.html)
- [ ] I have added tests that prove my fix/feature works
- [ ] Existing tests pass locally with my changes

## AI Assistance

If you used AI tools while preparing this PR, you are still the author and responsible for understanding, verifying, and defending your submission. Please engage with reviewers personally rather than through your agent during feedback and revisions. See our [AI Covenant](AI_COVENANT.md) for details.

PR #3399, opened 2026-04-16
Fix extra slots for jsonschema

### LINKED ISSUE #2238: JSON Schema "additionalProperties" default value not set correctly

gouttegd:
## Summary

Fixes #2238
Supersedes #2940

The PR does the same thing as PR #2940 from @sneakers-the-rat (the first commit is in fact a squash of that PR, minus the “monkey patch” to the meta model), with the following differences:

(a) it builds on top of #3388 instead of “monkey patching” the meta model;
(b) when a class does not have an explicit `extra_slots` slot, then we set the `additionalProperties` to a default value that is based on the global `not_closed` parameter (corresponding to the command line `--closed/--not-closed` option);
(c) the “top-level” `additionalProperties`, at the root of the schema, is set to the same value as the `additionalProperties` of the root class – this is necessary because the top-level `additionalProperties` is the only one that really matters for the root class: if it is set to `false`, then the JSON-Schema validator will reject any extra property at the root level, even if the root class itself has `additionalProperties` set to `true`.

@sneakers-the-rat If you agree with (b) and (c), feel free to cherry-pick the last two commits to your own #2940 PR. :)

## How was this tested?

New tests added to `test_jsonschemagen` (that is entirely @sneakers-the-rat ’s work, I claim zero credit here!)

## Areas of uncertainty

The entire test suite fails because of the meta model update in #3388.

## Checklist

- [x] ~~I~~ @sneakers-the-rat  has added tests that prove my fix/feature works

### DIFF
--- a/packages/linkml/src/linkml/generators/jsonschemagen.py
+++ b/packages/linkml/src/linkml/generators/jsonschemagen.py
@@ -295,6 +295,12 @@ class JsonSchemaGenerator(Generator, LifecycleMixin):
     def start_schema(self, inline: bool = False):
         self.inline = inline
 
+        top_additional_properties = self.not_closed
+        if self.top_class:
+            top_class_def = self.schemaview.get_class(self.top_class)
+            if top_class_def is not None:
+                top_additional_properties = self.get_additional_properties(top_class_def)
+
         self.top_level_schema = JsonSchema(
             {
                 "$schema": "https://json-schema.org/draft/2019-09/schema",
@@ -303,7 +309,7 @@ class JsonSchemaGenerator(Generator, LifecycleMixin):
                 "version": self.schema.version if self.schema.version else None,
                 "title": self.schema.title if self.title_from == "title" and self.schema.title else self.schema.name,
                 "type": "object",
-                "additionalProperties": self.not_closed,
+                "additionalProperties": top_additional_properties,
             }
         )
 
--- a/packages/linkml/src/linkml/generators/jsonschemagen.py
+++ b/packages/linkml/src/linkml/generators/jsonschemagen.py
@@ -311,15 +311,13 @@ class JsonSchemaGenerator(Generator, LifecycleMixin):
         cls = self.before_generate_class(cls, self.schemaview)
 
         subschema_type = "object"
-        additional_properties = False
         if self.is_class_unconstrained(cls):
             subschema_type = ["null", "boolean", "object", "number", "string"]
-            additional_properties = True
 
         class_subschema = JsonSchema(
             {
                 "type": subschema_type,
-                "additionalProperties": additional_properties,
+                "additionalProperties": self.get_additional_properties(cls),
                 "description": be(cls.description),
             }
         )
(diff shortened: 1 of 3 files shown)


sneakers-the-rat:
I am not sure why we need this? the monkeypatch was explicitly marked as being temporary and would be removed with the update of the metamodel? it was just to demonstrate that it worked until then.

idk i'm not super protective of my work getting credited usually, but this has been happening a lot to me recently where people will just copy/paste my PRs and then claim that they are superceded. usually not great to duplicate PRs either for keeping the history of things understandable. 

replied on #2940 re: the root validation issue, seems like it's worth discussion about the best path for that, but if we end up going this direction then i'd be happy to merge in your fix to that branch to give you credit.

gouttegd:
@sneakers-the-rat Not trying to steal any credit here (though I admit using the word “supersedes #2940” was inappropriate, and it does _not_ convey what was my intention – for that I apologise).

I’ve spent some time yesterday trying to fix the issue I’ve mentioned [here](https://github.com/linkml/linkml/pull/2940#issuecomment-4259767776), and since I’ve found what I believe is a proper way to do it, I wanted to have the fix out _somewhere_, because it’s easier to discuss a potential fix if there’s some code somewhere in a branch to see.

> seems like it's worth discussion about the best path for that, but if we end up going this direction then i'd be happy to merge in your fix

As long as the issue is fixed, I don’t mind in the slightest to discard this PR and to carry on with yours. :)

That’s what I meant by 

> If you agree with (b) and (c), feel free to cherry-pick the last two commits to your own #2940 PR

gouttegd:
Superseded by #3612

Issue #68, opened 2020-09-01
YAML Generator needs an "all" option


hsolbrig:
When generating _instances_ of a model, it would be useful to be able to emit all model elements in YAML whether they are filled or not.  We would use this for generating templates.

Note: This may also apply to JSON generators as well

Issue #67, opened 2020-09-01
Error diagnosis for malformed class is difficult to understand


hsolbrig:
```yaml
classes:
    MapVersion:
          this is some text
```
Produces a cryptic error message

sierra-moxon:
Need to figure out when this occurs and is still an issue - we also are working "<file>" to be more specific.

cmungall:
is there a test to reproduce this?

PR #3407, opened 2026-04-17
Deterministic RDF Serialization via pyoxigraph RDFC-1.0

### LINKED ISSUE #1943: Possibility for deterministic order of multivalue slots in RDF output?

matentzn:
Closes #1943, #696

Whoever reviews this, please carefully read the following as this is a very impactful PR. I think this should probably be tested a little manually as well instead of only relying on CI/CD.

## Summary

All RDF generators (OWL, RDF, SHACL, ShEx) and the rdflib dumper now produce
deterministic output with stable blank node labels and sorted triples, using
pyoxigraph's RDFC-1.0 canonicalization.

Every `graph.serialize(format=...)` call in the RDF serialization path has been
replaced with `canonicalize_rdf_graph(graph, output_format=...)`, a new utility
in `linkml_runtime.utils.rdf_canonicalize`.

## How it works

1. Serialize the rdflib `Graph` to N-Triples (most unambiguous transfer format)
2. Parse into pyoxigraph and build a `Dataset`
3. Canonicalize blank node labels with RDFC-1.0 (`dataset.canonicalize()`)
4. Sort triples by `(subject, predicate, object)` string representation
5. Serialize with pyoxigraph's Turtle/NT serializer, passing prefix bindings
   and base IRI from the original rdflib graph

## Dependency changes

- **Added**: `pyoxigraph >= 0.5.6` as a required dependency of `linkml-runtime`
  (single Rust binary wheel, zero Python deps)
- **Removed**: `morph-kgc` from the `tests` dependency group (conflicts with
  `pyoxigraph >= 0.5.6` since morph-kgc pins `pyoxigraph < 0.4.0`). The
  YARRRML tests use `pytest.importorskip("morph_kgc")` so they skip gracefully.

### 1. `xsd:string` normalization (by design, unfixable)

pyoxigraph follows RDF 1.1 where plain string literals and
`"text"^^xsd:string` are identical. The explicit `^^xsd:string` datatype
annotation is **erased at the Rust data-model level** during `Literal`
construction (see [oxigraph#798](https://github.com/oxigraph/oxigraph/issues/798))
and cannot be recovered. There is no flag, option, or workaround to preserve it.

**Impact**: Code that re-parses generated Turtle with rdflib will see
`Literal("x")` (datatype `None`) instead of `Literal("x", datatype=XSD.string)`.
One test (`test_ifabsent`) was updated to accept both forms. This is a semantic
no-op per RDF 1.1.

### 2. Non-standard RDF fallback

Some generators (e.g. SHACL with `include_annotations=True`) produce triples
with `Literal` predicates, which is technically invalid RDF. pyoxigraph
correctly rejects these. The utility catches `SyntaxError` and falls back to
rdflib's serializer for such graphs (with a warning).

### 3. Numeric short forms

pyoxigraph uses Turtle short forms: `42` instead of `"42"^^xsd:integer`,
`true` instead of `"true"^^xsd:boolean`, `1.23` instead of
`"1.23"^^xsd:decimal`. rdflib parses these back with the correct datatype,
so this is lossless and not a practical issue.

### 4. Base IRI / prefix collision

When a graph has both `@base` and a prefix whose namespace equals the base IRI
(e.g. rdflib's auto-bound `base:` prefix), pyoxigraph emits CURIEs like
`base:label` that rdflib's parser rejects. The utility skips any prefix whose
namespace matches the base IRI to avoid this.

## TODO:

- [ ] Snapshot regeneration needed (separate step - will do that now as a separate PR)

## Remaining `graph.serialize()` calls (intentionally not patched)

@amc-corey-cox (or any other reviewer) - we should all agree on these.

- `rdf_loader.py` — serializes to `json-ld` internally for loading, not user-facing RDF output
- `validate_jsonld_context.py` — debug logging only (`logger.info`)
- `test_issue_103.py` — serializes to `json-ld`, not RDF
- `test_rdfgen.py:105` — inside `@pytest.mark.skip("TODO")` (dead code)


tests/linkml/test_generators/test_shaclgen.py
```diff
@@ -376,7 +376,18 @@ def test_ifabsent(input_path):
...
+            # per RDF 1.1 (plain literals and xsd:string are equivalent).
+            # Accept either form for xsd:string typed values.
+            expected = Literal(default_value, datatype=datatype)
+            if (subject, SH.defaultValue, expected) in g:
+                return
+            if datatype and str(datatype) == "http://www.w3.org/2001/XMLSchema#string":
+                if (subject, SH.defaultValue, Literal(default_value)) in g:
+                    return
+            raise AssertionError(
+                f"Expected ({subject}, sh:defaultValue, {expected!r}) not found in graph"
```
matentzn: **The xsd:string normalization** -- the `test_ifabsent` change to accept both forms is reasonable, but I want to make sure we're not papering over something. If downstream consumers of linkml-generated RDF rely on explicit `^^xsd:string` annotations, this could bite them. The PR description correctly notes this is a semantic no-op per RDF 1.1, so I think it's fine, but worth a heads-up in release notes.

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
```diff
@@ -0,0 +1,133 @@
...
+    if ox_format is None:
+        logger.warning(
+            "pyoxigraph does not support format %r; falling back to rdflib serializer",
+            output_format,
+        )
+        return graph.serialize(format=output_format)
+
+    # 1. Transfer rdflib graph to pyoxigraph via N-Triples.
+    nt_data = graph.serialize(format="nt")
+    nt_bytes = nt_data.encode("utf-8") if isinstance(nt_data, str) else nt_data
```
matentzn, line 153: **Performance** -- the serialize-to-NT -> parse-into-pyoxigraph -> canonicalize -> sort -> serialize pipeline has some overhead. For typical linkml schemas this is probably negligible, but for very large graphs it could add up.

jdsika:
## Alternative approach: diff-stable deterministic output

We've been working on a related feature in our fork ([ASCS-eV/linkml#1](https://github.com/ASCS-eV/linkml/pull/1)) that takes a different angle on deterministic serialization, and wanted to share our perspective in case it's useful.

### The "small diff" requirement

Our primary use-case is **storing generated artifacts (OWL, SHACL, JSON-LD) in version control** and reviewing changes during pull requests. For this workflow, determinism alone is necessary but not sufficient — what matters is that **a small schema change produces a small, reviewable diff** in the generated output.

Pure RDFC-1.0 canonicalization with sequential blank-node IDs (`_:c14n0`, `_:c14n1`, …) guarantees identical output for identical input. But when the schema *does* change — even minimally — adding or removing a single triple early in canonical order can **renumber all subsequent blank nodes**, producing a large diff that touches unrelated parts of the file. The diff is technically correct, but it's unreadable: you can't tell what actually changed without re-diffing the parsed graphs.

Put differently: determinism without diff stability answers **"did anything change?"** (a binary check), but not **"what changed?"** (which is what a human reviewer needs). For reproducibility of the exact same LinkML model that binary check is sufficient, but for version-controlled artifacts where humans review changes, it's not.

### Our approach: WL structural hashing

Our implementation uses a three-phase hybrid pipeline:

1. **RDFC-1.0** (via pyoxigraph) canonicalizes the graph — same correctness guarantee as this PR
2. **Weisfeiler-Lehman structural hashing** replaces sequential `_:c14nN` IDs with content-based hashes derived from each blank node's multi-hop neighbourhood (predicates, literals, named nodes — not blank-node IDs). These hashes are stable: adding/removing unrelated triples only affects structurally involved blank nodes.
3. **rdflib re-serialization** produces idiomatic Turtle (inline blank nodes, `()` collection syntax, preserved `xsd:string` annotations)

Additionally, we sort class/slot expressions (`any_of`, `all_of`, `none_of`, `exactly_one_of`) in the OWL generator, so logically equivalent schemas produce identical OWL regardless of declaration order.

### Key design differences

| Aspect | This PR (#3407) | Our approach ([ASCS-eV/linkml#1](https://github.com/ASCS-eV/linkml/pull/1)) |
|--------|----------------|--------------|
| Activation | Always-on | Opt-in `--deterministic` flag |
| Blank-node IDs | Sequential `_:c14nN` | Content-based `_:b<sha256>` |
| Diff on schema change | Global renumbering possible | Only affected nodes change |
| pyoxigraph | Required dependency | Optional (lazy import) |
| Expression sorting | No | Yes (`any_of`/`all_of`/`none_of`/`exactly_one_of`) |
| Turtle style | pyoxigraph native | rdflib idiomatic (inline blanks, collections) |
| xsd:string | Erased (RDF 1.1) | Preserved |

### Not a competing PR

We're not suggesting our approach should replace this one — they solve overlapping but distinct problems. We wanted to flag the diff-stability angle since it's been the driving requirement in our workflows, and it might be relevant for other users who commit generated artifacts.

Happy to discuss further or collaborate on finding a combined solution.

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
```diff
@@ -0,0 +1,223 @@
...
+
+**Known limitations:**
+
+1. **xsd:string normalization**: pyoxigraph follows RDF 1.1, where plain
+   string literals and ``"text"^^xsd:string`` are identical.  The output
+   will never contain explicit ``^^xsd:string`` annotations.  Code that
+   re-parses the output with rdflib will see ``Literal("x")`` (datatype
+   ``None``) rather than ``Literal("x", datatype=XSD.string)``.
+
+2. **Non-standard RDF**: Graphs with literal predicates (e.g. SHACL
```
sneakers-the-rat, line 15: are literals allowed as predicates in RDF? i guess the question is "is SHACL wrong or is pyoxigraph wrong?" i am in general skeezed out by an abundance of fallbacks after being scarred by LLM code, so this is a code smell/curiosity question not a serious concern

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
```diff
@@ -0,0 +1,223 @@
...
+        (ox.Triple(q.subject, q.predicate, q.object) for q in quads),
+        key=lambda t: (str(t.subject), str(t.predicate), str(t.object)),
+    )
+
+    # 5. Collect prefixes for formats that support them.
+    base_iri = str(graph.base) if graph.base else None
+    prefixes: dict[str, str] | None = None
+    if ox_format in _PREFIX_FORMATS:
+        prefixes = {}
+        for prefix, namespace in graph.namespace_manager.namespaces():
```
sneakers-the-rat, line 190: this appears (e.g. in the tutorial examples above) to add all prefixes in the namespace manager even if they are not referenced in the graph. this seems like an amount of noise that would be good to avoid if we can (although `https://w3id.org/linkml/examples/personinfo` 404's)

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
```diff
@@ -0,0 +1,223 @@
...
+            # Skip prefixes whose namespace matches the base IRI to avoid
+            # pyoxigraph emitting CURIEs like `base:label` that conflict
+            # with the @base directive.
+            if base_iri and ns_str == base_iri:
+                continue
+            # Skip namespace IRIs that pyoxigraph rejects or that produce
+            # CURIEs rdflib cannot round-trip.  Valid namespace IRIs for
+            # prefix use should end with '/' or '#' and contain no query
+            # parameters or fragment-like characters in the middle.
+            if not _is_safe_prefix_iri(ns_str):
```
sneakers-the-rat, line 203: silent skips probably not great

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
```diff
@@ -0,0 +1,223 @@
...
+            format=ox_format,
+            prefixes=prefixes,
+            base_iri=base_iri,
+        )
+    except ValueError:
+        # pyoxigraph rejects prefixes with invalid IRIs (e.g. containing
+        # fragment-like characters such as double '#').  Retry without
+        # the offending prefixes by falling back to no prefixes, which
+        # still produces valid (if verbose) Turtle.
+        logger.warning("pyoxigraph rejected one or more prefix IRIs; serializing without prefix declarations")
```
sneakers-the-rat, line 219: might want to be a bit more selective with this check, idk if pyoxigraph uses a more specific error type, but might want to check the error text and `raise` if it's a different error than the invalid prefix IRI error

packages/linkml/src/linkml/generators/rdfgen.py
```diff
@@ -68,15 +70,7 @@ def end_schema(self, output: str | None = None, context: str = None, **_) -> str
...
-            # Binary-safe when -o/--output is used:
-            # delegate to RDFLib (Graph.serialize(destination=..., format=...)).
-            # Serializers that produce bytes write directly to the file; stdout stays empty.
-            fmt = "turtle" if self.format == "ttl" else self.format
-            try:
-                out = graph.serialize(format=fmt)
-            except UnicodeDecodeError:
-                graph.serialize(destination=output, format=fmt)
-                return ""
+            out = self._data(graph)
```
sneakers-the-rat, line 73: naming is odd, but pre-existing, this is not what i would expect a method named `_data` to do

packages/linkml/src/linkml/generators/rdfgen.py
```diff
@@ -44,7 +45,8 @@ def __post_init__(self):
         super().__post_init__()
 
     def _data(self, g: Graph) -> str:
-        return g.serialize(format="turtle" if self.format == "ttl" else self.format)
+        fmt = "turtle" if self.format == "ttl" else self.format
+        return canonicalize_rdf_graph(g, output_format=fmt)
```
sneakers-the-rat, line 49: can this be made parameterizable? like canonicalize it if `canonicalize=True`? i think it's probably the right default to canonicalize on serialization, but can also imagine someone would want to disable that if they don't care about canonicalization and it takes a long time

sneakers-the-rat:
just gave a brief skim rather than a proper review bc time reasons. In general i am down with the approach, think using RDFC is the right call, and don't mind the pyoxigraph dep (though as said elsewhere, it woudl be great to make most deps optional e.g. in an "rdf" group). 

can we get some top-level profiling info on how long the canonicalization takes on e.g. the person, metamodel, and biolink schemas with and without the canonicalization here? just total execution time is fine.

> But when the schema does change — even minimally — adding or removing a single triple early in canonical order can renumber all subsequent blank nodes, producing a large diff that touches unrelated parts of the file. The diff is technically correct, but it's unreadable: you can't tell what actually changed without re-diffing the parsed graphs.

similarly, can we get an example of what a diff looks like in the github PR diff viewer (which can be pretty dumb sometimes) with this problem? naively it doesn't sound like that big of a problem to me, since the biggest problem at least within linkml is the random reordering when the schema *doesn't* change, but if it really is unreadable i'd like to see how bad that can be. I still think we should go with the standard rather than rolling our own, but it's definitely still worth getting a picture of what the drawbacks are.

jdsika:
Hey @sneakers-the-rat, I have created two example PRs for you to see the diff stability problem at the use case of a simple change in our LinkML model (we are remodelling ASAM OpenLABEL as LinkML). Both PRs make the **exact same model change** — adding one string slot to one class.

**PR #67 uses @matentzn's exact implementation** from this PR (pinned to `f4001f19`). **PR #68 uses our independent WL hashing implementation** on our [develop fork](https://github.com/ASCS-eV/linkml). We have also rebased our WL feature on top of this PR's head in [`feat/deterministic-output-v2`](https://github.com/ASCS-eV/linkml/tree/feat/deterministic-output-v2), which layers the structural hashing as an opt-in `--deterministic` flag on top of the RDFC-1.0 base.

Additionally @matentzn — our rdflib re-serialization step also resolves the `xsd:string` erasure issue. Upstream RDFC-1.0 (pyoxigraph) normalizes `"value"^^xsd:string` to bare `"value"` per RDF 1.1, which loses explicit datatype annotations (only 19 `xsd:string` occurrences remain in PR #67 vs 192 in PR #68). Our `--deterministic` pipeline round-trips through rdflib which preserves them.

This is the agent-generated short summary you'll also find in the PR descriptions.

Best regards, Carlo

---

**Demo PRs (same model change, different serialization)**

- **[PR #67 — Upstream RDFC-1.0](https://github.com/ASCS-eV/ontology-management-base/pull/67)**: Uses @matentzn's `pyoxygraph-rdf` branch directly — sequential `_:c14nN` IDs
- **[PR #68 — WL Structural Hashing](https://github.com/ASCS-eV/ontology-management-base/pull/68)**: Uses our `--deterministic` flag — content-based `_:b<sha256>` IDs

**The model change**: +6 lines in YAML — add `scenarioComment` (string) to `AdminTag`.

**The artifact diff**:

| | PR #67 (RDFC-1.0) | PR #68 (WL hashing) |
|---|---:|---:|
| OWL | ~8400 lines churned | **15 lines** added |
| SHACL | ~5300 lines churned | **7 lines** added |
| Total | **+7284 / -6422** | **+22 / -0** |

PR #68's diff is **100% signal** — every line is about `scenarioComment`. PR #67's diff is **~0.2% signal** — the rest is `_:c14nN` → `_:c14nM` renumbering.

Our `--deterministic` flag builds on top of RDFC-1.0 (correctness layer) and adds WL hashing (stability layer) + rdflib re-serialization (readability layer). Rebased implementation: [ASCS-eV/linkml `feat/deterministic-output-v2`](https://github.com/ASCS-eV/linkml/tree/feat/deterministic-output-v2).

sneakers-the-rat:
It seems like most of the big changes are just from going from "not having RDFC applied" to "having RDFC applied," like

```turtle
openlabel_v2:OddDynamicElements a owl:Class ;
    rdfs:label "OddDynamicElements"@en ;
    rdfs:subClassOf [ a owl:Restriction ;
            owl:maxCardinality 0 ;
            owl:onProperty openlabel_v2:horizontalCurvesValue ],
        [ a owl:Restriction ;
            owl:allValuesFrom xsd:string ;
            owl:onProperty openlabel_v2:HorizontalStraights ],
        [ a owl:Restriction ;
            owl:minCardinality 0 ;
            owl:onProperty openlabel_v2:ParticulatesMarine ],
```

(abbreviated) goes to

```turtle
openlabel_v2:OddDynamicElements a owl:Class , linkml:ClassDefinition ;
	rdfs:label "OddDynamicElements" ;
	rdfs:subClassOf openlabel_v2:Odd , _:c14n100 , _:c14n103 , _:c14n105 , _:c14n113 , _:c14n130

_:c14n100 a owl:Restriction ;
	owl:minCardinality 0 ;
	owl:onProperty openlabel_v2:horizontalCurvesValue .
```

(abbreviated).

sorry to be a bother, but would it be possible to split that into 1) convert existing schema into RDFC form, and then 2) show the diff of just the change to the schema? that would be the relevant comparison to make: "once things are in RDFC form, how noisy are diffs?" not "how noisy is it to convert to RDFC"

matentzn:
@jdsika — I haven't dug into your WL hashing implementation yet, but I'm coming around to the view that RDFC alone may not be enough.

I put up a demo PR (draft, do-not-merge) that isolates the problem: #3481. Its base is RDFC-applied artifacts with no schema change; its head adds **one optional string slot** (`person_comment`) to the `Person` class in `examples/PersonSchema/personinfo.yaml` (+6 YAML lines) and regenerates. The PR diff is therefore exactly "once everything is in RDFC form, what does adding one slot look like?" — the comparison @sneakers-the-rat asked for in his [May 5 comment](https://github.com/linkml/linkml/pull/3407#issuecomment-4382363519).

https://github.com/linkml/linkml/pull/3481/files

**Result:** ~1100 changed lines across `personinfo.owl.ttl` and `personinfo.shacl.ttl`, of which 7 mention the new slot. The rest is blank-node renumbering: introducing one new blank node forces every `_:c14nN` later in canonical order to shift, and every triple that referenced those nodes changes too — including in classes structurally unrelated to the new slot.

**My preferred path**, but I'd like a sense from the team before committing:

- **This PR (#3407)** lands as the **serialization determinism** layer — idempotent output for an unchanged schema. That alone is a real fix for the "random reordering on regeneration" pain that motivated the work.
- A **follow-up PR** adds the **minimal-diff-under-schema-change** layer (WL hashing or equivalent). I'd support that follow-up, and @jdsika's implementation is a strong starting point.

Splitting it this way keeps #3407 reviewable and unblocks the determinism win, while making diff stability a focused conversation of its own. If folks think the two should land together, I'm open to it — but I'd want that decided explicitly rather than by default.

sneakers-the-rat:
Wow that is indeed pretty damn annoying and noisy, thanks for setting those demos up, both of you, super helpful to see.

Yeah I agree that they should be separated - having canonicalization is basically a "is the library proficient at expected RDF features" kind of thing, even if sadly it does not produce clean diffs (though it would solve the random wander diff noise).

I would be amenable to a minimal diff serialization form as an opt-in feature clearly labeled as such, as a separate change, if it is lossless and etc. After seeing how annoying that diff is.

jdsika:
I definitely agree with splitting and I would in fact opt for "no flag" as it is not really a wanted behaviour to have changes in output without changes in input and It reduces the code paths to maintain in the future.

sneakers-the-rat:
I think a bespoke serialization format being the default is probably not expected behavior for most people serializing RDF, even if it has desirable qualities, but maybe that's a discussion for the second pr

matentzn:
For the reviewers @amc-corey-cox @kevinschaper : Two things to "check":

1. will the proposed approach create a performance bottleneck for super large inputs (that was not there before)
2. Are we 100% convinced that the code will realise the "serialization determinism" claimed by myself - I only spot checked this

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
```diff
@@ -0,0 +1,223 @@
...
+        return graph.serialize(format=output_format)
+
+    dataset = ox.Dataset()
+    for triple in triples:
+        dataset.add(ox.Quad(triple.subject, triple.predicate, triple.object, ox.DefaultGraph()))
+
+    # 3. Canonicalize blank node labels with RDFC-1.0.
+    dataset.canonicalize(ox.CanonicalizationAlgorithm.RDFC_1_0)
+
+    # 4. Sort triples for deterministic ordering.
```
sneakers-the-rat, line 173: genuine q: is this necessary? i would have thought that canonicalization would render it as ordered. the question comes from the most recent comment - are we sure about serialization invariance - which we should be! or else that would be a problem for the w3c and pyoxigraph. but if we e.g. apply operations after the fact where eg. we reorder in a way that doesn't use the canonicalized bnode labels then we could undo that (again not saying that's what's happening here, asking as a question)

amc-corey-cox: Yes, I think it is — RDFC-1.0 nails down the bnode labels, but pyoxigraph's Turtle emitter doesn't promise lexicographic subject order across versions, so the explicit sort is cheap insurance for the determinism claim. Worth a one-line code comment here saying so, since otherwise it reads like dead work.
I may be wrong here but that is how I understand it.

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
amc-corey-cox: Concurring — and the tutorial example is concrete evidence: `examples/tutorial/tutorial01/data.ttl` goes from a 5-triple file with one prefix declaration to the same 5 triples plus 33 prefix bindings, including a duplicate `schema:`/`schema1:` pair (https vs http). An O(n) walk over `sorted_triples` to collect actually-referenced namespaces would tighten this and shrink the diff in the tutorials substantially.

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
amc-corey-cox: Adding to your fallback concern from a different angle: the `SyntaxError` catch around the pyoxigraph parse drops back to `graph.serialize()` and emits `logger.warning`, but a logger warning isn't visible by default in most CLI contexts. A user running e.g. `shaclgen --include-annotations` would silently get non-deterministic output and not know. Either bump to a stderr notice, or surface it in the generator that triggered it.

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
amc-corey-cox: Concurring. Matching on the error message text and re-raising any `ValueError` that doesn't look like an invalid-prefix-IRI complaint would mean a future serializer bug surfaces as a stack trace instead of silently dropping all prefixes from the output.

tests/linkml/test_generators/test_shaclgen.py
amc-corey-cox: +1 on the release-note ask. Accepting both forms in the test is fine, but downstream consumers who pattern-match on explicit `^^xsd:string` will see a behavior change without warning otherwise — worth a CHANGELOG entry on the linkml-runtime side calling it out.

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
```diff
@@ -0,0 +1,223 @@
...
+    round-trips through rdflib.
+    """
+    if not prefixes:
+        return turtle_text
+
+    # Match: a prefix name, ':', a local part (no whitespace or token
+    # delimiters), ending in ``\.``, followed by whitespace.  Use a
+    # negative lookbehind to avoid matching inside ``<...>`` or word
+    # characters that would make this a substring of something else.
+    pattern = re.compile(
```
amc-corey-cox, line 85: Regex surgery on serializer output is the kind of thing that bites once a year on an edge case. The negative lookbehind plus `test_iri_with_trailing_dot_round_trips` cover the basics, but worth one more test where a CURIE-shaped substring appears inside a string literal containing whitespace, just to pin down that it can't false-match.

amc-corey-cox (commented):
Reviewed at f4001f1. Scope-narrowing to "serialization determinism only" with WL hashing deferred is the right call — RDFC-1.0 is the standard and a bespoke default would surprise downstream users.

Implementation is sound and test coverage on the documented limitations is genuinely good. I left in-thread replies on the existing prefix-bloat, fallback, ValueError-selectivity, sort-step, and xsd:string-changelog threads rather than re-litigating them, plus one new inline about regex robustness.

Two structural items I'd want decided before merge:

1. Required vs optional `pyoxigraph` dep on `linkml-runtime`. The `morph-kgc` removal in this same PR is concrete evidence of the resolver pressure — any downstream that pins a different `pyoxigraph` is now unsatisfiable. Worth an explicit team decision before locking in.

2. Performance numbers for personinfo / metamodel / biolink. @sneakers-the-rat asked up top; small ask, would close the question.

Determinism claim: yes for unchanged graphs in-process. A subprocess-based test would be a stronger guarantee against future regressions where someone slips an `id()`-keyed iteration into the path, but that can be a follow-up.

`morph-kgc` YARRRML tests are now silently skipped via `importorskip` — worth either a pinned test env or a tracking issue rather than leaving as a TODO in the PR description.

amc-corey-cox:
Picked up the perf and determinism checks you flagged on May 7.

**Perf — fine.** Canonicalize vs rdflib serialize: 1.1× on personinfo (1.2k triples), 1.2× on metamodel (5.6k), 1.1× on biolink (15.7k). Not a bottleneck.

**Determinism — function is correct; end-to-end is partial.** `canonicalize_rdf_graph` is byte-stable across processes for any fixed input graph. OWL regeneration is byte-stable end-to-end on schemas without relative-IRI annotations. SHACL, RDFGen, and metamodel OWL still differ across processes, but the causes are all upstream of your code — captured in #3516. None of them are introduced by this PR; they're pre-existing, just newly visible.

amc-corey-cox (approved):
With my comment above this looks good to me now. Nice work @matentzn and thanks for all the help @sneakers-the-rat and @jdsika . This was a genuine team effort and it is much better for it.

amc-corey-cox:
Merged `main` into the PR branch to clear conflicts (pyproject.toml + uv.lock). Two changes from `main`'s `tests` dep group landed against the PR's removal of `morph-kgc`:

- Kept `duckdb>=1.5.2` (new on main, no conflict).
- Kept `morph-kgc` removed — matches the PR's stated intent; `morph-kgc 2.10.0` still pins `pyoxigraph<0.4.0,>=0.3.0`, incompatible with the PR's `pyoxigraph >= 0.5.6`. YARRRML tests still `pytest.importorskip("morph_kgc")` so they'll skip cleanly.

`uv.lock` regenerated. `tests/linkml_runtime/test_utils/test_rdf_canonicalize.py` passes locally (12/12).

PR #3412, opened 2026-04-17
cli: Accept -h as a short equivalent of --help.

### LINKED ISSUE #3411: Use `-h` as a short option for `--help`
The `linkml` command (and all its subcommands) should follow 50 years of Unix traditions and accept `-h` as the short option equivalent of `--help`.

In the long-term, this would save an unfathomable amount of time by preventing this kind of sequence:

```console
$ linkml -h
Usage: linkml [OPTIONS] COMMAND [ARGS]...
Try 'linkml --help' for help.

Error: No such option: -h

$ linkml --help
Usage: linkml [OPTIONS] COMMAND [ARGS]...

  LinkML: A flexible linked data modeling language

Options:
  -V, --version  Show the version and exit.
  --help         Show this message and exit.

Commands:
  convert   Converts instance data to and from different LinkML Runtime...
  dev       Helper tools for linkml development
  examples  Process a folder of examples and a folder of counter examples.
  fix       Apply schema-fixer commands.
  generate  Generate formats from a LinkML schema
  lint      Run linter on SCHEMA.
  sqldb     Run the LinkML SQL CLI.
  validate  Validate data according to a LinkML Schema
```

which happens to me about **every single time** I need to use a LinkML command.

**Additional context**
The only drawback of such a change is that it would reserve the `-h` option, which means that a LinkML command would need to find another letter if it wanted to use a short name for an option that starts with `h`.

I don’t see that as a big issue, since having a `-h` option that does anything else than displaying a help message would be a very bad idea anyway – most users would expect `-h` to display the help (because of the aforementioned 50 years of Unix traditions) and could be surprised by a `-h` option that does something else.

gouttegd:
## Summary

Fixes #3411

The use of `-h` as a short option equivalent of `--help` is very common, but unfortunately it is not the default behaviour of the Click library, which by default sets `--help` as the only help option.

In this PR we explicitly instruct Click to accept both `-h` and `--help` as the help option on the top-level (`linkml`) command – this automatically propagates down to every subcommand.

## How was this tested?

Manual testing that `linkml -h`, `linkml lint -h`, `linkml generate shacl -h`, and a few other commands do display the help.

## Areas of uncertainty

None.

## Checklist

## AI Assistance

No AI tools used.


dalito (approved):

matentzn (approved):
Looks great, thank you!

PR #3414, opened 2026-04-17
fix(pythongen): skip runtime-computed ifabsent bakes for the metamodel

### LINKED ISSUE #3413: pythongen bakes runtime-computed ifabsent defaults when regenerating meta.py
## Summary

When `pythongen` generates `meta.py` from `meta.yaml`, it bakes the _runtime-computed_ `ifabsent` directives (`default_range`, `class_curie`, `slot_curie`, `class_uri`, `slot_uri`) into `SlotDefinition` / `ClassDefinition` / `EnumDefinition` / `EnumBinding` / `SlotExpression` / `AnonymousSlotExpression` as **static Python dataclass defaults**. Those defaults then propagate to every `SlotDefinition` / `ClassDefinition` instance that `SchemaView` constructs from a user schema — poisoning user-schema semantics with the metamodel's own internal values.

## Symptom

In a freshly regenerated `meta.py`:

```python
class SlotDefinition(Definition):
    range: Optional[Union[str, ElementName]] = "string"              # was None
    slot_uri: Optional[Union[str, URIorCURIE]] = "linkml:slot_uri"   # was None

class ClassDefinition(Definition):
    class_uri: Optional[Union[str, URIorCURIE]] = "linkml:ClassDefinition"   # was None

class EnumDefinition(Definition):
    enum_uri: Optional[Union[str, URIorCURIE]] = "linkml:EnumDefinition"     # was None
```

Reproduced against PR #3388's regenerated `meta.py`:

```python
sv = SchemaView(range_local.yaml)   # schema with no default_range, no explicit ranges
sv.schema.default_range is None     # ✓
sv.get_slot("none_range").range     # → "string"  (expected: None)
```

`SchemaView.induced_slot()` has correct runtime resolution at `schemaview.py:1700`:

```python
if v is None and metaslot_name == "range":
    v = self.schema.default_range
```

— but it's dead code, because the dataclass default populates `range="string"` during instance construction before the check runs.

## Root cause

`packages/linkml/src/linkml/generators/common/ifabsent_processor.py:389-390`:

```python
def _map_default_range_special_case(self, default_value, slot, cls) -> str:
    return f'"{self.schema_view.schema.default_range}"'
```

When pythongen is generating `meta.py`, `self.schema_view.schema` _is_ the metamodel schema, and `meta.yaml:54` declares `default_range: string` as a load-bearing authoring default (25 core meta-slots like `name`, `title`, `description` rely on it to resolve to `string`). So the processor bakes the metamodel's own internal default as the Python class-level default for every slot/class it generates — including `SlotDefinition`, which represents every user-schema slot.

The same mechanism applies to `class_curie` / `slot_curie` / `class_uri` / `slot_uri` via `_map_curie_special_case` and `_map_uri_special_case`.

## Blast radius

On the PR #3388 branch (with the regenerated `meta.py`), this produces **741 failures** across `test_schemaview.py::test_slot_range` (112), `test_logical_model_transformer.py` (70), and the compliance suites (~500 across boolean/inheritance/inlined/uniqueness/core). Patching the four defaults back to `None` collapses all 741 failures to zero.

## Expected behavior

On the metamodel schema specifically, these directives must resolve at **runtime** (via `SchemaView.induced_slot()`), not at codegen time. On user schemas the existing "bake" behavior is correct and user-facing (see `tests/linkml/test_issues/test_issue_675.py`) — a user who writes `ifabsent: class_curie` on their own class's slot _wants_ the class's CURIE baked as the default.

## Related

- Downstream: PR #3388 (linkml-model 1.11.0rc1 bump) is blocked on this.
- Prior art: matentzn applied the same "skip runtime-computed ifabsent values" logic to the SHACL generator in `fb448d0c0` (Jan 2026).
- Proper long-term fix: #2522 — resolve these via `__post_init__` the way `default_ns` was handled in `f0d337bde`, removing the need for a metamodel special-case.

amc-corey-cox:
Closes #3413. Unblocks #3388 (linkml-model 1.11.0rc1 bump).

## Summary

- When `pythongen` generates code from the metamodel schema (id `https://w3id.org/linkml/meta`), skip baking the five runtime-computed `ifabsent` directives (`default_range`, `class_curie`, `slot_curie`, `class_uri`, `slot_uri`) as static dataclass defaults.
- Full root-cause analysis lives in #3413; short version: baking these directives against the metamodel cross-contaminates every user `SlotDefinition` / `ClassDefinition` instance with the metamodel's own values (e.g. every user slot ends up with `range="string"` and `slot_uri="linkml:slot_uri"`). `SchemaView.induced_slot()` resolves them correctly at runtime when the dataclass default is `None`.
- Scope is narrow: metamodel-only guard via `schema.id == META_URI`. User-schema `ifabsent: class_curie` etc. continues to bake (preserves `test_issue_675` behavior).
- Proper long-term fix is `__post_init__` resolution per #2522; this PR is the pragmatic unblocker.

## Implementation

- `PythonIfAbsentProcessor.map_custom_default_values` gets one extra guard that returns `(True, None)` for the five directives when the schema being generated is the metamodel. ~8 lines of logic.
- Reuses the existing `META_URI` constant from `linkml_runtime.utils.namespaces`.
- Inherited by `PydanticIfAbsentProcessor` — no separate fix needed there.

## Test plan

- [x] New parametrized test `test_metamodel_runtime_computed_ifabsent_returns_none` covering `SlotDefinition.range`, `SlotDefinition.slot_uri`, `ClassDefinition.class_uri`, `EnumDefinition.enum_uri` on a metamodel-shaped schema.
- [x] Regression test `test_runtime_computed_ifabsent_still_bakes_for_user_schemas` confirming user-schema baking is preserved.
- [x] Full local suite: `11268 passed, 0 failed, 2318 skipped`.
- [x] Existing `test_issue_675::test_ifabsent` still passes (user-facing class_curie bake).
- [x] `uv run ruff format` + `uv run ruff check` clean.
- [x] CI green on PR.
- [ ] After merge: regenerate `meta.py` in the linkml-model repo using this fix, re-bump PR #3388.


tests/linkml/test_generators/python/test_python_ifabsent_processor.py
```diff
@@ -520,6 +520,86 @@ def test_default_ns_returns_none():
...
+def test_metamodel_runtime_computed_ifabsent_returns_none(cls_name, slot_name):
+    """On the metamodel schema, runtime-computed ifabsent directives must not be
+    baked as static dataclass defaults. SchemaView.induced_slot() resolves these
+    per-instance at load time; baking would cross-contaminate every user-schema
+    slot/class with the metamodel's own URIs/range."""
+    schema_view = SchemaView(metamodel_schema)
+    processor = PythonIfAbsentProcessor(schema_view)
+    cls = schema_view.all_classes()[ClassDefinitionName(cls_name)]
+    slot = cls.attributes[slot_name]
+    assert processor.process_slot(slot, cls) is None
```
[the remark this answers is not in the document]
amc-corey-cox: Good catch — fixed in f7f5576b0. Added two attributes to the metamodel fixture (`direct_slot_uri` with `ifabsent: slot_uri`, `direct_class_uri` with `ifabsent: class_uri`) and two corresponding parametrize cases. All six directives in the guard now have coverage.

amc-corey-cox:
@matentzn The changes you see there in meta.py are in order to get the CI to pass for this PR. Obviously, we'd need to regenerate the metamodel and vendor it again to make sure it is identical. I just wanted to give you a clean CI for your review. I'm double checking right now that producing the meta.py from the metamodel is identical to this. Those fields changed are what were causing all of our errors in your other metamodel bump PR.

amc-corey-cox:
To clarify, the meta.py changed here is _not_ the vendored version but instead the snaphsot against how we generate the version so this is a necessary fix because the generation is intentionally changed.

amc-corey-cox:
@matentzn I'm going to need you or someone else to review this because I created the PR so I can't merge it until I have an approving review.

packages/linkml/src/linkml/generators/python/python_ifabsent_processor.py
```diff
@@ -4,9 +4,26 @@
...
 class PythonIfAbsentProcessor(IfAbsentProcessor):
+    # When generating meta.py from meta.yaml, these ifabsent directives must not be
+    # baked in as dataclass-level defaults: instances of SlotDefinition/ClassDefinition
+    # represent user-schema elements, so a static default would cross-contaminate every
+    # user instance with the metamodel's own values (e.g. every user slot would carry
+    # range="string" and slot_uri="linkml:slot_uri"). SchemaView.induced_slot() resolves
+    # these at runtime. For non-metamodel schemas, baking remains correct because the
+    # generated class corresponds 1:1 with a concrete user class.
+    # TODO: replace with __post_init__ resolution per https://github.com/linkml/linkml/issues/2522.
+    _METAMODEL_RUNTIME_COMPUTED_IFABSENT = (
```
matentzn, line 19: `_METAMODEL_RUNTIME_COMPUTED_IFABSENT` is defined as a class attribute on `PythonIfAbsentProcessor`, but the PR says `PydanticIfAbsentProcessor` inherits it for free. I am assuming `PydanticIfAbsentProcessor` also delegates to `map_custom_default_values`.

matentzn (approved):
One minor comment but otherwise this looks good to me.

packages/linkml/src/linkml/generators/python/python_ifabsent_processor.py
amc-corey-cox: Confirmed. `PydanticIfAbsentProcessor` only overrides `map_enum_default_value` (`pydantic_ifabsent_processor.py:10-13`); `map_custom_default_values` is inherited from `PythonIfAbsentProcessor` unchanged. pydanticgen invokes it via `ifabsent_processor.process_slot(slot, class_def)` at `pydanticgen.py:678`, which goes through the base template-method chain into our guard — so the same short-circuit fires for both generators.

amc-corey-cox:
@sierra-moxon I have verified this is what was in the vendored in file prior to our performing the new vendoring for these lines. The key is this is the post-generation file so the python generator would be normalizing and emitting the keys. Thanks for asking me to check it though!

PR #3417, opened 2026-04-18
fix(sqla): emit relationship() for non-inlined FK slots in 2.x declarative

### LINKED ISSUE #3416: gen-sqla --sqla-style 2: FK columns to string-identifier targets get Integer(), and non-inlined slots get no relationship()
Two follow-ups from #3231 that I need to fix.

When a slot's range is a class whose identifier is a string (e.g. `range: uriorcurie`, the common ontology-term case), the FK column on the source table comes out as `Integer()` even though the target PK is `Text()`.  Can be minimally reproduced with:

```yaml
classes:
  Term:
    attributes:
      uri:
        identifier: true
        range: uriorcurie
        required: true
  Event:
    attributes:
      id:
        identifier: true
        range: integer
      term:
        range: Term
```

`gen-sqla --declarative --sqla-style 2` produces:

```python
class Event(Base):
    __tablename__ = 'Event'
    id: Mapped[int] = mapped_column(Integer(), primary_key=True)
    term: Mapped[Optional[int]] = mapped_column(Integer(), ForeignKey('Term.uri'))

class Term(Base):
    __tablename__ = 'Term'
    uri: Mapped[str] = mapped_column(Text(), primary_key=True)
```

SQLite tolerates the type mismatch silently; Postgres would likely reject the DDL. The FK column's type isn't consulting the target identifier's range — it's just defaulting to `Integer`.

Second thing, probably related: when the source slot is left at `inlined: false` (default), you get the FK column but no `relationship(...)` attribute alongside it. Flipping the same slot to `inlined: true, inlined_as_list: true` makes gen-sqla emit both the FK column (now suffixed, e.g. `term_uri`) and `term: Mapped[Optional["Term"]] = relationship(foreign_keys=[term_uri])`. So ORM-side joins only work when `inlined` is true — but `inlined` is a serialization concern, the relationship attribute should be there either way.

kevinschaper:
Closes #3416.

## Summary

Two distinct bugs in `gen-sqla --sqla-style 2`:

1. **Missing `relationship()` for non-inlined FK slots** — ORM-side joins only worked when a slot was marked `inlined: true`, even though `inlined` is a serialization concern.
2. **FK column typed `Integer()` when the target's PK is `Text()`** — reproduces when the target class inherits from an abstract parent with no declared identifier. Silently accepted by SQLite, rejected by Postgres.

Both are fixed; regression tests added for each. Bug 1's fix is a **breaking change for 2.x ORM users** — see the breaking-change note below.

## Bug 1: missing `relationship()` attribute

**Before:**
```python
class Event(Base):
    term: Mapped[str | None] = mapped_column(Text(), ForeignKey("Term.uri"))
    # no relationship() — ORM joins silently unavailable
```

**After:**
```python
class Event(Base):
    term_uri: Mapped[str | None] = mapped_column(Text(), ForeignKey("Term.uri"))
    term: Mapped[Term | None] = relationship(foreign_keys=[term_uri])
```

When `template == DECLARATIVE_2X` and no FK policy is passed, default the relational transformer to `INJECT_FK_FOR_ALL_REFS`. This triggers the same rename + `original_slot` path already used for inlined slots, so the existing template branch emits the relationship attribute.

### ⚠️ Breaking change for 2.x ORM users with non-inlined FK slots

For a non-inlined slot `term: range Term` (PK `uri`):

- **Before:** `obj.term` returned the FK string value (`"GO:0001234"`).
- **After:** `obj.term` returns the related `Term` object; the FK string lives on `obj.term_uri`.

Constructors and queries that previously used the bare slot name as a string need to switch to the renamed column:

```python
# Before
Concept(term="GO:0001234")
session.query(Concept).filter_by(term="GO:0001234")

# After
Concept(term_uri="GO:0001234")            # raw FK
Concept(term=existing_term_obj)           # via relationship
session.query(Concept).filter_by(term_uri="GO:0001234")
```

This brings 2.x non-inlined FK slots in line with the existing behaviour for *inlined* FK slots (which were already renamed `<slot>_<pk>`), so the renaming convention is now uniform across all class-range slots in the 2.x ORM.

### Interaction with `gen-sqltables` (opt-in alignment)

The 2.x ORM rename means the SQLA model's FK column names (`term_uri`) no longer match the default `gen-sqltables` output (`term`) for non-inlined slots. Two ways to use them together:

1. **Default — ORM owns the schema.** Use `mod.Base.metadata.create_all(engine)` (or Alembic from the ORM) as the DDL source. This is what the new and updated integration tests do, and it requires no flags.
2. **Opt in to matching DDL** with the new `--rename-foreign-keys` flag on `gen-sqltables` (`SQLTableGenerator(..., rename_foreign_keys=True)`). With the flag on, `gen-sqltables` emits `term_uri` instead of `term` for non-inlined class-range slots, so the DDL pairs cleanly with the 2.x ORM. **Default off** so existing `gen-sqltables` consumers (raw SQL, Alembic migrations, BI dashboards) are not silently broken.

### Reproducing pattern

```yaml
classes:
  OntologyEntity:
    abstract: true          # no declared identifier
  Term:
    is_a: OntologyEntity
    attributes:
      term_uri:
        identifier: true
        range: uriorcurie   # string-typed PK
  Event:
    attributes:
      term:
        range: Term
```

**Before the fix:**
```python
term_term_uri: Mapped[int | None] = mapped_column(Integer(), ForeignKey("Term.term_uri"))
#               ^^^                                 ^^^^^^^^^
# FK is typed as int, but Term.term_uri is Text() — Postgres rejects, SQLite tolerates
```

### Root cause

`SQLTableGenerator.get_sql_range` consults `SchemaView.get_identifier_slot(target_class)` to pick the FK column's type. That method walks `class_induced_slots`, which combines the class's own attributes with every attribute inherited through `is_a`.

The relational transformer at `relmodel_transformer.py:224-226` auto-injects an integer `id` PK onto any class lacking a direct identifier — **including abstract parents**. After injection, `OntologyEntity.attributes` contains a synthetic `id: integer, identifier=true`. Every subclass then inherits it:

```
ExposureRoute direct attrs:   term_uri (identifier), term_label
ExposureRoute induced slots:  term_uri (identifier), term_label, id (identifier)  ← two identifiers!
SchemaView.get_identifier_slot("ExposureRoute") → id (range=integer)               ← wrong one wins
```

So the FK column gets typed against `id` (integer) even though the actual SQL table only has `term_uri` as its PK.

The `foreign_key` annotation itself is correct — the transformer writes it via `get_direct_identifier_attribute` (`relmodel_transformer.py:367`), which ignores inheritance and picks `term_uri`. Only `get_sql_range` used the inheritance-traversing lookup, creating the mismatch between *what the FK points at* and *how the FK column is typed*.

### Fix

Reuse `RelationalModelTransformer.get_direct_identifier_attribute` (the same lookup the transformer already uses to write the `foreign_key` annotation), and only fall back to the inheritance-walking `get_identifier_slot` when the class legitimately inherits its PK:

```python
if slot_range in sv.all_classes():
    fk = RelationalModelTransformer.get_direct_identifier_attribute(sv, slot_range)
    if fk is None:
        fk = sv.get_identifier_slot(slot_range)  # inheritance fallback
    if fk:
        return self.get_sql_range(fk, sv.schema)
    return Text()
```

Classes that inherit their identifier from a non-abstract parent (e.g. `NamedThing { id: identifier }` + `Person is_a: NamedThing`) still resolve correctly through the fallback.

## Test plan

- [x] `tests/linkml/test_generators/test_sqlalchemygen.py` — all passing
- [x] `tests/linkml/test_generators/test_sqltablegen.py` — all passing
- [x] New regression tests:
  - `test_sqla_2x_fk_column_type_matches_string_identifier_target` — bug 2 simple case
  - `test_sqla_2x_fk_type_when_target_inherits_from_identifier_less_abstract` — the pattern that actually reproduces bug 2
  - `test_sqla_2x_emits_relationship_for_non_inlined_fk` — bug 1
  - `test_sqla_2x_non_inlined_fk_orm_join_works` — end-to-end ORM join via `create_all` (default path)
  - `test_sqla_2x_non_inlined_fk_orm_join_works_with_aligned_ddl` — end-to-end ORM join via `gen-sqltables` DDL with `rename_foreign_keys=True` (cross-generator pairing)
  - `test_rename_foreign_keys_flag` — unit test on the new `SQLTableGenerator` flag
- [x] Manually verified against the real schema in the issue (zebrafish-toxicology-atlas-schema): all four affected FK columns (`route_term_uri`, `exposure_type_term_uri`, `phenotype_term_id_term_uri`, `chemical_id_uri`) now correctly type as `Text()` / `Mapped[str]`.


packages/linkml/src/linkml/generators/sqlalchemygen.py
```diff
@@ -69,6 +69,11 @@ def generate_sqla(
         sqltr = RelationalModelTransformer(self.schemaview)
         if foreign_key_policy:
             sqltr.foreign_key_policy = foreign_key_policy
+        elif template == TemplateEnum.DECLARATIVE_2X:
+            # In 2.x declarative mode we want a relationship() attribute alongside
+            # every FK column, which requires the transformer to rename the source
+            # slot (adding `original_slot`) even when the slot is not inlined.
+            sqltr.foreign_key_policy = ForeignKeyPolicy.INJECT_FK_FOR_ALL_REFS
```
amc-corey-cox, line 76: Switching this policy makes the 2.x ORM's column names diverge from what `gen-sqltables` emits for the same schema: ORM expects `term_uri`, DDL produces `term`. The test rewrite to `Base.metadata.create_all(engine)` is that divergence surfacing. Could you decide and document one of:

1. `gen-sqltables` should match (use `INJECT_FK_FOR_ALL_REFS` too), or
2. 2.x is explicitly an ORM-owns-schema path and mixing with `gen-sqltables` output is unsupported.

Either is defensible, but the answer drives two related things worth calling out in the PR / a code comment:

(a) the design rationale — 2.x is choosing "slot name = ORM relationship, not FK value," which is the existing behavior for inlined slots now extended to all FK slots; and

(b) this is a breaking API change for existing 2.x users with non-inlined FK slots — `obj.term` was the string, now it's the related object, and constructors/queries need `term_uri`. Worth a changelog note.

packages/linkml/src/linkml/generators/sqltablegen.py
```diff
@@ -363,8 +363,22 @@ def get_sql_range(self, slot: SlotDefinition, schema: SchemaDefinition = None):
...
+            # Prefer the target's *direct* identifier: an abstract parent without
+            # an identifier gets an auto-injected integer PK from the relational
+            # transformer, and that injection leaks into subclasses via
+            # class_induced_slots — masking a real identifier declared on the
+            # subclass. Fall back to the inherited identifier for the common
+            # case where the subclass legitimately inherits its PK.
+            target_class = sv.get_class(slot_range)
+            fk = None
+            if target_class is not None:
+                for a in target_class.attributes.values():
```
amc-corey-cox: nit: this reimplements `RelationalModelTransformer.get_direct_identifier_attribute`, which is already imported in this file. Could collapse to:

```python
fk = RelationalModelTransformer.get_direct_identifier_attribute(sv, slot_range)
if fk is None:
    fk = sv.get_identifier_slot(slot_range)
```

amc-corey-cox (changes requested):
Both fixes are correct and well-tested — diagnosis on bug 2 is especially thorough. One question before merging: the policy switch makes the 2.x ORM diverge from gen sqltables (different column names) and is a breaking API change for existing 2.x users with non-inlined FK slots. Could you pick a direction on the gen-sqltables interop question and capture the rationale + user-visible change somewhere (PR description, code comment, changelog)? Inline has the detail. Small nit on de-duplication alongside.

kevinschaper:
Thanks for the careful read — I went with **Option 2 (ORM owns the schema by default) plus an opt-in alignment flag**, and addressed the de-dup nit.

**Bug 2 nit** (3244205498): collapsed to `RelationalModelTransformer.get_direct_identifier_attribute(sv, slot_range)` with the inheritance fallback, as suggested.

**Interop question** (3244205493): default behaviour of `gen-sqltables` is unchanged — existing consumers (raw SQL, Alembic migrations, BI dashboards) are not silently broken. Users who run *both* generators can now opt in to matching column names with the new `--rename-foreign-keys` / `SQLTableGenerator(rename_foreign_keys=True)` flag, which switches `gen-sqltables` to `INJECT_FK_FOR_ALL_REFS`. New tests:

- `test_rename_foreign_keys_flag` — DDL-only assertion that the flag changes column naming as advertised.
- `test_sqla_2x_non_inlined_fk_orm_join_works_with_aligned_ddl` — cross-generator integration test pairing `gen-sqltables` DDL (flag on) with the 2.x ORM, proving they're actually compatible.

**Rationale & breaking-change note** are now in the PR description under "⚠️ Breaking change for 2.x ORM users with non-inlined FK slots" and "Interaction with `gen-sqltables` (opt-in alignment)" — including the `obj.term` → `obj.term_uri` semantics flip and migration examples.

Reasoning behind picking Option 2 over flipping the default in both: `gen-sqltables` predates the 2.x ORM by years and has a wider user base, so making the rename opt-in keeps the blast radius confined to people who consciously adopt the 2.x ORM. We can revisit flipping the default if/when 2.x becomes the recommended path.

amc-corey-cox (approved):
This looks good to me now.

PR #3425, opened 2026-04-20
fix(linkml): relmodel performance improvements #3353

### LINKED ISSUE #3353: Relmodel_transformer invalidates lru_cache destroying performance

noelmcloughlin:
## Summary
Replaces #3357 

Fixes #3353 which explains the performance issue being addressed, add debug logs, testing.

## Reviewer 
It may be easier to review commit by commit.

## How was this tested?
Local testing

## Checklist


amc-corey-cox (approved):
This looks good to me.

PR #3430, opened 2026-04-21
fix(validation_context): optimize logic around json schema #3429

### LINKED ISSUE #3429: Validation_context.py produces full schema per every class (slow), instead of per schema
**Describe the bug**
JsonSchemaGenerator` regenerates full schema per `target_class`which causes big O(n) slowdown on large schemas. This should be optimized so only one operation takes the performance hit.

**Cause**
The `linkml-run-examples` uses `Validator` class, which creates a `ValidationContext` per distinct `target_class`. 
https://github.com/linkml/linkml/blob/6439c90e13b83e3401222d3c3b37d5715b08cb81/packages/linkml/src/linkml/validator/validator.py#L120-L122

Each `ValidationContext.json_schema_validator()` call runs `JsonSchemaGenerator` to produce the entire JSON Schema for all classes from scratch - large schemas can have 100's of classes. 
https://github.com/linkml/linkml/blob/6439c90e13b83e3401222d3c3b37d5715b08cb81/packages/linkml/src/linkml/validator/validation_context.py#L45

There is a `@lru_cache` cache located at `ValidationContext() -> jjson_schema_validator` which caches per instance, but never hits across target classes - but even with caching the solution is suboptimal.
https://github.com/linkml/linkml/blob/6439c90e13b83e3401222d3c3b37d5715b08cb81/packages/linkml/src/linkml/validator/validation_context.py#L32

`1.10.0`

noelmcloughlin:
## Summary

Fixes #3429 

Generate JSON Schema **once** for a target class, cache the `$defs` (instead of per every class).
Then re-root per class by adjusting only the top-level `$ref` (for all other class)

## How was this tested?

Locally against big schema (700+ classes).

## Areas of uncertainty

No, I think its straightforward optimization at json schema level.

## Checklist

## AI Assistance

Assisted by LLM, diagnosis, writing unit tests


packages/linkml/src/linkml/validator/validation_context.py
```diff
@@ -42,14 +50,33 @@ def json_schema_validator(
...
+                # Uses a minimal wrapper with a $ref pointing to cached $defs.
+                # Each $defs entry has their own 'properties', 'required', etc,
+                # so delegating via `$ref` preserves the validation behaviour.
+                cached = _json_schema_cache[cache_key]
+                json_schema = JsonSchema(
+                    {
+                        "$schema": cached.get("$schema"),
+                        "$id": cached.get("$id"),
+                        "$defs": cached["$defs"],
+                        "$ref": f"#/$defs/{self._target_class}",
```
amc-corey-cox: The cache-hit wrapper delegates to `$defs.X` via `$ref`, but every class in `$defs` is generated with `additionalProperties: false`, while the fresh root uses `additionalProperties = not_closed`. With `closed=False` (the default for `JsonschemaValidationPlugin`), this produces a real divergence — an instance with extra root-level properties is *accepted* by a freshly-generated validator but *rejected* by the cache-hit one.

Repro using the two-class fixture from the new tests:

```
Instance: {"name": "Rex", "breed": "Lab", "extra_prop": "hello"}
Fresh validator:    0 errors  (root has additionalProperties: true)
Cache-hit wrapper:  1 error   ("Additional properties are not allowed")
```

Would it work to cache only `$defs` (and other class-independent pieces) and re-run the cheap "hoist top-class properties into the root + set `additionalProperties = not_closed`" step per `target_class`? I haven't gone deep enough into the generator to know whether there are other root-level fields (e.g. rule-derived `allOf`/`if`/`then`) that would also need preserving — happy to hear what makes sense here.

A regression test that builds two validators for the same class — one with a cold cache, one with a warm cache populated by a *different* class — and asserts identical errors on a shared instance set would catch this and any sibling cases.

packages/linkml/src/linkml/validator/validation_context.py
```diff
@@ -42,14 +50,33 @@ def json_schema_validator(
...
             not_closed = not closed
-            jsonschema_gen = JsonSchemaGenerator(
-                schema=self._schema,
-                mergeimports=True,
-                top_class=self._target_class,
-                not_closed=not_closed,
-                include_range_class_descendants=include_range_class_descendants,
-            )
-            json_schema = jsonschema_gen.generate()
+            cache_key = (id(self._schema), closed, include_range_class_descendants)
```
amc-corey-cox: Two related concerns with `id(self._schema)` as a module-level cache key:

1. The cache holds no reference to the schema. Once the originating `ValidationContext` is GC'd, CPython is free to reuse that `id()` for an unrelated `SchemaDefinition` later → silent false-positive hit that returns the wrong cached schema.
2. The cache is unbounded for the process lifetime, so long-running servers/notebooks that load many schemas leak memory.

A stable key like `(schema.id, schema.version)` (or a content hash) plus a bounded eviction policy would address both.

tests/linkml/test_validator/test_validation_context.py
```diff
@@ -0,0 +1,213 @@
...
+
+    import linkml.validator.validation_context as vc_module
+    from linkml.generators.jsonschemagen import JsonSchemaGenerator as OriginalGenerator
+
+    class CapturingGenerator(OriginalGenerator):
+        def __init__(self, *args, **kwargs):
+            calls.append(kwargs.copy())
+            super().__init__(*args, **kwargs)
+
+    monkeypatch.setattr(vc_module, "JsonSchemaGenerator", CapturingGenerator)
```
amc-corey-cox: Two test-quality items in this file:

- `monkeypatch` + `CapturingGenerator` is a mock that asserts on call kwargs; per the project convention of avoiding mocks, a behavioral test (e.g. assert the resulting validator rejects extras when `closed=True` and accepts them when `closed=False`) would be more durable and would also catch the `additionalProperties` regression noted on the source file.
- Several tests `_json_schema_cache.pop(...)` for setup but don't restore afterward, leaking module-level state across tests. An autouse fixture that clears the cache around each test would isolate them and let the tests stop reaching into private state.

amc-corey-cox (changes requested):
This is a great PR with a nice performance improvement. Thank you for bringing this to us. I see some things of concern. This is a little out of my depth so if I've misunderstood something here please feel free to push back.

packages/linkml/src/linkml/validator/validation_context.py
noelmcloughlin: Thanks @amc-corey-cox for the review and finding these issues.Definitely I will add the tests you mention since your simple test could have been added in first place to catch "additionalProperties" False vs not_closed. 

> other root-level fields (e.g. rule-derived allOf/if/then) that would also need preserving — happy to hear what makes sense here.

The root-level rules fields look good to me in JSON Generator `handle_class():  ``if / then / else` (lines 354-367) and `AllOf` (lines 386)  are added to the class_subschema. Then line 393 stores the entire dict (including if/then/AllOf/else) into $defs["Dog"].

noelmcloughlin: You are correct - some other inlined metadata are dropped like our validation bug (additional_properties). Three keys differ: `metamodel_version` and `version` are missing; and `title` shows the class's canonical name instead of the schema's name:

```bash
v = ctx.json_schema_validator(closed=False, ...)
print(v.schema["metamodel_version"])     # KeyError on cache-hit; "1.7.0" on fresh
print(v.schema["title"])                 # "Dog" on cache-hit; "two_class_test" on fresh
```

This is not a validator issue - so I am thinking about separate Issue/PR - but since the fix is small python list comprehension step I will include in this PR with comment - can be removed if you like.

noelmcloughlin: Issues and PRs have their own overhead - the code is here, lets try to fix here :-)

packages/linkml/src/linkml/validator/validation_context.py
noelmcloughlin: Thank you - I have dug into this and refactored the LRU cache lifecycle.  The key is stable now, four  tuples just to be sure.  A leak was uncovered (class with slot "required=true") so `_ROOT_METADATA_KEYS=` was introduced, and extended unit tests to try to harden the PR.

noelmcloughlin: **NEW ISSUE:** 

Extended cache key did not work --> one `test_compliance.py` failure ....

- `test_include_multiple_overlays` depends on the new `--include CLI` flag that merges overlay schemas.
- The merged SchemaDefinition inherits the base's id/name/version - so `(id, name, version, range_descendants)` cache key started aliasing the overlay-merged schema onto what an earlier test had cached the bare base schema.

**REVISED FIX:**  

- Use parallel `(id(schema), include_range_class_descendants)` cache key PLUS a `_schema_pins: dict[tuple, SchemaDefinition]` to hold a strong reference to each cached schema.

- While entry exists, CPython cannot recycle that id() onto an unrelated object. On LRU eviction, both the cache entry and the pin are dropped, so memory stays bounded.

amc-corey-cox (approved):
Thanks for the fixes @noelmcloughlin. This looks good to me now!

Issue #18, opened 2019-11-05
Improvements to markdown generator


deepakunni3:
- Differentiating between Entity, Association, Slot, Relation and Property during export from biolinkml. This could be handled by creating folders for each, and exporting the markdown into these folders.
- CURIE mappings from YAML should be in markdown export (currently missing).
- The title in each markdown should be camelcase instead of sentence case. For example, it should be `Class: NamedThing` instead of `Class: named thing`.

cmungall:
@joeflack4 can you fold these into the PR you are working on?

joeflack4:
@cmungall I added these to [markdowngen improvements master issue](https://github.com/linkml/linkml/issues/245). I also enumerated the bullet points there. I'm wondering if we should close this issue now that these have been added to there?

Issue #14, opened 2021-03-25
default_namespace must be specified


hsolbrig:
Currently, if you omit the default namespace, you don't know that it is needed until you get a generator specific message such as:

```python
    context = ContextGenerator(os.path.join(inputdir, 'uriandcurie.yaml')).serialize()
  File "/Users/solbrig/git/biolink/biolinkml/biolinkml/utils/generator.py", line 77, in serialize
    self.visit_schema(**kwargs)
  File "/Users/solbrig/git/biolink/biolinkml/biolinkml/generators/jsonldcontextgen.py", line 48, in visit_schema
    default_uri = self.namespaces[self.default_ns]
KeyError: None
```
Default namespace should be tested for (or a good default should be found) earlier in the process

hsolbrig:
Unable to reproduce.

turbomam:
Is "default_namespace" the same as `default_prefix`?

PR #3446, opened 2026-04-24
feat(generators): add --normalize-prefixes flag for well-known prefix names

jdsika:
> [!WARNING]
> **Blocked: depends on an unreleased `prefixmaps`, and this breaks the Docker build.**
>
> `packages/linkml/pyproject.toml` currently carries a git pin:
>
> ```
> "prefixmaps @ git+https://github.com/linkml/prefixmaps@7543515",
> ```
>
> plus `[tool.hatch.metadata] allow-direct-references = true` to permit it.
> The fix it depends on (linkml/prefixmaps#82 — the `gx` prefix mapping to a
> dead Graphite/Synaptica namespace instead of Gaia-X) is merged in the
> prefixmaps `main` branch but **has not been released**: the latest version on
> PyPI is still `0.2.6` (2024-10-17).
>
> This is not just a policy problem. The git pin is baked into the built wheel's
> metadata, so the **Docker image build fails**: `pip install` has to fetch the
> dependency from GitHub, and the base image has no `git`.
>
> ```
> ERROR: Cannot find command 'git' - do you have 'git' installed and in your PATH?
> ERROR: process "/bin/sh -c pip install /tmp/linkml-whl/*.whl" did not complete
>        successfully: exit code: 1
> ```
>
> That is the one failing check on this PR; every other job is green.
>
> **Do not merge until `prefixmaps` cuts a release.** Then the pin becomes
> `prefixmaps >= 0.2.8`, `allow-direct-references` is removed, and `uv.lock` is
> regenerated — both spots are marked `TODO(prefixmaps-0.2.8)`. The Docker
> failure resolves itself at that point, since the dependency comes from PyPI.

---

## Summary  Add an opt-in `--normalize-prefixes` flag to all generators that normalises non-standard prefix aliases to well-known names from a **static, version-independent prefix map** (derived from rdflib 7.x defaults, cross-checked against [prefix.cc](https://prefix.cc) consensus).  Refs: integration branch of the ENVITED-X pipeline (#14, superseded by #18). No dedicated tracker issue — split from a larger consolidated change per the contributor guidelines.  ## Motivation  Schemas often declare custom prefix aliases for well-known namespaces (e.g. `sdo` for `https://schema.org/`, `dce` for `http://purl.org/dc/elements/1.1/`). When generated artifacts are consumed by tools that expect conventional prefix names, this causes friction. This flag allows opt-in normalization without changing default behavior.  ## Changes  ### `generator.py` — Static prefix map and shared helper  - **`_WELL_KNOWN_PREFIX_MAP`**: Frozen `dict` literal of ~30 well-known namespace-to-prefix mappings. Eliminates rdflib version dependency (defaults changed between 6.x and 7.x). Includes both `http://schema.org/` and `https://schema.org/` → `schema` (linkml-runtime uses HTTP, W3C prefers HTTPS). - **`well_known_prefix_map()`**: Returns a copy of the static map (no longer creates `Graph()` at runtime). - **`normalize_graph_prefixes(graph, schema_prefixes)`**: Shared helper that rebinds non-standard aliases in an rdflib Graph to their standard names. Uses `graph.bind(..., override=True, replace=True)` for clean rebinding. - **`normalize_prefixes`** field on `Generator` base class + CLI flag.  ### `owlgen.py` — Explicit post-bind normalisation  - Binds schema prefixes normally (`override=True`, the default), then calls `normalize_graph_prefixes()` when flag is set. - **Replaces** previous `override=False` approach which made **all** 29 rdflib defaults sticky (not just well-known ones).  ### `shaclgen.py` — Same explicit approach  - Same pattern as OWL: bind normally → post-process with `normalize_graph_prefixes()`.  ### `jsonldcontextgen.py` — Three-case remap logic  - Initialises `_prefix_remap` in `__post_init__()` (replaces defensive `getattr`). - Handles three cases during remap:   1. Standard prefix not yet bound → rebind from old to new.   2. Standard prefix bound to different URI (stale binding) → remove stale, then rebind.   3. Standard prefix bound to same URI (duplicate) → just drop the alias. - Applies remap to both prefix declarations and CURIE generation in `_build_element_id()`.  ## How was this tested?  New `test_normalize_prefixes.py` with 18 tests across 5 test classes:  | Class | Tests | What it verifies | |-------|:-----:|------------------| | `TestOwlNormalizePrefixes` | 5 | sdo→schema, flag-off, dce→dc (graph bindings), custom prefix, http variant | | `TestShaclNormalizePrefixes` | 5 | Same coverage for SHACL generator | | `TestContextNormalizePrefixes` | 1 | http://schema.org/ edge case | | `TestWellKnownPrefixMap` | 6 | Static map integrity, copy safety, rdflib superset check | | `TestCrossGeneratorConsistency` | 1 | All three generators agree on sdo→schema |  Plus 3 existing context-gen tests retained unchanged.  ## Backward Compatibility  The flag defaults to **off** — existing behavior is fully preserved. No snapshot updates required.  ## Areas of uncertainty  - The well-known prefix map is static; it can drift from community usage (prefix.cc) and needs occasional curation. - Interaction with `--deterministic` prefix filtering is covered by tests, but combined use on very large graphs has only been exercised on the Gaia-X Trust Framework artifacts.  ## Checklist  - [x] My code follows the [contributor guidelines](https://linkml.io/linkml/maintainers/contributing.html) - [x] I have added tests that prove my fix/feature works - [x] Existing tests pass locally with my changes  ## AI Assistance  If you used AI tools while preparing this PR, you are still the author and responsible for understanding, verifying, and defending your submission. Please engage with reviewers personally rather than through your agent during feedback and revisions. See our [AI Covenant](https://github.com/ASCS-eV/linkml/blob/main/AI_COVENANT.md) for details.

### DIFF
--- a/packages/linkml/src/linkml/utils/generator.py
+++ b/packages/linkml/src/linkml/utils/generator.py
@@ -20,11 +20,12 @@ import logging
 import os
 import re
 import sys
+import types
 from collections.abc import Callable, Mapping
 from dataclasses import dataclass, field
 from functools import lru_cache
 from pathlib import Path
-from typing import ClassVar, TextIO, Union, cast
+from typing import TYPE_CHECKING, ClassVar, TextIO, Union, cast
 
 import click
 from click import Argument, Command, Option
@@ -58,6 +59,9 @@ from linkml_runtime.linkml_model.meta import (
 from linkml_runtime.utils.formatutils import camelcase, underscore
 from linkml_runtime.utils.namespaces import Namespaces
 
+if TYPE_CHECKING:
+    from rdflib import Graph
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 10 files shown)


packages/linkml/src/linkml/utils/generator.py
```diff
@@ -78,6 +82,138 @@ def _resolved_metamodel(mergeimports):
...
+    Both ``http`` and ``https`` variants of schema.org are included because
+    the linkml-runtime historically binds ``schema: http://schema.org/``
+    while rdflib (and the W3C) prefer ``https://schema.org/``.
+    """
+    return dict(_WELL_KNOWN_PREFIX_MAP)
+
+
+# Frozen, version-independent map: namespace URI → canonical prefix name.
+# Source: rdflib 7.x defaults, cross-checked against https://prefix.cc
+_WELL_KNOWN_PREFIX_MAP: types.MappingProxyType[str, str] = types.MappingProxyType(
```
matentzn: @kevinschaper when you review this please be super diligent on this part whatever you AI review assistant says. IMO this needs to go through https://github.com/linkml/prefixmaps.

jdsika: ha, I did not know about this repo. I will have a look as well.

jdsika: https://github.com/linkml/prefixmaps/pull/81

jdsika:
## Status: Draft — awaiting upstream dependency

This PR is moved back to **draft** pending the merge of an upstream fix:

- **linkml/prefixmaps#81** — Adds lowercase W3C/OGC prefixes (odrl, 	ime, dcmitype, wgs, xml) to the linked_data context

Once that lands and a new prefixmaps release is cut, we can evaluate whether to:
1. Replace the hardcoded `_WELL_KNOWN_PREFIX_MAP` with a runtime call to `load_context("merged").as_inverted_dict()`
2. Or keep the static map (rdflib 7.x defaults) but document prefixmaps as the canonical upstream source

Additionally:
- **linkml/prefixmaps#82** tracks the `gx`/`gax` Gaia-X namespace conflict (pending prefix.cc vote)
- A new `gx: https://w3id.org/gaia-x/development#` mapping has been proposed at prefix.cc

The implementation and tests in this PR are **complete and passing** — no code changes needed beyond the upstream resolution.

jdsika:
@kevinschaper do you think we are ready to release `prefixmaps  == 0.2.8` 

Is there anything still needed for the prefixmaps repo?

PR #3447, opened 2026-04-25
feat(contextgen): add @type: @vocab coercion for eligible enums

### LINKED ISSUE #2497: Enums not serialised to JSON-LD as anticipated
**Describe the bug**

Given the following linkml schema:

```yaml
id: http://example.com/broken-enum-jsonld
name: Broken-Enum-JSON-LD

prefixes:
  stuff: http://example.com/stuff/

enums:
  SomeEnum:
    description: Some Enum
    permissible_values:
      some_value:
        meaning: stuff:SomeValue
      some_other_value:
        meaning: stuff:SomeOtherValue

classes:
  SomeClass:
    tree_root: True
    class_uri: stuff:SomeClass
    description: Some Class
    attributes:
      some_slot:
        multivalued: True
        slot_uri: stuff:someSlot
        description: Some Slot
        range: SomeEnum
```

When I convert the following JSON representation (which is validated as correct by linkml) into JSON-LD using linkml,

```json
{
    "some_slot": [
        "some_value",
        "some_other_value"
    ]
}
```

I end up with the following JSON-LD representation which **does not correctly represent the meaning `SomeEnum` as I would expect it to in RDF**:

```json
{
  "some_slot": [
    "some_value",
    "some_other_value"
  ],
  "@type": "SomeClass",
  "@context": {
    "Broken-Enum-JSON-LD": "http://example.com/broken-enum-jsonld/",
    "skos": "http://www.w3.org/2004/02/skos/core#",
    "stuff": "http://example.com/stuff/",
    "@vocab": "http://example.com/broken-enum-jsonld/",
    "some_slot": {
      "@context": {
        "text": "skos:notation",
        "description": "skos:prefLabel",
        "meaning": "@id"
      },
      "@id": "stuff:someSlot"
    },
    "SomeClass": {
      "@id": "stuff:SomeClass"
    }
  }
}
``` 

Note that the URIs representing the `meaning` of the enum are not present anywhere in the document, that `some_slot` holds only the text values, but that the `@context` suggests that it should hold an object with `text`/`description`/`meaning` values.

Further, when I convert the JSON example to rdf/ttl I get the following output which aligns better with what I would expect in an RDF representation:

```ttl
@prefix stuff: <http://example.com/stuff/> .

[] a stuff:SomeClass ;
    stuff:someSlot stuff:SomeOtherValue,
        stuff:SomeValue .
```

linkml, version 1.8.6

jdsika:

## Summary

Adds JSON-LD 1.1 `@type: @vocab` coercion for enum-ranged slots whose permissible values are eligible (all have meanings, single namespace, text matches local name).

**Closes linkml/linkml#2497**

## Problem

LinkML's `visit_enum` is a no-op ÔÇö enum permissible values are not emitted in the generated JSON-LD context. Slots with enum ranges only get a structural `ENUM_CONTEXT` mapping (`{text: skos:notation, description: skos:prefLabel, meaning: @id}`), which requires structured object values.

In practice, JSON data files use **bare string** enum values (e.g. `"DrivableAreaType": "RoadTypeMotorway"`). Without context support, these remain plain strings and don't expand to IRIs.

## Approach

Uses JSON-LD 1.1 ┬º4.2.3 (type coercion via `@vocab`) combined with ┬º4.1.8 (scoped contexts):

```json
"DrivableAreaType": {
  "@id": "openlabel_v2:DrivableAreaType",
  "@type": "@vocab",
  "@context": {
    "@vocab": "https://w3id.org/ascs-ev/envited-x/openlabel/v2/",
    "text": "skos:notation",
    "description": "skos:prefLabel",
    "meaning": "@id"
  }
}
```

This supports **both** value forms simultaneously:

- **Bare string**: `"RoadTypeMotorway"` ÔåÆ expands to IRI via scoped `@vocab`
- **Structured object**: `{text, description, meaning}` ÔåÆ SKOS mappings still work

### Eligibility Criteria

An enum qualifies for `@type: @vocab` only when ALL conditions are met:

1. Every permissible value has a `meaning` IRI
2. All meaning IRIs share the same namespace prefix
3. Each value's `text` matches the local part of its `meaning` CURIE

Ineligible enums fall back to the existing `ENUM_CONTEXT` behavior unchanged.

### Collision handling

When many enum-typed slots share identical scoped `@context` blocks, the `pyld` JSON-LD processor has a scoped context cache pollution bug that causes `@type: @vocab` coercion to silently fail for all but the first term. We root-caused this to a stale `_uuid` cache key in `pyld`'s `_process_context()` and submitted a minimal fix upstream:

- **digitalbazaar/pyld#249**: fix: prevent scoped context cache pollution during context processing

Note that `rdflib`'s JSON-LD implementation is **not** affected ÔÇö the generated contexts are spec-compliant and work correctly with rdflib. The pyld bug only impacts consumers using the pyld library directly.

## Changes

| File | Change |
|------|--------|
| `packages/linkml/src/linkml/generators/jsonldcontextgen.py` | Added `_vocab_eligible_enum()` method (3-check eligibility) + modified `visit_slot` enum branch to emit `@type: @vocab` with scoped `@context` for eligible enums |
| `tests/linkml/test_generators/test_jsonldcontextgen.py` | 7 new tests: eligible enum (positive), 3 ineligible variants (text mismatch, no meaning, mixed namespaces), rdflib expansion verification (bare string + structured), kitchen_sink fallback |

## Testing

- **35 tests pass** in `test_jsonldcontextgen.py` (+ 3 xfailed for unrelated HTTPS schema.org issues)
- All existing context generator tests continue to pass ÔÇö ineligible enums fall back to `ENUM_CONTEXT` unchanged
- End-to-end rdflib expansion tests verify correct IRI expansion for both bare strings and structured objects
- Kitchen sink schema correctly falls back (`EmploymentEventType` has text Ôëá local: `HIRE` ÔåÆ `bizcodes:001`)

## Backward Compatibility

- Enums that don't meet the eligibility criteria produce **identical** output to before
- Eligible enums get strictly **more** information (`@type: @vocab` + scoped `@vocab`) ÔÇö the existing SKOS mappings are preserved in the scoped context
- No CLI flag required ÔÇö the feature activates automatically for qualifying enums

## References

- [JSON-LD 1.1 ┬º4.2.3: Type Coercion](https://www.w3.org/TR/json-ld11/#type-coercion) (`@type: @vocab`)
- [JSON-LD 1.1 ┬º4.1.8: Scoped Contexts](https://www.w3.org/TR/json-ld11/#scoped-contexts)
- linkml/linkml#2497: Enums not serialised to JSON-LD as anticipated
- digitalbazaar/pyld#249: pyld scoped context cache pollution fix (discovered while testing this feature)

### DIFF
--- a/packages/linkml/src/linkml/generators/jsonldcontextgen.py
+++ b/packages/linkml/src/linkml/generators/jsonldcontextgen.py
@@ -16,7 +16,7 @@ from rdflib import SKOS, XSD, Namespace
 from linkml._version import __version__
 from linkml.utils.deprecation import deprecated_fields
 from linkml.utils.generator import Generator, shared_arguments
-from linkml_runtime.linkml_model.meta import ClassDefinition, SlotDefinition
+from linkml_runtime.linkml_model.meta import ClassDefinition, EnumDefinition, SlotDefinition
 from linkml_runtime.linkml_model.types import SHEX
 from linkml_runtime.utils.formatutils import camelcase, underscore
 from linkml_runtime.utils.schemaview import SchemaView
@@ -300,6 +300,57 @@ class ContextGenerator(Generator):
             return True, next(iter(coercions))
         return False, None
 
+    def _vocab_eligible_enum(self, enum: EnumDefinition) -> tuple[bool, str | None]:
+        """Check if an enum qualifies for ``@type: @vocab`` context generation.
+
+        An enum is eligible when:
+
+        1. Every permissible value defines a ``meaning`` IRI.
+        2. All meaning IRIs share a single namespace prefix.
+        3. Each value's ``text`` equals the local part of its ``meaning`` CURIE.
+
+        When eligible, the generated context can use ``@type: @vocab`` with a
+        scoped ``@vocab`` set to the common namespace.  This lets bare string
+        values (e.g. ``"RoadTypeMotorway"``) expand to full IRIs — a JSON-LD
+        1.1 §4.2.3 / §4.1.8 compliant approach.
+
+        Returns ``(True, namespace_str)`` or ``(False, None)``.
+        """
+        if not enum.permissible_values:
+            return False, None
+
+        prefixes: set[str] = set()
+        for pv_text, pv in enum.permissible_values.items():
+            if not pv or not pv.meaning:
+                return False, None
+
+            meaning = str(pv.meaning)
+            if ":" not in meaning:
+                return False, None
+
+            prefix, local = meaning.split(":", 1)
+            prefixes.add(prefix)
+
+            if local != pv_text:
+                self.logger.debug(
+                    "Enum %s value '%s' text does not match meaning local name '%s'; falling back to ENUM_CONTEXT.",
+                    enum.name,
+                    pv_text,
+                    local,
+                )
+                return False, None
+
+        if len(prefixes) != 1:
+            return False, None
+
+        # Resolve the single prefix to a namespace URI
+        prefix = next(iter(prefixes))
+        namespace = self.namespaces.get(prefix)
+        if not namespace:
+            return False, None
+
+        return True, str(namespace)
+
     def visit_slot(self, aliased_slot_name: str, slot: SlotDefinition) -> None:
         if self.exclude_imports and slot.name not in self._local_slots:
             return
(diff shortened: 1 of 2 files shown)


kevinschaper (approved):
looks good, thanks @jdsika!

PR #3449, opened 2026-04-25
feat(generators): add --default-language flag for language-tagged literals

jdsika:
## Summary

Adds a `--default-language` CLI option to both `gen-owl` and `gen-shacl` that emits BCP 47 language-tagged string literals (e.g. `"Person"@en`) for human-readable annotations.

This enables ontology producers to comply with **RDF 1.1 §3.3** (language-tagged strings as `rdf:langString`) and **OWL 2 Structural Specification §10** (annotation axioms accept literal values, including `rdf:langString`) without manual post-processing. The LinkML metamodel already has an `in_language` metaslot, but no generator currently uses it — this PR connects the dots.

## Problem

LinkML generators currently emit all string literals as plain `xsd:string` values, even for human-readable annotations like `rdfs:label`, `rdfs:comment`, `sh:name`, and `sh:description`. This prevents downstream consumers from:

- Filtering labels by language in SPARQL (`FILTER(lang(?label) = "en")`)
- Supporting multilingual ontologies where the same concept has labels in multiple languages
- Complying with W3C best practices for language-tagged metadata

### Real-world motivation

In the [ENVITED-X](https://2go4it.2go4it.de/product/envited-x/) ontology ecosystem (Gaia-X-compliant ontologies for the automotive domain, maintained by [ASCS e.V.](https://2go4it.2go4it.de/product/ascs-ev/)), all OWL and SHACL artifacts are generated from LinkML schemas. The generated ontologies are used by European data space participants who require language-tagged metadata for multilingual support (German/English at minimum). Currently, language tags must be added via manual post-processing, which is fragile and doesn't scale.

### 1. Only human-readable strings are tagged

A frozenset `_LANGUAGE_TAGGABLE_RANGES` guards which metaslot ranges are eligible:

```python
_LANGUAGE_TAGGABLE_RANGES: frozenset[str] = frozenset({"string", "ncname"})
```

Technical types (URI, integer, boolean, datetime) are never tagged. This prevents semantic errors where e.g. constraint values in `owl:oneOf` or `sh:in` would change from `xsd:string` to `rdf:langString`, breaking SHACL validation matching.

### 2. Element-level `in_language` overrides the default

`_resolve_language()` checks the element's `in_language` first, then falls back to the generator-level default. This leverages the existing LinkML metamodel without introducing new metaslots. Both generators (owlgen + shaclgen) delegate to a shared `LanguageTagResolver` from `linkml.utils.language_tags`.

### 3. BCP 47 syntactic validation (RFC 5646 §2.1 / §2.2.9)

The shared module provides a `BCP47_RE` regex compiled from the RFC 5646 §2.1 ABNF grammar (covering `langtag`, `privateuse`, and `grandfathered` productions). *Well-formedness* in the sense of RFC 5646 §2.2.9 (Classes of Conformance) is what we check — registry-level *validity* is out of scope for a code generator. Tags that don't match emit a **warning** (not error), since:

- Private-use tags (`x-...`) are valid per spec
- rdflib has its own lax validator that rejects truly malformed tags
- Our regex catches tags that pass rdflib but don't conform to strict BCP 47

This follows the principle of "warn on dubious, fail on impossible" — consistent with how rdflib itself handles edge cases.

### 4. Enum constraint values are NOT tagged

Permissible value literals in `owl:oneOf` restrictions are constraint data, not labels. Their `rdfs:label` IS tagged, but the literal value itself remains `xsd:string`. Same for `sh:in` constraint lists. This is critical for correct SHACL validation:

```turtle
# Constraint value — must remain plain xsd:string:
owl:oneOf ("Red" "Green" "Blue")
# Label — can be language-tagged:
<http://example.org/Red> rdfs:label "Red"@en .
```

### 5. Whitespace normalisation

Empty and whitespace-only values are normalised to `None`, preventing emission of empty language tags.

### 6. Architectural alignment between generators

Both `gen-owl` and `gen-shacl` import the same `LanguageTagResolver` from the new `linkml.utils.language_tags` module — the regex, well-formedness check, and resolution order live in one place, so a future bug fix lands once. Adopting the same pattern in another generator is now a two-line change (add a `default_language` field, instantiate the resolver in `__post_init__`).

### gen-owl (`owlgen.py`)
- New `default_language: str | None` field
- `_LANGUAGE_TAGGABLE_RANGES` frozenset (`string`, `ncname`)
- `_resolve_language()` delegates to `LanguageTagResolver`
- `_literal()` helper creates properly tagged `Literal` objects
- `add_metadata()` tags string-range annotation literals only; the catch-all `else` branch (reached by enum-ranged metaslots such as `pv_formula`, `obligation_level`, `alias_predicate`) always emits a plain `xsd:string` literal
- `add_enum()` PV labels respect language tags
- New `--default-language` Click CLI option

### gen-shacl (`shaclgen.py`)
- New `default_language: str | None` field
- `_resolve_language()` delegates to `LanguageTagResolver`
- NodeShape `rdfs:label` / `rdfs:comment` get language tags
- PropertyShape `sh:name` / `sh:description` get language tags via `prop_pv_text()`
- `_add_annotations()` tags string annotation values (non-string types keep XSD datatype)
- Numeric literals (`sh:order`, `sh:minCount`, etc.) are never tagged
- New `--default-language` Click CLI option

### Shared module (`linkml/utils/language_tags.py`)
- `BCP47_RE` — RFC 5646 §2.1 ABNF (langtag / privateuse / grandfathered)
- `is_well_formed_bcp47()` — RFC 5646 §2.2.9 well-formedness predicate
- `LanguageTagResolver` — validates the generator-level default once at construction; per-element `in_language` tags are validated lazily, with at most one warning per distinct malformed tag (so one bad tag stamped on hundreds of elements yields a single warning, not hundreds)

### Tests
- **OWL**: 13 tests — tagged labels, backward-compat, URI ranges (strong negative assertion), enum-ranged catch-all branch (regression test for review comment #2), `in_language` override, annotations, empty/whitespace normalisation, BCP 47 warning (valid + invalid), `in_language` BCP 47 warning, dedup
- **SHACL**: 13 tests — NodeShape, PropertyShape, plain literals, numeric guards, annotations, empty/whitespace normalisation, `in_language` override, BCP 47 warning (valid + invalid), `in_language` BCP 47 warning, dedup

## Usage

```bash
# Tag all human-readable strings with English:
gen-owl --default-language en schema.yaml > ontology.owl.ttl
gen-shacl --default-language en schema.yaml > shapes.shacl.ttl

# German ontology:
gen-owl --default-language de schema.yaml > ontologie.owl.ttl

# Element-level override (in LinkML schema):
classes:
  Vehicle:
    in_language: de       # overrides --default-language for this class
    attributes:
      name:
        range: string

# No flag = backward compatible (no language tags):
gen-owl schema.yaml > ontology.owl.ttl
```

### Example output

Without `--default-language`:
```turtle
<http://example.org/Vehicle> rdfs:label "Vehicle" ;
    rdfs:comment "A motorised vehicle" .
```

With `--default-language en`:
```turtle
<http://example.org/Vehicle> rdfs:label "Vehicle"@en ;
    rdfs:comment "A motorised vehicle"@en .
```

## Standards compliance

| Standard | Requirement | Status |
|----------|-------------|--------|
| **RDF 1.1 §3.3** (Literals) | `rdf:langString` vs `xsd:string` are different datatypes | ✅ Correctly distinct |
| **OWL 2 Structural Specification §5.5** (Annotation Properties) and **§10** (Annotations) | Annotation property values accept `rdf:langString` literals | ✅ |
| **SHACL §2.3.2.1** (`sh:name` / `sh:description`) | Values are `xsd:string` or `rdf:langString` literals | ✅ |
| **BCP 47 / RFC 5646** | Well-formedness per §2.2.9 against the §2.1 ABNF | ✅ Warning on non-conforming tags (de-duplicated) |

## Backward compatibility

- Default is `None` — **existing behaviour is completely unchanged**
- All pre-existing OWL+SHACL tests continue to pass
- Empty strings and whitespace-only values normalised to `None` (safe no-op)
- SPARQL note: `"Vehicle"` ≠ `"Vehicle"@en` per RDF 1.1 — this is intentional and documented

## Changes since last review (commit `507c3ee8`)

Resolves all five line-level comments from @amc-corey-cox:

1. **De-duplicated BCP 47 warnings** (comment 1). Validation now happens once at construction for the default tag, and at most once per distinct element-level `in_language` value. New `LanguageTagResolver` in `linkml/utils/language_tags.py` holds the `_warned: set[str]` state. Two new regression tests (`test_default_language_bcp47_warning_is_deduplicated`, one per generator) pin the behaviour.

2. **Gated the catch-all branch in `OwlSchemaGenerator.add_metadata`** (comment 2). Enum-ranged metaslot values are now emitted as plain `xsd:string`, never `rdf:langString` — preserves `sh:in` and `owl:oneOf` matching for downstream consumers. The originally cited example `status="testing"` was illustrative; `status` actually has range `uriorcurie` and takes the URIRef branch. The real catch-all victims are enum-ranged metaslots such as `pv_formula`, `obligation_level`, and `alias_predicate`. A new test (`test_default_language_does_not_tag_enum_ranged_metaslot_in_catchall_branch`) monkey-patches `pv_formula`'s slot URI so it reaches the branch and asserts the emitted value carries no language tag.

3. **`_BCP47_RE` `ClassVar` discrepancy** (comment 3). Resolved by extraction: the inline regex no longer exists in `shaclgen.py`.

4. **Extracted the duplicated regex + resolver into a shared module** (comment 4). `linkml.utils.language_tags` exposes `BCP47_RE`, `is_well_formed_bcp47()`, and `LanguageTagResolver`. Both generators import from this single location, so future bug fixes land once.

5. **Strengthened the URI-range test** (comment 5). The previous test asserted only "at least one `@de` label exists". It now sweeps the whole graph and asserts no language-tagged literal appears outside an allowlist of human-readable predicates (`rdfs:label`, `rdfs:comment`, `skos:definition`, `skos:prefLabel`, `skos:altLabel`, `skos:editorialNote`, `skos:note`, `skos:example`, `dcterms:title`, `dcterms:description`) — a strong negative. Also asserts `bibo:status` (uriorcurie range) emits a URIRef.

Cosmetic: in both generators, `self._language_resolver = LanguageTagResolver(...)` is now assigned before `super().__post_init__()` so any parent-class hook that might invoke `_resolve_language` during initialisation is safe.

## Testing

- **123 language-tag tests** pass in the gen-owl + gen-shacl suites (one new owl test from this round). Broader `tests/linkml/test_generators/` suite: **744 passed, 30 skipped, 3 xfailed**. All Slow Tests on Ubuntu and Windows × Python 3.10–3.14 are green on CI.
- **3 rounds of adversarial review** completed across standards compliance, code correctness, RDF semantics, compiler properties, and performance.


packages/linkml/src/linkml/generators/owlgen.py
```diff
@@ -234,6 +235,73 @@ def _present(values: Iterable[_T | None]) -> list[_T]:
...
+        if element is not None:
+            element_lang = getattr(element, "in_language", None)
+            if element_lang and element_lang.strip():
+                tag = element_lang.strip()
+                if not self._BCP47_RE.match(tag):
+                    logger.warning("in_language value %r is not a well-formed BCP 47 tag (RFC 5646 §2.1)", tag)
+                return tag
+        tag = (self.default_language or "").strip() or None
+        if tag is not None and not self._BCP47_RE.match(tag):
+            logger.warning("--default-language value %r is not a well-formed BCP 47 tag (RFC 5646 §2.1)", tag)
```
amc-corey-cox: `_resolve_language` runs on every call (per element + per annotation), so a malformed tag emits one warning per call — potentially hundreds per run.

For `default_language` you could validate once in `__post_init__` and stash the resolved value. For `in_language` (per-element) it's trickier.

Can you use an existing pattern in linkml for de-duped warnings? If not, please don't invent infrastructure just for this — leaving as-is is fine.

packages/linkml/src/linkml/generators/owlgen.py
```diff
@@ -350,7 +423,7 @@ def add_metadata(self, e: Definition | PermissibleValue, uri: URIRef) -> None:
                     # else:
                     #    logger.debug(f"Skipping {uri} {metaslot_uri} => {v}")
                 else:
-                    obj = Literal(v)
+                    obj = Literal(v, lang=lang) if lang else Literal(v)
```
amc-corey-cox: This `else` is reached for ranges that aren't types/subsets/classes — most notably enum-ranged metaslots (e.g. `status` has range `LifeCycleEnum`). With `--default-language en`, `status="testing"` becomes `"testing"@en`, which the `_LANGUAGE_TAGGABLE_RANGES` allowlist above was meant to prevent. Is this branch reachable for non-string values in practice? If yes, worth gating by the allowlist too.

packages/linkml/src/linkml/generators/shaclgen.py
```diff
@@ -74,13 +75,65 @@ class ShaclGenerator(Generator):
...
     generatorversion = "0.0.1"
     valid_formats = ["ttl"]
     file_extension = "shacl.ttl"
     visit_all_class_slots = False
     uses_schemaloader = False
 
+    # Syntactic validator for BCP 47 language tags (RFC 5646 §2.1 ABNF).
+    # Each group maps 1:1 to an ABNF production: language, script, region,
+    # variant, extension, privateuse, and grandfathered (irregular + regular).
+    _BCP47_RE: re.Pattern[str] = re.compile(
```
amc-corey-cox: Missing `ClassVar` vs. owlgen.py:252. Without it, dataclass treats `_BCP47_RE` as an instance field — it lands in `__init__` and becomes per-instance overridable. One-word fix; worth evening out with owlgen while the two are side-by-side.

packages/linkml/src/linkml/generators/shaclgen.py
```diff
@@ -74,13 +75,65 @@ class ShaclGenerator(Generator):
...
+        r"|en-GB-oed|i-ami|i-bnn|i-default|i-enochian|i-hak|i-klingon"
+        r"|i-lux|i-mingo|i-navajo|i-pwn|i-tao|i-tay|i-tsu"
+        r"|sgn-BE-FR|sgn-BE-NL|sgn-CH-DE"
+        r"|art-lojban|cel-gaulish|no-bok|no-nyn|zh-guoyu"
+        r"|zh-hakka|zh-min|zh-min-nan|zh-xiang"
+        r")$",
+        re.ASCII,
+    )
+
+    def _resolve_language(self, element=None) -> str | None:
```
amc-corey-cox, line 95: nit: `_BCP47_RE` and `_resolve_language` are duplicated verbatim with owlgen. Your PR description calls out architectural alignment between generators as a goal, which argues for extracting these into a shared helper. Not a blocker — flagging because the first bug fix to one will get missed in the other.

tests/linkml/test_generators/test_owlgen.py
```diff
@@ -824,3 +824,271 @@ def test_children_are_mutually_disjoint(
...
+    g = Graph()
+    g.parse(data=owl, format="turtle")
+
+    # Verify labels do get the tag
+    labels = list(g.objects(EX.Vehicle, RDFS.label))
+    assert Literal("Vehicle", lang="de") in labels
+
+    # Verify integer/boolean metaslots (if any) don't get tags
+    # The schema title should be tagged (string range)
+    assert any(isinstance(o, Literal) and o.language == "de" for o in g.objects(None, RDFS.label)), (
```
amc-corey-cox: This assertion only checks the positive case (at least one `@de` label exists). The test's stated intent — verifying URI/integer/boolean metaslots aren't tagged — isn't actually tested: a regression where URI ranges *do* get tagged would still pass. Please add an explicit negative assertion (e.g. that uriorcurie metaslot objects are `URIRef`, not language-tagged `Literal`) or rename the test.

amc-corey-cox (changes requested):
Nice feature with thorough tests. A few asks before approval plus a couple of nits and a question. Looks good overall. Thanks!

packages/linkml/src/linkml/utils/language_tags.py
```diff
@@ -0,0 +1,116 @@
...
+    r"(?:(?:[A-Za-z]{2,3}(?:-[A-Za-z]{3}){0,3})|[A-Za-z]{4}|[A-Za-z]{5,8})"
+    r"(?:-[A-Za-z]{4})?"
+    r"(?:-(?:[A-Za-z]{2}|\d{3}))?"
+    r"(?:-(?:[A-Za-z\d]{5,8}|\d[A-Za-z\d]{3}))*"
+    r"(?:-[0-9A-WY-Za-wy-z](?:-[A-Za-z\d]{2,8})+)*"
+    r"(?:-x(?:-[A-Za-z\d]{1,8})+)?"
+    # privateuse
+    r"|x(?:-[A-Za-z\d]{1,8})+"
+    # grandfathered (irregular)
+    r"|en-GB-oed|i-ami|i-bnn|i-default|i-enochian|i-hak|i-klingon"
```
jdsika, line 48: @amc-corey-cox special emphasis on the "i-klingon" here -  apparently it is still valid but deprecated and replaced by "tlh" - things you learn while working on linkml PRs :)))

jdsika:
Hey @amc-corey-cox, thank you for the thorough review. I hope commit `507c3ee8` addresses your concerns — here is the AI summary of the second commit:

---------------  SUMMARY of second commit  -------------

**Crosswalk to your five comments:**

1. **De-duplicate BCP 47 warnings.** A new `LanguageTagResolver` class in `linkml/utils/language_tags.py` validates the generator-level default tag once at construction and remembers per-element `in_language` tags it has already warned about (`_warned: set[str]`). One bad tag stamped on hundreds of elements now yields a single warning, not hundreds. Pinned by two new regression tests (`test_default_language_bcp47_warning_is_deduplicated`, one per generator).

2. **Catch-all `else` branch in `owlgen.add_metadata` tags enum values.** Fixed: the branch now always emits a plain `xsd:string` literal. The example in your comment (`status="testing"`) is illustrative — `status` actually has range `uriorcurie` and takes the URIRef branch — but the underlying concern is real for genuinely enum-ranged metaslots (`pv_formula`, `obligation_level`, `alias_predicate`). A new test (`test_default_language_does_not_tag_enum_ranged_metaslot_in_catchall_branch`) monkey-patches `pv_formula`'s slot URI so it actually reaches that branch and asserts no language tag is applied.

3. **Missing `ClassVar` on `_BCP47_RE` in `shaclgen`.** Resolved by extraction — the inline regex is gone from `shaclgen.py` entirely.

4. **Regex + resolver duplicated across the two generators.** Extracted to `linkml/utils/language_tags.py` (exports `BCP47_RE`, `is_well_formed_bcp47()`, `LanguageTagResolver`). Both generators now import from one place, so future bug fixes land once.

5. **URI-range test only checks the positive case.** Rewritten with a strong negative assertion: the test sweeps the whole graph and asserts no language-tagged literal appears outside an allowlist of human-readable predicates (`rdfs:label`, `rdfs:comment`, `skos:definition`, `skos:prefLabel`, `skos:altLabel`, `skos:editorialNote`, `skos:note`, `skos:example`, `dcterms:title`, `dcterms:description`). Also asserts `bibo:status` emits a URIRef.

**Standards refs in code/comments now correctly distinguish RFC 5646 §2.1 (ABNF) from §2.2.9 (well-formedness).** Each production of the §2.1 grammar — `langtag`, `extlang`, `script`, `region`, `variant`, `extension`/`singleton` (correctly excluding `X`/`x`), `privateuse`, and all 26 grandfathered tags (17 irregular + 9 regular) — is covered by `BCP47_RE`. The PR description's standards-compliance table also fixes the OWL 2 section (was `§6.3`, now correctly cites Structural Specification §5.5 + §10).

**Cosmetic:** `self._language_resolver = LanguageTagResolver(...)` is assigned before `super().__post_init__()` in both generators, so any parent-class hook that might call `_resolve_language` during initialisation is safe.

**Tests:** 123 language-tag tests pass locally (`test_owlgen.py` + `test_shaclgen.py`); broader `tests/linkml/test_generators/` suite is 744 passed / 30 skipped / 3 xfailed. CI is green across Ubuntu/Windows × Python 3.10–3.14 including Slow Tests.

The stale `> Depends on: #3447` line has also been removed from the PR description — #3447 touches only `jsonldcontextgen.py`, has no code-level relationship to this PR, and is already merged.

Happy to add more grandfathered-tag coverage or anything else you'd like.

amc-corey-cox (approved):
Thanks for the thorough follow-up — all five points addressed cleanly, with a regression test pinning each.

  Two things worth calling out: on comment #2, my original status="testing" example was actually wrong (status is uriorcurie and takes the URIRef branch) — you correctly tracked down the real victims, the enum-ranged metaslots like pv_formula/obligation_level, and fixed those. And the shared-module extraction for #4 went past a simple copy-paste dedup into a single source for the regex, well-formedness check, and resolution order, so the next bug fix only has to land once. 

Nice architectural calls - thanks for the great work.

Issue #2464, opened 2024-12-18
SHACL generator: support LinkML rules


elichad:
**What is your feature request?**
The SHACL generator should support LinkML [rules](https://linkml.io/linkml/schemas/advanced.html#rules). Currently they are not considered.

**How important is this feature to you?** Select from the options below:
• Medium - can do work without it; but it's important (e.g. to save time or for convenience)

**Additional context**
I have yet to determine if workarounds are available for my specific use case, or if this is absolutely blocking for me.

mih:
Possibly related: Slots' `ifabsent` is also not supported by the SHACL generator and raises `NotImplemented`.

elichad:
@mih that sounds like a separate issue to me - there are many LinkML features not supported by the SHACL generator currently, so I think it's better to track them separately. 

Also, yours sounds like an easier fix than mine :) There's already some infrastructure for converting `ifabsent`, it might just be missing support for whatever datatype you're using: linkml/generators/shacl/shacl_ifabsent_processor.py

(I'm not a LinkML dev but I've been poking around the SHACL generator code myself)

jsheunis:
> there are many LinkML features not supported by the SHACL generator currently, so I think it's better to track them separately

> There's already some infrastructure for converting ifabsent, it might just be missing support for whatever datatype you're using

Agreed. I created an issue here: https://github.com/linkml/linkml/issues/2502

jsheunis:
Coming back to this issue to up it again. We're working with tools to generate forms from SHACL shapes, and so-called conditionals are essential for an intuitive UX in such forms, e.g. if a user selects Option A in Question 1, then Questions 8-10 should be mandatory, otherwise not. Another quite simple case is to support something similar to LinkML's string serialization and ifabsent processors, but with those same conditionals exported to SHACL (e.g. if the user does not complete field X, generate its value by concatenating the values of fields Y and Z).

From my limited search, it seems that LinkML would need to support [SHACL advance features](https://www.w3.org/TR/shacl-af) to some extent in order to allow the export of SHACL that include expressions, functions, and/or rules. Quoting from these docs:

> SHACL functions can also be used declaratively in frameworks such as the SHACL [node expressions](https://www.w3.org/TR/shacl-af/#dfn-node-expressions) which are used in [SHACL rules](https://www.w3.org/TR/shacl-af/#dfn-shacl-rules). In those scenarios they may be used to perform data transformations such as string concatenation.

[...]

PR #3451, opened 2026-04-27
feat(gen-shacl): generate sh:sparql constraints from LinkML rules

jdsika:

## Summary

Implement SHACL-SPARQL constraint generation for the boolean-guard pattern commonly used in conditional validation rules. When a LinkML class has `rules:` blocks with preconditions (`value_presence: PRESENT`) and postconditions (`equals_string: true`), the generator now emits `sh:SPARQLConstraint` nodes on the corresponding `sh:NodeShape`.

This is a first step toward solving #2464 by supporting the most common rule pattern. Unrecognised patterns are silently skipped (logged at DEBUG level), so this is fully backward-compatible.

### 1. Boolean-guard pattern
Precondition: value_presence PRESENT on slot A
Postcondition: equals_string true on slot B
Semantics: If A has a value, B must be true.

### 2. Exclusive-value pattern (NEW)
Precondition: equals_string SomeValue on slot X
Postcondition: maximum_cardinality N on the SAME slot X
Semantics: If slot X contains value V, then X may have at most N values (mutual exclusion).

For N=1: emits a FILTER-based SPARQL query
For N>1: emits a COUNT-based subquery with HAVING clause
Enum values resolved to full IRIs via meaning field when available.

## Features

- New `_add_rules()` method translates recognised rule patterns to SPARQL

- **Boolean-guard pattern**: precondition `value_presence: PRESENT` + postcondition `equals_string: true`

- Rule `description` mapped to `sh:message` on the constraint

- Deactivated rules are skipped

- Warnings emitted for `bidirectional`/`open_world` rule flags

- New `--emit-rules/--no-emit-rules` CLI flag (default: enabled)

- Full URI references in SPARQL queries (no PREFIX declarations needed)

## Standards Compliance

- W3C SHACL Section 5 (SPARQL-based Constraints)

- Pre-bound `$this` variable per Section 5.3.1

- Validated by pyshacl with `advanced=True`

## Tests

18 tests in `test_shaclgen.py` covering boolean-guard generation, sh:message, deactivated rules, unsupported patterns, --no-emit-rules flag, multiple rules per class, exclusive-value pattern (9 tests), bidirectional skip, elseconditions warning, and end-to-end pyshacl validation.

## Motivation

We maintain [ASCS-eV/ontology-management-base](https://github.com/ASCS-eV/ontology-management-base), which uses LinkML to generate SHACL for the OpenLabel ontology. The handcrafted v1 SHACL has 17 `sh:sparql` constraints for conditional validation. With this PR, `gen-shacl` generates all 17 constraints automatically from the LinkML `rules:` blocks.

Refers to #2464 (partial - boolean-guard pattern)

## Changes since review (commit `e92dcb38a`)

Addresses all 6 comments from @amc-corey-cox:

1. **bidirectional rules** — now **skip-and-warn** (`continue`) instead of emit-and-warn, preventing silent semantic divergence when users don't read logs.
2. **elseconditions** — added explicit `logger.warning()` when the else branch is dropped, so schema authors know only the forward branch is emitted.
3. **PresenceEnum wrapping** — kept `PresenceEnum(PresenceEnum.PRESENT)`; it is **not redundant**. `PresenceEnum.PRESENT` is a `PermissibleValue`, but parsed schemas return `PresenceEnum` instances. The wrapping ensures type-compatible comparison. Added an explanatory comment in code.
4. **xsd:boolean vs xsd:string** — SPARQL now uses `str(?flag) != "true"` which works regardless of whether data stores the flag as `xsd:boolean` or `xsd:string` (avoids SPARQL type-error coercion to false).
5. **Unused fixture** — deleted `boolean_guard_rules.yaml` (tests use inline schema strings).
6. **End-to-end pyshacl test** — `test_rule_boolean_guard_pyshacl_end_to_end` validates both a conforming instance (passes) and a violating instance (correctly flagged) with `pyshacl.validate(advanced=True)`.


packages/linkml/src/linkml/generators/shaclgen.py
```diff
@@ -283,10 +300,228 @@ def st_node_pv(p, v):
...
+                    cls.name,
+                )
+
+            if getattr(rule, "open_world", False):
+                logger.warning(
+                    "Rule in class %r has open_world=true; "
+                    "SHACL operates under closed-world assumption. "
+                    "The constraint is emitted but may not match open-world semantics.",
+                    cls.name,
+                )
```
amc-corey-cox, line 463: Question on the contract here: when `bidirectional` or `open_world` is set, we emit the forward / closed-world reading plus a warning. That's defensible (partial validation is still useful), but a user who doesn't read logs gets silent semantic divergence from the source rule. Intentional, or would you rather skip-and-warn in these cases?

packages/linkml/src/linkml/generators/shaclgen.py
```diff
@@ -283,10 +300,228 @@ def st_node_pv(p, v):
...
+
+    def _rule_to_sparql(self, sv, cls: ClassDefinition, rule) -> str | None:
+        """Convert a ``ClassRule`` to a SPARQL SELECT query string.
+
+        Returns ``None`` when the rule does not match any supported pattern.
+        """
+        pre = getattr(rule, "preconditions", None)
+        post = getattr(rule, "postconditions", None)
+        if not pre or not post:
+            return None
```
amc-corey-cox, line 500: Rules with `elseconditions` are accepted here and only the forward branch is emitted — the else branch is silently dropped (the test at `test_rule_with_elseconditions_emitted` locks in this behavior). Same question as the bidirectional/open_world case: is silent partial emission the intended contract, or should this log a warning so a schema author isn't surprised?

packages/linkml/src/linkml/generators/shaclgen.py
```diff
@@ -283,10 +300,228 @@ def st_node_pv(p, v):
...
+        # preconditions: exactly one slot with value_presence PRESENT
+        # postconditions: exactly one slot with equals_string "true"
+        if len(pre_slots) == 1 and len(post_slots) == 1:
+            pre_slot_name = next(iter(pre_slots))
+            post_slot_name = next(iter(post_slots))
+
+            pre_cond = pre_slots[pre_slot_name]
+            post_cond = post_slots[post_slot_name]
+
+            is_value_present = getattr(pre_cond, "value_presence", None) == PresenceEnum(PresenceEnum.PRESENT)
```
amc-corey-cox, line 517: Minor: `PresenceEnum(PresenceEnum.PRESENT)` re-wraps an already-enum value. `PresenceEnum.PRESENT` (or just `"PRESENT"`) should work directly.

packages/linkml/src/linkml/generators/shaclgen.py
```diff
@@ -283,10 +300,228 @@ def st_node_pv(p, v):
...
+        """
+        flag_uri = self._slot_uri(sv, flag_slot_name, cls)
+        value_uri = self._slot_uri(sv, value_slot_name, cls)
+
+        return (
+            f"SELECT $this WHERE {{\n"
+            f"    OPTIONAL {{ $this <{flag_uri}> ?flag . }}\n"
+            f"    OPTIONAL {{ $this <{value_uri}> ?value . }}\n"
+            f"    FILTER (\n"
+            f"        ( !BOUND(?flag) || ?flag != true ) &&\n"
```
amc-corey-cox: `?flag != true` compares as `xsd:boolean`. But the source rule says `equals_string: "true"` — i.e. the flag is a *string*. If the data stores the flag as `xsd:string` (as the LinkML rule actually requires), this comparison raises a SPARQL type error → coerced to false → no violation reported. Worth either documenting the caveat or normalising the comparison (e.g. `str(?flag) != "true"`).

tests/linkml/test_generators/input/shaclgen/boolean_guard_rules.yaml
```diff
@@ -0,0 +1,70 @@
+id: https://example.org/boolean-guards
```
amc-corey-cox: This fixture isn't referenced by any test — the test suite inlines duplicate copies of the same schema as Python strings. Either point a test at this file or delete it.

tests/linkml/test_generators/test_shaclgen.py
```diff
@@ -1160,3 +1160,772 @@ def test_nodeidentifier_range_produces_blank_node_or_iri():
...
+    g = _parse_shacl(_RULES_SCHEMA_YAML)
+
+    shape = EX_RULES.Environment
+    sparql_nodes = list(g.objects(shape, SH.sparql))
+    assert len(sparql_nodes) >= 1
+
+    for node in sparql_nodes:
+        query_text = str(list(g.objects(node, SH.select))[0])
+        # prepareQuery validates SPARQL syntax; $this is a valid variable name
+        prepareQuery(query_text)
```
amc-corey-cox, line 2255: `prepareQuery` only checks SPARQL syntax. The PR description claims the output is validated by pyshacl with `advanced=True`, but no test exercises that path. Could you add at least one end-to-end test: build an RDF instance that violates the rule, run pyshacl against the generated SHACL, and assert the violation is flagged (and a clean instance passes)?

amc-corey-cox (changes requested):
Two pattern translations (boolean-guard, exclusive-value) → sh:SPARQLConstraint. Clean extension to ShaclGenerator, good test structure. Main concern: the spec-conformance claim in the description isn't backed by an end-to-end pyshacl test. A few smaller correctness/contract questions inline.

amc-corey-cox (approved):
On the PresenceEnum comment — you were right to push back: a parsed schema yields a PresenceEnum instance that doesn't equal the bare PresenceEnum.PRESENT PermissibleValue, so the wrapping is required and my suggested simplification would've silently broken the match.

Thanks, and good call on the clarifying comment.

This looks good to me now.

jdsika:
I will resolve the merge conflict in the morning!

amc-corey-cox:
I won't get back to this until Monday - but I will get it merged then if the conflicts are fixed.
Thanks!

jdsika:
@amc-corey-cox are you ready to merge? I am building up a pile of PRs here :)

amc-corey-cox:
Sorry for the delay @jdsika - I went out of town right after I posted that and it took me longer to get back on top of things than I expected. I'm getting this in now as soon as all the tests pass after merging main in.

Issue #3458, opened 2026-04-29
pythongen: Unsafe prefix names produce invalid or broken Python — no validation against keywords, builtins, or non-identifier characters


HendrikBorgelt:
---
name: Bug report 
about: Create a report to help us improve
title: 'pythongen: Unsafe prefix names produce invalid or broken Python — no validation against keywords, builtins, or non-identifier characters'
labels: bug, generator-python, generator-dataclasses, generator-pydantic
assignees: ''

---

**Describe the bug**
The gen-python (and gen-pydantic) generators do not validate that schema prefix names produce valid, safe Python identifiers before emitting them as module-level variable names or attribute-access expressions in generated code. This causes NameError, SyntaxError, or silent shadowing of Python builtins at import time of the generated module — with no warning at generation time.

**Version of LinkML you are using**
1.9.0

**Please provide a schema (and if applicable, a data file) that replicates the issue**

For Reproducing the schema you can right now just use a prefix with a dot in its prefix label.

The report down below goes slightly more into detail as to what caused this error. Sorry for it beeing LLM generated and therefore sounding very blunt. Also Issue #3376 allready mentions the issue, but only for a more limited example.

### Schema prefix (dotted name, e.g. after [Allotrope prefix renaming](https://github.com/nfdi-de/chem-dcat-ap/pull/135)

```yaml
prefixes:
  allotrope.equipment: http://purl.allotrope.org/ontologies/equipment#AFE_
```

### Generated output in `chem_dcat_ap.py`

```python
# Namespace declaration — dot correctly replaced with underscore ✓
ALLOTROPE_EQUIPMENT = CurieNamespace('allotrope_equipment', 'http://purl.allotrope.org/ontologies/equipment#AFE_')

# Usage site — dot interpreted as attribute access on an undefined object ✗
class Reactor(Device):
    class_class_uri: ClassVar[URIRef] = ALLOTROPE.EQUIPMENT["0000153"]
    #                                   ^^^^^^^ NameError: name 'ALLOTROPE' is not defined
```

### Error at import time

```
NameError: name 'ALLOTROPE' is not defined
```

## Root Cause

Two separate locations are responsible, and their behaviour is inconsistent
with each other.

### 1. `linkml/generators/pythongen.py` — `gen_namespaces()`

The `gen_namespaces()` method applies `.` → `_` and `-` → `_` substitution
when **declaring** the namespace variable:

```python
# linkml/generators/pythongen.py — gen_namespaces()
[...]

PR #3459, opened 2026-04-29
added python inbuild checking to pythongen as well as test

### LINKED ISSUE #3376: Pydantic generator may produce unloadable code if a slot has the same name as a Python type
**Describe the bug**
If a schema defines a slot (or an attribute) whose name happens to be the same as the name of a Python type that also happens to be used within the schema, the code produced by the Pydantic generator cannot be loaded – Pydantic throws an error as soon as one tries to import the generated module.

See a minimum example and the corresponding error below.

Tip of the main branch as of April 9th, 2026 (`v1.11.0-rc1-17-g7a99df91b`).

Given the following schema:

```yaml
id: https://example.org/foo
name: foo
prefixes:
imports:
  - linkml:types
default_range: string

classes:

  Date:
    attributes:
      date:
        range: date
        required: true
      type:
        required: true
```

(Note the `date` attribute, which has the same name as the `date` type in Python’s `datetime` package), if you generate Pydantic code with:

```console
$ linkml generate pydantic foo.yaml > foo.py
```

and then try to load the `foo` module, you get the following error:

```console
$ python -m foo
pydantic.errors.PydanticUserError: Error when building FieldInfo from annotated attribute. Make sure you don't have any field name clashing with a type annotation.

For further information visit https://errors.pydantic.dev/2.12/u/unevaluable-type-annotation
```

This is because the generate module contains the following:

```python
class Date(ConfiguredBaseModel):
    [...]
    date: date = Field(default=..., json_schema_extra = { "linkml_meta": {'domain_of': ['Date']} })
```

Pydantic [does not allow](https://pydantic.dev/docs/validation/latest/errors/usage_errors/#unevaluable-type-annotation) this situation where the type annotation of a field has the same name as the field itself.

HendrikBorgelt:
## Summary

Fixes #3376 & #3458 

# Fix: `pythongen` — unsafe prefix names produce broken or invalid Python

## Problem

The `gen-python` generator does not validate that schema prefix names produce
valid, safe Python identifiers before emitting them as module-level variable
names.  This causes two distinct failure classes in generated code:

**1. `NameError` at import time — dot-separated prefixes (primary bug)**

A prefix such as `allotrope.equipment` is a valid CURIE prefix (dots are legal
NCName characters per the W3C XML spec) and can appear in real-world schemas.
The generator declares it correctly:

```python
# gen_namespaces() — declaration  ✓
ALLOTROPE_EQUIPMENT = CurieNamespace('allotrope_equipment', '...')
```

But `curie_for(pythonform=True)` in `linkml_runtime/utils/namespaces.py`,
which is called for every class/slot URI expression, did not apply the same
substitution.  It emitted the raw uppercased prefix, producing Python
attribute-access syntax instead of a variable name:

```python
# curie_for(pythonform=True) — usage site  ✗
class Reactor(Device):
    class_class_uri: ClassVar[URIRef] = ALLOTROPE.EQUIPMENT["0000153"]
    #                                   ^^^^^^^ NameError: 'ALLOTROPE' is not defined
```

The two code paths had never been made consistent with each other.

**2. Silent builtin shadowing — prefixes that are Python builtin names**

A prefix such as `float`, `int`, `str`, `list`, or `type` generates a
module-level assignment `FLOAT = CurieNamespace(...)`.  This silently replaces
Python's built-in `float` type for all code in that module, causing subtle
runtime errors that are difficult to trace.

**3. `SyntaxError` at parse time — Python keyword prefixes**

A prefix such as `in`, `for`, `class`, or `def` can cause a `SyntaxError`
when the generated identifier is used in certain syntactic positions.

---

## Root Cause

Two separate code paths produced the Python variable name for a prefix, and
their transformation logic was inconsistent:

| Code path | Location | Transformation applied |
|---|---|---|
| Declaration | `pythongen.py` `gen_namespaces()` | `.replace('.','_').replace('-','_').upper()` |
| Usage | `namespaces.py` `curie_for(pythonform=True)` | `.upper()` only — **no substitution** |

Neither path checked for keyword or builtin clashes.

---

### `linkml/generators/pythongen.py`

Adds a module-level helper `_safe_python_identifier(prefix: str) -> str` that
applies all transformations in a single, testable place:

1. Replace `.` and `-` with `_`, uppercase.
2. Append `_NS` if the result clashes with a Python keyword, soft keyword
   (Python ≥ 3.12), or builtin name.
3. Prepend `NS_` if the result starts with a digit.

The helper is then used consistently in every declaration and usage site inside
the generator: `gen_namespaces()`, `python_uri_for()`, `gen_class_meta()`, and
`gen_type_meta()`.

### `linkml_runtime/utils/namespaces.py`

Adds `_prefix_to_python_var(prefix: str) -> str`, a mirror of the helper
above, and calls it inside `curie_for(pythonform=True)` to replace the
previous raw `.upper()`.  This ensures the usage-site expression always matches
the declaration-site variable name produced by `gen_namespaces()`.

The naming convention follows PEP 8:
> *single_trailing_underscore_: used by convention to avoid conflicts with
> Python keyword, e.g. `class_`*

`_NS` suffix is used (rather than a bare `_`) to make the intent clear and
to avoid accidental conflicts with other generated names.

---

## Tests added

`tests/linkml/test_generators/test_pythongen_safe_prefixes.py`

- **Unit tests for `_safe_python_identifier`** — parametrised over every
  unsafe category: dot-separated, hyphenated, Python builtins (`float`, `int`,
  `str`, `list`, `dict`, `set`, `type`, `id`, `object`), hard keywords (`in`,
  `for`, `class`, `not`, `def`), soft keywords, and leading-digit prefixes.
- **Unit tests for `_prefix_to_python_var`** — same parametrised table,
  asserting that the two helpers always produce identical output.  A mismatch
  between them is exactly the bug this PR fixes.
- **Integration tests** — generate and compile Python from
  `tests/linkml/test_generators/input/unsafe_prefix_names.yaml`, a schema that
  declares prefixes from every problematic category.  Assertions cover:
  - the generated module compiles without `NameError` or `SyntaxError`
  - dotted prefixes use `_` in both declarations and expressions
  - builtin prefixes use the `_NS` suffix in both declarations and expressions
  - keyword prefixes use the `_NS` suffix
  - already-safe prefixes are not altered

---

Important!! The FAQ can be updated https://linkml.io/linkml/faq/modeling.html#why-are-my-class-names-translated-to-camelcase

## Checklist

## AI Assistance

AI was used to assist in this PR submission. I have reviewed and checked the code, in accordance with AI_COVENANT.md.


packages/linkml/src/linkml/generators/pythongen.py
```diff
@@ -37,6 +38,24 @@
 logger = logging.getLogger(__name__)
 
 
+def _safe_python_identifier(prefix: str) -> str:
```
kevinschaper: **Structural ask before merge.** This helper and `_prefix_to_python_var` in `linkml_runtime/utils/namespaces.py` are the same function, and the test `test_prefix_to_python_var_matches` exists specifically to enforce "they must always agree." That's exactly the divergence-prone setup this PR is trying to protect against — if Python 3.14 adds a new soft keyword and someone updates one helper but not the other, the original bug class quietly returns and the test passes (because both helpers are equally wrong).

Suggest defining the helper once in `linkml_runtime.utils.namespaces` (the lower-level package) and importing it here:

```python
from linkml_runtime.utils.namespaces import _prefix_to_python_var as _safe_python_identifier
```

The packages are in the same UV workspace and pythongen already imports from linkml-runtime, so this is essentially free. The mirror test then becomes a self-evident tautology you can delete.

packages/linkml_runtime/src/linkml_runtime/utils/namespaces.py
```diff
@@ -10,6 +12,28 @@
 from linkml_runtime.utils.uri_validator import validate_uri
 from linkml_runtime.utils.yamlutils import TypedNode
 
+
+def _prefix_to_python_var(prefix: str) -> str:
```
kevinschaper, line 16: Suggest making this the single source of truth — see the comment on `_safe_python_identifier` in `pythongen.py`. This is the lower-level package; pythongen can import it directly.

packages/linkml/src/linkml/generators/pythongen.py
```diff
@@ -37,6 +38,24 @@
...
+    """Return a safe module-level Python variable name for a CURIE prefix.
+    Uppercases and replaces ``.`` / ``-`` with ``_``, appends ``_NS`` on
+    keyword/builtin clashes, prepends ``NS_`` on leading-digit prefixes."""
+    name = prefix.upper().replace(".", "_").replace("-", "_")
+    lower = name.lower()
+    clashes = keyword.iskeyword(name) or keyword.iskeyword(lower)
+    if not clashes and hasattr(keyword, "issoftkeyword"):
```
kevinschaper: Nit: linkml requires Python ≥ 3.10 where `keyword.issoftkeyword` is always present, so the `hasattr` guard doesn't buy anything. Could simplify to `clashes = clashes or keyword.issoftkeyword(name) or keyword.issoftkeyword(lower)`. Not blocking.

kevinschaper (commented):
Thanks for the fix @HendrikBorgelt!

I threw in a couple of small requested tweaks.

kevinschaper (changes requested):
putting it into "request changes" mode

packages/linkml_runtime/src/linkml_runtime/utils/namespaces.py
HendrikBorgelt: Thanks for spotting this. I implemented this as requested see: https://github.com/linkml/linkml/pull/3459/changes#diff-9552fca5599ddfd9ad93f6f3b47e60f7be8c9cb417a569ad9ccb282fb86306e3R36

HendrikBorgelt:
Thanks @kevinschaper for the adivese. I have implemented and tested everything, so I hope it is now working fine.

HendrikBorgelt:
@noelmcloughlin , I saw your issue #3316 was resolved, which partially overlaps with this PR. As such, I would also like you to comment on whether the slightly more extensive prefix/keyword check here would cause any issues with what you have requested and was merged with #3317.

noelmcloughlin:
Hi @HendrikBorgelt  thanks for the PR and appreciate reaching out to me.  #3317 was specifically covering PEP8 keyword convention and covers `slots` ( see the unit tests) so was handled inside pythongen.py.

You could rebase and use this PR solution instead - ensuring test cases from #3317 pass (and full test suite) to prove outcome is good.

kevinschaper (approved):

PR #3462, opened 2026-04-29
Updates LinkML model to 1.11.0-rc2 in runtime

matentzn:


amc-corey-cox:
Hey @matentzn — I pushed a commit that:
- Reverts the `permissivle` → `permissible` typo fix in `meta.yaml` (it wasn't in the upstream linkml-model release, so the vendoring test fails on it)
- Regenerates the test snapshots for `meta.owl`, `meta.ttl`, `meta.json`, `meta.sqla.py`

The remaining CI failures are all `test_vendored_files_match_upstream` for `meta.yaml` and `meta.context.jsonld`. These compare the vendored files against what `w3id.org` serves, which redirects to GitHub Pages. The raw `linkml_model/` files on `gh-pages` are only updated by the release workflow (not on merge to main), so this test will fail until the 1.11.0 release is cut on linkml-model.

I think the model is ready for release — can you cut 1.11.0 on linkml-model so we can get this PR to green?

PR #3467, opened 2026-04-30
fix(generators): panderagen cyclic dependency; polaris missing types

### LINKED ISSUE #3466: Pandergen: Cyclic dependency in sorting; Polars: missing linkml types
**Describe the bug**

Two issues encountered causing failed tests on PR #3420 and #3434

1. The `panderagen/class_handler_base.py` generator adds dependencies for hierarchy (ok) and adds dependencies by association (bug).  The bug manifests when a Cyclic dependency pattern is not detected and adds `cn --> slot.range` creating cycle with `child --> cn` and ValueError.
https://github.com/linkml/linkml/blob/86419ae238fd42ae3c53c14f9df9ccec4719921f/packages/linkml/src/linkml/generators/panderagen/class_handler_base.py#L48
```
tests/linkml/test_compliance/test_import_compliance.py::test_import_metamodel[False-dataframe_polars_schema]
ValueError: Cyclic dependency detected: common_metadata -> structured_alias -> ex...
```

2. `polars_schema_dataframe_generator -> PolarsSchemaDataframeGenerator`'s type map only covers basic XSD types and can't handle the metamodel's types (uri, uriorcurie, etc).
https://github.com/linkml/linkml/blob/86419ae238fd42ae3c53c14f9df9ccec4719921f/packages/linkml/src/linkml/generators/panderagen/polars_schema/polars_schema_dataframe_generator.py#L19
```
tests/linkml/test_compliance/test_import_compliance.py::test_import_metamodel[True-dataframe_polars_schema]
ValueError: Cyclic dependency detected: common_metadata -> structured_alias -> ex...
```

**Analysis**
Fixing seems trivial:

1. When processing associations, skip any cn ->slot.range where cn is already an ancestor of slot.range (i.e. slot.range is a descendant of cn). In that case the hierarchy edge slot.range --> cn already exists, so adding the reverse would create a cycle. The hierarchy ordering is enough to emit parent before the child.

2. Update the Polars generator's type map, the best matches (with cavets) are:

-  "shex:iri":                           "pl.Utf8",   # objectidentifier
-  "shex:nonLiteral":               "pl.Utf8",  # nodeidentifier
-  "linkml:DateOrDatetime":  "'pl.Utf8",   # No direct Polars equivalent, not ideal, very permissive.

**DataOrDateTime**
Using 'pl.Utf8' is not idea but the alternatives have caveat's so best to use simple case for now.
- "pl.Date"	  Matches dates, but loses sub-day precision
- "pl.Datetime"	  Most general — pl.Date values can be cast to it - could be better default but even Linkml's [DateOrDatetime](https://linkml.io/linkml-registry/DateOrDatetime/) uses `base:str` so Utf8 may be best.

DependencySorter
1.10.0

noelmcloughlin:
## Summary

Fixes #3466

## How was this tested?

Local Unit testing
Added test coverage with sample schema

## Areas of uncertainty

Check if `pl.Utf8` is best choice for Polars type map.

## Checklist


kevinschaper (changes requested):
Thanks for the fix @noelmcloughlin! 

The type-map additions look right, and the cycle-skip in `add_dependencies_by_association` does break the sort cycle. One concern: on the new `cyclic_model.yaml`, `serialize()` succeeds but the emitted code raises `NameError: name 'ChildStruct' is not defined` at import — `ParentDict` references `ChildStruct` before `ChildStruct` is bound. So the failure mode moves from codegen-time to import-time rather than going away.

Two options if you want to address it here: (a) emit in two passes — declare every `*Dict = {}` first, populate after all are declared, then bind `*Struct = pl.Struct(*Dict)`; or (b) keep this PR scoped to the type-map and the sort-cycle fix and open a follow-up for the forward-reference case. Either is fine with me — I'd lean toward (a) since you've got the context loaded, but happy to merge as-is with a tracking issue if you'd prefer.

Also a small drive-by while you're in the type map: line 24 has `"xsd:int": "pl:Int32"` (colon, should be a dot), it's not introduced in this PR but it would be awesome if you can sneak it into this PR. 

One last thing: would you mind tightening the cyclic test to `compile_dataframe_model(...)` instead of just substring-checking the serialized output? That'd catch the import-time issue above and any future regressions.

noelmcloughlin:
Hi @kevinschaper  thanks for the detailed review and finding gaps.  I updated the PR accordingly - and expanded the test coverage.  Here is CHANGELOG:

- polars_schema_dataframe_generator.py: Added `dict_range(field_range)` static method that converts XStruct -> pl.Struct(XDict) and pl.List(XStruct) -> pl.List(pl.Struct(XDict)) using a regex sub.

- panderagen_polars_schema/class.jinja2:  Split the single render_schema macro into three targeted macros:
  --pass 1-- declare_empty_dict - emits XDict = {}
  --pass 2-- populate_dict - emits XDict = {...} using generator.dict_range() so only *Dict names are needed
  --pass 3-- bind_struct_and_schema - emits XStruct = pl.Struct(XDict) and X = pl.Schema(XDict)

- panderagen_polars_schema/slots.jinja2: Added `render_slot_dict_range` macro that calls `generator.dict_range()` instead of emitting field.range directly.

- panderagen_polars_schema/polars_schema.jinja2:  Restructured to call the three macros in three separate loops.

- test_panderagen_polars.py:
   -- `test_parent_slot_range_child_no_cycle()` now calls `compile_dataframe_model()`
   ---- checks for expected attributes, catching both codegen-time and import-time failures.
   ---- skip Python 3.10 due to typing.Optional limitation.
  -- new `test_dict_range()` directly tests the regex with every distinct input class (coverage gap)
  
- cyclic_model.yaml: Added attribute `child_refs` (multivalued, inlined) so the test schema --> exercises pl.List(XStruct) -> pl.List(pl.Struct(XDict)) path

noelmcloughlin:
Claude found another bug - investigating.

 "_PR #3467 doesn't fully fix it — the parent-child cycle gets killed, but the metamodel has another cycle (common_metadata → extension → structured_alias → extension) the PR doesn't catch. Let me combine the PR fix with my allow_cycles patch on top._"

noelmcloughlin:
Ready for review

kevinschaper (approved):
thanks @noelmcloughlin!

PR #3473, opened 2026-05-02
fix(shaclgen): emit sh:minCount/maxCount 0 for zero cardinality values

jdsika:

## Summary

Fix a Python truthiness bug in the SHACL generator that prevents `sh:maxCount 0` and `sh:minCount 0` from being emitted when `maximum_cardinality: 0`, `minimum_cardinality: 0`, or `exact_cardinality: 0` is set in a LinkML schema.

## Problem

In `shaclgen.py`, the cardinality checks used Python truthiness:

```python
if s.minimum_cardinality:    # 0 is falsy in Python!
    prop_pv_literal(SH.minCount, s.minimum_cardinality)
elif s.exact_cardinality:    # 0 is falsy in Python!
    ...
if s.maximum_cardinality:    # 0 is falsy in Python!
    prop_pv_literal(SH.maxCount, s.maximum_cardinality)
elif s.exact_cardinality:    # 0 is falsy in Python!
    ...
```

Since `0` evaluates as `False` in Python, setting `maximum_cardinality: 0` (which should emit `sh:maxCount 0`, meaning "this property MUST NOT appear") produced no output at all. The same bug affected `minimum_cardinality: 0` and `exact_cardinality: 0`.

## Fix

Changed all four cardinality checks to explicit `is not None` comparisons:

```python
if s.minimum_cardinality is not None:
    prop_pv_literal(SH.minCount, s.minimum_cardinality)
elif s.exact_cardinality is not None:
    ...
if s.maximum_cardinality is not None:
    prop_pv_literal(SH.maxCount, s.maximum_cardinality)
elif s.exact_cardinality is not None:
    ...
```

This distinguishes "not set" (`None`) from "explicitly set to zero" (`0`), and matches the pattern already used in **the OWL generator** (`owlgen.py`) and the **doc generator** (`docgen.py`) for the same attributes.

## Precedence: explicit `minimum_cardinality` vs `required`

The generator resolves `sh:minCount` with an `elif` cascade in which an explicit `minimum_cardinality` is authoritative and `required` is only the fallback:

```python
if s.minimum_cardinality is not None:      # explicit cardinality wins
    prop_pv_literal(SH.minCount, s.minimum_cardinality)
elif s.exact_cardinality is not None:
    prop_pv_literal(SH.minCount, s.exact_cardinality)
elif s.required and not s.identifier:      # fallback
    prop_pv_literal(SH.minCount, 1)
```

A side effect of the fix: a slot with **both** `required: true` **and** `minimum_cardinality: 0` now emits `sh:minCount 0` (previously the truthiness bug skipped the `0` and the `required` fallback produced `sh:minCount 1`). This is intentional and consistent with the rest of LinkML:

- **`owlgen.py` uses the identical ordering** — `if slot.minimum_cardinality is not None: min_card = N ... elif slot.required: min_card = 1` — so OWL already emits `owl:minCardinality 0` for this combination. This change makes the SHACL output agree with the OWL output.
- The LinkML metamodel documents `minimum_cardinality` as "the minimum number of entries **for a multivalued slot**", so `required: true` + `minimum_cardinality: 0` is a **schema-authoring contradiction**. When it occurs, the explicit, more specific constraint (`minimum_cardinality: 0`) is emitted rather than the `required` shorthand.
- Per the W3C SHACL spec, [`sh:minCount 0`](https://www.w3.org/TR/shacl/#MinCountConstraintComponent) is vacuously satisfied, so emitting it is harmless and simply reflects the schema faithfully. Schemas that want `required`-style enforcement in SHACL should not set `minimum_cardinality: 0`.

## Use Case

This is needed for modeling class hierarchies where subclasses restrict inherited properties. `slot_usage` with `maximum_cardinality: 0` is the idiomatic way in LinkML to express "this inherited slot is not applicable on this subclass". Without this fix, the SHACL output silently omitted the constraint while the OWL output correctly emitted `owl:maxCardinality 0`, leading to inconsistent generated artifacts.

### Minimal reproduction

```yaml
classes:
  Parent:
    slots:
      - some_slot
  Child:
    is_a: Parent
    slot_usage:
      some_slot:
        maximum_cardinality: 0
```

**Before fix:** `gen-shacl` produces no `sh:maxCount` for `some_slot` on `Child`.
**After fix:** `gen-shacl` correctly produces `sh:maxCount 0`.

## Testing

Added fixture classes to `tests/linkml/test_generators/input/shaclgen/cardinality.yaml` and regression tests to `test_shaclgen.py`:

- `test_zero_maximum_cardinality_emits_maxcount` — `maximum_cardinality: 0` → `sh:maxCount 0`
- `test_zero_exact_cardinality_emits_both_counts` — `exact_cardinality: 0` → `sh:minCount 0` + `sh:maxCount 0`
- `test_zero_minimum_cardinality_emits_mincount` — `minimum_cardinality: 0` → `sh:minCount 0`
- `test_explicit_minimum_cardinality_overrides_required` — `required: true` + `minimum_cardinality: 0` → `sh:minCount 0` (documents the precedence, consistent with owlgen)

All existing tests pass (61 passed in `test_shaclgen.py`); non-zero cardinalities are unaffected by the `is not None` check.


jdsika:
That one seems straight forward to me @amc-corey-cox 

Use case is clear?

I want to explicitly disable inheritance of certain slots from a parent class - that was my case which uncovered this bug (?)

turbomam:
I approve of this. Thanks.

Changing `if s.minimum_cardinality:` to `is not None` is correct because 0 is legal for all of these and the old test could not tell "set to zero" apart from "not set at all". The added cases cover all three cardinality slots.

The same problem exists outside the SHACL generator. `SchemaView.induced_slot` uses the same test when copying inherited values, so a child slot cannot override an inherited `required: true` with `required: false`, and a parent's `minimum_value: 0` is lost entirely. I have written that up as 

- https://github.com/linkml/linkml/issues/3845

`pydanticgen/array.py` tests dimension cardinalities the same way in six places, which may be worth a look

amc-corey-cox (approved):
Following @turbomam lead - this looks good to me, too.

PR #3485, opened 2026-05-07
fix(shaclgen): emit sh:pattern for pattern constraints inside any_of

jdsika:

## Summary

The SHACL generator (`shaclgen.py`) currently ignores `pattern` constraints specified inside `any_of` branches. When a slot uses `any_of` with a branch that has `pattern:` (with or without `range:`), the generated SHACL produces an empty blank node `[ ]` instead of `[ sh:pattern "..." ]`.

This means any value trivially satisfies the branch (an empty node has no constraints), making the `sh:or` validation too permissive.

## Root Cause

In `shaclgen.py` (the `for any in s.any_of:` loop, ~L325–373), branch translation dispatches solely on `any.range`:
- If range maps to a class → `sh:class`
- If range maps to a type → `sh:datatype`  
- If range maps to an enum → `sh:in`
- Else → empty BNode (via `add_simple_data_type(func, None)`)

The loop **never reads `any.pattern`** from the `AnonymousSlotExpression`, so pattern-only branches (no range) produce empty constraint nodes, and range+pattern branches lose the pattern constraint.

Note: the top-level `s.pattern` check (~L394) sits in the `else:` arm of `if s.any_of:`, so it only runs when `any_of` is absent and cannot cover patterns inside branches.

## Fix

After the if/elif/else dispatch that creates each branch BNode, add:

```python
if any.pattern:
    g.add((range_list[-1], SH.pattern, Literal(any.pattern)))
```

This correctly handles all three cases:
1. **Pattern-only branch** (no range): empty node gets `sh:pattern` → `[ sh:pattern "^..." ]`
2. **Range + pattern**: node gets both `sh:datatype` and `sh:pattern` → `[ sh:datatype xsd:string ; sh:pattern "^..." ]`  
3. **Range-only** (no pattern): no-op — existing behavior preserved

This mirrors the existing top-level handling (`if s.pattern: prop_pv(SH.pattern, Literal(s.pattern))`), so both paths emit `sh:pattern` consistently.

## Test Coverage

A 3-class test schema (`tests/linkml/test_generators/input/shaclgen/any_of_pattern.yaml`) exercises:
- `PatternOnlyBranch`: enum + URI + pattern-only (no range) in `any_of`
- `RangeWithPattern`: `range: string` + `pattern` on same branch
- `MixedBranches`: combination of range-only, pattern-only, and range+pattern branches

Two tests consume it:
- `test_any_of_with_pattern()` — asserts on the generated RDF triples (which branch node carries `sh:pattern`, and that range-only branches do not).
- `test_any_of_with_pattern_pyshacl_end_to_end()` — behavioural guard: runs `pyshacl` over conforming and non-conforming instances for all three classes. Without the fix a pattern-only branch serialises as an empty shape `[ ]`, which every value node trivially satisfies, so `sh:or` accepts anything and the non-conforming assertions fail.

Both tests were verified to fail on `main` without the one-line change and pass with it. All existing tests continue to pass.

## Real-World Use Case

This fix is needed for **SPDX license validation** in the [Gaia-X / ENVITED-X](https://gaia-x.eu/) ecosystem. The `gx:license` slot uses `any_of` to accept:
1. Known SPDX identifiers (enum)
2. Arbitrary URIs (`range: uri`)
3. Custom license refs matching `^LicenseRef-[a-zA-Z0-9\-\.]+$` (pattern-only branch)

Without this fix, branch (3) generates an empty SHACL node that accepts *anything*, making validation meaningless. The corresponding upstream model change is tracked in [service-characteristics fix/spdx-license-ref-pattern](https://gitlab.com/ascs-ev/service-characteristics/-/merge_requests).

## Specification References

- [SHACL W3C Rec §4.6.2 (sh:pattern)](https://www.w3.org/TR/shacl/#PatternConstraintComponent): defines pattern as a string constraint on value nodes
- [LinkML Spec §5.6.4 (pattern)](https://linkml.io/linkml-model/latest/docs/pattern/): `The ordinal string value of the slot must match the given pattern`
- [LinkML Spec §5.6.14 (any_of)](https://linkml.io/linkml-model/latest/docs/any_of/): `holds if at least one of the expressions hold`

`sh:pattern` is a value-node string constraint, so it is legal (and idiomatic) on the branch shapes inside `sh:or`; for literals it matches the lexical form, for IRIs the IRI string.

---

*This fix was developed and validated using the Gaia-X credential model as a real-world test case. AI-assisted development (GitHub Copilot).*

### DIFF
--- a/packages/linkml/src/linkml/generators/shaclgen.py
+++ b/packages/linkml/src/linkml/generators/shaclgen.py
@@ -343,6 +343,11 @@ class ShaclGenerator(Generator):
 
                             add_simple_data_type(st_node_pv, r)
                             range_list.append(st_node)
+                        # Propagate pattern constraint to the branch node.
+                        # A branch may combine range + pattern (e.g. range: string
+                        # with pattern: "^...") or specify pattern alone (no range).
+                        if any.pattern:
+                            g.add((range_list[-1], SH.pattern, Literal(any.pattern)))
                     Collection(g, or_node, range_list)
                 else:
                     prop_pv_literal(SH.hasValue, s.equals_number)
--- a/packages/linkml/src/linkml/generators/shaclgen.py
+++ b/packages/linkml/src/linkml/generators/shaclgen.py
@@ -359,6 +359,11 @@ class ShaclGenerator(Generator):
 
                             add_simple_data_type(st_node_pv, r)
                             range_list.append(st_node)
+                        # Propagate pattern constraint to the branch node.
+                        # A branch may combine range + pattern (e.g. range: string
+                        # with pattern: "^...") or specify pattern alone (no range).
+                        if any.pattern:
+                            g.add((range_list[-1], SH.pattern, Literal(any.pattern)))
                     Collection(g, or_node, range_list)
                 else:
                     prop_pv_literal(SH.hasValue, s.equals_number)
--- a/tests/linkml/test_generators/input/shaclgen/any_of_pattern.yaml
+++ b/tests/linkml/test_generators/input/shaclgen/any_of_pattern.yaml
@@ -1,0 +1,59 @@
+id: https://w3id.org/linkml/examples/any_of_pattern
+name: test_any_of_pattern
+description: >-
+  Test schema for pattern constraints inside any_of branches.
+  Exercises three cases: (1) pattern-only branch (no range),
+  (2) range + pattern on the same branch, (3) mixed branches
+  where some have pattern and some do not.
+prefixes:
+  linkml: https://w3id.org/linkml/
+  ex: https://w3id.org/linkml/examples/any_of_pattern/
+imports:
+  - linkml:types
+default_range: string
+default_prefix: ex
+
+enums:
+  LicenseEnum:
+    permissible_values:
+      MIT:
+      Apache-2.0:
+      GPL-3.0-only:
+
+classes:
+  PatternOnlyBranch:
+    description: >-
+      A class where one any_of branch specifies only a pattern
+      (no range). The generated SHACL sh:or should contain a
+      node with sh:pattern but no sh:datatype or sh:class.
+    attributes:
+      license:
+        any_of:
+          - range: LicenseEnum
+          - range: uri
+          - pattern: "^LicenseRef-[a-zA-Z0-9\\-\\.]+$"
+
+  RangeWithPattern:
+    description: >-
+      A class where an any_of branch combines range + pattern.
+      The generated SHACL sh:or node should have both sh:datatype
+      and sh:pattern.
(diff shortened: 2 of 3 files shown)


amc-corey-cox (approved):

amc-corey-cox:
@jdsika or @rmessaou - sorry for the delay on this one. I've approved it - if you can resolve the conflict and merge main this will auto-merge. Thanks for the good work!

jdsika:
@amc-corey-cox rebased onto current `main` (`ddb7a28`) — the merge conflict is resolved and the PR is `MERGEABLE` again. Head is now `0c334af`.

**Conflict resolution.** `shaclgen.py` merged cleanly; the only conflict was in `test_shaclgen.py`, and it was purely additive — `main` had appended the `--emit-rules` / `sh:sparql` test block at EOF while this branch appended its own test there. Both sides are kept in full, nothing was dropped or reworded.

**Re-validated against current `main`:**

- The one-line propagation now sits at `shaclgen.py` L362–366 and mirrors the existing top-level handling at L394–395 (`if s.pattern: prop_pv(SH.pattern, Literal(s.pattern))`), so the `any_of` path and the plain path emit `sh:pattern` identically.
- Spec-wise this is a plain [`sh:pattern`](https://www.w3.org/TR/shacl/#PatternConstraintComponent) value-node string constraint on a branch shape inside `sh:or` — legal and idiomatic; literals match on lexical form, IRIs on the IRI string.
- Guarded by `if any.pattern:`, so range-only branches are bit-for-bit unchanged.

**Test coverage strengthened.** In addition to the existing triple-level assertions I added `test_any_of_with_pattern_pyshacl_end_to_end`, which runs `pyshacl` over conforming and non-conforming instances for all three classes. This is the behavioural guard the fix deserves: without the change, the pattern-only branch serialises as an empty shape `[ ]`, which every value node trivially satisfies, so `"NotALicenseRef"` *conforms*. I confirmed both tests fail on `main` without the one-line change and pass with it.

**Independent real-world regression check.** I ran the patched generator over a production LinkML model outside this repo (~1,300 lines, 10 classes, 60+ slots, closed node shapes, rules, `--message-template`, `--default-language`). The emitted SHACL is **byte-identical** with and without the patch, and the generated shapes still accept all 7 gold-standard instances and reject all 9 negative fixtures — confirming the change is a strict no-op for schemas that do not use `pattern` inside `any_of`.

**Local verification:** `tests/linkml/test_generators` → **1644 passed, 48 skipped, 3 xfailed**; `ruff check` and `ruff format --check` clean.

The PR description has been updated so the line references and the test-coverage section match the rebased code. Ready to merge from my side — thanks for the review!

PR #3490, opened 2026-05-07
fix: respect id/key slot_usage required=false overrides #3489

### LINKED ISSUE #3489: json-schema: inherited identifier cannot be suppressed (required: false)
**Describe the bug**

A slot marked `identifier: true` is unconditionally added to the `required` array of every class that inherits it.  

Overriding the child slot with `slot_usage: { required: false }` alone does not suppress the slot.  The identifier (`id`) still appears in `required` unless `identifier: false` is also set.

1.10.0

**Reproducer**
CUT +PASTE into your terminal / virtualenv:
```bash
# 1) minimal schema
cat > /tmp/schema2.yaml << 'YAML'
id: https://example.org/test
name: test
prefixes:
imports:
  - linkml:types

classes:
  NamedThing:
    abstract: true
    slots:
      - id

  Person:
    is_a: NamedThing
    tree_root: true
    slots:
      - name
    slot_usage:
      id:
        required: false        # does not suppress
        # identifier: false     # need this to force suspression

slots:
  id:
    identifier: true
    range: string

  name:
    range: string
YAML

# 2) Generate JSON Schema and inspect required fields for Person
uv run gen-json-schema /tmp/schema2.yaml | python3 -c '
import json, sys
s = json.load(sys.stdin)
print("required:", s["$defs"]["Person"].get("required"))
'
```

**Actual output:**
```
required: ['id']
```

**Expected output:**
```
required: None
```

noelmcloughlin:
## Summary

Fixes #3489 

## How was this tested?

Ran local full tests
Added 2 unit test

## Checklist

## AI Assistance
LLM assisted

### DIFF
--- a/packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
+++ b/packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
@@ -1718,7 +1718,15 @@ class SchemaView:
                 setattr(induced_slot, metaslot_name, v)
         if induced_slot.inlined_as_list:
             induced_slot.inlined = True
-        if induced_slot.identifier or induced_slot.key:
+        # Check if current class explicitly overrides 'required' in slot_usage
+        has_explicit_required_override = False
+        if cls is not None and cls.slot_usage and slot_name in cls.slot_usage:
+            slot_usage_def = cls.slot_usage[slot_name]
+            # Check if 'required' is explicitly set in this slot_usage
+            if hasattr(slot_usage_def, "required") and slot_usage_def.required is not None:
+                has_explicit_required_override = True
+        # Only force required=True for identifiers if not explicitly overridden in slot_usage
+        if (induced_slot.identifier or induced_slot.key) and not has_explicit_required_override:
             induced_slot.required = True
         if mangle_name:
             mangled_name = f"{camelcase(class_name)}__{underscore(slot_name)}"
--- a/tests/linkml_runtime/test_utils/test_schemaview.py
+++ b/tests/linkml_runtime/test_utils/test_schemaview.py
@@ -3472,3 +3472,93 @@ def test_induced_slot_domain_of_no_duplicates() -> None:
 
     induced = view.induced_slot("s1", "A")
     assert induced.domain_of.count("A") == 1
+
+
+@pytest.mark.parametrize(
+    ("class_name", "expected_required"),
+    [
+        ("NamedThing", True),  # parent: identifier always required
+        ("ChildNoOverride", True),  # no slot_usage override: still required
+        ("ChildRequiredFalse", False),  # required: false in slot_usage: optional
+        ("ChildRequiredTrue", True),  # required: true in slot_usage: still required
+    ],
+)
+def test_identifier_slot_required_override(class_name: str, expected_required: bool) -> None:
+    """Test that identifier slots respect required: false in slot_usage overrides.
+
+    A child class should be able to override required=true for an inherited
+    identifier slot by setting required: false in its slot_usage.
+    """
+    schema_str = """
+id: https://example.org/test
+name: test
+prefixes:
+  linkml: https://w3id.org/linkml/
+imports:
+  - linkml:types
+classes:
+  NamedThing:
+    abstract: true
+    slots:
+      - id
+  ChildNoOverride:
+    is_a: NamedThing
+  ChildRequiredFalse:
+    is_a: NamedThing
+    slot_usage:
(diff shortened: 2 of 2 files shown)


packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
```diff
@@ -1718,7 +1718,15 @@ def induced_slot(
...
-        if induced_slot.identifier or induced_slot.key:
+        # Check if current class explicitly overrides 'required' in slot_usage
+        has_explicit_required_override = False
+        if cls is not None and cls.slot_usage and slot_name in cls.slot_usage:
+            slot_usage_def = cls.slot_usage[slot_name]
+            # Check if 'required' is explicitly set in this slot_usage
+            if hasattr(slot_usage_def, "required") and slot_usage_def.required is not None:
+                has_explicit_required_override = True
+        # Only force required=True for identifiers if not explicitly overridden in slot_usage
+        if (induced_slot.identifier or induced_slot.key) and not has_explicit_required_override:
```
amc-corey-cox, line 1729: I don't think this resolves #3489's reproducer. The issue is JSON-Schema output (`gen-json-schema` -> `required: ['id']`), but the generator force-requires the identifier independently of `induced_slot.required`, at `jsonschemagen.py:895`: `value_required = slot.required or slot == class_id_slot or ...`. I ran the issue's exact reproducer on this branch and it still emits `required: ['id']`, so this change flips the runtime value but never reaches the symptom. The fix also needs to reach that generator line.

packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
```diff
@@ -1718,7 +1718,15 @@ def induced_slot(
                 setattr(induced_slot, metaslot_name, v)
         if induced_slot.inlined_as_list:
             induced_slot.inlined = True
-        if induced_slot.identifier or induced_slot.key:
+        # Check if current class explicitly overrides 'required' in slot_usage
+        has_explicit_required_override = False
```
amc-corey-cox, line 1722: This detects the override only on the directly-queried class (`cls.slot_usage`), so inherited overrides are missed. With `NamedThing -> Child (slot_usage id.required: false) -> GrandChild`, I get `Child.id.required = False` but `GrandChild.id.required = True` -- the inherited override is clobbered back.

Since `induced_slot.required` has already merged ancestor + self `slot_usage` by this point (the loop above), a simpler and inheritance-correct guard avoids re-walking `slot_usage` entirely:

```python
if (induced_slot.identifier or induced_slot.key) and induced_slot.required is None:
    induced_slot.required = True
```

Minor: the `hasattr(slot_usage_def, 'required')` check is redundant -- `SlotDefinition` always has a `required` attribute.

tests/linkml_runtime/test_utils/test_schemaview.py
```diff
@@ -3472,3 +3472,93 @@ def test_induced_slot_domain_of_no_duplicates() -> None:
...
+@pytest.mark.parametrize(
+    ("class_name", "expected_required"),
+    [
+        ("NamedThing", True),  # parent: identifier always required
+        ("ChildNoOverride", True),  # no slot_usage override: still required
+        ("ChildRequiredFalse", False),  # required: false in slot_usage: optional
+        ("ChildRequiredTrue", True),  # required: true in slot_usage: still required
+    ],
+)
+def test_identifier_slot_required_override(class_name: str, expected_required: bool) -> None:
```
amc-corey-cox, line 3486: These assert `induced_slot(...).required` (the runtime value), not the JSON Schema output the issue is actually about. The #3489 reproducer (`gen-json-schema` -> `required: ['id']`, expected `None`) isn't covered, and it still fails on this branch (see the schemaview.py comment). Worth adding a test that mirrors the issue: generate JSON Schema and assert the identifier is absent from `required`. A grandchild case (inherited `slot_usage` override) would also catch the regression noted above.

amc-corey-cox (changes requested):
Doesn't yet fix #3489 — the JSON Schema output is unchanged (verified: still required: ['id']); the identifier is force-required in jsonschemagen.py:895, not via the induced_slot value this changes. Also regresses on inherited slot_usage overrides. Details inline.

##### Note - I had Claude help me with this review and run the tests on the JSON Schema output. I stand by it's output but if you disagree with the analysis please push back and show me how it is wrong.

noelmcloughlin:
Cancelling due to feedback in #3489

PR #3505, opened 2026-05-10
fix(enumerations): ensure enum constructor idempotency in certain contexts

### LINKED ISSUE #3497: enumerations.py is not idempotent on enum instances
**Describe the bug**

Any LinkML Runtime class whose `slot_usage` refines a slot's range to a **subclass enum** produces a `__post_init__` that crashes the moment that slot is populated.

*Cause*

1. `__init__` handles `PermissibleValue` and `Curie` arguments but not existing `EnumDefinitionImpl` subclass instances as shown below.
https://github.com/linkml/linkml/blob/7fad7d245518fa71512a96dcdc4faa3298324bc9/packages/linkml_runtime/src/linkml_runtime/utils/enumerations.py#L37-L47

2. So execution falls through to `key not in self.__class__` ...
https://github.com/linkml/linkml/blob/7fad7d245518fa71512a96dcdc4faa3298324bc9/packages/linkml_runtime/src/linkml_runtime/utils/enumerations.py#L28-L30

This membership check requires `item` to be hashable but dataclass generated *enum subclasses* are not (?) causing `TypeError`.  

From https://docs.python.org/3/library/dataclasses.html#module-contents 
```
   ... etc ...
If eq and frozen are both true, by default @dataclass will generate a __hash__() method
for you. If eq is true and frozen is false, __hash__() will be set to None, marking it 
unhashable (which it is, since it is mutable).
   ... etc ...
```
But  `Frozen=True` is not used for python dataclasses:
https://github.com/search?q=repo%3Alinkml%2Flinkml%20frozen%3DTrue&type=code

**Impact**
This bug is related to #3506 causing pytest failures:
```
```

1.10.0

**Reproducer**
Sample Project:
```bash
git clone https://github.com/lmodel/ocsf.git
cd ocsf/
uv sync
```
*REPL*
```python
from ocsf.datamodel.ocsf import AuthenticationActivityIdEnum, ActivityIdEnum
inst = AuthenticationActivityIdEnum("LOGON")
ActivityIdEnum(inst)
# TypeError: cannot use 'ocsf.datamodel.ocsf.AuthenticationActivityIdEnum'
# as a dict key (unhashable type: 'AuthenticationActivityIdEnum')
```

noelmcloughlin:
## Summary

Fixes #3497 

treat an existing `EnumDefinitionImpl` instance as a no-op fast path (or extract/re-resolve), so generated code is robust to the duplicate cast patterns.

## How was this tested?

Workarounds
Unit Tests / local tests

## Checklist

## AI Assistance

LLM assisted unit tests


amc-corey-cox (approved):
This looks good to me!

PR #3507, opened 2026-05-10
fix(schemaloader):  deep slot_usage chains, mangled name-swap orphans

### LINKED ISSUE #3506: SchemaLoader leaves orphan parent slots (process_slot_usages)
**Describe the bug**

When a class declares `slot_usage`, the loader synthesises a new mangled slot (e.g. `Authentication_activity_id`) and intends to replace the bare slot name in `cls.slots`. 
https://github.com/linkml/linkml/blob/1c5f68e43a6960ec1066521f0d5bb112cebde21a/packages/linkml/src/linkml/utils/schemaloader.py#L931-L932

For a deep inheritance chain (e.g. `Object ->  ... ->  BaseEvent ->  IamEvent ->  Authentication`), the proximal parent's slot was *itself* mangled at a higher level (it's `BaseEvent_activity_id`, not `activity_id`), so the lookup misses and the bare `activity_id` is never removed.  The loaded schema can end up with **two slot entries pointing at the same Python attribute, with two different ranges**.

**Impact**
This bug is related to #3497 causing pytest failures:
```
```

1.10.0

**Reproducer**
```bash
git clone https://github.com/ocsf/ocsf-schema.git
cd ocsf-schema
uv sync
```
*REPL*
```python
from linkml.utils.schemaloader import SchemaLoader
loader = SchemaLoader("src/ocsf/schema/ocsf.yaml")
loader.resolve()
cls = loader.schema.classes["Authentication"]
[s for s in cls.slots if "activity" in s]
# ['activity_id', 'Authentication_activity_id']  <-- duplicate!
```

noelmcloughlin:
## Summary

Fixes #3506

- In `process_slot_usages`, also probe for the *root* slot name (and any intermediate mangled names from ancestors) when deciding whether to replace an entry in `cls.slots` - avoiding orphan slot names.

## How was this tested?

- New unit tests
- Bioloink and graphql tests had bug (regenerated snapshots)

## Areas of uncertainty

Comments added to help reviewers.

## Checklist

## AI Assistance

LLM Assisted


kevinschaper (approved):
good find!  thanks @noelmcloughlin!

PR #3509, opened 2026-05-10
fix(pythongen): dedupe domain_slots by python field name (alias)

### LINKED ISSUE #3508: pythongen should guard prevent two cast blocks in __post_init__
**Describe the bug**
`gen_postinits` iterates `self.domain_slots(cls)` twice, emitting an `isinstance` + cast block for each. Combined with Bug #3506, two cast blocks end up in`__post_init__` referring to the same Python attribute - both `Dog_sound` (required, range `DogSoundEnum`) AND the orphan `sound` (optional, range `AnimalSoundEnum`) flow through.

```python
# (block 1 — required, from slot_usage)
if not isinstance(self.sound, DogSoundEnum):
    self.sound = DogSoundEnum(self.sound)
# ...
# (block 2 — optional, from the orphan parent slot)
if self.sound is not None and not isinstance(self.sound, AnimalSoundEnum):
    self.sound = AnimalSoundEnum(self.sound)
```
Block 2's `isinstance(..., AnimalSoundEnum)` is `False` (subclass enums don't inherit from the parent enum class), so it re-enters the cast. When #3506 is fixed, `domain_slots` no longer returns the duplicate; the second cast block disappears. 

But **as a defence-in-depth measure,** `gen_postinits` could de-duplicate by aliased Python field name before  emitting.

**Version**
1.10.0

**Reproducer**
See #3506 reproducer.

**Impact**
This bug is related to #3506 and #3497 causing pytest failures:
```
```

noelmcloughlin:
## Summary

Fixes #3508

- As a defence-in-depth measure, `gen_postinits` can de-duplicate by aliased Python field name before emitting.
- The full list from "domain_slots(cls)" is now de-duplicated by Python field name (self.slot_name(slot.name), resolving alias.

## How was this tested?

New unit tests

## Areas of uncertainty

This PR may be overkill - #3507 may be enough.

## Checklist

## AI Assistance
LLM assistance

### DIFF
--- a/packages/linkml/src/linkml/generators/pythongen.py
+++ b/packages/linkml/src/linkml/generators/pythongen.py
@@ -798,12 +798,35 @@ version = {'"' + self.schema.version + '"' if self.schema.version else None}
                     post_inits.append(self.gen_postinit(cls, slot))
         else:
             pkeys = []
+
+        # De-duplicate domain_slots by Python field name to prevent emitting two
+        # cast blocks for the same attribute.  This can happen when an ancestor
+        # leaves an orphan bare slot in cls.slots alongside the class-scoped
+        # mangled slot (#3506).  Both resolve to the same Python field name via
+        # slot_name() / alias.  When two slots share a field name, the required
+        # one (i.e. the class-scoped slot_usage refinement) takes priority over
+        # the optional orphan so we keep the most-specific cast.
+        best_slot_for_field: dict[str, object] = {}
         for slot in self.domain_slots(cls):
+            field_name = self.slot_name(slot.name)
+            existing = best_slot_for_field.get(field_name)
+            if existing is None or (slot.required and not existing.required):
+                best_slot_for_field[field_name] = slot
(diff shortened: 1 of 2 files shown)


amc-corey-cox:
Thanks for this! I dug in a bit — #3507 fixes the root cause upstream, so orphan slots no longer reach `gen_postinits` and `domain_slots` doesn't return duplicates anymore. I confirmed the new regression test passes even with the `pythongen.py` change reverted, which means it's exercising #3507's fix rather than the dedup here. Given that, I think #3507 is enough and I'll close this.

If you hit a case where duplicates still slip through, please reopen — it would need a test that actually forces a duplicate through `domain_slots` (e.g. via monkeypatch) to exercise the dedup path, since the current one doesn't. And I'd lean toward having the generator *warn/raise* on a field-name collision rather than silently pick a winner, so we surface the underlying bug instead of masking it. Really appreciate the defence-in-depth instinct here.

PR #3512, opened 2026-05-13
Openapi 3.1

Silvanoc:
## Summary

Adds support for OpenAPI v3.1.0 to the OpenAPI generator.

It also fixes the declared `valid_formats`, what may break downstream code it any would be relying on it. But it should not be an issue, since no downstream code should be relying on it yet.

## How was this tested?

New specific tests added.

## Areas of uncertainty

Use of Pydantic and tests.

## Checklist

## AI Assistance


Silvanoc:
@noelmcloughlin sorry, tests are not yet ready.  Therefore I'm moving this PR back to `draft`. `openapigen.py` should be ready though, just in case you want to have a preview at it.

Silvanoc:
I realize that:

1. We need a better way to get variability into the LinkML schemas and OpenAPI templates being used for the tests.
2. We have already too many LinkML schemas and OpenAPI templates for the tests. Adding more for OpenAPI v3.1.0 will make the whole directory too messy IMO.

Since many of the LinkML schemas and OpenAPI templates are very small and/or very similar, I'm considering the option of embedding/inlining them into `test_openapigen.py`. @noelmcloughlin any strong opinion on it? If not, I'll provide a PR with a proposal.

noelmcloughlin:
Just a (hopefully) quick and easy question.  So from https://openapi-spec-validator.readthedocs.io/en/latest/

Is the intent to support all three [OpenAPI 2.0 (aka Swagger)](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md), [OpenAPI 3.0](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md) [OpenAPI 3.1](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md) and [OpenAPI 3.2](https://spec.openapis.org/oas/v3.2.0.html). Does the spec have development releases so 3.0.2 is dev only?

What happens when it increments - main impact is regression tests (v3.0.5 schema + template)?

Will the validator support 3.0.1 and 3.0.3 and (future) 3.0.4 releases / or they do some kind of n-2 support lifecycle so latest semver version, plus previous, and one before that? 

`from openapi_spec_validator import OpenAPIV30SpecValidator, OpenAPIV31SpecValidator, validate`

Silvanoc:
> Is the intent to support all three [OpenAPI 2.0 (aka Swagger)](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md), [OpenAPI 3.0](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md) [OpenAPI 3.1](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md) and [OpenAPI 3.2](https://spec.openapis.org/oas/v3.2.0.html). Does the spec have development releases so 3.0.2 is dev only?

Not my intention. My intention is to add coverage for what is needed. Initially I only needed v3.0.3 and therefore I started with it. My project moved to v3.1.0, so I now need support for it. The first commit of this PR should make the addition of new versions easier than before. But I cannot say anything about the effort required to cover other released versions and I won't try it as of now.

> What happens when it increments - main impact is regression tests (v3.0.5 schema + template)?

I personally have seen a lot of v3.0.3 and v3.1.0, but not many APIs using other versions. Therefore I would consider the increments as they are requested. I wouldn't try to pro-actively tackle them.

> Will the validator support 3.0.1 and 3.0.3 and (future) 3.0.4 releases / or they do some kind of n-2 support lifecycle so latest semver version, plus previous, and one before that?
> 
> `from openapi_spec_validator import OpenAPIV30SpecValidator, OpenAPIV31SpecValidator, validate`

It's an external dependency and I would say that we don't try to support more than the validator supports. I wouldn't try either to add support to the validator for a version someone asks us to support, but the validator doesn't.

noelmcloughlin:
Okay thanks.  I was just curious to get a feel for the versioning increments - but sounds good to me.

Issue #3516, opened 2026-05-14
Generator-level nondeterminism leaks past RDFC-1.0 canonicalization


amc-corey-cox:
While verifying #3407 I found three sources of process-level nondeterminism that survive `canonicalize_rdf_graph` because they're upstream of it. Not blockers for #3407 — pre-existing on `main`, just newly visible.

1. **SHACL** — `_build_ignored_properties` (`shaclgen.py:476`) emits a Python `set()` as an `rdf:first`/`rdf:rest` list, so element order varies with `PYTHONHASHSEED`. The graphs are not RDF-isomorphic across runs, so RDFC-1.0 can't normalize them. Fix is one-line: sort before emitting.

2. **RDFGen** — injects `linkml:generation_date` timestamp; JSON-LD parse step also picks up dict-ordering noise.

3. **Metamodel OWL** — `bibo:status testing` annotations serialize to scheme-less `<testing>` IRIs. pyoxigraph rejects → `canonicalize_rdf_graph` falls back to rdflib (non-deterministic). 32 such triples in the metamodel.

Each reproduces by running the relevant generator 3× in fresh processes and diffing.

davieshq:
Confirming all 3 leak paths reproduce at LinkML 1.11.0 stable (PyPI release 2026-05-13) across 6 schemas with varied topology (simple classes, abstract+leaves, role-bearers, state-machines, cross-module-references) — Python 3.12.x on Ubuntu 24.04/WSL2.

One empirical datapoint that might refine bullet #1: **`PYTHONHASHSEED=0 gen-shacl <schema>` does NOT mitigate the non-determinism** in our tests (3-run SHA-256 still shows 3 distinct hashes, even with the env-var exported in a fresh process before each invocation). Possible explanations:

- Non-determinism source is downstream of Python's `hash()` randomisation — e.g., rdflib's blank-node naming, or pyoxigraph fallback dict ordering when canonicalisation is skipped
- Env var isn't propagating into the rdflib/pyoxigraph subprocess context
- Additional non-determinism source in `_build_ignored_properties` beyond the bare `set()` iteration

If the one-line `sort()` at `shaclgen.py:476` doesn't fully resolve gen-shacl in isolation when applied, this might explain why. Happy to share full reproducer + 18 SHA-256 hashes (2 schemas × 3 generators × 3 runs) if useful.

[...]

PR #3518, opened 2026-05-14
fix(shaclgen): sort sh:ignoredProperties for cross-process determinism

amc-corey-cox:
Fixes finding 1 of #3516.

`ShaclGenerator._build_ignored_properties` iterated a Python `set()` into an `rdf:first`/`rdf:rest` list, so element order varied with `PYTHONHASHSEED`. The resulting graphs were not RDF-isomorphic across processes, and RDFC-1.0 canonicalization (added in #3407) cannot normalize that because list element order is graph content.

Verified: with this fix, SHACL output for `personinfo` and `biolink-model` is byte-identical across three fresh Python processes.

The other two findings in #3516 (RDFGen timestamp, metamodel `<testing>` IRI fallback) remain open.


noelmcloughlin:
Hi @amc-corey-cox   Thanks for the PR, the explanation and implementation looks more deterministic, worth doing.

One suggestion - check we have sufficient use cases to test "rdf-isomorphism":

- See criteria here [graph-isomorphism](https://w3c.github.io/rdf-concepts/spec/#graph-isomorphism)
- And concrete [RDF-isomorphic test case - section 2.2](https://www.w3.org/TR/rdf11-testcases/)  ----> scroll to "2.2 Evaluation Tests"

noelmcloughlin:
A related aside regarding non-determinism...  there is a project called [reproducible builds](https://reproducible-builds.org/) that offers guardrails and ideas.

I am hitting "non-determinism" issues with some LinkML generators:

- non-deterministic output / timestamps breaks quality gates  (maybe https://github.com/linkml/linkml/issues/3516 is related)
- Multiple outputs instead of a single tarball/zipball (persistent Identifier needs singleton, see [likml .htaccess example](https://github.com/perma-id/w3id.org/blob/master/linkml/.htaccess)).

I have not raised issue - it feels a wider concern (linkml goverance, linkml-project-copier, SSSOM integration, PIDs).

noelmcloughlin (approved):
LGTM - can be merged if unit tests capture rdf-isomorphism

PR #3523, opened 2026-05-15
fix(typedbgen): use class_induced_slots for slot iteration (#3519)

### LINKED ISSUE #3519: gen-typedb: inline `attributes:` are declared but not owned
---
name: Bug report
about: gen-typedb drops `owns` for classes defined with inline `attributes:`
title: "gen-typedb: inline `attributes:` are declared but not owned"
labels: bug
---

**Describe the bug**

`gen-typedb` (the `linkml.generators.typedbgen.TypeDBGenerator`) correctly emits
`attribute <name>, value <type>;` declarations for slots defined inline via
`attributes:` on a class, but the resulting entity block contains **no `owns`
clauses** for those attributes. The same generator handles classes whose
attributes are declared via top-level `slots:` references correctly.

Because the entity does not own the attribute type, TypeDB rejects any
`insert` that tries to attach the attribute to the entity:

```
[INF11] Type-inference was unable to find compatible types
```

The inline-`attributes:` pattern is very common in real LinkML schemas — for
example, the [dismech](https://w3id.org/monarch-initiative/dismech) schema uses
it for `ProteinStructure`, `DiseaseClassifications`, and others — so any
non-trivial schema is likely affected.

`linkml==1.11.0` (generator reports itself as `linkml-typedb-generator v0.1.0`).
Reproduced on `main` as of this report.

Minimal repro schema:

```yaml
id: https://example.org/gen-typedb-attributes-bug
name: gen_typedb_attributes_bug
description: >-
  Minimal LinkML schema demonstrating that gen-typedb drops attribute
  ownership for classes that define their attributes inline via `attributes:`
  rather than top-level `slots:`. Both classes below should produce identical
  `owns` declarations in the generated TypeQL.

prefixes:
  ex: https://example.org/gen-typedb-attributes-bug/

default_prefix: ex
default_range: string

imports:
  - linkml:types

slots:
  shared_name:
    range: string
    required: true
  shared_description:
    range: string

classes:
  ClassWithSlots:
    description: Attributes declared via top-level `slots:` reference.
    slots:
      - shared_name
      - shared_description

  ClassWithAttributes:
    description: Attributes declared inline. gen-typedb currently drops these.
    attributes:
      inline_name:
        range: string
        required: true
      inline_description:
        range: string
```

Run:

```bash
uvx --from 'linkml>=1.11.0' gen-typedb schema.yaml
```

**Actual output**

```typeql
define

  # Attribute types
  attribute shared-name, value string;
  attribute shared-description, value string;
  attribute inline-name, value string;          # declared
  attribute inline-description, value string;   # declared

  # Entity types
  entity classwithslots,
      owns shared-name @card(1..1),             # present
      owns shared-description;                  # present

  entity classwithattributes;                   # BUG: no owns clauses
```

**Expected output**

```typeql
  entity classwithattributes,
      owns inline-name @card(1..1),
      owns inline-description;
```

**Likely cause**

`TypeDBGenerator._entity_and_relation_defs` (and the related helpers around
`packages/linkml/src/linkml/generators/typedbgen.py:400-470`) iterates the
"direct slot names" of a class via `sv.get_class(class_name).slots`, which is
empty for `attributes:`-only classes. The attribute *type* declarations earlier
in the generator use `sv.class_induced_slots(class_name)` (around line 324),
which is why the `attribute …, value …;` lines are emitted while the matching
`owns …` lines are dropped.

The fix is to source the direct slot list from `class_induced_slots(...)` (or
equivalently, the union of `class.slots` and `class.attributes`) when computing
`direct_slot_names` for `owns`/`relates`/`plays` emission, while continuing to
subtract ancestor slots.

**Pytest repro**

A self-contained pytest file (also included in this issue) reproduces the bug
against any installed `linkml`:

```python
# test_typedb_inline_attributes.py
import re
import subprocess
import sys
from pathlib import Path

import pytest

SCHEMA = Path(__file__).parent / "schema.yaml"


def generated_tql() -> str:
    result = subprocess.run(
        [sys.executable, "-m", "linkml.generators.typedbgen", str(SCHEMA)],
        capture_output=True, text=True, check=True,
    )
    return result.stdout


def _entity_block(tql: str, name: str) -> str:
    m = re.search(rf"^\s*entity {re.escape(name)}\b[^;]*;", tql, re.MULTILINE | re.DOTALL)
    assert m, f"entity {name!r} not found in:\n{tql}"
    return m.group(0)


def test_attributes_are_declared(generated_tql):
    for name in ("shared-name", "shared-description", "inline-name", "inline-description"):
        assert f"attribute {name}" in generated_tql


def test_class_with_slots_owns_its_attrs(generated_tql):
    block = _entity_block(generated_tql, "classwithslots")
    assert "owns shared-name" in block
    assert "owns shared-description" in block


def test_class_with_attributes_owns_inline_attrs(generated_tql):
    block = _entity_block(generated_tql, "classwithattributes")
    assert "owns inline-name" in block, (
        "Expected `owns inline-name` on classwithattributes, but the entity "
        f"block has no owns declarations at all:\n{block}"
    )
    assert "owns inline-description" in block
```

`test_class_with_attributes_owns_inline_attrs` currently fails on
`linkml==1.11.0`; the other two pass.

kevinschaper:
@GullyBurns I tried out a quick TypeDB experiment for one of my projects and ran into aspects of the schema getting missed when they were defined as class attributes rather than on the top level slot definitions. Using induced_slot lets schemaview handle the attribute vs slot definitions and merge them, and I _think_ the subtraction pattern from rustgen will give the right results. Do you have a few minutes to try this branch out to make sure it doesn't cause any surprises for you? 

## Summary

- Fixes #3519: `gen-typedb` dropped `owns` clauses for classes that declared their slots inline via `attributes:`, leaving the generated TypeQL unusable (any `insert ... has ...` was rejected with `[INF11] Type-inference was unable to find compatible types`).
- Root cause: every "iterate slots on a class" call site in `typedbgen.py` (`_build_name_maps`, `_collect_entity_defs`, `_collect_relation_defs`) used `class.slots` directly, which is empty for `attributes:`-only classes. The attribute *type* declarations were correct because that path already used `class_induced_slots`; only the entity/relation `owns`/`relates`/`plays` emission was broken.
- Fix: switch every direct slot iteration to `sv.class_induced_slots(...)`, matching the universal pattern across pydanticgen, rustgen, dbmlgen, shaclgen, docgen, javagen, jsonschemagen, erdiagramgen, etc. The "direct slots only" requirement (TypeDB inherits capabilities from supertypes, so redeclaring triggers `[SVL42]`) becomes `class_induced_slots(C)` minus the union of `class_induced_slots(ancestor).names` — the same pattern rustgen uses for its poly-trait inheritance subtraction.
- Inline `attributes:` are now handled correctly as a side effect: `class_induced_slots()` is designed to return the union of inherited slots, top-level `slots:` references, and inline `attributes:` declarations.

## Before / after

The included regression fixture (`tests/linkml/test_generators/input/typedb_inline_attributes.yaml`) declares one class via `slots:` and one via `attributes:`. On `main`:

```typeql
entity classwithslots,
    owns shared-name @card(1..1),
    owns shared-description;

entity classwithattributes;     # BUG: no owns clauses
```

After this PR both classes get identical `owns` declarations. The same fix surfaces several previously-missing `owns`/`plays` clauses in `kitchen_sink.yaml` output (e.g. `hasaliases owns aliases @card(0..)`, `fakeclass owns test-attribute`, plus `plays` chains involving `dataset`, `person`, `company`).


GullyBurns:
@kevinschaper Apologies for the delay. I'll take a look asap.

GullyBurns:
This looks great to me. I notice that there's a failing ruff check. Claude tells me that you should run `pre-commit run 
--all-files` and commit the result. 

I should mention that I'm an strong supporter of the typedb system, but am not a core developer (or even a deep expert) with a direct link to the system. I really wanted to facilitate the first implementation of the generator, but feel it would likely be better to directly connect with Joshua Send (TypeDB's CTO), who is now getting involved with the LinkML community for technical matters regarding LinkML support (cc- @flyingsilverfin)

tests/linkml/test_generators/test_typedbgen.py
```diff
@@ -278,6 +280,37 @@ def test_name_collision_attr_gets_suffix(tmp_path):
     assert "owns person-attr" in output
 
 
+def test_inline_attributes_produce_owns(input_path):
```
amc-corey-cox, line 283: This regression test only covers scalar (`string`-ranged) inline attributes, i.e. the `owns` path. But the diff also rewrites the object-ranged emission — `relates`, `plays`, and the first pass of `_collect_relation_defs` — and none of that is exercised here, since the fixture has no class-ranged attribute. That's the half of the change most likely to regress, running with zero coverage.

Could you add a fixture (or extend this one) with an inline `attributes:` entry whose `range:` is a class, and assert the resulting `plays`/`relates` clauses? That would also likely account for the 2 uncovered partials Codecov flags on this patch — they're plausibly on exactly this untested path.

packages/linkml/src/linkml/generators/typedbgen.py
```diff
@@ -501,18 +515,22 @@ def _collect_relation_defs(self, sv: SchemaView, rel_names: dict[str, str]) -> l
...
         # relation types, so they are skipped here.
+        all_class_names = set(sv.all_classes().keys())
         slot_pairs: dict[str, list[tuple[str, str]]] = {}
         for class_name in sv.all_classes():
             if sv.get_class(class_name).represents_relationship:
                 continue  # object-ranged slots handled as 'relates' in entity/relation defs
-            for slot_name in sv.get_class(class_name).slots or []:
-                induced = sv.induced_slot(slot_name, class_name)
-                if induced.range not in sv.all_classes():
+            ancestor_slot_names = self._ancestor_slot_names(sv, class_name)
```
amc-corey-cox: Semantics change worth confirming is intentional: the old `_collect_relation_defs` iterated `class.slots` directly with **no** ancestor subtraction, whereas this now applies `_direct_induced_slots` (which subtracts ancestor slot names). So an object-ranged slot that's redeclared/inherited on a subclass will now be collected only at the class that introduces it, not per-subclass.

This looks correct for TypeDB (roles inherit from supertypes, so the relation should be defined once), but it's a real behavioral change rather than a pure refactor — a one-line confirmation that it's deliberate would be good, especially since the domain expert hasn't reviewed yet.

packages/linkml/src/linkml/generators/typedbgen.py
```diff
@@ -501,18 +515,22 @@ def _collect_relation_defs(self, sv: SchemaView, rel_names: dict[str, str]) -> l
...
+            for induced in self._direct_induced_slots(sv, class_name, ancestor_slot_names):
```
amc-corey-cox: Nit: `_collect_entity_defs` already builds `ancestor_slot_names_cache` and `direct_slots_cache`, but this loop recomputes both via fresh `_ancestor_slot_names` / `_direct_induced_slots` calls per class. `class_induced_slots` isn't free, so hoisting those caches to shared scope would avoid the duplicate traversal. Not important — the generator runs once — purely cleanliness.

amc-corey-cox (changes requested):
@kevinschaper the only thing really necessary is the coverage gap but it would good to confirm the behavioral change. Also, a small nit.

tests/linkml/test_generators/test_typedbgen.py
amc-corey-cox: Addressed in af11a3c1c — the fixture now includes `inline_owner` (range `ClassWithSlots`) and `test_inline_attributes_with_class_range_produce_relation` asserts the `relation` block plus `relates`/`plays` on both entities. The object-ranged path is now covered. Resolving.

packages/linkml/src/linkml/generators/typedbgen.py
amc-corey-cox: Confirmed intentional — the rationale (collect once at the introducing class; per-subclass redeclaration is redundant and rejected by TypeDB [SVL42]) is now documented in the `_collect_relation_defs` docstring. Resolving.

packages/linkml/src/linkml/generators/typedbgen.py
amc-corey-cox: Addressed — `_build_direct_slots_cache` now builds the per-class cache once in `serialize()` and both collectors share it, so the traversal happens once. Resolving.

amc-corey-cox (approved):
This looks good to me now! Thanks!

PR #3524, opened 2026-05-17
Hardening for the pyoxigraph RDF canonicalizer

matentzn:
Follow-ups to the deterministic-RDF-serialization work (`canonicalize_rdf_graph` in `linkml_runtime/utils/rdf_canonicalize.py`). Four small, independently testable fixes plus the corresponding tests.

(I was working on these when https://github.com/linkml/linkml/pull/3407 was open but forgot to set PR to draft again while doing so, sorry)

## Summary

- **Literal-safe CURIE rewriting.** The trailing-escaped-dot expansion (`_expand_trailing_dot_curies`) previously ran a regex over the full Turtle string. A CURIE-shaped substring inside a string literal — e.g. a literal containing `"see ex:other\. for context"` — would false-match and get rewritten into `<...>` form, producing un-parseable Turtle. The rewrite is now driven by `_iter_turtle_structural_spans`, a small tokenizer that yields structural spans only and skips string literals (single- and triple-quoted, both delimiters, with backslash escapes), `<...>` IRI refs, and `#…` line comments.
- **Warnings instead of logging.** The three degraded paths (unsupported format, non-standard RDF fallback, pyoxigraph prefix rejection) used `logging.warning`, which is silent for library API consumers unless logging is explicitly configured. They now emit a new `RDFCanonicalizationWarning` (a `UserWarning` subclass) via `warnings.warn`, so callers see them by default and can filter with `warnings.filterwarnings("ignore", category=RDFCanonicalizationWarning)`.
- **Drop unused `@prefix` declarations.** rdflib's `NamespaceManager` auto-binds ~30 well-known vocabularies (`foaf`, `skos`, `dcterms`, …) on every `Graph`. Without filtering, every canonicalized output was padded with `@prefix` lines for namespaces that never appeared in the body. `_filter_prefixes_to_used` keeps only prefixes whose namespace is referenced by some IRI in the graph (subject, predicate, object, or literal datatype). The `examples/tutorial/tutorial01/data.ttl` fixture rebake demonstrates this — 28 unused declarations are gone.
- **Narrow the `ValueError` swallow.** The retry-without-prefixes path used a bare `except ValueError`, which would silently drop *all* prefix declarations on any `ValueError`, masking unrelated bugs (e.g. an invalid base IRI, or a future pyoxigraph serializer regression). It now only swallows messages starting with `"Invalid prefix"` (the empirically verified pyoxigraph 0.5.x signal); other `ValueError`s propagate.

## Why each change is needed

| Change | Failure it prevents |
| --- | --- |
| Tokenizer-aware CURIE rewrite | Round-trip corruption when a literal contains a CURIE-shaped substring with a trailing escaped dot. The naive regex would rewrite the literal's body into an angle-bracket IRI form, producing Turtle that no longer parses back to the same graph. |
| `RDFCanonicalizationWarning` | The pyoxigraph rejection / non-standard-RDF / unsupported-format paths return non-canonicalized output. With `logging.warning`, library callers (e.g. someone using `linkml-runtime` from a notebook or another tool) had no signal that the output's determinism guarantee was downgraded. |
| Unused-prefix filter | Output bloat — tutorial files were ~30 lines longer than necessary; downstream diffs over canonicalized RDF were noisier than they should be. |
| Narrower `except ValueError` | A future pyoxigraph release could raise `ValueError` from `ox.serialize` for an unrelated reason (e.g. invalid `base_iri`). With the broad catch, that bug would silently strip all prefix declarations rather than failing loudly. |

`test_fallback_on_invalid_rdf` updated to assert the warning fires.

## Files changed

- `packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py` — four fixes above.
- `tests/linkml_runtime/test_utils/test_rdf_canonicalize.py` — new tests.
- `examples/tutorial/tutorial01/data.ttl` — fixture rebake (drops 28 unused `@prefix` lines).

## Notes / open questions

- The tutorial fixture `data.ttl` also picks up a trailing newline. Cosmetic, likely from pyoxigraph's serializer.
- The literal-protection fix is the most behaviorally important — it's a latent data-corruption bug whenever someone's Turtle has a literal containing a `prefix:local\.` shape. Worth a separate commit if reviewers want them split.


amc-corey-cox (approved):
This looks good to me. Thank you @matentzn for this excellent work.

amc-corey-cox:
Pushed a small commit (d150e305f) rebaking `examples/tutorial/tutorial04/{data.ttl,data-semantic.ttl}` — the `Slow Tests` job was failing on `test_tutorial[tutorial04]` because the unused-prefix filter slims every serialized Turtle, but only tutorial01's fixture had been rebaked. Regenerated both via the tutorial's own `linkml-convert -t rdf` commands (38 → 5/6 prefixes); full tutorial suite passes locally with `--with-slow`. The third turtle block in tutorial04 uses `compare_rdf` (graph isomorphism) so it was unaffected. Revert if you'd rather handle it yourself.

PR #3527, opened 2026-05-17
fix(excel): class names longer than 31 chars

### LINKED ISSUE #3526: excelgen crash on class names longer than 31 chars
**Describe the bug**

Any LinkML class whose name exceeds 31 characters reliably triggers a crash in `excelgen`. This is common in schemas auto-generated from Java APIs, where fully-qualified-style names regularly run 35–50 characters.

The `create_workbook_and_worksheets` creates an openpyxl worksheet with `workbook.create_sheet(cls_name)` and openpyxl silently
**truncates** worksheet titles to 31 characters (the Excel hard limit) and emits a `UserWarning`. Then `add_columns_to_worksheet` looks the worksheet up by the **un-truncated** class name:
```python
worksheet = workbook[worksheet_name]   # full cls_name, > 31 chars
```

causing `KeyError: 'Worksheet <FullClassName> does not exist.'`.

The same `cls_name` is also used by `column_enum_validation`, so any class with both a long name and an enum-typed slot would also fail.

1.11.0

## Reproducer

Create `/tmp/excel.yaml`
```yaml
id: https://example.org/excelgen-bug
name: excelgen_bug
prefixes:
  ex: https://example.org/
default_prefix: ex
imports:
  - linkml:types

classes:
  AClassNameThatIsDefinitelyLongerThan31Chars:
    description: 35+ char class name to overflow Excel's 31-char sheet title limit.
    attributes:
      id:
        identifier: true
        range: string
      label:
        range: string
```

Run:

```bash
gen-excel /tmp/excel.yaml -o /tmp/bug.xlsx
```

Otuput:

```bash
.../openpyxl/workbook/child.py:99: UserWarning: Title is more than 31 characters.
  Some applications may not be able to read the file
  warnings.warn("Title is more than 31 characters. Some applications may not be able to read the file")
  File ".../bin/gen-excel", line 8, in <module>
    sys.exit(cli())
  ...
  File ".../linkml/generators/excelgen.py", line 102, in add_columns_to_worksheet
    worksheet = workbook[worksheet_name]
  File ".../openpyxl/workbook/workbook.py", line 287, in __getitem__
    raise KeyError("Worksheet {0} does not exist.".format(key))
KeyError: 'Worksheet AClassNameThatIsDefinitelyLongerThan31Chars does not exist.'
```

The same crash surfaces transitively through `gen-project` whenever any class in the schema has a name longer than 31 characters.

noelmcloughlin:
## Summary

Fixes #3526

Compute a deterministic, collision-free sheet title once per class and use it for **both** `create_sheet` and the subsequent lookup. 

Keep a`cls_name -> sheet_title` map.

## Checklist

## AI Assistance

LLM assisted unit tests

### DIFF
--- a/packages/linkml/src/linkml/generators/excelgen.py
+++ b/packages/linkml/src/linkml/generators/excelgen.py
@@ -1,3 +1,4 @@
+import hashlib
 import logging
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


sujaypatil96 (approved):
Thanks for implementing a solution for the case when the class name length is >= 31 characters. The collision free/safe title name generation logic looks good and is an effective truncation strategy! definitely better than what we have already where the generation simply crashes/errors out, so love the improvement 🚀

PR #3533, opened 2026-05-19
fix(processing): hoist compact-key list items, error on missing-pk items (#3529)

### LINKED ISSUE #3529: ReferenceValidator mangles compact-key list input to inlined-dict fields as {None: ...}
When \`ReferenceValidator.normalize()\` converts a compact-key list to an inlined-dict field (a multivalued, inlined collection that pydantic types as \`dict[str, X]\`), it produces a single-entry dict keyed by \`None\` instead of hoisting the compact-key name.

Reproducer (linkml_runtime 1.11.0):

\`\`\`python
from linkml_runtime.processing.referencevalidator import ReferenceValidator
from linkml_runtime.utils.introspection import package_schemaview

normalizer = ReferenceValidator(package_schemaview('linkml_map.datamodel.transformer_model'))
normalizer.expand_all = True

# permissible_value_derivations is typed as dict[str, PermissibleValueDerivation]
for label, pv in [
    ('dict-keyed',       {'red': {'sources': ['a']}}),
    ('compact-key-list', [{'red': {'sources': ['a']}}]),
    ('explicit-list',    [{'name': 'red', 'sources': ['a']}]),
]:
    obj = {'enum_derivations': {'T': {'populated_from': 'S', 'permissible_value_derivations': pv}}}
    result = normalizer.normalize(obj)
    print(f'{label}: {result["enum_derivations"]["T"]["permissible_value_derivations"]}')
\`\`\`

Output:

\`\`\`
dict-keyed:       {'red': {'sources': ['a'], 'name': 'red'}}
compact-key-list: {None: {'red': {'sources': ['a']}}}        ← bug
explicit-list:    {'red': {'name': 'red', 'sources': ['a']}}
\`\`\`

Expected: \`compact-key-list\` normalizes equivalently to \`dict-keyed\` — the outer single-key dict's key becomes the dict key (and is injected as \`name\` on the inner value), giving \`{'red': {'name': 'red', 'sources': ['a']}}\`.

Fix area: the list→dict conversion path for inlined collections, where each list item is expected to either have an explicit \`name\` field or be a compact-key dict whose single key is the name.

Origin: linkml/linkml-map#250. Worked around locally by pre-expanding compact-key list items before invoking the normalizer; a canary test there will fail (expected-failure flipping to pass) when this is fixed, signaling that the workaround can be removed.

amc-corey-cox:
Closes #3529.

## Problem

`ReferenceValidator.normalize()` silently produced `{None: ...}` for two related shapes of list input to an inlined-dict slot:

- Compact-key list items (`[{key: {...}}]`) — the YAML idiom for a list of identified things.
- Expanded items written without their identifier (`[{sources: ['a']}]`).

In both cases `v.get(pk_slot_name)` returned `None`, and that `None` became a dict key — producing output that no downstream consumer can use.

## Fix

In `ensure_expanded_dict` and `ensure_compact_dict` (List → Dict branches):

1. **Hoist compact-key items where unambiguous.** `[{key: {...}}]` is semantically equivalent to `{key: {...}}` (and to `[{<pk>: key, ...}]`) when the key isn't a slot of the range class, the value is a dict, and a pk is defined. Hoist in those cases.
2. **Error otherwise.** Items the heuristic can't safely hoist (no pk on the range, key collides with a slot name, value is scalar/null, multi-key item, etc.) emit a `DictCollectionFormConstraint` validation result. Never `{None: ...}`.

The slot-name-collision refusal is deliberate (no-footgun): if a user accidentally indented an attribute name into compact-key position, hoisting would silently rescue the typo into wrong-but-valid data. The user has unambiguous alternatives (`{key: {...}}` or `[{name: 'key', ...}]`) when they really mean the hoist.

See the [issue thread](https://github.com/linkml/linkml/issues/3529) for the full design discussion (semantic-equivalence argument, why A+B-fallback over strict rejection, CollectionForm conformance).

## Out of scope

`ensure_simple_dict` has the same family of bug (line 728 crashes with `KeyError` on list-input items missing pk). Fixing that needs the same treatment but is a separate path; will file as a follow-up.

## Test plan

- [x] 10 new pytest cases in `test_referencevalidator.py`: hoist happy path, mixed lists, multi-item, CompactDict variant, all four refusal shapes, slot-name collision refusal, non-colliding key still hoists.
- [x] All existing `tests/linkml_runtime/test_processing/` pass (21 passed, 1 skipped).
- [x] Full `tests/linkml_runtime/` suite passes (1736 passed, no regressions).
- [x] `ruff format` + `ruff check` clean.


tests/linkml_runtime/test_processing/test_referencevalidator.py
```diff
@@ -1158,3 +1158,139 @@ def test_metamodel(self):
...
+    rv = _normalizer(_COMPACT_KEY_SCHEMA_YAML)
+    report = Report()
+    out = rv.normalize({"items": input_items}, report=report)
+    assert out["items"] == expected_items
+    assert report.errors() == []
+
+
+def test_compact_key_list_hoists_to_compact_dict_when_not_expanding():
+    """The same input also normalizes correctly when the canonical form is CompactDict."""
+    rv = _normalizer(_COMPACT_KEY_SCHEMA_YAML, expand_all=False)
```
noelmcloughlin, line 1248: Add a new `parametrize("expand_all", [True, False])` unit test covering normalize of list `items: ["a", 1]` which should produce `{}` with 2 errors and no "None" key, i.e.
```
assert `mydict["items"] == {}
assert `None not in mydict["items"]
```

noelmcloughlin: LTGM - If you can fix the copilot `type` nitpicks then you can ping me for approval.

kevinschaper (changes requested):
Putting into a proper request changes state for the co-pilot type nits

PR #3536, opened 2026-05-20
feat(typedbgen): implement TypeDB CTO's mapping recommendations

### LINKED ISSUE #3534: feat(typedbgen): implement TypeDB CTO's mapping recommendations for richer schema generation
## Context

TypeDB's CTO (@flyingsilverfin) [reviewed the LinkML↔TypeDB mapping](https://github.com/linkml/linkml/pull/3382#issuecomment-4256296715) and outlined how TypeDB 3.x's data model maps almost directly onto LinkML's schema language. The current `gen-typedb` generator (introduced in #3382, with a fix in #3523) handles the basics well but leaves several TypeDB annotations and structural mappings on the table.

This issue tracks the gap between what the generator currently emits and what TypeDB 3.x supports. **All six feature suggestions below originate from @flyingsilverfin's comment.** The implementation notes (LinkML source fields, TypeDB target syntax, and trade-off analysis) were authored by @GullyBurns with assistance from Claude Code.

## What's already working

- `is_a` → `sub` (entity and relation inheritance)
- Primitive slots → `attribute <name>, value <type>`
- `identifier: true` → `@key`
- `required: true` (singular) → `@card(1..1)`
- `multivalued: true` → `@card(0..)`
- `abstract: true` → `@abstract`
- `represents_relationship: true` classes → `relation` types with `relates` roles

### 1. Enums → `@values(...)` annotation

> "primitive enums become `@values`" — @flyingsilverfin

**Current:** Enums are emitted as `attribute X, value string;` with a comment listing permitted values.
**Proposed:** Emit `@values("val1", "val2", ...)` on the attribute declaration, which TypeDB 3.x enforces at the schema level.

**Implementation notes:**
- **LinkML source:** `EnumDefinition.permissible_values`
- **TypeDB target:** `attribute status, value string @values("employed", "unemployed", "student");`

### 2. `minimum_value`/`maximum_value` → `@range(min..max)`

> "min/max value becomes `@range(min..max)` (or leave one open if not set)" — @flyingsilverfin

**Current:** Not implemented — these LinkML slot properties are ignored.
**Proposed:** When a slot has `minimum_value` and/or `maximum_value`, emit `@range(min..max)` on the `owns` or `value` declaration.

**Implementation notes:**
- **LinkML source:** `SlotDefinition.minimum_value`, `SlotDefinition.maximum_value`
- **TypeDB target:** `attribute age, value integer @range(0..150);`

### 3. Precise cardinality via `@card(min..max)`

> "min/max/exact cardinality can be `@card(min, max)`" — @flyingsilverfin

**Current:** Only boolean flags are used (`required`, `multivalued`, `identifier`). Slots with `minimum_cardinality: 2, maximum_cardinality: 5` would just get `@card(0..)`.
**Proposed:** Read `minimum_cardinality` and `maximum_cardinality` from the induced slot and emit precise `@card(min..max)`.

**Implementation notes:**
- **LinkML source:** `SlotDefinition.minimum_cardinality`, `SlotDefinition.maximum_cardinality`
- **TypeDB target:** `owns authors @card(1..10)`

### 4. Attribute subtyping via slot `is_a`

> "this also works for reusable primitive slots (eg. Attributes in TypeDB can subtype other ones!)" — @flyingsilverfin

**Current:** All attributes are declared flat — `attribute email, value string;` with no hierarchy.
**Proposed:** When a LinkML slot has `is_a` pointing to another scalar slot, emit `attribute email sub contact-info;` to leverage TypeDB's attribute type inheritance.

**Implementation notes:**
- **LinkML source:** `SlotDefinition.is_a`
- **TypeDB target:** `attribute email sub contact-info;`

### 5. Auto-detect entity vs. relation

> "entities if they don't contain any slots that range over other classes; relations if they contain slots that do range over classes" — @flyingsilverfin

**Current:** A class must be explicitly annotated with `represents_relationship: true` to become a TypeDB `relation`. Otherwise it's always an `entity`, even if all its slots range over other classes.
**Proposed:** Consider auto-inferring `relation` for classes where all (or a majority of) slots range over other classes, with `represents_relationship` as an explicit override.

**Implementation notes:** This is a heuristic and could surprise users. Might be better as an opt-in generator flag (e.g., `--infer-relations`) rather than a default behavior change.

### 6. Mixin support

> "I believe mixins are just composition of reusable schema definitions" — @flyingsilverfin

**Current:** Mixins are not structurally represented. Mixin slots flow through `class_induced_slots` (so `owns` clauses work), but the mixin's identity is lost in the TypeDB output.
**Proposed:** Investigate mapping mixins to TypeDB's type composition patterns. This may be limited by TypeDB's single-inheritance model for types, but mixin *slots* could be emitted as shared attribute/role declarations.

## Out of scope (for now)

These were noted by @flyingsilverfin as not yet supported or hard to map:

- **Boolean combination of slot types** — no clear TypeDB equivalent
- **Arrays / ordered slot contents** — TypeDB list types are "coming soon"
- **Default values** — TypeDB doesn't support these yet
- **Subsets** — unclear mapping

## Related

- #3382 — Original TypeDB generator PR
- #3519 / #3523 — Fix for inline `attributes:` being dropped
- @flyingsilverfin's comment: (https://github.com/linkml/linkml/pull/3382#issuecomment-4256296715)

GullyBurns:
## Summary

Implements the TypeDB 3.x schema mapping improvements suggested by @flyingsilverfin in #3382 and tracked in #3534.

- Enums → `@values(...)` annotation (instead of comments)
- `minimum_value`/`maximum_value` → `@range(min..max)`
- Precise `@card(min..max)` from `minimum_cardinality`/`maximum_cardinality`
- Attribute subtyping via slot `is_a` → `sub`
- Mixin slot composition
- (Optional) Auto-detect entity vs. relation via `--infer-relations` flag

## Test plan

- [ ] Existing `typedbgen` tests pass (`uv run pytest tests/linkml/test_generators/test_typedbgen.py`)
- [ ] New tests for each feature: enums with `@values`, `@range`, precise `@card`, attribute `sub`, mixin slots
- [ ] Kitchen sink schema output reviewed for regressions
- [ ] Generated TypeQL validated against TypeDB 3.x syntax

Closes #3534

### DIFF
--- a/packages/linkml/src/linkml/generators/typedbgen.py
+++ b/packages/linkml/src/linkml/generators/typedbgen.py
@@ -361,6 +361,10 @@ class TypeDBGenerator(Generator):
     def _build_owns_stmt(self, slot_tname: str, induced: SlotDefinition) -> str:
         """Return a TypeQL ``owns`` statement with the appropriate cardinality annotation.
 
+        Uses ``minimum_cardinality`` / ``maximum_cardinality`` / ``exact_cardinality``
+        when set, falling back to the boolean flags (``required``, ``multivalued``,
+        ``identifier``) for backwards compatibility.
+
         :param slot_tname: the safe TypeDB attribute name for the slot
         :param induced: the induced SlotDefinition carrying cardinality/identity metadata
         :return: a TypeQL ``owns`` clause string (without trailing punctuation)
@@ -368,6 +372,12 @@ class TypeDBGenerator(Generator):
         owns = f"owns {slot_tname}"
         if induced.identifier:
             owns += " @key"
+        elif induced.exact_cardinality is not None:
+            owns += f" @card({induced.exact_cardinality}..{induced.exact_cardinality})"
+        elif induced.minimum_cardinality is not None or induced.maximum_cardinality is not None:
+            lo = induced.minimum_cardinality if induced.minimum_cardinality is not None else 0
+            hi = str(induced.maximum_cardinality) if induced.maximum_cardinality is not None else ""
+            owns += f" @card({lo}..{hi})"
         elif induced.required and not induced.multivalued:
             owns += " @card(1..1)"
         elif induced.multivalued:
--- a/packages/linkml/src/linkml/generators/typedbgen.py
+++ b/packages/linkml/src/linkml/generators/typedbgen.py
@@ -330,7 +330,7 @@ class TypeDBGenerator(Generator):
                     continue  # object range → relation, not attribute
                 seen[slot_name] = f"attribute {slot_name}, value {value_type};"
 
-        # Enums used in slot ranges also need attribute declarations with comments
+        # Enums used in slot ranges get @values(...) annotations to enforce permitted values.
         enum_attr_lines: list[str] = []
         enum_slots_seen: set[str] = set()
         for class_name in sv.all_classes():
@@ -340,10 +340,11 @@ class TypeDBGenerator(Generator):
                     enum_slots_seen.add(slot_name)
                     enum_def = sv.get_enum(induced_slot.range)
                     permitted = list(enum_def.permissible_values.keys()) if enum_def else []
-                    comment = f"# Enum: {_typedb_name(induced_slot.range)} (permitted values: {', '.join(permitted)})"
+                    comment = f"# Enum: {_typedb_name(induced_slot.range)}"
                     enum_attr_lines.append(comment)
                     seen.pop(slot_name, None)
-                    enum_attr_lines.append(f"attribute {slot_name}, value string;")
+                    values_str = ", ".join(f'"{v}"' for v in permitted)
+                    enum_attr_lines.append(f"attribute {slot_name}, value string @values({values_str});")
 
         result = list(seen.values()) + enum_attr_lines
         return result
--- a/packages/linkml/src/linkml/generators/typedbgen.py
+++ b/packages/linkml/src/linkml/generators/typedbgen.py
@@ -368,6 +368,19 @@ class TypeDBGenerator(Generator):
                     values_str = ", ".join(f'"{v}"' for v in permitted)
                     enum_attr_lines.append(f"attribute {slot_name}, value string @values({values_str});")
 
+        # Attribute subtyping: slots with is_a pointing to another scalar slot
+        # emit "attribute <child>, sub <parent>;" instead of a flat value declaration.
+        # TypeDB inherits the value type from the parent, so no value clause is needed.
+        all_slots = sv.all_slots()
+        for slot_name_orig, safe_name in attr_names.items():
+            if safe_name not in seen:
+                continue
+            slot_def = all_slots.get(slot_name_orig)
+            if slot_def and slot_def.is_a:
+                parent_safe = attr_names.get(slot_def.is_a)
+                if parent_safe and parent_safe in seen:
+                    seen[safe_name] = f"attribute {safe_name}, sub {parent_safe};"
+
         result = list(seen.values()) + enum_attr_lines
         return result
 
--- a/packages/linkml/src/linkml/generators/typedbgen.py
+++ b/packages/linkml/src/linkml/generators/typedbgen.py
@@ -384,13 +384,20 @@ class TypeDBGenerator(Generator):
         result = list(seen.values()) + enum_attr_lines
         return result
 
-    def _ancestor_slots(self, sv: SchemaView, class_name: str) -> set[str]:
-        """Return the set of slot names inherited from all ancestors (excluding self)."""
+    def _ancestor_slot_names(self, sv: SchemaView, class_name: str) -> set[str]:
+        """Return the set of slot names inherited from non-mixin ancestors (excluding self).
+
+        Mixin ancestors are skipped because TypeDB has single ``sub`` inheritance
+        only — there is no mixin/trait mechanism.  Mixin-contributed slots must be
+        re-declared on each consuming class as ``owns`` / ``plays``.
+        """
         result: set[str] = set()
         for ancestor in sv.class_ancestors(class_name)[1:]:  # skip self
-            ancestor_cls = sv.get_class(ancestor)
-            if ancestor_cls and ancestor_cls.slots:
-                result.update(ancestor_cls.slots)
+            ancestor_def = sv.get_class(ancestor)
+            if ancestor_def and ancestor_def.mixin:
+                continue  # mixin slots must be re-declared; TypeDB has no mixin inheritance
+            for slot in sv.class_induced_slots(ancestor):
+                result.add(slot.name)
         return result
 
     def _build_owns_stmt(self, slot_tname: str, induced: SlotDefinition) -> str:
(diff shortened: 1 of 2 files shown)


GullyBurns:
## Item #5 (auto-detect entity vs. relation): decided to skip

After analysis, we've decided **not** to implement automatic inference of entity vs. relation types. Here's why.

### The suggestion

@flyingsilverfin suggested that classes could be automatically classified as TypeDB `relation` types (vs. `entity`) based on whether their slots range over other classes:

> "entities if they don't contain any slots that range over other classes; relations if they contain slots that do range over classes"

### Why it doesn't work in practice

We analyzed the `kitchen_sink.yaml` schema (the standard test schema used across all LinkML generators) and found that **any heuristic produces false positives**:

| Class | Object-ranged slots | Scalar slots | Actual intent |
|---|---|---|---|
| `FamilialRelationship` | `related_to→Person` | `cordialness`, `type`, etc. | Relation ✓ |
| `Dataset` | `persons→Person`, `companies→Company`, `activities→activity`, `code_systems→CodeSystem` | *(none)* | **Container entity** ✗ |
| `WithLocation` | `in_location→Place` | *(none)* | **Mixin/trait** ✗ |
| `Person` | 5 class-ranged slots | 7 scalar slots | Entity ✓ |
| `Company` | `ceo→Person` | `id`, `name`, `aliases` | Entity ✓ |

- **"All slots are class-ranged → relation"** misclassifies `Dataset` (a root container) and `WithLocation` (a mixin).
- **"2+ class-ranged, 0 scalar → relation"** still catches `Dataset`.
- **Adding exclusions** (skip `tree_root`, skip classes named `*Dataset`, etc.) is fragile and schema-dependent.

Even `FamilialRelationship` — the most obvious relation in the schema — has *mixed* slots (both class-ranged and scalar), so a strict "all slots are class-ranged" rule would miss it.

### The right approach

The existing `represents_relationship: true` annotation is the correct design. It puts the **modeling decision in the schema author's hands**, which is where it belongs. Whether a class represents a relationship or an entity is a semantic choice that can't be reliably inferred from slot ranges alone.

The generator already handles `represents_relationship` correctly:
- Object-ranged slots → `relates` roles
- Scalar slots → `owns` on the relation
- Range classes → `plays` declarations

No code changes needed for this item.

GullyBurns:
## Item #6 (mixin support): implementation approach

@flyingsilverfin suggested:
> "I believe mixins are just composition of reusable schema definitions"

This is correct — LinkML mixins contribute slots to consuming classes without requiring inheritance. However, **TypeDB has single `sub` inheritance only** — there's no mixin/trait/interface mechanism. 

The current generator has two bugs related to this:

1. **Mixin classes are emitted as orphaned entity types** — e.g., `entity hasaliases;` appears in the output but nothing can `sub` it (since `Person` may already have an `is_a` using its one `sub` slot).
2. **Mixin slots are silently dropped** — the ancestor-subtraction logic treats mixin ancestors the same as `is_a` ancestors. Since there's no `sub` path in TypeDB, these slots vanish. For example, `Person` mixes in `HasAliases` but never gets `owns aliases`.

**Fix approach:** 
- Skip mixin classes in entity/relation output (they're a LinkML modeling convenience, not TypeDB types)
- Don't subtract mixin-ancestor slots during the "direct slots" calculation — instead, re-declare them on each consuming class as `owns`/`plays`

This matches TypeDB's model: the only way to share capabilities across unrelated types is to declare the same `owns`/`plays` on each type independently.

flyingsilverfin:
I'd say for item 6, the way i expect them to be modeled is by redeclaring them on each consuming class, correct!

Item 5 is tricky, lemme outline some thoughts

flyingsilverfin:
Here's some thoughts I came up with while I was scanning the kitchen sink yaml:

1) Multiple ranges for an attribute/slot
There's definitely some patterns TypeDB can't express 'properly' and I'd say we should error if encountered. I'm going off the kitchen sink yaml:
```
  AnyOfMix:
    attributes:
      attribute4:
        any_of:
          - range: integer
          - range: Person
          - range: EmploymentEventType
```

If we turn this into a Relation with a role `attribute4`, which is played by Person - we hit the limit that attributes can't play roles. If we turn it into an attribute, we can't have it be filled by multiple types. Outcome: TypeDB can't support any_of unless it's purely a set of class-valued ranges and we turn the class into a Relation, or we have a single primitive valued range that can be a direct attribute ownership. 

I think there's probably more complex conventions we can pick up (the `attribute4` always becomes a relation, and then we give it a set of owned attributes for the primitives or a role played by the possible class-based ranges - but I think we can not go there for now).



2) equals_string
```
  EqualsString:
    attributes:
      attribute5:
        equals_string: "foo"
```
Easy: 
```
entity equals_string, owns attribute5 @values("foo");
```
Lots of our constraints can be applied on the attribute or the ownership

3) Mixins
Don't 'store' mixins (for now - in the future TypeDB will have interfaces or traits and you can then of import them there). For now - just flatten and redeclare the ownerships or roles

4) Relations vs Entities vs Roles

My instinct is that it's still a _good_ representation in TypeDB terms to convert any class that has an attribute _or_ a slot that has a range that is a class into a Relation. (this is Option 1)

This does lead to some weirdness, such as `Concept` in the kitchen sink being a relation!

However, it does compose well. If you have a slot or attribute that values over a class, you're making a role type and giving that class's TypeDB representation a `plays` of that role.

The other somewhat arbitrary split (Option 2) we could do is that any class that defines an _inline_ attribute that is ranging over a class, is a relationship itself with those attributes as roles (and the ranging classes as playing those roles). However, a _slot_ ranging over a class can be converted to a binary relation with a `source` and `target` or some other convention-based role names.

The 'simplest' model (Option 3) is to flatten all classes into entities, and all slots and inline attributes ranging over classes, into binary relationships with convention-based roles, as above. However, this takes away the real power of n-ary relationships or relationships on relationships in TypeDB! Entirely usable that way, but less native to the system then. 

The question I have here is really more of a LinkML modeling question, maybe @kevinschaper can help or pull someone in that knows better - how are relationships meant to be modeled in LinkML? In the call today I heard that few people actually use the `represents_relationship`! Perhaps after hearing my presentation someone can help explain it in TypeDB terms a little bit :)

Note: re. the `Dataset` class. I could see this as an entity or a relation to be honest. The main difference in TypeDB is that a relationship cannot exist without at least one linked role player - only entities can exist independently. So I suppose the AI's suggestion of a 'marker' in LinkML for relationships is the most reliable signal, otherwise we have to pick some convention as I outlined the 3 options above.

sierra-moxon:
I think this could be better: https://linkml.io/linkml/faq/modeling.html#how-do-i-represent-relationships-in-linkml
Lots of flexibility in how relationships can be modeled in LinkML, which is what you are picking up on! :) 

examples:
NMDC schema relies heavily on inlining.
Biolink relies heavily on modeling relationships as "association" classes.

I would say that this kind of modeling makes my data modeler's heart weep a bit 😄 : 

```
        any_of:
          - range: integer
          - range: Person
          - range: EmploymentEventType
 ```

flyingsilverfin:
That's a pretty small docs! Is there a more extensive guide anywhere @sierra-moxon ?

sierra-moxon:
definitely did not mean to close this PR!  reopened!  apologies.   think most LinkML models I've seen in practice, lean towards inlining and class ranges and/or "identifiers" to dictate "relationships" (vs. the biolink strategy).  But it does, kind of, depend (as the doc says) on the modeler's use cases.

flyingsilverfin:
Do you think one of the three defaults we could use is the "most sensible" @sierra-moxon ?

PR #3537, opened 2026-05-21
feat(jsonschemagen): emit `readOnly: true` for slots with `readonly` set

### LINKED ISSUE #3528: `JsonSchemaGenerator`: emit `readOnly: true` for slots with a `readonly` meta slot value

Have `JsonSchemaGenerator` emit JSON Schema [`readOnly: true`][js-readonly]
for any slot whose [`readonly`][lk-readonly] meta slot has a non-empty
value.

JSON Schema's `readOnly` is well-specified: a non-validating annotation
indicating the instance value is managed by the owning authority and
should not be modified by consumers. LinkML's `readonly` is defined more
vaguely ("If present, slot is read only. Text explains why."), but its
intent is close enough that emitting `readOnly: true` is a reasonable
default mapping. Neither keyword affects validation, so the mapping is
safe. The LinkML `readonly` value is a free-text reason with no JSON
Schema equivalent; it could optionally be preserved as a custom
annotation (e.g., `x-readonly-reason`), or simply dropped 
for the JSON Schema target.

`readOnly` is part of the default JSON Schema 2019-09 meta-data
vocabulary, so no dialect change is needed.

**How important is this feature to you?**

Important — we want the LinkML schema to be the single source of
truth. The available workarounds (post-process the generated JSON
Schema, or subclass `JsonSchemaGenerator`) mean that publishing the
LinkML schema alone is not enough: anyone who needs the correct JSON
Schema also has to receive and run our post-processor, which
undermines the schema-as-source-of-truth story. With this mapping
built in, upstream-Pydantic pipelines (e.g.
`Field(..., json_schema_extra={"readOnly": True})` as used in
[`dandischema.models`][dandi-models], round-tripped through
[`pydantic2linkml`][p2l]) would preserve `readOnly` end-to-end without
any custom plumbing on the consumer side.

**Additional context**

Discussion / verification of the semantic overlap, including a
demonstration that neither keyword affects validation behavior:
dandi/pydantic2linkml#68.

If this proposal is accepted, I am happy to submit the PR.

[js-readonly]: https://json-schema.org/draft/2019-09/json-schema-validation#rfc.section.9.4
[lk-readonly]: https://linkml.io/linkml-model/latest/docs/readonly/
[dandi-models]: https://github.com/dandi/dandischema/blob/master/dandischema/models.py
[p2l]: https://github.com/dandi/pydantic2linkml

candleindark:
## Summary

Map LinkML's `readonly` meta slot to JSON Schema's `readOnly` annotation in `JsonSchemaGenerator`. When a slot has a non-empty `readonly` value, the generated JSON Schema property now carries `"readOnly": true`. Absent or empty values produce no change. The free-text reason is dropped — JSON Schema has no equivalent.

Closes #3528.

## Test plan

- [x] New test `test_slot_readonly_emits_read_only` with fixture `tests/linkml/test_generators/input/jsonschema_slot_readonly.yaml` covering: non-empty `readonly` → `readOnly: true`; empty-string `readonly` → keyword omitted; absent `readonly` → keyword omitted; inherited slot via `is_a` → subclass property also carries `readOnly: true`.
- [x] Existing `tests/linkml/test_generators/test_jsonschemagen.py` and `test_jsonschemagen_examples.py` suites pass.
- [x] `tests/linkml/test_scripts/test_gen_json_schema.py` snapshot tests pass (no metamodel-derived snapshot drift).
- [x] Broader sweep `uv run pytest tests/linkml/ -k "jsonschema or json_schema"`: 868 passed, 90 skipped.

### DIFF
--- a/packages/linkml/src/linkml/generators/jsonschemagen.py
+++ b/packages/linkml/src/linkml/generators/jsonschemagen.py
@@ -829,6 +829,8 @@ class JsonSchemaGenerator(Generator, LifecycleMixin):
         prop.add_keyword("description", slot.description)
         if self.title_from == "title" and slot.title:
             prop.add_keyword("title", slot.title)
+        if getattr(slot, "readonly", None):
+            prop.add_keyword("readOnly", True)
 
         own_constraints = self.get_value_constraints_for_slot(slot)
 
--- a/tests/linkml/test_generators/input/jsonschema_slot_readonly.yaml
+++ b/tests/linkml/test_generators/input/jsonschema_slot_readonly.yaml
@@ -1,0 +1,24 @@
+id: https://example.org/readonly-test
+name: readonly_test
+prefixes:
+  linkml: https://w3id.org/linkml/
+imports:
+  - linkml:types
+default_range: string
+
+slots:
+  managed:
+    readonly: assigned by the server
+  empty_reason:
+    readonly: ""
+  plain:
+    description: not read only
+
+classes:
+  Base:
+    slots:
+      - managed
+      - empty_reason
+      - plain
+  Sub:
+    is_a: Base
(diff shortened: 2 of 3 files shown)


kevinschaper (approved):
thanks @candleindark!

PR #3550, opened 2026-05-21
fix(rustgen): emit typed PyO3 conversions for Anything

kervel:
## Summary

`anything.rs.jinja` rendered both directions of the `Anything` <-> Python bridge with debug-stringifying fallbacks:

* `IntoPyObject::into_pyobject` only handled `Unit`/`Bool`/`String`/ `Seq`/`Map`. Every other `serde_value::Value` variant fell into a catch-all that did `format!("{:?}", other)` and returned the result as a `PyString`. So `Value::U64(42)` surfaced in Python as the string `"U64(42)"`, `Value::Bytes(b"hi")` as `"Bytes([104, 105])"`, etc.
* `FromPyObject::extract_bound` only extracted `&str` and `bool`, in that order. Python ints and floats fell through to the stringifying fallback at the bottom. The `&str`-before-`bool` order was also wrong: `True` extracts as the string `"True"` rather than `Value::Bool(true)`.

Replace both fallbacks with explicit arms:

* `value_to_py` now handles `U8`/`U16`/`U32`/`U64`/`I8`/`I16`/`I32`/ `I64`/`F32`/`F64`/`Char`/`Bytes`/`Option`/`Newtype` with the right PyO3 conversions. Strings/bool/sequences/maps unchanged. The match is now exhaustive over `serde_value::Value`, so a future variant addition will fail the build rather than silently stringify.
* `py_to_value` now tries `bool`, then `i64`, then `f64`, then `&str`. `bool` first because Python `bool` is a subclass of `int`; `i64` before `f64` so integers don't widen to floats.

## How was this tested?

This code already lived in `rust-linkml-core` for a long time i forgot to port it to the generator.

## Checklist

## AI Assistance

(yes i did)

### DIFF
--- a/packages/linkml/src/linkml/generators/rustgen/templates/anything.rs.jinja
+++ b/packages/linkml/src/linkml/generators/rustgen/templates/anything.rs.jinja
@@ -41,12 +41,21 @@ impl<'py> FromPyObject<'py> for Anything {
                 return Ok(Value::Unit);
             }
 
-            // Try simple primitives first
-            if let Ok(s) = o.extract::<&str>() {
-                return Ok(Value::String(s.to_string()));
-            }
+            // Try simple primitives first.
+            // Order matters: Python's `bool` is a subclass of `int`, so `bool`
+            // must be tried before `i64`. Likewise, `i64` before `f64` so we
+            // don't widen integers to floats.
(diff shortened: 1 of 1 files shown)


kevinschaper (approved):
thanks @kervel!

PR #3555, opened 2026-05-22
don't leave files open 2: this time, they're errors

sneakers-the-rat:
from the producers of: [dont leave files open](https://github.com/linkml/linkml/pull/3207), now playing:

### *this time, they're errors*

---

Continues: #3207, #2288, #1939, #1954
Part of: #2985 

I'm a little bit like [this guy](https://www.youtube.com/watch?v=ymHyOmlUlP0) when it comes to warnings in the tests and now specifically leaving files open.

I think it's fair to say:
- we should not ship code that leaves files open
- we should not have to suffer an avalanche of warnings about leaving files open that prevent us from seeing any important information from warnings

so I am making a proposal that we start making some classes of warnings emitted during testing errors. 

ideally, the only kind of error that would be unhandled and emitted without error are deprecation warnings from other packages, because those are the kind of warnings-as-output that are actually useful during testing. but all the other warnings are indications that something is wrong and should be fixed - 
particularly when they are emitted from our own code. 
particularly *particularly* when they are emitted from code that we ship in the package.

let us begin by `/voteban`ning `ResourceWarning`s, which are strictly programming mistakes, and they are mistakes that can have consequences.

i am opening this as a draft because my zen garden-based attention span for cleaning up the linkml tests needs to be given a cap, but i draft as an offering and will return.

## Checklist

## AI Assistance

I would never let a robot take away the joy of cleaning, crafting, or coding from me.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,14 +2,12 @@ import os
 import re
 import shutil
 import sys
-import warnings
 from abc import ABC, abstractmethod
 from collections.abc import Callable
 from importlib.abc import MetaPathFinder
 from importlib.metadata import version
 from pathlib import Path
 
-import docker
 import pytest
 import requests_cache
 from _pytest.assertion.util import _diff_text
@@ -253,6 +251,7 @@ def pytest_addoption(parser):
     parser.addoption(
         "--with-rustgen", action="store_true", help="Include tests marked as rustgen (Rust codegen/maturin)"
     )
+    parser.addoption("--with-docker", action="store_true", help="include docker tests")
 
 
 @pytest.hookimpl(tryfirst=True)
@@ -282,6 +281,12 @@ def pytest_collection_modifyitems(config, items: list[pytest.Item]):
             if item.get_closest_marker("network"):
                 item.add_marker(skip_network)
 
+    if not config.getoption("--with-docker"):
+        skip_docker = pytest.mark.skip(reason="need --with-docker option to run")
+        for item in items:
+            if item.get_closest_marker("docker"):
+                item.add_marker(skip_docker)
+
     # Group compliance tests on a single xdist worker - they share
     # mutable module-level caches in helper.py that are not safe to split.
     for item in items:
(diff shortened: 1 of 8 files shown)


tests/conftest.py
```diff
@@ -300,13 +305,6 @@ def pytest_collection_modifyitems(config, items: list[pytest.Item]):
             if item.get_closest_marker("pydanticgen_npd"):
                 item.add_marker(skip_npd)
 
-    # skip docker tests when docker server not present on the system
-    if not _docker_server_running():
-        skip_docker = pytest.mark.skip(reason="Docker server not running on host machine")
-        for item in items:
-            if item.get_closest_marker("docker"):
-                item.add_marker(skip_docker)
```
sneakers-the-rat, line 308: regarding the docker changes, one might note that the only tests that have the docker marker also have the kroki marker, the the kroki marker is
- [excluded](https://github.com/linkml/linkml/blob/b40f736baa28aa1693085ebc6890780387f2ac29/.github/workflows/main.yaml#L90)
- [from every](https://github.com/linkml/linkml/blob/b40f736baa28aa1693085ebc6890780387f2ac29/.github/workflows/main.yaml#L94)
- [test condition](https://github.com/linkml/linkml/blob/b40f736baa28aa1693085ebc6890780387f2ac29/.github/workflows/main.yaml#L150)

so there is no reason to open a docker connection when starting the tests since those tests are never run anyway. this is not my known dislike of forcing every computing environment to be running a docker engine or be running in a docker engine speaking, i do not think these tests are run, but if they are, they can be run by adding the `--with-docker` flag explicitly, which is kinder than attempting to patch into someone's container environment when it might not be expected.

tests/linkml/test_notebooks/test_examples.py
```diff
@@ -12,6 +12,8 @@
 
 logger = logging.getLogger(__name__)
 
+pytestmark = pytest.mark.notebooks
```
sneakers-the-rat, line 15: i added these because the notebook tests take a very long time and also make a bunch of temporary files and i always want to exclude them and wanted a clean handle to do so

sneakers-the-rat:
well that's certainly concerning that the tests pass on python 3.11 and 3.12 but not 3.13 or 3.14. differences in handling how files are opened and closed by python version give me the heebie jeebies because a) what and b) why and c) oh no

PR #3556, opened 2026-05-22
test(snapshots): refresh markdowndatadict_personinfo metamodel-version

noelmcloughlin:
## Summary

The Metamodel ([https://w3id.org/linkml/meta.yaml](https://w3id.org/linkml/meta.yaml)) was updated to 1.11.0 but the markdowndatadict snapshot was never regenerated, so it kept asserting against the old "metamodel version: 1.7.0" string and failed any --with-network run.

Regenerated via:
  uv run pytest tests/linkml/test_generators/test_markdowndatadictgen.py \\
    --with-network --generate-snapshots

## How was this tested?

This interfered with another PR I was testing.

## Checklist

- [x] My code follows the [contributor guidelines](https://linkml.io/linkml/maintainers/contributing.html)
- [ ] I have added tests that prove my fix/feature works
- [x] Existing tests pass locally with my changes

Issue #1832, opened 2024-01-19
Protobuf generator generates dubious output


ielis:
**Describe the bug**

LinkML generates Protocol Buffers with `proto3` syntax. The syntax prescribes that field numbers must be [unique](https://protobuf.dev/programming-guides/proto3/#assigning) among all fields of that message. However, the LinkML output sets all field numbers to `0`.

**To reproduce**

Start with the person info example:
```yaml
id: https://w3id.org/linkml/examples/personinfo
name: personinfo
prefixes:                                 
  linkml: https://w3id.org/linkml/
  schema: http://schema.org/
  personinfo: https://w3id.org/linkml/examples/personinfo/
  ORCID: https://orcid.org/
imports:
  - linkml:types
default_range: string
  
classes:
  Person:
    class_uri: schema:Person             
    slots:   ## specified as a list
     - id
     - full_name
     - aliases
     - phone
     - age
    id_prefixes:
      - ORCID
  Container:
    attributes:
      persons:
        multivalued: true
        inlined_as_list: true
        range: Person

# slots are first-class entities in the metamodel
# declaring them here allows them to be reused elsewhere
slots:
  id:
    identifier: true
  full_name:
    required: true
    description:
      name of the person
    slot_uri: schema:name
  aliases:
    multivalued: true
    description:
      other names for the person
  phone:
    pattern: "^[\\d\\(\\)\\-]+$"
    slot_uri: schema:telephone 
  age:
    range: integer
    minimum_value: 0
    maximum_value: 200

```

Next, generate the proto files:
```shell
gen-proto personinfo.yaml | less
```

which produces the following output:
```
syntax="proto3";
 package
// metamodel_version: 1.7.0
message Container
 {
 repeated  person persons = 0
 }
message Person
 {
  string id = 0
  string fullName = 0
 repeated  string aliases = 0
  string phone = 0
  integer age = 0
 }
```

**Expected behavior**
The fields should be set with unique numbers.

noelmcloughlin:
I will raise PR to fix.  The generator emits corrupt files (field number 0 is reserved, and a bare `package` is syntax error) and protoc cannot compile.

Issue #455, opened 2021-11-04
All generator CLIs should accept an --output argument and handle output args in the same way


cmungall:
Looking at this together with @sujaypatil96 

Currently each gen-X method will write to stdout, so the idiomatic usage is:

```bash
gen-X my.yaml > my.X_format
```

This is fine, but we should strive to follow https://clig.dev/ and have an explicit `-o/--output`

Some wrinkles to be aware of:

The way the generator framework currently works is that each generator `print()`s its output. This is sometimes done in a streaming fashion; sometimes all at once, at end_schema time.

E.g. in owlgen we have:

```python
    def end_schema(self, output: Optional[str] = None, **_) -> None:
        data = self.graph.serialize(format='turtle' if self.format in ['owl', 'ttl'] else self.format).decode()
        if output:
            with open(output, 'w') as outf:
                outf.write(data)
        else:
            print(data)
```

Note that end_schema signature is over-ridden to give an output arg (this isn't accessible via the CLI though!)

jsonschemagen also prints its full payload in end_schema -- however, it doesn't override the signature of end_schema:

```python
   def end_schema(self, **_) -> None:
        # create more lax version of every class that is used as an inlined dict reference;
        # in this version, the primary key/identifier is optional, since it is used as the key of the dict
        for cls_name, (id_slot, cls_name_lax) in self.optional_identifier_class_map.items():
            lax_cls = deepcopy(self.schemaobj['$defs'][cls_name])
            lax_cls.required.remove(id_slot)
            self.schemaobj['$defs'][cls_name_lax] = lax_cls
        print(as_json(self.schemaobj, sort_keys=True))
```

in contrast, graphqlgen will stream the output:

```python
def visit_class(self, cls: ClassDefinition) -> bool:
        etype = 'interface' if (cls.abstract or cls.mixin) and not cls.mixins else 'type'
        mixins = ', '.join([camelcase(mixin) for mixin in cls.mixins])
        print(f"{etype} {camelcase(cls.name)}" + (f" implements {mixins}" if mixins else ""))
        print("  {")
        return True
```

in all cases, each generator `print`s output

the parent generator class takes care of routing this to a buffer, which is later exported:

```python
   def serialize(self, **kwargs) -> str:
        """
        Generate output in the required format

        :param kwargs: Generater specific parameters
        :return: Generated output
        """
        output = StringIO()
        with redirect_stdout(output):
            self.visit_schema(**kwargs)
        ....

           self.end_schema(**kwargs)
[...]

PR #3557, opened 2026-05-22
fix(protobuf): fix broken generator (part 1)

noelmcloughlin:
## Summary

Partly addresses #1832

The **protogen.py** generator needs work:

1. Its broken (#1832) and needs fixing.
2. It does not support `wire` format but should be separate PR
3. Need to handle multi-file (#3556 772)

But we also need to consider #455 and additionally I have some confusion about multi-file - its breaks persistent identifiers (maybe multi-file output should be tarballed).

## How was this tested?

Locally
Added unit tests

## Areas of uncertainty

Since `genproto.py` is broken this PR should be merged.  As followup PR's arrive we should find real life testers as it is hard for reviewers to review code changes on broken tools nobody is using

## Checklist

## AI Assistance

Claude wrote the code.  But I reviewed proto3 documentation and fixed few gaps and reviewed best I could. (see uncertainty above).


packages/linkml/src/linkml/generators/protogen.py
```diff
@@ -18,52 +108,249 @@ class ProtoGenerator(Generator):
 
     # ClassVars
     generatorname = os.path.basename(__file__)
-    generatorversion = "0.1.1"
+    generatorversion = "0.1.2"
```
kevinschaper: Non-blocking: this is effectively a complete rewrite of a broken generator, but the version bump is `0.1.1 → 0.1.2`. Consider `0.2.0` (or `1.0.0`) so downstream users have a visible signal that previous output was unusable and they should regenerate.

packages/linkml/src/linkml/generators/protogen.py
```diff
@@ -1,13 +1,103 @@
...
+    - Any underscore immediately followed by a digit is dropped (``foo_2bar``
+      becomes "foo2bar") since digits aren't allowed after ``_``.
+    - Leading and trailing underscores are stripped.
+    - An empty result, or one starting with a digit, is prefixed with ``N`` so
+      the identifier matches ``[A-Za-z][A-Za-z0-9_]*`` and never starts with
+      ``_``.
+    """
+    cleaned = re.sub(r"[^A-Za-z0-9_]", "_", (value or "").strip())
+    cleaned = re.sub(r"_+", "_", cleaned)
+    cleaned = re.sub(r"_(\d)", r"\1", cleaned)
```
kevinschaper: Non-blocking edge case: this silently merges `foo_2bar` and `foo2bar` into the same identifier `foo2bar`. Two distinct LinkML names that differ only by an underscore-before-digit would collide here. Two options if you'd like to harden:

1. Document the collapse explicitly in the docstring so it's visible to schema authors, or
2. Use a non-destructive transform — e.g. `foo_2bar` → `foo_N2bar` (reusing the leading-digit prefix rule) — so distinguishability is preserved.

Not blocking, but worth at least the docstring note since the surprise would only show up at protoc time with a confusing duplicate-field error.

kevinschaper (approved):
Thanks @noelmcloughlin! I added a couple of inline comments

packages/linkml/src/linkml/generators/protogen.py
noelmcloughlin: both comments applied - thanks for review.

Issue #772, opened 2022-05-12
allow proto schema generation by yaml file/domain instead of one large .proto file


sierra-moxon:
**Describe the solution you'd like**
Instead of generating one large .proto file out of many LinkML yaml files, allow user to optionally have one .proto file per domain with import statements in the .proto file that reflect the import statments in the individual yaml files.

For GA4GH - @julesjacobsen

**When will use cases depending on this become relevant?** Select from the options below:
• Mid-term - 2-4 months

**Additional context**
https://github.com/phenopackets/phenopacket-schema/blob/master/src/main/proto/phenopackets/schema/v2/phenopackets.proto

noelmcloughlin:
Could leverage reusable design of #3686

PR #3559, opened 2026-05-22
feat(protogen): proto3 as a wire format (part a)

noelmcloughlin:
## Summary

**proto3 as a wire format** for LinkML data requires following:

1. A `.proto` schema that captures the LinkML semantics needed for round-trip:
--  cardinality, inlined vs reference, oneof choices, map keys, imports, deprecation.
2. A loader / dumper pair so a LinkML object can be serialised to proto3 wire bytes and read back.
3. Stable field numbers across schema revisions; wire compatibility depends on numbers never silently moving

This PR begins the work:
- non-inlined class refs resolving to the identifier scalar of the range class. instead of nested message.
- identifier slot pinned to field 1 with // identifier annotation when no explicit ranks exist.
- // required comment emitted for required slots.

References:
https://github.com/linkml/linkml/issues/772
[phenopackets.proto](https://github.com/phenopackets/phenopacket-schema/blob/master/src/main/proto/phenopackets/schema/v2/phenopackets.proto)

## How was this tested?

Full suite
Added wire specific proto tests

## Areas of uncertainty

I am not expert on proto wire format - have reviewed [docs](https://protobuf.dev/programming-guides/encoding/) but accuracy of implementation follows adopters (I have use case)

## Checklist

## AI Assistance
LLM assisted


kevinschaper (changes requested):
It looks like this PR needs some tidying up to remove the notebook and generated model files that are mixed in (I'm guessing those changes aren't actually necessary here?)

Also, do we need to actually bring grpcio-tools in for the tests so that pytest.importorskip("grpc_tools") isn't always skipped on CI?

noelmcloughlin:
I've noticed these notebooks, generated by some pattern in local testing. Can they be git excluded?

PR #3570, opened 2026-05-25
fix(docgen): omit empty Cardinality table in slot pages

### LINKED ISSUE #3569: Docgen emits empty Cardinality table that swallows the next heading
The docgen Jinja templates render the "Cardinality and Requirements" section as a header plus an empty table (header row + separator, no body rows) and no trailing blank line. The Python-Markdown `tables` extension then consumes the next line as a table row, so the following `### Slot Characteristics` heading is rendered as literal text inside a one-cell row of the empty table instead of as a section heading.

## Context

The bug lives in the docgen templates in this repo (under `linkml/generators/docgen/`). The symptom is visible in the generated `linkml-model` docs — every slot/class page that has no cardinality data shows it. See https://linkml.io/linkml-model/dev/docs/readonly/ under "Cardinality and Requirements": the next heading ("Slot Characteristics") is pulled into the empty table cell and the rest of the page renders inside what the parser thinks is still a table.

This is adjacent to linkml/linkml#3520 (broader 1.11.x docs rendering breakage). That one was worked around in `linkml-model` PR #255 (commit `bfd963c7`) by adding `attr_list` and `md_in_html` to `markdown_extensions` in `mkdocs.yml`. Once that landed, this empty-table case became the next visible failure.

Two clean fixes, either is fine:

(a) skip the "Cardinality and Requirements" section entirely when the table body would be empty
(b) always emit a blank line after the table separator regardless of whether body rows are present

## Examples

From `docs/source/docs/readonly.md` in `linkml/linkml-model` (lines 45-53):

```
### Cardinality and Requirements

| Property | Value |
| --- | --- |
### Slot Characteristics

| Property | Value |
| --- | --- |
| Inherited | Yes |
```

Rendered output: the "Cardinality and Requirements" table shows a single cell containing the literal text `### Slot Characteristics`, and the actual Slot Characteristics section never appears as a heading. Reproducible on any slot/class page in the linkml-model dev docs.

## Related

- linkml/linkml#3520 — broader 1.11.x docs rendering bug
- linkml/linkml-model#255 (`bfd963c7`) — `attr_list` + `md_in_html` workaround in `mkdocs.yml`
- https://linkml.io/linkml-model/dev/docs/readonly/ — live example of the misrender

matentzn:
When a slot has no cardinality data, the slot template was still emitting the `### Cardinality and Requirements` heading plus an empty `| Property | Value | / | --- | --- |` table. Python-Markdown's `tables` extension then consumed the next `### Slot Characteristics` heading as a one-cell row of that empty table, breaking the rest of the page (visible in every slot/class page in the [linkml-model dev docs](https://linkml.io/linkml-model/dev/docs/readonly/)).

Guard the section behind a `has_cardinality` check, matching the existing `has_slot_chars` / `has_basic_constraints` pattern in the same template. Added an assertion in `test_core_element_properties` against the existing `base_property` fixture (an `inherited: true` slot with no cardinality data) to lock the fix in.

Closes #3569.

### DIFF
--- a/packages/linkml/src/linkml/generators/docgen/slot.md.jinja2
+++ b/packages/linkml/src/linkml/generators/docgen/slot.md.jinja2
@@ -105,7 +105,11 @@ Alias: {{ element.alias }}
 {%- if element.is_grouping_slot %}
 | Is Grouping Slot | Yes |
 {%- endif %}
+{%- set has_cardinality = element.required or element.recommended
+    or element.multivalued or element.minimum_cardinality is not none
+    or element.maximum_cardinality is not none or element.exact_cardinality is not none %}
 
+{%- if has_cardinality %}
 ### Cardinality and Requirements
 
 | Property | Value |
@@ -128,6 +132,7 @@ Alias: {{ element.alias }}
 | Exact Cardinality | {{ element.exact_cardinality }} |
 {%- endif %}
 
+{% endif %}
 {%- if element.multivalued and (element.list_elements_unique is not none or element.list_elements_ordered is not none) %}
 ### List/Collection Properties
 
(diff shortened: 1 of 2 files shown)


dalito (approved):
This looks good to me.

sujaypatil96 (approved):
Looks good to me! 🚀

PR #3575, opened 2026-05-27
fix(prefixes): ensure curiemaps are not lost in merged linkml schemas

### LINKED ISSUE #3574: prefixes declared by sub-schema are dropped during (merge) import
**Describe the bug**
When a parent schema imports a child schema that declares additional prefixes, generators emit prefix bindings only for the prefixes declared on the **root** schema. 

Prefixes declared exclusively on imported sub-schemas were silently dropped.

For `gen-python` the symptom is a `NameError` at import time of the generated module, because permissible-value `meaning:` URIs from imported enums expand to `URIRef(str(<PREFIX_VAR>))` where `<PREFIX_VAR>` was never defined:

-    ` NameError: name 'ISO3166' is not defined`

For other generators the symptom is silently malformed output: unbound CURIE prefixes in SHACL / OWL graphs, missing `prefixes:` entries in YARRRML mappings, and an incomplete `prefixes:` block in the merged `gen-yaml` output.

1.11.1

**Cause**
Two independent code paths discarded imported prefixes:

1. **`SchemaLoader` path** (generators with `uses_schemaloader = True`): 

`linkml.utils.mergeutils.merge_namespaces` contained a commented-out assignment so the absolute-prefix branch never wrote imported prefixes onto the target schema.

2. **`SchemaView` path** (generators with `uses_schemaloader = False`):

   `SchemaView` exposes imported prefixes via `namespaces()` /`imports_closure()`, but never copies them onto  `schemaview.schema.prefixes`. Generators that iterate `schema.prefixes.values()` directly (to drive `graph.namespace_manager.bind(...)`, etc.) therefore saw only the root prefixes.

## Reproducer

`/tmp/child.yaml`:

  ```yaml
  id: https://example.org/child
  name: child
  prefixes:
    ex: https://example.org/child/
    iso3166: https://www.iso.org/iso-3166-country-codes.html#
  default_prefix: ex
  default_range: string
  imports:
    - linkml:types

  enums:
    CountryStyleEnum:
      permissible_values:
        Country:
          meaning: "iso3166:"
  ```

`/tmp/parent.yaml`:

  ```yaml
  id: https://example.org/parent
  name: parent
  prefixes:
    parent: https://example.org/parent/
  default_prefix: parent
  default_range: string
  imports:
    - linkml:types
    - ./child

  classes:
    Holder:
      attributes:
        style:
          range: CountryStyleEnum
  ```

Error:

```bash
gen-python /tmp/parent.yaml > parent_out.py
python -c "import importlib.util; \
  spec=importlib.util.spec_from_file_location('m','parent_out.py'); \
  m=importlib.util.module_from_spec(spec); spec.loader.exec_module(m)"
# NameError: name 'ISO3166' is not defined
```

noelmcloughlin:
## Summary

Fixes #3574

Prefixes defined exclusively by a sub-schema must be retained during mergeimports

## How was this tested?

New tests
full test suite

## Areas of uncertainty

To isolate snapshot noise review:
-  https://github.com/linkml/linkml/commit/1569726b3d74747b357e9feb8121c105fea8ade6 
-  https://github.com/linkml/linkml/pull/3575/changes/cdcca32964d3c1dadef20f9a64e01d113176194d

## Checklist

## AI Assistance

LLM assisted, but understand the code changes


rbm1996:
Ran `materialize_prefixes()` directly (verbatim) against a schema that triggers a related-but-distinct staleness bug in `SchemaView.namespaces()` itself (`@lru_cache(None)`, called from inside `imports_closure()`, can cache an incomplete snapshot before the full closure loads). It correctly adds the missing prefix to `schema.prefixes`, but `namespaces()` was already poisoned before this function runs, and since it never calls `.cache_clear()`, the mutation is invisible to `expand_curie()`/`get_uri(expand=True)` — so `ShaclGenerator._add_type()` (which doesn't read `schema.prefixes` directly) still emits the malformed CURIE. Details and a runnable repro in #3933. Might be worth adding `schemaview.namespaces.cache_clear()` right after the mutation here, or clearing it as part of `materialize_prefixes()` itself.

packages/linkml/src/linkml/utils/mergeutils.py
```diff
@@ -99,16 +100,28 @@ def merge_namespaces(target: SchemaDefinition, mergee: SchemaDefinition, namespa
...
         if (
             prefix.prefix_prefix in target.prefixes
             and target.prefixes[prefix.prefix_prefix].prefix_reference != prefix.prefix_reference
         ):
             raise ValueError(f"Prefix: {prefix.prefix_prefix} mismatch between {target.name} and {mergee.name}")
+        if prefix.prefix_prefix not in target.prefixes:
+            # Propagate prefixes declared only in imported sub-schemas into the
+            # root schema's prefixes dict. Generators iterating target.prefixes
+            # need these so emitted content referencing the prefix resolves.
+            target.prefixes[prefix.prefix_prefix] = Prefix(prefix.prefix_prefix, prefix.prefix_reference)
```
kevinschaper: This block is what makes the pre-existing `raise` three lines up reachable, and it turns a warning into a hard failure for schema trees that load today.

`merge_namespaces` has always had that mismatch check, but it compared against `target.prefixes`, which never received imported prefixes — the propagation was the commented-out code you're restoring here. So the guard was effectively dead. Now that imported prefixes land in `target.prefixes`, a parent importing two siblings that declare the same prefix differently stops loading.

Minimal repro — `top.yaml` imports `c1.yaml` and `c2.yaml`, both declaring `shared:` with different references:

```
clean origin/main :  MERGED OK; shared = None       (loads, warns only)
this PR           :  ValueError: Prefix: shared mismatch between top and c2
```

Both builds emit the pre-existing `... is already mapped to ... - Overriding with mapping to ...` warning; only this one then dies.

I think raising is probably *right* — silently picking one of two conflicting definitions is worse, and that's the bug class this PR is fixing. But trailing-slash vs hash variants of the same prefix are common in real import trees, so this is a breaking change for existing schemas and it isn't mentioned in the description, has no test, and has no migration note.

Could you make it an explicit decision? Three defensible options: keep the raise and document it as breaking; warn-and-keep-first; or gate the strictness behind a flag. Whichever you pick, it wants a test that pins the chosen behaviour — right now nothing exercises the error path.

tests/linkml/test_utils/test_schemaloader.py
```diff
@@ -179,21 +179,35 @@ def test_importmap_in_memory_dict_not_mutated():
 
 def test_importmap_mixes_dict_and_path_values(input_path):
     """A single importmap can contain both a dict and a file-path entry."""
+    # Use a base dict whose prefixes do not collide with import_test_4.yaml
+    # (which declares ``base: http://example.org/``); the in-memory base
+    # schema declares ``base: http://example.org/base/``.
+    base_dict_no_conflict = {
```
kevinschaper: This is the same finding as my comment on `mergeutils.py`, and it's the evidence that the breaking change bites in practice.

This test previously covered sibling imports where the in-memory base declares `base: http://example.org/base/` and `import_test_4.yaml` declares `base: http://example.org/`. Under this PR that combination raises, so the fixture was swapped for a non-colliding one — the comment says so directly.

Two problems with resolving it this way:

- The scenario the test used to cover is no longer covered by anything, and nothing tests the new `ValueError` path either. The behaviour change is invisible in the test suite in both directions.
- `CLAUDE.md` is explicit: *"Do not 'fix' issues by changing or weakening test conditions."*

Worth knowing this test arrived in #3706, which merged three days ago — so this is fresh coverage being narrowed rather than an old fixture being tidied.

Suggest restoring the original fixture and either asserting the new `ValueError` (if raising is the decision) or asserting it still loads (if warn-and-pick-first is). Either way the conflict case should stay exercised.

packages/linkml/src/linkml/utils/mergeutils.py
kevinschaper, line 124: This block is what makes the pre-existing mismatch check above it start firing — but the mechanism is across calls rather than within this one, so it's worth spelling out.

`merge_namespaces` runs once per imported schema. On main the propagation you're restoring here is commented out, so `target.prefixes` only ever holds the *target's own* prefixes. The check at the top of the pass therefore compares against a dict that never contains an import's prefixes, and effectively never fires.

With this block in place, merging the first sibling writes its prefix into `target.prefixes`; the **next** call — merging the second sibling — then hits that check and raises. So the guard becomes reachable for a new class of schemas: a parent importing two siblings that declare the same prefix differently.

Verified both sides, with `top.yaml` importing `c1.yaml` and `c2.yaml` where both declare `shared:` with different references:

```
current origin/main :  loads OK, shared -> None      (prefix silently dropped)
this PR             :  ValueError: Prefix: shared mismatch between top and c2
                         raised from merge_namespaces
```

Note that main's behaviour there is itself the bug you're fixing — the prefix is lost entirely. So this isn't a regression so much as the fix surfacing a conflict that was previously being swallowed.

I think raising is probably right; silently picking one of two conflicting definitions is worse. But it is a breaking change for schema trees that load today, trailing-slash vs hash variants of the same prefix are common in real imports, and it isn't mentioned in the description, has no test, and has no migration note.

Could you make it an explicit decision? Three defensible options: keep the raise and document it as breaking; warn and keep the first definition; or gate the strictness behind a flag. Whichever you pick wants a test pinning it — nothing currently exercises the error path.

tests/linkml/test_utils/test_schemaloader.py
kevinschaper, line 185: This is the same finding as my comment on `mergeutils.py`, and it's the evidence that the breaking change bites in practice.

This test previously covered sibling imports where the in-memory base declares `base: http://example.org/base/` and `import_test_4.yaml` declares `base: http://example.org/`. Under this PR that combination raises, so the fixture was swapped for a non-colliding one — the comment says so directly.

Two problems with resolving it this way:

- The scenario the test used to cover is no longer covered by anything, and nothing tests the new `ValueError` path either. The behaviour change is invisible in the test suite in both directions.
- `CLAUDE.md` is explicit: *"Do not 'fix' issues by changing or weakening test conditions."*

Worth knowing this test arrived in #3706, which merged three days ago — so this is fresh coverage being narrowed rather than an old fixture being tidied.

Suggest restoring the original fixture and either asserting the new `ValueError` (if raising is the decision) or asserting it still loads (if warn-and-pick-first is). Either way the conflict case should stay exercised.

kevinschaper (changes requested):
Mostly I'm putting this into request changes mode because of the extra generated notebooks etc that likely don't need to be part of the PR. 

I attached a couple of Claude-generated inline comments, and I'm not actually familiar enough with this part of the codebase to be 100% confident that they're valid.

Issue #3581, opened 2026-05-28
CI failing: idna 3.17 broke `uts46data` import in test_rewrite_rules


kevinschaper:
## Bug

Every CI run since `2026-05-28` (including main itself, e.g. [run 26597970992](https://github.com/linkml/linkml/actions/runs/26597970992) on commit `d24624c`) fails the same 4 tests across every Python version:

```
FAILED tests/linkml/test_rewrite_rules/test_rewrite_rules.py::test_rewrite_rules[meta_entries]
FAILED tests/linkml/test_rewrite_rules/test_rewrite_rules.py::test_rewrite_rules[vocab_entries]
FAILED tests/linkml/test_rewrite_rules/test_rewrite_rules.py::test_rewrite_rules[meta_model_entries]
FAILED tests/linkml/test_rewrite_rules/test_rewrite_rules.py::test_rewrite_rules[meta_vocab_entries]
```

with:

```
ImportError: cannot import name 'uts46data' from 'idna.uts46data'
```

## Cause

[`idna 3.17`](https://pypi.org/project/idna/3.17/) was released on `2026-05-28`. It restructured `idna.uts46data` so the legacy `from idna.uts46data import uts46data` symbol no longer resolves. Some transitive dep used by `tests/linkml/test_rewrite_rules/` (most likely `requests` / `urllib3` / `rdflib` via `idna`) still relies on the old shape.

Locally with `idna==3.11` pinned, the failures do not reproduce.

## Impact

- Every PR shows red checks (e.g. #3550 is currently red for this reason, not for anything in the PR).
- The failures predate any PR opened today — they exist on `main` too.

## How to verify

```
uv pip install 'idna==3.16'
uv run pytest tests/linkml/test_rewrite_rules/ -x
```

kevinschaper:
Resolved by upstream dependency updates — closing.

**Root cause:** the `ImportError: cannot import name 'uts46data' from 'idna.uts46data'` came from `url_normalize` (a transitive dep pulled in via the requests-cache network layer), which used the legacy `from idna.uts46data import uts46data` symbol that idna 3.17 removed on 2026-05-28.

**Fix:** `url_normalize` has since been updated (now 2.2.1 in `uv.lock`) to call `idna.encode(..., uts46=True)` instead, which is compatible with idna 3.17. No LinkML code change was required.

**Verification (2026-07-22):**
- Installed: `idna 3.17`, `url_normalize 2.2.1`.
[...]

PR #3582, opened 2026-05-28
fix(deps): bump idna to 3.17 in uv.lock (#3581)

kevinschaper:
## Problem

CI failed on every Linux job (Windows passed) with:

```
ImportError: cannot import name 'uts46data' from 'idna.uts46data'
  at .venv/.../idna/core.py:334
```

Meant to be an immediate unblock of #3581, but doesn't genuinely solve it.

## Root cause (confirmed via CI instrumentation)

The lock pinned idna **3.11**, but the notebook test `test_redo_notebook[ShExPrimerModel.ipynb]` executes a cell:

```
!uv pip install linkml --upgrade --disable-pip-version-check -q
```

`--upgrade` ignores the lock and pulls **idna 3.17** (latest as of 2026-05-28) into the **shared** CI venv. idna 3.17 restructured `uts46data.py` and dropped the legacy `uts46data` symbol. A later partial `uv` restore left 3.11's `core.py` (line 334: `from .uts46data import uts46data`) beside 3.17's `uts46data.py`, so the rewrite-rules test failed when `url_normalize` called `idna.encode(uts46=True)`. This is also why only Linux failed — notebook tests are skipped on Windows (`sys.platform == "win32"`).

A teardown probe caught the file changing mid-session: `243725 -> 234325 bytes` (= idna 3.17's `uts46data.py`), header `# This file is automatically generated by tools/idna-data`.

## Fix

Bump the lock to **idna 3.17** so the base install and the notebook's `--upgrade` converge on one consistent version — no more skew. idna 3.17 is compatible with our code; the 4 `test_rewrite_rules` tests pass on it locally.

## Follow-up

The underlying issue — notebook tests running `!uv pip install … --upgrade` against the shared test venv — is tracked separately and can recur on a future dep release.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1246,11 +1246,11 @@ wheels = [
 
 [[package]]
 name = "idna"
-version = "3.11"
+version = "3.17"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/6f/6d/0703ccc57f3a7233505399edb88de3cbd678da106337b9fcde432b65ed60/idna-3.11.tar.gz", hash = "sha256:795dafcc9c04ed0c1fb032c2aa73654d8e8c5023a7df64a53f39190ada629902", size = 194582, upload-time = "2025-10-12T14:55:20.501Z" }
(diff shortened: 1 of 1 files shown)


matentzn:
Am I stupid or would you have to pin this in toml as well to make the change persistent?

kevinschaper:
> Am I stupid or would you have to pin this in toml as well to make the change persistent? 

You are not stupid, possibly this PR is. I just wanted a totally uncontroversial boring change that would un-block the approved PRs in the queue. Pinning in the pyproject.toml is more effective but also will affect anyone who imports linkml. 

I don't think idna actually needs to be pinned to one version or the other, I think it's really just the change that happens in the notebooks that causes the problem.

matentzn (approved):

PR #3587, opened 2026-05-29
feat(rustgen): emit serde skip_serializing_if for optional/list/mapping fields

kervel:
## Summary

This makes generated rust classes behave like linkml yaml dumper: don't emit values if they are not set. I intentionally didn't go for pydantic model (exclude_unset) because that would require to implement my own set tracking mechanism which would make everything much heavier ...

## How was this tested?

* unit test in this repo
* rust-linkml-core updated with code generated by this MR, verified working correctly.

## Areas of uncertainty

* exclude_unset semantics would be nice but i think for a rust generator its out of scope and would require keeping an extra flag for every single attribute, which would destroy the ability to create idiomatic rust structs.

## Checklist

## AI Assistance

Yes i did, but i reviewed by hand.


kevinschaper (approved):
Thanks @kervel!

PR #3588, opened 2026-05-29
fix(rustgen): per-key merge for optional map fields

kervel:
The generated Merge impls are used by the rust linkml core to turn multiple slotdefinitions into a slotview, amongst other. They give you an equivalent of pythons 'update' method. That feature already existed, this PR fixes missing merging for optional dictionaries.

## Summary

Optional map-valued fields (Option<HashMap<K, V>>) were merged with overwrite_except_none, which replaces the whole inherited map when an override is present. Map fields should merge per-key so that, e.g., a slot_usage override of annotations folds into the inherited annotations instead of dropping them.

Optional maps now use a generated option_map_overwrite strategy (right wins per key, left-only keys preserved, None on the right is a no-op); the helper is emitted into the crate when needed. Required maps keep merge::hashmap::overwrite and non-map fields are unchanged.

## How was this tested?

See Kapernikov/rust-linkml-core#98

rust-linkml-core already uses code with this patch.
a unit test was added.

## Areas of uncertainty

None

## Checklist

## AI Assistance

Yes, but reviewed / edited myself.


kevinschaper (approved):
Thanks @kervel, sorry for the delay in reviewing

PR #3589, opened 2026-05-29
javagen: Enforce typing constraints for refined inherited slots.

### LINKED ISSUE #3530: Generated Java code does not enforce refined type constraints
The code generated by the Java generator does not enforce the typing constraint when a derived class refines a parental attribute so that it expects a more specialised range.

To illustrate, consider the following schema:

```yaml
id: https://example.org/test
name: test
prefixes:
imports:
  - linkml:types

slots:
  bar:
    range: Bar

classes:

  Foo:
    slots:
      - bar

  Bar:
    attributes:
      name:
        range: string

  ExtendedFoo:
    is_a: Foo
    slot_usage:
      bar:
        range: SpecializedBar

  SpecializedBar:
    is_a: Bar
    attributes:
      length:
        range: integer
```

The class `Foo` has a `bar` slot whose range is set to the `Bar` class (i.e. that attribute expects a value that is an instance of the `Bar` class). The derived class `ExtendedFoo` refines the inherited `bar` slot so that its range is, within the derived class, set to the `DerivedBar` class (itself derived from `Bar`).

This is of course an artificial example, but for a real-world example of this situation, you need look no further than LinkML’s own meta-schema, where the `definition` class has a `is_a` slot whose range is set to `definition` (to represent the “parent” definition of the present definition). In several of the derived classes, the `is_a` slot is refined so that its range is set to a more specialised subclass of `definition`; for example, in `class_definition`, `is_a` is refined to expect only a `class_definition` (expressing the fact that a class definition can only have another class definition as parent), while in `slot_efinition` `is_a` is refined to expect only a `slot_definition` (likewise).

Such a specialisation of an inherited slot is currently completely ignored by the Java generator (and at least some of the generators for other statically typed object-oriented programming languages).

In Java, the model above would yield the following `Foo` class:

```java
public class Foo {
    private Bar bar;

    public void setBar(Bar value) {
        bar = value;
    }

    public Bar getBar() {
        return bar;
    }
}
```

while the `ExtendedFoo` class would simply look like this:

```java
public class ExtendedFoo extends Foo {
}
```

This does _not_ allow to enforce the constraint that the `bar` slot, within the context of the `ExtendedFoo` class, should only contain a `SpecializedBar` value. There is nothing preventing user code to assign a `Bar` value to the slot:

```java
ExtendedFoo f = new ExtendedFoo();
f.setBar(new Bar());
```

In turn, this means that whenever the slot is queried, there is no guarantee that the value returned by the `getBar()` accessor will indeed be a `SpecializedBar` and not any other type of `Bar`. Any “naive” use of the accessor like this:

```java
SpecializedBar b = (SpecializedBar) f.getBar();
```

could cause an “invalid cast” error, because downcasting is not possible in Java.

Since the refined type constraint is explicitly stated in the original LinkML schema, it _should_ be enforced by the generated code, without leaving the user exposed to possible misuse and subsequent casting errors.

I believe that, for the `ExtendedFoo` class, the generated code should look like the following:

```java
public class ExtendedFoo extends Foo {

    // Overloaded write accessor that accepts the specialised type; not
    // strictly necessary, but makes the intention clearer
    public void setBar(SpecializedBar value) {
        // No type checking necessary, we already know the value is of
        // the correct type
        super.setBar(value);
    }

    // Overridden write accessor (called instead of the write accessor
    // in the parental Foo class) that accepts the non-specialised type.
    @Override
    public void setBar(Bar value) {
        // Explicit dynamic type checking
        if ( !(value instanceof SpecializedBar) ) {
            throw new IllegalArgumentException("Invalid value for the 'bar' slot");
        }
        super.setBar(value);
    }

    // Overridden read accessor that returns the specialised type
    @Override
    public SpecializedBar getBar() {
        // This cast is safe because only a SpecialisedBar could have
        // been effectively assigned to the slot
        return (SpecializedBar) super.getBar();
    }
}
```

gouttegd:
## Summary

Fixes #3530 

This PR updates the Java code generator so that it can generate code that can correctly deal with the case of “refined inherited slots”.

It does that in three steps:

(A) First, the `OOCodeGen` class is augmented with a `get_refined_ranges` method that allows to query, for a given slot in a given class:

* whether the class is refining the range of the slot compared to its parent class (with parameter `upwards=True`), and if so, to obtain the list of successive refinements that occurs in the ancestor hierarchy (up to the first class that originally defines the slot);
* whether the slot is refined in any subclass (with parameter `upwards=False`), and if so, to obtain the list of refinements that occurs in all subclasses (down to the leaf classes of the class tree).

(B) The `OOField` class, representing a field in a class, is augmented with an additional member variable named `refined_ranges`. When the field is an _inherited_ field, that variable is set with the list of refinements as obtained with `get_refined_ranges(upwards=True)`. This provides the code templates with a way to know whether an inherited field (which would normally be completely ignored by the Java code generator, since it has been dealt with when the code for the parent class that defines it has been generated) is being refined in the current class and therefore requires special treatment.

(C) Lastly, the `org.incenp.linkml` template variant is updated to use the new features provided by `OOCodeGen` to generate suitable accessors that can enforce the type constraints of any refined inherited slot. For single-valued slot, the generated code follows the idea presented in the original comment of issue #3530. For multi-valued slots, it follows the idea presented in [this comment](https://github.com/linkml/linkml/issues/3530#issuecomment-4499677430).

## How was this tested?

Unit tests have been added to the existing `test_javagen.py` module to cover parts (A) and (B) described above.

The actual code generation part (C) has been tested as part of the LinkML-Java runtime (https://github.com/gouttegd/linkml-java/pull/2).

## Areas of uncertainty

If any Java expert has a better idea for how to deal with multivalued refined inherited slots – in particular, an idea that does _not_ involve wildcard generics –, I am all ears.

## Checklist

## AI Assistance

None.


kevinschaper (approved):
thanks @gouttegd, really nice improvement to OOCodeGen

Issue #3592, opened 2026-05-29
EnumDefinitionImpl.__eq__ returns False for matching strings (ValueError)


noelmcloughlin:
Related to #1203

**Describe the bug**
As mentioned in #3590,   `EnumDefinitionImpl` inherits `YAMLRoot.__eq__` method. As a result a cooked enum instance does not compare equal to the raw permissible-value text even though `str(instance) == "TEXT"`:

```python
PreTradeCategoryEnum("INDICATION") == "INDICATION"                                  # False
```

Same as #3590, the issue lies inside `generatedDataClass.__post_init__` > `self._normalize_inlined_as_list/dict` -> `YAMLRoot._normalize_inlined()` call chain.

https://github.com/linkml/linkml/blob/d24624c32876160d85f31a52a3c867ce5b440cf3/packages/linkml_runtime/src/linkml_runtime/utils/yamlutils.py#L111

When a keyed-slot loader compares the cooked-side key (already promoted to the enum type - see #3590) against the raw YAML key (still a string) we get Value Errors:

```
ValueError: Slot: trade_capture_report_identifier_rules -
  attribute identifier_role value (TRADE_REPORT_ID)
  does not match key (TRADE_REPORT_ID)
```

**Visually same but types differ.**

### Reproducer

```bash
python - <<'PY'
from linkml_runtime.utils.enumerations import EnumDefinitionImpl
from linkml_runtime.linkml_model.meta import (
    EnumDefinition, PermissibleValue,
)

class E(EnumDefinitionImpl):
    _defn = EnumDefinition(name="E")
    A = PermissibleValue(text="A")

x = E("A")
assert str(x) == "A"
assert x == "A", "enum instance should compare equal to its text"
PY
```

Observed: `AssertionError`.
1.11.1

Issue #3590, opened 2026-05-29
EnumDefinitionImpl instances are unhashable (TypeError)


noelmcloughlin:
Related to #1203 and maybe fixed by https://github.com/linkml/linkml/pull/3593

**Describe the bug**

The `EnumDefinitionImpl` <--`YAMLRoot` <-- `jsonObj` inheritance receives `__eq__` from imported jsonasobj2 library but we never defined `__hash__` method, so instances become unhashable **(i.e. __hash__=None)** causing Error.

The issue lies inside `generatedDataClass.__post_init__` > `self._normalize_inlined_as_list/dict` -> `YAMLRoot._normalize_inlined()` call chain.

https://github.com/linkml/linkml/blob/d24624c32876160d85f31a52a3c867ce5b440cf3/packages/linkml_runtime/src/linkml_runtime/utils/yamlutils.py#L111

In LinkML a slot can be marked as the key or identifier of its range class.  When such a class appears inline in a parent slot, the inlined collection can be represented as a dictionary keyed by that field - this is called 'inlined-as-dict-slot, and slot is "keyed" (thus `_normalized_inline`).  This code removes duplicates by adding key to "set".
```python
cooked_keys = set()
...
if keyed:
    if key in cooked_keys:
        raise ValueError(...)
    cooked_keys.add(key)
```
But when the keyed-slot key is enum-typed we get:
```bash
TypeError: cannot use 'PreTradeCategoryEnum' as a set element
    (unhashable type: 'PreTradeCategoryEnum')
```

### Reproducer

```bash
python - <<'PY'
from linkml_runtime.utils.enumerations import EnumDefinitionImpl
from linkml_runtime.linkml_model.meta import (
    EnumDefinition, PermissibleValue,
)

class E(EnumDefinitionImpl):
    _defn = EnumDefinition(name="E")
    A = PermissibleValue(text="A")

print(hash(E("A")))
PY
```

**Version of LinkML you are using**
1.11.1

noelmcloughlin:
Reopening due to https://github.com/linkml/linkml/pull/3597 as I need to review

Issue #1203, opened 2023-01-04
Make it easier to work with enumerations in LinkML Runtime


jbussemaker:
**Is your feature request related to a problem? Please describe.**
Working with enumerations (subclasses of `EnumDefinitionImpl`) can be frustrating, especially when already used to how the native Python Enum classes work. For example, value comparison and hashing (for lookups) is either not supported or awkward to use.

**Describe the solution you'd like**
I would like to have support for:
1. Equality comparison between `EnumDefinitionImpl` (with a `code`), `PermissibleValues` and strings
2. Support for hashing of `PermissibleValues`
3. Support stringification
4. Checking whether some enumeration value is a valid value using `in`

For example, consider the following enumeration (automatically generated):
```python
class EnumValues(EnumDefinitionImpl):
    A = PermissibleValues(text="A")
    B = PermissibleValues(text="B")
```

Then the following tests should pass:
```python
# 1. Equality comparison
assert EnumValues.A == EnumValues("A")
assert EnumValues("A") == EnumValues.A
assert EnumValues.A == "A"
assert "A" == EnumValues.A
assert EnumValues("A") == "A"
assert "A" == EnumValues("A")

# 2. Hashing support
assert "A" in {EnumValues.A, EnumValues.B}
assert EnumValues.A in {EnumValues.A, EnumValues.B}
assert EnumValues("A") in {EnumValues.A, EnumValues.B}
assert "A" in {EnumValues("A"), EnumValues("B")}
assert EnumValues.A in {EnumValues("A"), EnumValues("B")}
assert EnumValues("A") in {EnumValues("A"), EnumValues("B")}

# 3. Stringification
assert str(EnumValues.A) == "A"
assert str(EnumValues("A")) == "A"

# 4. Checking valid values
assert EnumValues.A in EnumValues
assert EnumValues("A") in EnumValues
assert "A" in EnumValues
```

**How important is this feature?** Select from the options below:
• Medium - can do work without it; but it's important (e.g. to save time or for convenience)

**When will use cases depending on this become relevant?** Select from the options below:
• Short-term - 2-4 weeks

**Additional context**
Suggested modifications for `EnumDefinitionImpl` (apply `__contains__` also to `EnumDefinitionMeta`):
```python
def __eq__(self, other):
    if isinstance(other, EnumDefinitionImpl):
        return other.code == self.code
    if isinstance(other, PermissibleValues):
        return other == self.code
    return vars(self) == vars(other)

def __hash__(self):
[...]

PR #3593, opened 2026-05-29
fix(enumerations): walk MRO in EnumDefinitionMeta __contains__ / __getitem__

### LINKED ISSUE #3591: EnumDefinitionMeta ValueError - MRO ignored by __contains | __getitem methods
Related to #1203

**Describe the bug**

These two `EnumDefinitionMeta` methods only look up permissible values in `cls.__dict__` which bypasses [Method Resolution Order](https://docs.python.org/3/glossary.html#term-method-resolution-order) entirely (risking ValueErrors):

```python
def __contains__(cls, item) -> bool:
    return item in cls.__dict__

def __getitem__(cls, item):
    return cls.__dict__[item]
```

`EnumDefinitionImpl.__init__` then uses `__contains__` to validate codes:

```python
if key not in self.__class__ and self._defn.code_set:
    ...
elif key not in self.__class__:
    raise ValueError(
        f"Unknown {self.__class__.__name__} enumeration code: {key}"
    )
```

`pythongen` emits empty wrapper subclasses for identifier slots whose range is an enum:

```python
class FIXDatatypeDatatypeName(FIXDatatypeName):
    pass
```

The wrapper inherits the parent enum's permissible values via the MRO but `__contains__` does not walk it, so every lookup against the wrapper fails:
```
ValueError: Unknown FIXDatatypeDatatypeName enumeration code: Country
```

### Reproducer

```bash
python - <<'PY'
from linkml_runtime.linkml_model.meta import (
    EnumDefinition, PermissibleValue,
)

class Parent(EnumDefinitionImpl):
    _defn = EnumDefinition(name="Parent")
    A = PermissibleValue(text="A")

class Wrapper(Parent):
    pass

print(Parent("A"))   # works
print(Wrapper("A"))  # ValueError: Unknown Wrapper enumeration code: A
PY
```

1.11.1

truffle-dev:
## Summary

Fixes #3591.

`EnumDefinitionMeta.__contains__` and `__getitem__` looked only at `cls.__dict__`, so subclasses inherited no permissible values from their parent enumeration class. `pythongen` emits empty wrapper subclasses for identifier slots whose range is an enum:

```python
class FIXDatatypeDatatypeName(FIXDatatypeName):
    pass
```

The wrapper inherits the parent's permissible values via the MRO, but the metaclass operators did not walk it, so every lookup against the wrapper raised:

```
ValueError: Unknown FIXDatatypeDatatypeName enumeration code: Country
```

This walks `cls.__mro__` in both operators. `__getitem__` now raises `KeyError` for missing keys, matching the implicit behaviour of the prior `cls.__dict__[item]` lookup; tests cover that path so the contract stays explicit.

## How was this tested?

`tests/linkml_runtime/test_utils/test_enumerations.py` (new) parametrises across a parent, an empty wrapper subclass (the reporter's shape), and a grandchild that defines its own value. It covers:

- `key in cls` finds inherited values; returns False for missing keys.
- `cls[key]` returns the nearest ancestor's `PermissibleValue`; raises `KeyError` for missing keys.
- `cls("inherited_code")` (the reporter's repro) succeeds on the wrapper and grandchild.

Stash-bisect confirms the new tests fail without the source change (6 fail / 12 pass) and pass with it (18/18). Full `tests/linkml_runtime/` suite: 1766 passed, 271 skipped, 2 xfailed locally — no regressions.

## Areas of uncertainty

- `EnumDefinitionMeta.__setitem__` is not touched. The duplicate-key guard there also only checks `cls.__dict__` (so it cannot detect inherited collisions), but `cls.__dict__[key] = value` already raises `TypeError: 'mappingproxy' object does not support item assignment`, so the write path itself is dead in the base class. Out of scope for this fix; happy to address in a follow-up if the maintainers would like the duplicate guard widened to the MRO and the assignment switched to `setattr(cls, key, value)`.

## Checklist


noelmcloughlin:
Can be linked to https://github.com/linkml/linkml/issues/3591

noelmcloughlin (approved):
LGTM - I have similar fix in my parallel PR

When #3582 is merged tests should pass

truffle-dev:
Thanks for the cross-link. The PR body opens with `Fixes #3591` so the GitHub auto-link should fire on merge; the issue sidebar may just need a maintainer to associate it.

For visibility: your #3596 also walks the MRO for `__contains__` / `__getitem__` on the same file, as part of a broader runtime fix that covers #3590, #3592, and the #1203 workaround. This PR is narrowly scoped to #3591 only (plus the matching `__getitem__` `KeyError` contract).

Happy either way — this can land first as the focused fix and #3596 follow with the rest, or I'll close in favor of #3596 if maintainers prefer the broader landing.

noelmcloughlin:
Narrow scope is good - we can merge this PR. Thanks!!

truffle-dev:
For reference: the test failure is `ImportError: cannot import name 'uts46data' from 'idna.uts46data'`, which fires across other unrelated PR branches in the same window — e.g. docs/clarify-oocodegen-granularity at job 78641424366 fails identically on Python 3.13. Looks like an upstream idna packaging issue that landed after main last ran at d24624c (2026-05-29T03:29Z); every PR-branch CI run since 2026-05-29T16:11Z has hit the same import. Nothing in this diff touches idna or the rewrite-rules test — the change is scoped to `EnumDefinitionMeta.__contains__` and `__getitem__` walking the MRO.

truffle-dev:
Rebased onto upstream/main `f9798a5` to pick up [`e270b6f`](https://github.com/linkml/linkml/commit/e270b6fc7) (bump idna 3.11→3.17 in uv.lock), which should clear the ubuntu `ImportError: cannot import name 'uts46data' from 'idna.uts46data'` failure on 3.10-3.14 that was unrelated to this change. New head: `0bd82ffac`. CI is re-running.

PR #3596, opened 2026-05-29
fix(enumerations): fix _hashable_, _eq_, and MRO contains_/_getitem_

noelmcloughlin:
## Summary

Fixes `linkml-runtime` issues  #3590 #3591 #3592

- Ships only the instance-level `EnumDefinition _eq_ / _ne_ / _hash_` and MRO woak into `EnumDefinitionMeta`
- It coves **instance-form** half of #1203 but full fix is separate PR.

## How was this tested?

Full local testing
Added new unit tests based on examples in #1203

## Checklist

## AI Assistance

LLM assisted. I have reviewed the PR a few times so understand what we are trying to fix.


packages/linkml_runtime/src/linkml_runtime/__init__.py
```diff
@@ -3,8 +3,17 @@
...
 from linkml_runtime.utils.curienamespace import CurieNamespace
+
+# Install equality / hashing / string patches on the metamodel's
+# ``PermissibleValue`` dataclass so that enums behave intuitively at runtime
+# (see linkml/linkml#1203).  Done here because both ``enumerations`` and
+# ``linkml_model.meta`` are guaranteed to be fully loaded by this point.
+from linkml_runtime.utils.enumerations import _patch_permissible_value
 from linkml_runtime.utils.schemaview import SchemaView
 
+_patch_permissible_value()
```
amc-corey-cox: This installs a process-wide mutation of the public `PermissibleValue` metamodel class — `__eq__`, `__ne__`, `__hash__`, and `__str__` get redefined at import time for *every* importer of `linkml_runtime`, not just enum call sites. Two implications worth weighing:

- It narrows `PermissibleValue` equality from field-wise to text-only globally, so two permissible values with the same `text` but different `meaning`/`description` now compare equal (and hash-collide). I couldn't find an in-repo path that relies on the field-wise behavior, so I don't think anything here breaks — but the change isn't opt-in, and it reaches downstream consumers (SSSOM-py, etc.) that import `linkml_runtime`. `str(pv)` also changes shape globally (verbose → just the text).
- The three filed bugs this PR targets — #3590 (unhashable), #3591 (MRO), #3592 (`== str`) — are all about the *instance* form (`E("A")`) and are fully fixed by the `EnumDefinitionImpl.__eq__/__hash__` + MRO changes alone, without this patch. The patch only adds the *bare-attribute* half of #1203 (`MyEnum.A == "A"`).

Notably, #3597 appears to drop this entirely: it reaches the same #1203 contract by promoting bare `PermissibleValue` class attrs into real `EnumDefinitionImpl` instances (`_promote_permissible_values`), and its docstring says it "obsoletes the historical `_patch_permissible_value` runtime monkey-patch." So this reads as a temporary bridge that the very next PR removes.

Would it make sense to trim the global patch from this PR — keeping the instance-level `__eq__`/`__hash__` + MRO walk, which cleanly fix #3590/3591/3592 — and let #3597's promotion deliver the bare-attribute ergonomics? That avoids ever shipping a global mutation of the metamodel, and the two PRs stop being competing strategies for the same problem.

One practical note if you do trim: several assertions in `test_issue_1203.py` exercise the *bare attribute* `EnumValues.A` (equality with a string, `hash(...)`, `in {...}`, `str(...)`) and depend on this patch — they'd belong with #3597, where promotion makes `EnumValues.A` an instance. The instance-form assertions and the `test_linkml_issue_723.py` changes pass under the instance-level `__eq__` alone.

amc-corey-cox (commented):
This PR fixes three real, filed bugs — #3590 (instances unhashable), #3591 (MRO ignored in __contains__/__getitem__), #3592 (E("A") !="A") — all of which concern the instance form and are fully covered by the new EnumDefinitionImpl.__eq__/__hash__ plus the MRO walk.That part is solid.
My one substantive concern is the _patch_permissible_value() monkey-patch installed at import time (see inline). It's a process-wide mutation of the public PermissibleValue metamodel class — global __eq__/__hash__/__str__, not opt-in, reaching every downstream importer. It isn't needed for any of the three bug fixes; it only adds the bare-attribute half of #1203 (MyEnum.A == "A"). And #3597 already delivers that same behavior more cleanly by promoting members to real instances, explicitly stating it "obsoletes" this patch.
 So the non-patch content here is essentially a subset of #3597, and the patch is a bridge #3597 throws away. Rather than merge a temporary global mutation, should we close this PR and review/merge #3597 instead — folding in the instance-level __eq__/__hash__ + MRO fixes (identical here) and letting promotion handle the rest? If we'd prefer a fast bug-fix release ahead of the larger architecture series, the alternative is to trim the patch from this PR and ship just Layer 1 + MRO. Either way avoids shipping the global patch.
(Heads-up: #3593 from @truffle-dev overlaps the same MRO lines and will need deconfliction whichever lands first.)

amc-corey-cox (changes requested):
See my comment review above - mis-clicked.

packages/linkml_runtime/src/linkml_runtime/__init__.py
noelmcloughlin: Hi @amc-corey-cox  thanks for reviewing - your suggestion makes sense as #3597 is correct path but I was unsure if there would be pushback - but probably nobody will push back on correct fix path.

Could you merge #3593 asap and I will start from that baseline.

noelmcloughlin:
Hi @amc-corey-cox  

> (Heads-up: #3593 from @truffle-dev overlaps the same MRO lines and will need deconfliction whichever lands first.)

We can merge that PR first - I reviewed/approved.

truffle-dev:
Thanks @noelmcloughlin. To keep that ordering cheap I have refreshed #3593 onto current main (clean rebase, no conflicts with the 10 commits since f9798a5; CI re-running now). Once it lands I am happy to help rebase the MRO-adjacent lines here if anything snags.

packages/linkml_runtime/src/linkml_runtime/__init__.py
noelmcloughlin: Hi @amc-corey-cox  I have removed the global patch and addressed the comments.  This PR is now narrow focus and full fix for Issue 1203 will come in separate PR waiting this merge.

truffle-dev (approved):
Following up on the deconfliction thread: I checked this branch against main now that #3593 is in. The rebase landed clean — `__getitem__` and the `__contains__` MRO walk come through unchanged, and the new PermissibleValue / EnumDefinitionImpl coercion layers above the walk rather than duplicating it, so empty wrapper subclasses keep working for the coerced forms too. The `__eq__`/`__hash__` pair both keying on `_code.text` keeps the hash-equality contract consistent with bare strings. From the #3593 side, no snags.

noelmcloughlin:
I have reduced scope. Patch is gone.

amc-corey-cox (approved):
@noelmcloughlin thanks for the rework and extra effort on this. It looks good to me now.

PR #3609, opened 2026-06-04
Fix jsonschema extraslots allowed

### LINKED ISSUE #3608: `extra_slots.allowed` not handled correctly by the JSON-Schema generator
**Describe the bug**
PR #2940 brought support for `extra_slots.allowed` in the JSON-Schema generator.

However, as [noted in the discussion](https://github.com/linkml/linkml/pull/2940#issuecomment-4259767776) of that PR, the case where the class that is marked with `extra_slots.allowed` happens to be the root class of the schema is not handled correctly.

Tip of the main branch as of June 4th, 2026 (after PR #2940 has been merged).

As provided in the discussion in PR #2940:

```yaml
prefixes:
imports:
  - linkml:types
default_range: string

classes:

  Foo:
    attributes:
      foo_name:
      bar:
        range: Bar
    tree_root: true
    extra_slots:
      allowed: true

  Bar:
    attributes:
      bar_name:
```

Note how: (1) the `Foo` class is configured to allow extra slots, and (2) that same class is also marked as the root of the tree.

The following instance data _should_ be valid (there is an extra slot at the root level, which is precisely what the schema above should allow):

```yaml
foo_name: the name of foo
bar:
  bar_name: the name of bar
```

but the validator disagrees:

```
[ERROR] [sample.yaml/0] Additional properties are not allowed ('extra' was unexpected) in /
```

See [this comment](https://github.com/linkml/linkml/pull/2940#issuecomment-4266306472) for a discussion about the cause of the issue. I have a fix already in the pipes, I am creating this issue for reference.

gouttegd:
## Summary

Fixes #3608 and #3611

This PR implements the fix discussed and agreed upon in the comments of PR #2940 (from [this comment](https://github.com/linkml/linkml/pull/2940#issuecomment-4266306472) onward.

Basically, the gist of it is that the value of the top-level `additionalProperties` in the generated JSON-Schema should depend on the `extra_slots.allowed` property of the root class: if the root class is set to allow “extra slots“, then setting `additionalProperties: true` in the definition of the object (which is what the JSON-Schema generator currently does) **is not enough**, because what matters for the root class is _only_ the top-level `additionalProperties`, whose value must therefore be set according to the wishes of the schema author.

Additionally (and again as agreed upon on #2940), the _default_ value of `additionalProperties` is set according to the global `closed` setting of the JSON-Schema generator. (Arguably this global setting is no longer really relevant now that `extra_slots` gives schema authors a much finer control about allowing “extra slots” / “additional properties”, but since it exists, it should be honoured when possible.)

## How was this tested?

Tested locally against the schema provided in #3608. I plan to add that schema to the test suite later on.

## Checklist

- [ ] I have added tests that prove my fix/feature works – Not yet

## AI Assistance

None.

### DIFF
--- a/examples/tutorial/tutorial01/personinfo.json
+++ b/examples/tutorial/tutorial01/personinfo.json
@@ -1,7 +1,7 @@
 {
     "$defs": {
         "Person": {
-            "additionalProperties": false,
+            "additionalProperties": true,
             "description": "",
             "properties": {
                 "age": {
--- a/packages/linkml/src/linkml/generators/jsonschemagen.py
+++ b/packages/linkml/src/linkml/generators/jsonschemagen.py
@@ -920,7 +920,7 @@ class JsonSchemaGenerator(Generator, LifecycleMixin):
         if self.is_class_unconstrained(cls):
             return True
         elif not cls.extra_slots:
-            return False
+            return self.not_closed
         elif cls.extra_slots.allowed is not None:
             return cls.extra_slots.allowed
         elif cls.extra_slots.range_expression:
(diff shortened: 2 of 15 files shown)


gouttegd:
Superseded by #3612

PR #3612, opened 2026-06-05
Fix JSONSchemaGenerator’s setting of `additionalProperties`

### LINKED ISSUE #3608: `extra_slots.allowed` not handled correctly by the JSON-Schema generator

### LINKED ISSUE #3611: `generate json-schema --not-closed` does not do what it is supposed to do
(Related to #3608, but it is in fact a distinct issue, independent of the handling of `extra_slots.allowed`.)

The JSONSchema generator has a global `not_closed` option (exposed on the command line as `--[not-]closed`). It is documented as follows:

> If not closed, then an open-ended set of attributes can be instantiated for any object

`not_closed` is True by default, meaning the generator is supposed to generate schemas that allow for “an open-ended set of attributes for any object”.

However this is not what happens. Consider the following LinkML schema:

```yaml
id: https://example.org/extra-sub
prefixes:
imports:
  - linkml:types
default_range: string

classes:

  Foo:
    attributes:
      foo_name:
      bar:
        range: Bar
    tree_root: true

  Bar:
    attributes:
      bar_name:
```

and consider the resulting JSONSchema, first with `not_closed=False` (so we want a _closed_ schema, not allowing anything that has not been defined):

```json
{
    "$defs": {
        "Bar": {
            "description": "",
            "properties": {
                "bar_name": {
                    "type": [
                        "string",
                        "null"
                    ]
                }
            },
            "title": "Bar",
            "type": "object"
        },
        "Foo": {
            "description": "",
            "properties": {
                "bar": {
                    "anyOf": [
                        {
                            "$ref": "#/$defs/Bar"
                        },
                        {
                            "type": "null"
                        }
                    ]
                },
                "foo_name": {
                    "type": [
                        "string",
                        "null"
                    ]
                }
            },
            "title": "Foo",
            "type": "object"
        }
    },
    "$id": "https://example.org/extra-sub",
    "$schema": "https://json-schema.org/draft/2019-09/schema",
    "description": "",
    "metamodel_version": "1.11.0",
    "properties": {
        "bar": {
            "anyOf": [
                {
                    "$ref": "#/$defs/Bar"
                },
                {
                    "type": "null"
                }
            ]
        },
        "foo_name": {
            "type": [
                "string",
                "null"
            ]
        }
    },
    "title": "extra-sub",
    "type": "object",
    "version": null
}
```

All `additionalProperties` are set to `false`, so the schema will indeed not allow any unknown slot. So far so good.

But now consider the resulting JSONSchema with `not_closed=True` (the default behaviour):

```json
{
    "$defs": {
        "Bar": {
            "description": "",
            "properties": {
                "bar_name": {
                    "type": [
                        "string",
                        "null"
                    ]
                }
            },
            "title": "Bar",
            "type": "object"
        },
        "Foo": {
            "description": "",
            "properties": {
                "bar": {
                    "anyOf": [
                        {
                            "$ref": "#/$defs/Bar"
                        },
                        {
                            "type": "null"
                        }
                    ]
                },
                "foo_name": {
                    "type": [
                        "string",
                        "null"
                    ]
                }
            },
            "title": "Foo",
            "type": "object"
        }
    },
    "$id": "https://example.org/extra-sub",
    "$schema": "https://json-schema.org/draft/2019-09/schema",
    "additionalProperties": true,
    "description": "",
    "metamodel_version": "1.11.0",
    "properties": {
        "bar": {
            "anyOf": [
                {
                    "$ref": "#/$defs/Bar"
                },
                {
                    "type": "null"
                }
            ]
        },
        "foo_name": {
            "type": [
                "string",
                "null"
            ]
        }
    },
    "title": "extra-sub",
    "type": "object",
    "version": null
}
```

The only difference is that the _top-level_ `additionalProperties` has been set to `true`. **This is not enough** to “allow for an open-ended set of properties on **any object**”. This will only allow for an open-ended set of properties **on the root object**, not on any object “below” it. This will _not_ allow an unknown property in the `Bar` object.

To really allow unknown properties on any object, when `not_closed` is True _all_ the `additionalProperties` declarations under `$defs` must be set to `true` as well, not only the top-level one.

gouttegd:
## Summary

Fixes #3608, #3611

This PR first (commit 1) first makes sure that, for every definition (`$defs`) in the generated schema, the value of `additionalProperties` is set to the global `not_closed` setting if the class does not have an explicit `extra_slots` setting. This fixes #3611, by making the `not_closed` parameter count for every single class in the schema and not only for the top-level class.

Of note, this change requires several updates to the test suite (commit 2–5), because the previous (incorrect) behaviour of the JsonschemaGenerator had been baked in at several places in the test suite.

Then (commit 6), the PR makes sure that, when generating the top-level `additionalProperties`, the value of `extra_slots` in the declared top class, if any, is taken into account, thereby fixing #3608.

## How was this tested?

New tests added, existing tests updated.

## Checklist

## AI Assistance

None.


packages/linkml/src/linkml/generators/jsonschemagen.py
```diff
@@ -434,6 +434,12 @@ def __post_init__(self):
     def start_schema(self, inline: bool = False):
         self.inline = inline
 
+        top_additional_properties = self.not_closed
+        if self.top_class:
+            top_class_def = self.schemaview.get_class(self.top_class)
+            if top_class_def is not None:
```
sneakers-the-rat, line 441: Seems like we should raise if there is supposed to be a top class but there isn't?

packages/linkml/src/linkml/generators/jsonschemagen.py
```diff
@@ -920,7 +926,7 @@ def get_additional_properties(self, cls: ClassDefinition) -> bool | JsonSchema:
         if self.is_class_unconstrained(cls):
             return True
         elif not cls.extra_slots:
-            return False
+            return self.not_closed
```
sneakers-the-rat, line 930: I don't think this is correct, the change above is to respect the extra slots spec at the top level, and not_closed I believe is about additionalProperties at the top level (e.g. for a top-level class, can that class have extra slots. For a top-level schema, can that schema have extra objects).

I don't think it should apply to every class. It seems like the old intention of having that switch apply to everything is replaced by being able to declare it on a per-class basis. In general I don't think generators should have knobs on them that change things that are expressed in the schema - they can have knobs that change how things outside the schema are declared, or to accommodate things that can't be expressed in the schema, but imo we shouldn't be overriding schemas when generating.

packages/linkml/src/linkml/generators/jsonschemagen.py
```diff
@@ -434,6 +434,12 @@ def __post_init__(self):
     def start_schema(self, inline: bool = False):
         self.inline = inline
 
+        top_additional_properties = self.not_closed
+        if self.top_class:
+            top_class_def = self.schemaview.get_class(self.top_class)
+            if top_class_def is not None:
+                top_additional_properties = self.get_additional_properties(top_class_def)
```
sneakers-the-rat, line 442: As mentioned in the last issue, I think we should change the behavior here so we can warn if there is a mismatch: 

- set the default value is `not_closed` to `None` so we can tell if its being explictly set,
- if unset, defaults True
- if top class, use top class's value instead
- if not_closed set and top class declares extra_slots, warn if their values differ (I.e. user tried to declare not_closed explicitly, warn that is overridden by the schema)

sneakers-the-rat (commented):
Reviewing on phone so sorry if a bit incomplete, will come back later at keyboard.

I think the basic change is good, needed to align the behavior of the generator and the extra_slots spec. Good spotting the problem. This problem comes from how the generator added additional params that changed the behavior of a schema, which previously couldn't express how to handle extra slots.

Now that schema can do this, I don't think the generator should be able to override that. The default value when a class does not declare extra_slots is to treat it as False. The change to use the generator param if undeclared is now a violation of the described meaning of extra_slots in the meta schema. The param is still useful because it captures something that is not expressible - how to handle extra props at the top level when the top level is a schema (schemas can't declare `extra_slots`) - but IMO I don't think it should override class level extra_slots.

However if there is some reason I'm missing for why we'd want to do that, I'm open to disagreement.

Edit: if we want to make a schema itself open ended, or to set a schema wide default for extra_slots, we should add terms to the meta schema IMO, which would remove the need for this prop altogether

sneakers-the-rat:
am not a json schema expert, but seems unrelated to this PR - would be worth raising a separate issue to discuss!

packages/linkml/src/linkml/generators/jsonschemagen.py
gouttegd: > and not_closed I believe is about additionalProperties at the top level

Not according to how the `not_closed` option is documented in the `jsonschemagen.py` module:

```python
not_closed: bool | None = True
"""If not closed, then an open-ended set of attributes can be instantiated for any object"""
```

There is nothing to suggest this should only apply to the top-level (though the current implementation, in effect, does only apply this setting to the top-level, as I explain in #3611).

> I don't think it should apply to every class.

The comment explicitly says “for any object”. I think the _intention_ is clear, even if the implementation was not correct. By taking that setting into account for every class, this PR merely makes that option behave _as it has been supposed to behave all along_.

> It seems like the old intention of having that switch apply to everything is replaced by being able to declare it on a per-class basis. In general I don't think generators should have knobs on them that change things that are expressed in the schema

Yes, I fully agree that this option no longer makes any sense now that schema authors can explicitly set `extra_slots.allowed` on a per-class basis.

However since the option has existed for a long while, I’d rather leave to the LinkML core developers to decide what they want to do with it (knowing that removing it altogether, or changing its default value to `False` – which whould be a much more sensible default IMHO –, would be a breaking change).

This PR (1) makes the option behave as it should always have been, even before the introduction of `extra_slots` and (2) makes it play nice with `extra_slots`, by ensuring the option does _not_ override what the schema authors have explicitly stated in their schema, since the option is only taken into account when `extra_slots` is not explicitly being used. This I believe is the least disruptive approach.

Again, I’d rather leave any more disrupting approach (such as deprecating `not_closed`) to the core devs.

sneakers-the-rat: Yes, what I'm saying is that it shouldnt behave that way, not that it isn't what the comment says. Previously that behavior would have been ok because there was nothing in the spec that allowed someone to control that from within a schema. Now that there is, it becomes a generator overriding schema properties, which is the problem. It would not be a breaking change if it never performed this function, which is part of the reason I am suggesting we don't make this change.

gouttegd: > Now that there is, it becomes a generator overriding schema properties, which is the problem.

Problem or not, this is not an isolated case. There are other situations where a generator can override the schema. To cite two:

(A) In the very same JSONSchemaGenerator, there is an option `include_range_class_descendants`. If set to `False` (which is the default), it leads the generator to produce a schema that will forbid the use of subclasses whenever a parent class is expected. That is, if I declare a slot to be of range `Foo`, the schema will _not_ accept instance data where the value of the slot is a _subclass_ of `Foo` – it will only accept `Foo` itself. How is that not an overriding of the intentions of the schema’s author?

(B) The Pydantic generator has a `extra_fields` option, which is very similar to the `not_closed` option of the JSONSchemaGenerator (and supposedly it was added for the very same reason: because before the introduction of `extra_slots`, there was no way for schema authors to declare, directly from within the schema, whether unknown properties were allowed or forbidden).

What’s funny about this one is that, in the Pydantic generator, the `extra_fields` option applies to _all classes_, not just the top-level. Exactly as the `not_closed` option of the JSONSchema generator was _supposed_ to work.

Maybe allowing the generators to override the contents of a schema is a bad, but I would argue that different generators using similar options to allow overriding the contents of a schema **in different ways** (with the option in one generator affecting only the top-level while the similar option in another generator affecting all the classes) is much, much worse.

Again, this PR **fixes** the JSONSchema generator to make the `not_closed` option behave (1) like it has been intended to behave and (2) like the similar option `extra_fields` in the Pydantic generator behaves.

That’s all I am willing to do in this PR.

> what I'm saying is that it shouldnt behave that way

Feel free to make the case that #3611 is in fact not a bug, that the current behaviour is the normal one, and that the only bug is in the documentation of `not_closed` (which should not say that the option applies to all classes). Then while you are at it you can similarly make the case that the `extra_fields` option of the Pydantic generator is similarly ill-inspired.

sneakers-the-rat: > I would argue that different generators using similar options to allow overriding the contents of a schema in different ways is much, much worse.

I think they are both bad to do!

My position is pretty simple - generators should not be able to override the schema. There are cases where they do, in this case and in pydantic because the schema didn't support something previously. But now that the schema does, they shouldn't.

I find this conversation somewhat exhausting so I'm going to bow out. Ive made my review.

gouttegd: My position is simple as well: I set out to fix a bug (two bugs, in fact). I am not considering doing any deeper change without input (and agreement!) from some LinkML core devs about what _they_ think should be the correct behaviour and what we should do with the `not_closed` option.

cmungall (approved):
Regarding generator-specific overrides, I created a discussion here https://github.com/orgs/linkml/discussions/3647, that I think should guide us moving forward.

This is a tricky one though, the PR fixes what I think it accurately categorized as a bug; a bug that has been around pre metamodel changes to allow extra slots. But there is a good pragmatic argument for repurposing it in a way that composes nicely with the new schema level extra slots:

> The param is still useful because it captures something that is not expressible - how to handle extra props at the top level when the top level is a schema (schemas can’t declare extra_slots) - but IMO I don’t think it should override class level extra_slots

This is a good point, and I'd suggest that we handle this by:

- extending the metamodel to allow extra slots at the schema level
- provide a *separate* option, *consistently labeled across generators* as a stopgap

I'll make a separate issue for this.

I'm approving this PR, as it fixes a bug (even if it's a bug that implements semantics that could be useful)

packages/linkml/src/linkml/generators/jsonschemagen.py
Silvanoc: @sneakers-the-rat @gouttegd I haven't read all the exchange, but it sounds to me as there might be some points to discuss about the intended behavior. I've found it in different places in the project and having some clarify always
helps on the long-run. Would you open a discussion or bug-report for it?

gouttegd: > Would you open a discussion or bug-report for it?

Err, can do, but what is it exactly you’d like to discuss?

The exchange here between @sneakers-the-rat and myself was not so much about the _behaviour_ of the `not_closed` option than about its mere _existence_. I think we both agreed that the existing behaviour was incorrect, we disagreed on whether it should be fixed or whether the option should be deprecated entirely because it no longer makes sense now that `extra_slots` is available in the meta-model.

So do you want to discuss about whether the JSON-Schema generator should have a `not_closed` option at all, or about what the behaviour of that option should be?

Happy to discuss either way, but it might be better that _you_ open the ticket so that you can clearly state what your concerns are. :)

Silvanoc: There is possibly no need to discuss anything, and simply a **change request issue** to remove `not_closed` is needed. If it's a meaningful thing to do, then I wouldn't let the conclusion die with this PR.

PR #3613, opened 2026-06-07
fix(panderagen): clear error for unmapped type uri without typeof (#3594)

### LINKED ISSUE #3594: panderagen AttributeError on types whose uri is not in TYPE_MAP, and no typeof
---

**Describe the bug**
This is `dataframe_generator.py` bug:   `PanderaDataframeGenerator.map_type` looks the type's `uri` up in a hard-coded `TYPE_MAP`. When no URI is provied, it falls back into`self.schemaview.get_type(t.typeof)` loop, checking for truthy "uri", not never checking if "typeof" is defined.

https://github.com/linkml/linkml/blob/d24624c32876160d85f31a52a3c867ce5b440cf3/packages/linkml/src/linkml/generators/panderagen/dataframe_generator.py#L75

If a type declares only `uri:` and `base:` (no `typeof:`) and the URI is outside the [default TYPE_MAP](https://github.com/linkml/linkml/blob/d24624c32876160d85f31a52a3c867ce5b440cf3/packages/linkml/src/linkml/generators/panderagen/pandera/pandera_dataframe_generator.py#L36) - then `t.typeof` is `None`, and `get_type(None)` returns `None` and recursion `map_type(None) call eventually blows back:
```
AttributeError: 'NoneType' object has no attribute 'uri'
```

The  stack trace deep inside the generator with no indication of which type triggered it due to lack of type checking.

### Reproducer

`/tmp/schema.yaml`:

```yaml
id: https://example.org/repro
name: repro
prefixes:
  ex: https://example.org/repro/
default_prefix: ex
default_range: string
imports:
  - linkml:types

types:
  PositiveCount:
    uri: xsd:positiveInteger
    base: int
    minimum_value: 1

classes:
  Holder:
    attributes:
      n:
        range: PositiveCount
```

REPL:

```bash
gen-pandera /tmp/schema.yaml

# File ".../linkml/generators/panderagen/dataframe_generator.py",
#    line 80, in map_type
#    if t.uri:
# AttributeError: 'NoneType' object has no attribute 'uri'
```

1.11.1

gaoflow:
## Problem

`gen-pandera` crashes with an opaque `AttributeError: 'NoneType' object has no attribute 'uri'` deep inside the generator when a schema declares a type whose `uri` is outside the generator's `TYPE_MAP` and that has no `typeof` to chain from. The traceback gives no hint which type triggered it.

Reproducer (`schema.yaml`):

```yaml
id: https://example.org/repro
name: repro
prefixes:
default_prefix: ex
default_range: string
imports: [linkml:types]
types:
  PositiveCount:
    uri: xsd:positiveInteger   # not in TYPE_MAP
    base: int                  # no typeof
classes:
  Holder:
    attributes:
      n: {range: PositiveCount}
```

```console
$ gen-pandera schema.yaml
...
  File ".../panderagen/dataframe_generator.py", line 80, in map_type
    if t.uri:
AttributeError: 'NoneType' object has no attribute 'uri'
```

## Cause

`map_type` falls back to `map_type(self.schemaview.get_type(t.typeof))` whenever the `uri` lookup misses. When `t.typeof` is `None`, `get_type(None)` returns `None` and the recursive `map_type(None)` blows up at `if t.uri:`. The same happens if `typeof` names a type that doesn't resolve.

## Fix

Only chain up to the parent type when `typeof` is defined **and** resolves to a real type; otherwise fall through to a clear `ValueError` that names the offending type and its `uri`/`typeof`. This keeps the existing behaviour for mapped URIs and valid `typeof` chains, and fails fast with an actionable message for the unmapped case:

```
ValueError: Type 'PositiveCount' cannot be mapped to a pandera type (uri='xsd:positiveInteger', typeof=None)
```

## Tests

`tests/linkml/test_generators/panderagen/test_panderagen_map_type.py` — parametrised over unmapped-uri-without-typeof types (asserts the clear `ValueError`, red before / green after) plus a regression case confirming a `typeof` chain to a mapped base type still resolves. Full `panderagen` suite passes (53).

Closes #3594.

One open question: for a type that carries a Python `base` (e.g. `base: int`) but an unmapped `uri`, the generator could alternatively fall back to that `base` so the schema generates instead of erroring. I kept this PR to the crash fix since `TYPE_MAP` is a curated uri→type map; happy to add a `base` fallback if you'd prefer that direction.


kevinschaper (approved):
Nice fix, thanks @gaoflow!

PR #3614, opened 2026-06-07
fix(linter): don't flag None range on any_of/exactly_one_of slots (#3477)

### LINKED ISSUE #3477: linkml-lint no_undeclared_ranges false-positives on slots with any_of declarations
## Bug

\`linkml-lint\`'s \`no_undeclared_ranges\` rule reports \"range 'None' is not defined\" on slots whose range is expressed through \`any_of\` rather than via a top-level \`range:\` attribute. The \`any_of\` expression *is* a valid range declaration, but the lint rule doesn't recognize it as one.

## Minimal reproduction

\`\`\`yaml
id: http://example.org/test
name: test
prefixes:
imports:
  - linkml:types
classes:
  Foo:
    attributes:
      bar:
        description: A value that is either an integer or a string.
        any_of:
          - range: integer
          - range: string
\`\`\`

\`\`\`
$ linkml-lint test.yaml
  error    Class 'Foo' slot 'bar' range 'None' is not defined.  (no_undeclared_ranges)
\`\`\`

The slot's range is fully specified by \`any_of\`. \`linkml-validate\` correctly accepts both integer and string values for \`bar\` against this schema.

## Why this matters in practice

The natural way to express union types in LinkML is via slot-level \`any_of\`. Combining it with a top-level \`range:\` triggers a separate generator bug (#3111 — JSON Schema emits both \`type:\` and \`anyOf\` and they conflict). So the \`any_of\`-only formulation is currently the *only* working way to express a union range. That makes the lint false-positive especially painful — it's reported on every working union-typed slot.

## Reproduces on

linkml/linkml main (commit 373cba1) and linkml/linkml-runtime main (1.10.0).

## Suggested fix direction

The \`no_undeclared_ranges\` rule should treat a slot as having a declared range if any of \`any_of\`, \`all_of\`, \`exactly_one_of\`, or \`none_of\` is non-empty (each entry providing its own \`range:\`).

## Real-world consumer

Hit while authoring linkml/schema-automator#201 (data dictionary format spec).

gaoflow:
## Problem

`linkml-lint`'s `no_undeclared_ranges` rule reports `range 'None' is not defined` on a slot whose range is expressed through `any_of` (or `exactly_one_of`) instead of a top-level `range:`, even though the boolean expression fully specifies the range.

```yaml
classes:
  Foo:
    attributes:
      bar:
        any_of:
          - range: integer
          - range: string
```

```console
$ linkml-lint test.yaml
  error    Class 'Foo' slot 'bar' range 'None' is not defined.  (no_undeclared_ranges)
```

## Cause

`SchemaView.slot_range_as_union(slot)` returns the slot's own (unset) top-level range as `None` **alongside** the concrete ranges gathered from `any_of`/`exactly_one_of` — e.g. `[None, 'integer', 'string']`. The rule then checks `None` against the set of defined elements and flags it.

## Fix

Drop the `None` entry from the range set only when the union also contains concrete ranges (i.e. when a boolean expression supplied them). This:

- removes the false positive for `any_of`/`exactly_one_of` slots,
- keeps flagging a genuinely undefined range named inside the expression (e.g. `spreadsheet`), now by its real name rather than as `None`,
- preserves the existing behaviour that a slot with **no** range at all and no `default_range` is still reported (covered by `test_missing_default_range_undeclared_ranges`).

## Tests

Added parametrised (`any_of`/`exactly_one_of`) tests to `test_rule_no_undeclared_ranges.py`: one asserting no false positive when all expression ranges are defined, one asserting an undefined expression range is still flagged by name. Red before / green after; full linter suite passes (161).

Closes #3477.


amc-corey-cox (approved):
This looks good to me!

PR #3622, opened 2026-06-10
Implement linkml arrays for JSON Schema generator

### LINKED ISSUE #1889: Add support for linkml:array in jsonschemagen
See #1888

sneakers-the-rat:
## Summary

Fix: #1889 

What up its yer blob, me, back again with another "implementing linkml arrays"

- move some of the pydantic-specific array generation stuff to a common module
- move the array testcase fixtures to the conftest file
- copy paste baby
- now it works

## How was this tested?

exactly how pydanticgen arrays are tested!

## Areas of uncertainty

there are probably some weird cases about how the inner dtype is handled - what i did is just like "assumed that the current method for getting a slot's range as a schema did that" and then "put that inside of arrays." idk just @ me about stuff idk this generator at all.

## Checklist

## AI Assistance

i have to eat bowls full of dogshit all day long already why would i bring a bowl full of dogshit to my zen garden project

### DIFF
--- a/packages/linkml/src/linkml/generators/jsonschemagen.py
+++ b/packages/linkml/src/linkml/generators/jsonschemagen.py
@@ -961,7 +961,7 @@ class JsonSchemaGenerator(Generator, LifecycleMixin):
         ).schema_
         return self.top_level_schema
 
-    def serialize(self, as_json: bool = True, **kwargs) -> str | JsonSchema:
+    def serialize(self, **kwargs) -> str:
         if self.materialize_patterns:
             logger.info("Materializing patterns in the schema before serialization")
             self.schemaview.materialize_patterns()
--- a/packages/linkml/src/linkml/generators/jsonschemagen.py
+++ b/packages/linkml/src/linkml/generators/jsonschemagen.py
@@ -8,9 +8,12 @@ from typing import Any
 
 import click
 from jsonasobj2 import as_dict
+from pydantic import Field
 
 from linkml._version import __version__
 from linkml.generators.common import build
+from linkml.generators.common.array import ArrayRangeGenerator, ArrayRepresentation
+from linkml.generators.common.build import RangeResult
 from linkml.generators.common.lifecycle import LifecycleMixin
 from linkml.generators.common.subproperty import get_subproperty_values
 from linkml.generators.common.type_designators import get_type_designator_value
@@ -19,8 +22,10 @@ from linkml.utils.helpers import get_range_associated_slots
 from linkml_runtime.linkml_model.meta import (
     AnonymousClassExpression,
     AnonymousSlotExpression,
+    ArrayExpression,
     ClassDefinition,
     ClassDefinitionName,
+    DimensionExpression,
     EnumDefinition,
     Example,
     PermissibleValue,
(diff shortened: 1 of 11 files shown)


packages/linkml/src/linkml/generators/jsonschemagen.py
```diff
@@ -673,7 +678,8 @@ def get_type_info_for_slot_subschema(
         slot_is_inlined = self.schemaview.is_inlined(slot)
         if slot.range in self.schemaview.all_types().keys():
             schema_type = self.schemaview.induced_type(slot.range)
-            (typ, fmt) = json_schema_types.get(schema_type.base.lower(), ("string", None))
+            if schema_type.base:
+                (typ, fmt) = json_schema_types.get(schema_type.base.lower(), ("string", None))
```
sneakers-the-rat, line 740: not all types have bases, e.g. when there are custom types

packages/linkml/src/linkml/generators/jsonschemagen.py
```diff
@@ -739,26 +745,17 @@ def get_value_constraints_for_slot(self, slot: SlotDefinition | AnonymousSlotExp
         return constraints
 
     def get_subschema_for_slot(
-        self, slot: SlotDefinition | AnonymousSlotExpression, omit_type: bool = False, include_null: bool = True
+        self,
+        slot: SlotDefinition | AnonymousSlotExpression,
+        omit_type: bool = False,
+        include_null: bool = True,
+        cls: ClassDefinition | None = None,
```
sneakers-the-rat, line 810: this might be less than optimal, but i needed to be able to access the outer class's name so that i could deduplicate top-level defs for the recursive infinite dimension array definition for slots with the same name but in different classes

tests/linkml/test_generators/input/arrays/any_shape.yaml
```diff
@@ -5,10 +5,10 @@ imports:
   - linkml:types
 
 classes:
-  AnyType:
+  AnyType_:
```
sneakers-the-rat, line 8: i am not really sure what's going on with how to specify "Any" as a type, but this made the json schema generator choke for some reason, and then it turns out that AnyType was a self reference to the class not "any" so i added a special case in pydanticgen to treat a range as "Any" when it doesn't otherwise match any of the prior conditions and is "Any"

tests/linkml/test_generators/test_jsonschemagen.py
sneakers-the-rat, line 1: i could probably collapse these all down into one gigantic parameterize statement, i typically find breaking them up by name is a bit easier to make sense of. would need to also change how the fixtures that load the schemas work but that's also no biggie. lmk if that's preferable.

sneakers-the-rat (commented):
mostly p self explanatory because it's exactly the same thing as the list of lists generator for pydantic, except json schema flavored.

packages/linkml/src/linkml/generators/jsonschemagen.py
```diff
@@ -888,6 +885,11 @@ def get_subschema_for_slot(
                 ),
             )
 
+        if isinstance(slot, SlotDefinition) and slot.array:
+            # TODO: this is currently too lax, in that it will validate ANY array.
+            # see https://github.com/linkml/linkml/issues/2188
```
sneakers-the-rat: whoops this should get removed, will do that when i'm back at keyboard

kevinschaper:
Should this one drop to draft state?

sneakers-the-rat:
Not that I'm aware of? Looking now I forgot to remove that comment, but that's it. Will do that and rebase now

kevinschaper (approved):

Issue #3560, opened 2026-05-23
in linkml-convert, relative imports are not resolved relative to the importing schema's directory


paulmillar:
**Describe the bug**

The `imports` slot is a list of further LinkML schemata to import.  Relative paths should be resolved relative to the importing schema.  In newer versions of LinkML, the `linkml-convert` command resolves imports relative to the current directory.

```console
(linkml) paul@monkeywrench:~/linkml-issue$ ls -R
.:
data  schema

./data:
test-data.json

./schema:
my-schema.yaml  utils.yaml
(linkml) paul@monkeywrench:~/linkml-issue$ 
```

The `my-schema.yaml` file imports the `utils.yaml` file:

```console
(linkml) paul@monkeywrench:~/linkml-issue$ grep -3 utils schema/my-schema.yaml 

imports:
  - linkml:types
  - utils

default_range: string

(linkml) paul@monkeywrench:~/linkml-issue$ 
```

The `linkml-lint` and `linkml-validate` work as expected:

```console
(linkml) paul@monkeywrench:~/linkml-issue$ linkml-lint schema/my-schema.yaml 
✓ No problems found
(linkml) paul@monkeywrench:~/linkml-issue$ linkml-validate -s schema/my-schema.yaml  data/test-data.json 
No issues found
(linkml) paul@monkeywrench:~/linkml-issue$ 
```

However, the `linkml-convert` command tries to resolve the import relative to the current directory:

```console
(linkml) paul@monkeywrench:~/linkml-issue$ linkml-convert -t yaml -s schema/my-schema.yaml data/test-data.json 
Traceback (most recent call last):
[...]
FileNotFoundError: [Errno 2] No such file or directory: '/home/paul/linkml-issue/utils.yaml'
(linkml) paul@monkeywrench:~/linkml-issue$ 
```

Creating a symbolic link in the current directory that points to the `utils.yaml` file provides a work-around:

```console
(linkml) paul@monkeywrench:~/linkml-issue$ ln -s schema/utils.yaml 
(linkml) paul@monkeywrench:~/linkml-issue$ linkml-convert -t yaml -s schema/my-schema.yaml data/test-data.json 
name: The canonical example
items:
- name: Item 1
  actions:
  - 'action #1'
  - 'action #2'
  - 'action #3'
- name: Item 2
  actions:
  - 'action #3'
  - 'action #4'

(linkml) paul@monkeywrench:~/linkml-issue$ 
```

[...]

PR #3628, opened 2026-06-10
fix(linkml-validate): resolve relative imports

### LINKED ISSUE #3499: SchemaView.imports_closure() resolves relative imports against root schema directory
## Bug Description

In `linkml_runtime/utils/schemaview.py`, the `imports_closure` method (around line ~468) calls `self.load_import(sn)` without passing `from_schema`. Inside `load_import`, when `from_schema` is `None`, it defaults to `self.schema` (the root schema). This means relative imports like `./types` declared in sub-schemas are resolved relative to the root schema's source directory instead of the sub-schema's directory.

## Affected Code Path

- `imports_closure` line ~467-468:
  ```python
  if sn not in self.schema_map:
      self.schema_map[sn] = self.load_import(sn)
  ```
  — no `from_schema` passed

- `load_import` line ~369:
  ```python
  if from_schema.source_file and not is_absolute_path(sname):
  ```
  — when `from_schema` is `None`, defaults to root schema

## Impact

Custom importmap overrides that load sub-schemas with relative internal imports will fail with `FileNotFoundError`.

## Minimal Reproduction

1. Create a schema that imports `./sub/child` where `./sub/child.yaml` imports `./types`
2. Use an importmap to redirect the parent schema to a custom directory
3. `imports_closure()` will fail to resolve `./types` because it looks relative to the root schema's directory

## Suggested Fix

Pass `from_schema=self.schema_map.get(sn)` to `load_import` at line 468, or track the source schema for each entry in `schema_map`.

### LINKED ISSUE #3627: linkml.validator.validate_file - does not resolve relative imports
**Describe the bug**

In modular schema design, `validate_file` raises `FileNotFoundError: '<CWD>/other.yaml'`. 

The relative import is resolved against the **process working directory**, not against the directory of file passed in.

1.10.1

**Reproduction**

```bash
mkdir -p /tmp/relimports/schemas && cd /tmp/relimports

cat > schemas/base.yaml << 'EOF'
id: https://example.org/base
name: base
prefixes:
imports:
  - linkml:types
default_range: string
classes:
  Thing:
    attributes:
      id:
        identifier: true
EOF

cat > schemas/main.yaml << 'EOF'
id: https://example.org/main
name: main
prefixes:
imports:
  - linkml:types
  - ./base
EOF

cat > data.yaml << 'EOF'
id: x
EOF

# Run from /tmp/relimports, NOT from /tmp/relimports/schemas:
uv run python - << 'PY'
from linkml.validator import validate_file
validate_file('data.yaml', 'schemas/main.yaml', target_class='Thing')
PY
# -> FileNotFoundError: [Errno 2] No such file or directory:
#    '/tmp/relimports/base.yaml'
```

noelmcloughlin:
## Summary

Fixes #3627 and #3560

- Ensure relative imports work in linkml validate and across loaders (#3627)
- Fixes linkml-convert relative imports (#3560)

Does not fix #3499 - follow-up PR needed

## How was this tested?

Full suite local test
added tests
raised two related issues

## Checklist

## AI Assistance
LM assisted


packages/linkml/src/linkml/validator/__init__.py
```diff
@@ -25,7 +25,10 @@ def _get_default_validator(
         if isinstance(schema, dict):
             schema = SchemaDefinition(**schema)
         elif isinstance(schema, str):
+            schema_source = schema
             schema = yaml_loader.load(schema, target_class=SchemaDefinition)
+            if "\n" not in schema_source:
+                schema.source_file = schema_source
```
sneakers-the-rat: this seems like a bit of a hack, like this seems like it should be done in the loader itself so that everything that gets loaded from a path has its `source_file` set correctly. 

it also seems like we're doubling down on prior hacks and throwing away information: e.g. if something is a path and it gets passed in as a `Path`, we shouldn't need to first transform that to a string in the first place, but then by this point we extra shouldn't need to look for the presence of a newline to tell a path apart from a schema as a string. This appears to be inherited from our dear old friend `hbread` which has not historically been a huge friend to code clarity.

noelmcloughlin: Hi @sneakers-the-rat   thanks for review - yeah I think this was a quick fix approach but I'll think more on this. Expect delayed response.

matentzn:
Converted back to draft as per https://github.com/linkml/linkml/pull/3628#discussion_r3421746822

packages/linkml/src/linkml/validator/__init__.py
noelmcloughlin: Ok. The PR feedback was good. I had to lookup hbreader in pypi.org but no documentation there - I guess hbreader should be replaced.

Anway, updated Changeset as follows:
- Yamlloader.load_any() now emits source_file to the constructor (from metadata.source_file)
- Loader._read_source() casts os.Pathlike to str due to hbreader bug - so loader now sets source_file to absolute resovled path not raw string.
- Validator.__init__() newline hack is gone
- Schemaview.load_schema_wrap() removed the prior hack mentioned
- raw_loader:  handle Metadata=false fall-through (only used by two tests)

sneakers-the-rat: yeah it's [here](https://github.com/linkml/hbreader/). i've only ever seen it used in linkml, i think it's in-house, but i generally agree, i don't think it's a good idea to have an opaque "put anything in and then get something out" function at the bottom of i/o, since different kinds of i/o should be handled... differently. this is sort of a very simple example of why. especially given that all it is is pretty much a wrapper around `open().read()` in most cases.

packages/linkml_runtime/src/linkml_runtime/loaders/yaml_loader.py
```diff
@@ -45,8 +45,19 @@ def load_any(
...
     ) -> YAMLRoot | list[YAMLRoot]:
+        if metadata is None:
+            metadata = FileInfo()
         data_as_dict = self.load_as_dict(source, base_dir=base_dir, metadata=metadata)
-        return self._construct_target_class(data_as_dict, target_class)
+        result = self._construct_target_class(data_as_dict, target_class)
+        # When the source was a file (or URL), ``hbread`` resolves and records it on the metadata.
+        # Propagate that to ``source_file`` (e.g. on a SchemaDefinition) so relative imports can be
+        # resolved against it. Inline string sources leave ``source_file`` unset, so this naturally
+        # distinguishes a path from schema text without inspecting the source itself.
```
sneakers-the-rat: i'm not going to belabor this PR, and I don't think it's fair to ask you to rewrite the whole loader here, but i still think it's fragile/unnecesarily implicit that we start out having a path, and we should still have one here (the `source` type annotation is incorrect and inconsistent with the parent method currently, but this escapes type checkers because `root.load()` is untyped), but rather than directly using the information we already have, we rely on hbreader to fill it in and retrieve it again from an implicit mutation of a passed object. if we go this route, it would be great to get an explicit test for "when a schema is loaded with the yaml loader, its `source_file` attribute is correctly assigned"

packages/linkml/src/linkml/utils/rawloader.py
```diff
@@ -108,6 +108,11 @@ def _name_from_url(url) -> str:
             schema.source_file_date = src_date
         schema.source_file_size = schema_metadata.source_file_size
         schema.generation_date = datetime.now().strftime(DATETIME_FORMAT)
+    else:
+        # ``metadata=False`` means no source-file metadata should be emitted. The loader now records
+        # ``source_file`` (the resolved path) whenever the source is a file, so clear it here to
+        # honor the flag and keep output free of machine-specific paths.
+        schema.source_file = None
```
sneakers-the-rat: i get how this is just fitting in with the existing expectations, but this strikes me as a serialization-time concern, not a loading-time concern. ideally we would *always* have the metadata present on the in-memory object, and then this flag would affect how the thing was serialized - which is *i believe* the intent. not a request for change on this PR, just noting more broadly that this is another reason i think we should not have so many loading classes for schema (rawloader, yaml loader, schemaloader, schemaview, are there more?) so that we can make this behavior more consistent.

packages/linkml_runtime/src/linkml_runtime/loaders/loader_root.py
```diff
@@ -175,6 +176,12 @@ def _read_source(
         if isinstance(source, dict):
             return source
 
+        # A ``Path`` (or other ``os.PathLike``) is a file location, but ``hbread`` would mistake it
+        # for stringable data and read its text as the source. Coerce to ``str`` so it is detected
+        # and resolved as a filename.
```
sneakers-the-rat: wow that's... terrible... the hbreader thing, not your impl.

packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
```diff
@@ -209,14 +209,10 @@ def load_schema_wrap(path: str, **kwargs: dict[str, Any]) -> SchemaDefinition:
     from linkml_runtime.loaders.yaml_loader import YAMLLoader
 
     yaml_loader = YAMLLoader()
+    # When ``path`` is a file or URL the loader records it on ``schema.source_file``, which is
+    # necessary for resolving relative imports. A yaml string leaves ``source_file`` unset, so there
+    # should be no expectation of relative imports working for inline schema text.
```
sneakers-the-rat:
```suggestion
```

mild recommendation, take or leave: i think this would probably be confusing as a comment in all contexts except this PR - this would be more a comment on the docstring of the method, if needed at all. ideally, to the user, resolving paths "just works," and `SchemaDefinition` already has `source_file` which should have its own documentation. again not a strong opinion, feel free to ignore if you like the comment.

sneakers-the-rat (commented):
thanks for hanging in there, i appreciate working through the historic codebase, keeping changes minimal while still trying to find a path that's not compounding hacks. tricky!

two things i think would be good to have for this PR, both of them are tests:
- an explicit test for the yaml loader, "when we load a schema file, the `source_file` attribute is correctly populated for absolute and relative paths"
- as with schemaview, we should have tests for nested series of relative paths: if it works for schema all in the same directory, great, but it should also work with relative paths in different directories (at least subdirectories, we don't have any kind of directory traversal protection anywhere and i'm not sure it's necessary for a tool like this, in any case that's a separate concern). there's an example test [here](https://github.com/linkml/linkml/blob/b8fe8b7da89cf1332aece90bdccaa0a4243932dc/tests/linkml_runtime/test_utils/test_schemaview.py#L496), and there's already a set of testing schema with relative imports [here](https://github.com/linkml/linkml/tree/main/tests/linkml_runtime/test_utils/input/imports_relative) so you can just reuse that without needing to write a new one.

rest of comments are just comments/suggestions/musings, take or leave

packages/linkml/src/linkml/utils/rawloader.py
noelmcloughlin: Thanks - noted - https://github.com/linkml/linkml/issues/3699

packages/linkml_runtime/src/linkml_runtime/loaders/yaml_loader.py
noelmcloughlin: Thanks again - I raised https://github.com/linkml/linkml/issues/3700 to capture general point that known path is disappearing into magician's hat to reappear via passed/mutatted object.

I will fix the type mismatch (`Path`)!

packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
noelmcloughlin: done

packages/linkml_runtime/src/linkml_runtime/loaders/yaml_loader.py
```diff
@@ -38,15 +40,26 @@ def load_as_dict(
...
+        if metadata.source_file:
+            for target in result if isinstance(result, list) else [result]:
+                if hasattr(target, "source_file") and not target.source_file:
```
amc-corey-cox: This fires on every YAML load, not just schemas — any target class with a `source_file` attribute gets the loader's path injected. I checked with a plain pydantic model and it comes back holding an absolute path, which `linkml-convert` would then write out.

`source_file` only exists on `SchemaDefinition` in the metamodel, so gating on `isinstance(target, SchemaDefinition)` (lazy import to dodge the circular) costs nothing.

packages/linkml_runtime/src/linkml_runtime/loaders/yaml_loader.py
```diff
@@ -38,15 +40,26 @@ def load_as_dict(
...
+                    target.source_file = str(metadata.source_file)
```
amc-corey-cox: `hbread` absolutizes, so `SchemaView('x/main.yaml').schema.source_file` goes from `x/main.yaml` to `/home/you/.../x/main.yaml`.

`schemaloader.py:703` deliberately basenames this — "Make the source file relative if it is locally generated" — so machine paths here look like something the project already rules out rather than a free choice. Could we record the path the loader was handed, and only absolutize when `base_dir` is in play?

tests/linkml_runtime/test_loaders_dumpers/test_loaders.py
```diff
@@ -39,6 +40,49 @@ def test_yaml_loader():
...
+    assert schema_from_path.source_file == str(schema_path)
+
+
+def test_yaml_loader_leaves_source_file_unset_for_string():
+    """Loading inline schema text leaves ``source_file`` unset (no path to record)."""
+    schema = yaml_loader.load(_SOURCE_FILE_SCHEMA, target_class=SchemaDefinition)
+    assert schema.source_file is None
+
+
+def test_yaml_loader_does_not_overwrite_existing_source_file(tmp_path):
```
amc-corey-cox: This locks in a precedence change I think is a regression — the guard is `not target.source_file` in `yaml_loader.py:60`.

`load_schema_wrap` used to overwrite `source_file`; now a value declared inside the schema file wins, and linkml itself emits one under `--metadata`. Relative imports then resolve against a stale directory — I get a `FileNotFoundError` here where main loads fine. Worse, `linter.py:197` dumps the schema back to `source_file` on `--fix`, so `linkml-lint --fix` can write to the wrong file.

The metamodel calls the slot `readonly: supplied by the schema loader`, so I'd let the resolved path win and invert this test.

tests/linkml_runtime/test_loaders_dumpers/test_loaders.py
```diff
@@ -39,6 +40,49 @@ def test_yaml_loader():
...
+    schema_path = tmp_path / "schema.yaml"
+    schema_path.write_text(_SOURCE_FILE_SCHEMA)
+
+    # str path
+    schema = yaml_loader.load(str(schema_path), target_class=SchemaDefinition)
+    assert schema.source_file == str(schema_path)
+
+    # Path object (parent annotation allows Path; loader coerces os.PathLike)
+    schema_from_path = yaml_loader.load(schema_path, target_class=SchemaDefinition)
+    assert schema_from_path.source_file == str(schema_path)
```
amc-corey-cox, line 68: Good to have these. `hbread` also sets `source_file` for URLs and open file handles — both newly affected, neither covered. The URL case is the one I'd want a test for, since `_get_imports_closure` runs `os.path.dirname()` on it.

packages/linkml/src/linkml/utils/rawloader.py
amc-corey-cox, line 134: Question, not a change request: does `metadata=False` mean "don't record" or "don't emit"? The flag's deprecation note frames it as a generator output header, which reads as a serialization concern — clearing at load time leaves the in-memory object lossy for anything downstream that wants the path.

Smaller: this also drops a `source_file` the schema file itself declared, which used to survive.

packages/linkml_runtime/src/linkml_runtime/loaders/loader_root.py
```diff
@@ -175,6 +176,12 @@ def _read_source(
...
+        if isinstance(source, os.PathLike):
```
amc-corey-cox, line 182: Nit: `_read_source` (line 163) and `load_source` still say `source: str | dict | TextIO`, though the body now handles `os.PathLike`.

amc-corey-cox (changes requested):
Moving source_file population into the loader is the right layer, and I'm glad to see the "\n" not in path heuristic go — the tests requested earlier all landed and hold up.

Two things I'd want resolved before merge, both silent-wrong-behavior rather than errors: source_file is now injected into any target class that has that attribute, not just schemas, and a source_file declared inside a schema file now outranks the path the loader actually resolved (which breaks relative imports and can send linkml-lint --fix at the wrong file).

packages/linkml/src/linkml/utils/rawloader.py
noelmcloughlin: This PR is more tricky than I expected.   Here is this loader, `metadata=False` means "don't record" but the flag has inconsistent meaning across loaders (see https://github.com/linkml/linkml/issues/3699#issuecomment-5198794319).

Thanks for spotting the introduced bug ("_drops a source_file the schema file itself declared, which used to survive_") - will add commit to fix and test.

I found a new bug during testing (not fixed here): https://github.com/linkml/linkml/issues/3852

noelmcloughlin:
Hi @amc-corey-cox @sneakers-the-rat 

I think I have fixed all PR comments (let me know if you want commits squashed).

Two new issues raised:

- https://github.com/linkml/linkml/issues/3699
- https://github.com/linkml/linkml/issues/3852

Related issues by other people:

- https://github.com/linkml/linkml/issues/3560   (**fixed**)
- https://github.com/linkml/linkml/issues/3499    (not fixed - its not trivial / not one-liner fix)

amc-corey-cox (approved):
Looks good to me.

Issue #1341, opened 2023-03-16
gen-project creates an open jsonschema by default


turbomam:
one has to add

```yaml
--generator-arguments 'jsonschema: {not_closed: false}'
```

to get a closed jsonschema

cmungall:
Should we switch behavior for 1.5?

- yes 👍 
- no 👎 
- other - add comments

pkalita-lbl:
I'll register a tepid "no" vote here. By default JSON Schema allows for additional properties -- that is, its default is open. For example, this data:

```json
{
  "a": 1,
  "b": 2
}
```

is perfectly valid under this schema:

```json
{
  "type": "object",
  "properties": {
    "a": { "type": "number" }
  },
  "required": ["a"]
}
```

It's not until you opt-in to closed-ness by way of the `additionalProperties` keyword that the same data becomes _invalid_:

```json
{
  "type": "object",
  "properties": {
    "a": { "type": "number" }
  },
  "required": ["a"],
  "additionalProperties": false
}
```

So in that sense `JsonSchemaGenerator` is inline with JSON Schema's own convention (open by default, closed by opt-in). Personally I think that's good, which is where my "no" vote comes from. 

But I also think it's a fair question to ask: is that too much of a bad surprise for LinkML users who maybe aren't as familiar with the ins-and-outs of JSON Schema? If so, I could be persuaded to say "yes" instead.

sierra-moxon:
I vote no, but maybe we echo jsonschema `additionalProperties` keyword in our cli to make it more obvious how to close a linkml jsonschema.

PR #3630, opened 2026-06-10
Align extra_slots support across pythongen/pydanticgen, per-class enabled

### LINKED ISSUE #3629: Inconsistent extra_slots support - Pythongen and Pydanticgen
**Describe the bug**

LinkML supports inconsistent open/closed semantics across three formats:  JSON, Python, Pdyantic.

[extra_slots](https://linkml.io/linkml-model/latest/docs/extra_slots/) per class, allowi additional properties: 

```yaml
classes:
  FlexiblePerson:
    extra_slots:
      allowed: true          #  not default
    attributes:
        name:
```
The JSON schema generator already honors this (see #1341) but the Python generators ignore entirely:
- gen-python - always strict, passing undeclared key raised 'TypeError' env when class is open.
- gen-pydantic - as single global `extra_fields` setting (forbid/allow/ignore) to evyer class - not per-class.

1.11.1

noelmcloughlin:
## Summary

Fixes #3629 

- Both generators read `cls.extra_slots.allowed` and emit per-class behavior that matches schema.

## How was this tested?

Added unit tests

## Areas of uncertainty

- Needs soaking by "open worlder's" as we enable feature that is not default.
- Expect future bugs on edge cases.
- Maybe #1341 should be closed also - tangential stale issue.

## Checklist

- [x]I have added tests that prove my fix/feature works

## AI Assistance
LLM assisted

### DIFF
--- a/packages/linkml/src/linkml/generators/pydanticgen/pydanticgen.py
+++ b/packages/linkml/src/linkml/generators/pydanticgen/pydanticgen.py
@@ -36,6 +36,7 @@ from linkml.generators.pydanticgen.template import (
 from linkml.utils.deprecation import deprecation_warning
 from linkml.utils.generator import shared_arguments
 from linkml_runtime.linkml_model.meta import (
+    AnonymousSlotExpression,
     ClassDefinition,
     ElementName,
     SchemaDefinition,
@@ -77,6 +78,7 @@ DEFAULT_IMPORTS = (
     + Import(
         module="typing",
         objects=[
+            ObjectImport(name="Annotated"),
             ObjectImport(name="Any"),
             ObjectImport(name="ClassVar"),
             ObjectImport(name="Literal"),
@@ -485,7 +487,9 @@ class PydanticGenerator(OOCodeGenerator, LifecycleMixin):
                 raise ValueError(f"could not find suitable element in {clist} that does not ref {slist}")
         return slist
 
-    def _get_extra_for_class(self, cls: ClassDefinition) -> Literal["allow", "forbid", "ignore"] | None:
+    def _get_extra_for_class(
+        self, cls: ClassDefinition
+    ) -> Literal["allow", "forbid", "ignore"] | PydanticAttribute | None:
         """
         Determine the per-class ``extra`` value for the Pydantic ``ConfigDict``
         based on the class's ``extra_slots`` metadata.
@@ -493,9 +497,13 @@ class PydanticGenerator(OOCodeGenerator, LifecycleMixin):
         Returns ``None`` if the class does not override the generator-wide default,
         so the inherited base-model configuration is used.
         """
-        if not cls.extra_slots or cls.extra_slots.allowed is None:
+        if not cls.extra_slots:
             return None
-        return "allow" if cls.extra_slots.allowed else "forbid"
+        elif cls.extra_slots.range_expression:
+            result = self.generate_slot(cls.extra_slots.range_expression, cls)
+            return result.attribute
+        else:
+            return "allow" if cls.extra_slots.allowed else "forbid"
 
     def generate_class(self, cls: ClassDefinition) -> ClassResult:
         # Handle union_of classes by creating a type alias instead of a class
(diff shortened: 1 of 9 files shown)


sneakers-the-rat:
On the pydanticgen part, there's a parallel conversation happening in a other PR re: json schema gen, and I think that the option for the generator to override the schema should be removed (previously it was not overriding the schema because there was no way to control extra slots in the schema). 

Implementing extra slots also requires handling the expression form where a type is set for `__pydantic_extra__`. Should be simple to do with the existing methods and template objects, since the slot expression is the same type as all the others.

https://pydantic.dev/docs/validation/latest/concepts/models/#extra-data

sneakers-the-rat:
I think the feature might be contained enough to do python and pydanticgen in one PR, but with the addition of handling the expression form, it might balloon a bit. If it becomes unwieldy, In general I would favor splitting it by generator rather than doing boolean form first and then expression form for both, just because I think we should ship the complete feature at a time. Might be worth doing that now and splitting PR  before further work done, but up to you.

Re: the requested deprecation/pydanticgen part, I don't want to take work off your plate that youve already started, but I was planning on doing it this week anyway, and would be happy to help with that if its proving tricky. If you do the implementation, i can handle doing the deprecation notice for the generator param (unless you are interested in checking out how that works, at which point ofc explore!)

Just offering help if its wanted and a bit of guidance so it doesn't feel like I'm asking for huge changes without help, but again not trying to be overbearing or discouraging by taking it from you.

tests/linkml/test_generators/test_pydanticgen.py
sneakers-the-rat, line 1: added test cases for the target behavior in pydanticgen

tests/linkml/test_generators/test_extra_slots_per_class.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def _extract_class_block(code: str, class_name: str) -> str:
+    """Return the source block for ``class_name`` up to the next top-level class."""
+    marker = f"class {class_name}("
+    start = code.index(marker)
+    rest = code[start + len(marker) :]
+    next_class = rest.find("\nclass ")
+    end = start + len(marker) + (next_class if next_class != -1 else len(rest))
+    return code[start:end]
```
sneakers-the-rat, line 145: cheap enough and definitely more functionally airtight to just validate against real data and real classes, see added test cases

tests/linkml/test_generators/test_extra_slots_per_class.py
sneakers-the-rat, line 1: split into pythongen and pydanticgen tests, put in `test_pythongen` and `test_pydanticgen`, or else put in the `python` subfolder - want to avoid one-off test modules in `test_generators`

sneakers-the-rat:
here I did the `range_expression` part of pydanticgen. so for the test schema:

```yaml
classes:
  Base:
    attributes:
      x:
        range: integer

  Other:
    attributes:
      y:
        range: string

  ExtraAllowed:
    is_a: Base
    extra_slots:
      allowed: true

  ExtraNotAllowed:
    is_a: Base
    extra_slots:
      allowed: false

  ExtraString:
    is_a: Base
    description: "Only extra fields that are strings are valid"
    extra_slots:
      range_expression:
        range: string

  ExtraClass:
    is_a: Base
    description: "Only extra fields that match a class are valid"
    extra_slots:
      range_expression:
        range: Other

  ExtraAnyOf:
    is_a: Base
    description: "Union of allowed types"
    extra_slots:
      range_expression:
        any_of:
          - range: string
          - range: integer

  ExtraCardinality:
    is_a: Base
    description: "Limit cardinality of multivalued extra slots"
    extra_slots:
      range_expression:
        range: integer
        multivalued: true
        maximum_cardinality: 5

  AnyClass:
    class_uri: linkml:Any
    description: "The Any type implicitly allows any extra slots"

```

 the generated pydantic looks like this (removed the meta declarations for simplicity's sake)

```python

class Base(ConfiguredBaseModel):
    x: Optional[int] = Field(default=None)


class Other(ConfiguredBaseModel):
    y: Optional[str] = Field(default=None)


class ExtraAllowed(Base):
    model_config = ConfigDict(extra = "allow")

    x: Optional[int] = Field(default=None)


class ExtraNotAllowed(Base):
    model_config = ConfigDict(extra = "forbid")

    x: Optional[int] = Field(default=None)


class ExtraString(Base):
    """
    Only extra fields that are strings are valid
    """
    __pydantic_extra__: dict[str, Annotated[str, Field(...)]]
    model_config = ConfigDict(extra = "allow")

    x: Optional[int] = Field(default=None)


class ExtraClass(Base):
    """
    Only extra fields that match a class are valid
    """
    __pydantic_extra__: dict[str, Annotated[Other, Field(...)]]
    model_config = ConfigDict(extra = "allow")

    x: Optional[int] = Field(default=None)


class ExtraAnyOf(Base):
    """
    Union of allowed types
    """
    __pydantic_extra__: dict[str, Annotated[Union[int, str], Field(...)]]
    model_config = ConfigDict(extra = "allow")

    x: Optional[int] = Field(default=None)


class ExtraCardinality(Base):
    """
    Limit cardinality of multivalued extra slots
    """
    __pydantic_extra__: dict[str, Annotated[list[int], Field(..., max_length=5)]]
    model_config = ConfigDict(extra = "allow")

    x: Optional[int] = Field(default=None)
```

sneakers-the-rat (changes requested):
just reviewing the pydanticgen part, did not review pythongen part.

all of that looks good to me, my only requested changes are for the tests. For pydanticgen, i'm pretty sure that the test cases ported over from jsonschemagen in `test_pydanticgen` cover all the cases in `test_extra_slots_per_class`, so we might be able to just remove those, but someone else double check that.

- split test cases for python and pydanticgen and put them in their respective test modules, or put the new test module in `tests/test_generators/python` to avoid adding new single-purpose test modules in `test_generators` which is already large
- want to avoid string comparison tests whenever possible when functional tests that use live objects are cheap enough to do (they are here, especially when the generation and module compilation is done in a module-scoped fixture and reused, see tests in `test_pydanticgen`)

packages/linkml/src/linkml/generators/pydanticgen/pydanticgen.py
```diff
@@ -650,8 +693,13 @@ def generate_slot(self, slot: SlotDefinition, cls: ClassDefinition) -> SlotResul
                     result.attribute.range = f"dict[str, {simple_dict_value}]"
                 else:
                     result.attribute.range = f"dict[{collection_key}, {result.attribute.range}]"
-        if not (slot.required or slot.identifier or slot.key) and not slot.designates_type:
+
+        if isinstance(slot, AnonymousSlotExpression):
+            if not slot.required:
+                result.attribute.range = f"Optional[{result.attribute.range}]"
+        elif not (slot.required or slot.identifier or slot.key) and not slot.designates_type:
```
sneakers-the-rat, line 700: (self-commenting to explain, this is my code and not @noelmcloughlin 's so blame is with me):
somewhat awkward, the first leg is not `and`'ed because that should be "if we are an anonymous slot expression, make us optional if we are not required. otherwise if we are not an anonymous slot expression (and these other conditions) ... "

this method is way overloaded and should be split up, but not attempting here. easiest path to do the `range_expression` implementation by reusing most of the template and most of the logic in this method.

noelmcloughlin:
Thanks @sneakers-the-rat for the valuable review and additional commits.  Let me know if this is urgent for you - will make time - otherwise I may be delayed returning to this PR.

PR #3641, opened 2026-06-10
fix(rdf parse): drop unresolvable per-module @context refs before

### LINKED ISSUE #3635: gen-rdf raises HTTP 404s, per-module JSON-LD @context refs
**Describe the bug**

gen-jsonld` emits relative refs (`./other.context.jsonld`, etc.) in the per-module `@context` array. 

`gen-rdf` then asks rdflib to dereference each one, rdflib resolves it against the schema `@base` URI.

The resulting HTTPS GETs return 404 because those context files are not published. The merged `@context` already contains all prefix bindings, so the relative module refs are redundant and can be safely dropped.

1.11.1

**Reproducer**

```bash
mkdir -p /tmp/relimports/schemas && cd /tmp/relimports

cat > schemas/base.yaml << 'EOF'
id: https://example.org/base
name: base
prefixes:
imports:
  - linkml:types
default_range: string
classes:
  Thing:
    attributes:
      id:
        identifier: true
EOF

cat > schemas/main.yaml << 'EOF'
id: https://example.org/main
name: main
prefixes:
imports:
  - linkml:types
  - ./base
EOF

cat > data.yaml << 'EOF'
id: x
EOF

cd /tmp/relimports/schemas
uv run gen-rdf main.yaml > /tmp/out.ttl
# -> rdflib raises BadSyntax / HTTPError 404 for
#    https://example.org/base.context.jsonld
```

noelmcloughlin:
## Summary

Fixes #3635

- Add a helper method that drops string entries from the @context array whose URI scheme is not in  {http, https, file}.

## How was this tested?

added tests

## Checklist

## AI Assistance
LLM assisted


noelmcloughlin:
I found `rdf_canonicalize.py's rdflib fallback` bugs related to https://github.com/linkml/linkml/pull/3703 and https://github.com/linkml/linkml/issues/3760 but will address in different issue/fix - this PR is narrow scope.

kevinschaper (approved):
Thanks @noelmcloughlin!

amc-corey-cox:
I pushed a small commit (0eb6ae0) to fix the CI failure — hope that's okay. The offline network guard that recently landed on main (#3897) blocks unmarked tests from touching the network, and the new rdfgen test trips it: rdflib fetches the absolute `@context` URLs (like `types.context.jsonld`) that survive your relative-ref filter. Marking the test `@pytest.mark.network` makes those fetches get served from local vendored files instead. Same fix we just applied on #3696. Verified the rdfgen test file passes locally with the guard active.

PR #3649, opened 2026-06-16
docs: Clarify the documentation about unique keys & identifiers.

### LINKED ISSUE #3329: Unique keys & identifiers should be better specified documented and clarified
The current specification and documentation is particularly confusing and imprecise regarding the expectations that surrounds identifier slots, key slots, and unique keys.

## Review

The [documentation about slots](https://linkml.io/linkml/schemas/slots.html#identifiers) says that “if a slot is declared as an ‘identifier’” (with “identifier” pointing to [the spec about the `identifier` slot](https://linkml.io/linkml-model/latest/docs/identifier/)), “then it serves as a unique key for members of that class. It can also be used for inlining as a dict in JSON serialisations. […] A class must not have more than one identifier (asserted or derived). `identifier` marks the _primary identifier_.”

The [spec about the `identifier` slot](https://linkml.io/linkml-model/latest/docs/identifier/) says that setting this slot to `true` “means the key slot(s) uniquely identifies the elements.” And already here some seeds of confusion are planted: Why does the spec about `identifier` mention “key slots”? Slots marked with `key: true` are _not_ the same thing as slots marked with `identifier: true`. And why the plural? It is already established that there can be only one `identifier` slot within a class – a constraint further stated on that same page: “a given domain [where “domain” is to be interpreted as “class”] can have at most one identifier”.

The confusion goes on as another comment says “a key slot is automatically required. Identifiers cannot be optional”. Again, why talking a “key slot” here? Especially since, two lines below, the same page explicitly states that “identifiers and keys are mutually exclusive”, definitely confirming that, in LinkML context, “identifier” and “key” refer to two different things.

Then, the [documentation about inlining](https://linkml.io/linkml/schemas/inlining.html) says that “if the range class does not declare an `identifier`” (with “identifier” again point to [the spec about the `identifier` slot](https://linkml.io/linkml-model/latest/docs/identifier/)), “then `inlined` is always true”. In other words, _referencing_ objects by their names (the opposite of inlining) is only possible when those objects have an “identifier” – OK, that makes perfect sense.

But further, the same page says that “inlining as a dictionary is only possible if the range of the slot is a class that has an identifier slot”. This I believe is _not_ true if we take “identifier slot” to mean “a slot marked with `identifier: true`”. For dict inlining, either an “identifier slot” (again, taken to mean “slot marked with `identifier: true`”) _or_ a “key slot” (taken to mean “slot marked with `key: true`”) will do. I believe this is confirmed by the [`test_inlined_compliance.py` test](https://github.com/linkml/linkml/blob/main/tests/linkml/test_compliance/test_inlined_compliance.py), which among other things checks dict inlining of classes that have a “key slot” but no “identifier slot”.

The [page about constraints](https://linkml.io/linkml/schemas/constraints.html#unique-key) then mentions “unique keys”, to say that “a class can declare `unique_keys`, a set of slots that are unique for members of that class. Identifiers are special cases of unique keys.” Which “identifiers” are we talking about here? Both “identifier slots” and “key slots” are arguably special cases of unique keys.

It should be quite clear by this point that, in several places, the doc/the spec is using “key” and “identifier” _interchangeably_, which may lead people to believe that they are more or less the same thing, thereby creating even more confusion (“why do have both an `identifier` slot and a `key` slot if identifiers are the same thing?”). Neither the documentation nor the spec, in their current state, are helpful to clear that confusion.

In fact, the clearest statement that clearly disambiguates “keys” and “identifiers” is to be found in a comment in the [`test_uniqueness_compliance.py` file](https://github.com/linkml/linkml/blob/main/tests/linkml/test_compliance/test_uniqueness_compliance.py), which says that “identifiers should be globally unique, but keys are only required to be unique within the object that hold them”.

I would like to suggest that it shouldn’t take digging through the comments of the test suite to get the best explanation about what is arguably a key feature (pun intended) of LinkML¹.

## Suggestion

I think that either the doc or the spec should explain, once and for all _in a single place_, everything there is to know about “unique keys”, “identifier slots”, and “key slots”.

Per my understanding (which may or may not be correct – did I mention that the current doc is confusing?), this could be explained as follows:

### Unique keys
* A class can have “unique keys”, which uniquely identify instances of that class within a certain scope.
* Unless overridden below, the uniqueness scope is that of the container (the multi-valued slot) that contains instances of the identified class. That is, when a multi-valued slot has a range set to a class that has one or more “unique keys”, there cannot be two instances with the same values for the keys throughout the entire list.
* There are two types of “unique keys”: “compound keys” and “singular keys”.

#### Compound keys
* “Compound keys” are defined through the `unique_keys` slot.

#### Singular keys
* “Singular keys” have the additional constraints, compared to “compound keys”, that
    * there can be only such key in a given class;
    * they are automatically _required_, even if the slot holding them is not explicitly marked as such.
* “Singular keys” are required for dict inlining.
* There are two types of “singular keys”: “singular unique keys” and “identifiers”.

##### Singular unique keys
* A “singular unique key” is defined by marking a slot with `key: true`. The corresponding slot is known as the “key slot” of the class that holds it.

##### Identifiers
* An “identifier” is defined by marking a slot with `identifier: true`. The corresponding slot is known as the “identifier slot” of the class that holds it.
* Identifiers have the additional constraint, compared to “singular unique keys”, that their values must be globally unique (there cannot be two instances with the same identifier _everywhere_). In other words, their scope is global, instead of being limited to an enclosing container.

(Note that since, in the description above, both “keys” and “identifiers” are considered to be subtypes of “singular keys”, and we have already said that there can be only one “singular key” in any given class, it naturally follows that a class cannot have both a “key slot” and an “identifier slot”.)

And once the above has been explained once and for all somewhere, the entire doc/spec should be checked to ensure that the correct terms are used. For example, the page about inlining should say that only a “singular key” (which can be indifferently a “singular unique key” or an “identifier”) is needed for dict inlining – not specifically an “identifier”.

---
¹Key feature because it impacts several important aspects whose understanding is needed to be able to (de)serialise LinkML instance data correctly, such as referencing and inlining.

gouttegd:
## Summary

Partially fixes #3329 (fully fixing the issue will also require amending the inline documentation of the `key` and `identifier` slots; this will need to happen in `linkml-model`, not here).

This commit amends the `constraints.md` documentation page to (hopefully) clearly explain the distinction between:

- a "compound unique key" (as defined through the `unique_keys` slot);
- a "singular unique key" (as defined by marking a slot with `key: true`);
- and an "identifier" (as defined by marking a slot with `identifier: true`).

Then other parts of the documentation are updated to ensure they use the same terminology as used in `constraints.md`, and fix some inconsistencies. Notably, the page about inlining used to say that an identifier slot was required for dict inlining, which is not true: dict inlining is also possible with a key slot (aka a singular unique key).

## How was this tested?

I tried reading the updated documentation to my cat. It’s unclear that she understood any better, but that may be due to the fact that, well, she is a cat, and smart as she is, her overall understanding of LinkML is limited.

## Areas of uncertainty

(1) I am not certain that I myself correctly understood all the subtleties behind `unique_keys`, `key: true` and `identifier: true`, due to the confusing state of the documentation (as outlined in #3329) that I am precisely trying to fix.

(2) The terminology proposed in the updated `constraints.md`:

* **unique keys** (encompasses everything: `unique_keys`, `key: true`, and `identifier: true`)
    * **compound unique keys** (corresponds precisely to `unique_keys`)
    * **singular keys** (encompasses both `key: true` and `identifier: true`)
        * **singular unique key** (corresponds precisely to `key: true`)
        * **identifier** (corresponds precisely to `identifier: true`)

could probably be better. In particular, it’s definitely not great that what the documentation is calling “unique key” does _not_ directly correspond to what the LinkML meta model calls `unique_keys`.

However our hands are kinda tied here, since we cannot change the LinkML meta model (if we could I would love to rename `unique_keys` to `compound_keys` but that’s obviously not an option).

Open to suggestions for a better terminology.

## Checklist

- ~~[ ] I have added tests that prove my fix/feature works~~ documentation fix only
- ~~[ ] Existing tests pass locally with my changes~~ documentation fix only

## AI Assistance

None.


amc-corey-cox (approved):
This looks good to me.

Issue #3623, opened 2026-06-10
Audit and harden supply-chain controls on linkml org and release pipeline


matentzn:
A worm like Shai-Hulud or one of its mini variants has already chewed through parts of our wider ecosystem (pyphetools, phenopackets store, embiggen and other graph pipelines). linkml is a foundational dependency for a lot of bioinformatics and semantic-web tooling. If a trojanized version of `linkml` or `linkml-runtime` ever lands on PyPI, the blast radius is enormous. I want to use this issue to audit what we have, decide what to add, and close the gaps.

I'm not a security specialist, just want to start the discussion.

Recent real-world incidents that motivate this:

- jFrog write-up on Shai-Hulud (self-propagating npm worm that steals maintainer tokens and republishes trojanized packages): https://research.jfrog.com/post/shai-hulud-the-second-coming-remediation-guidance/
- Socket on mini Shai-Hulud (Miasma / Hades) hitting bioinformatics and MCP developers, including packages adjacent to ours: https://socket.dev/blog/mini-shai-hulud-miasma-and-hades-worms-target-bioinformatics-and-mcp-developers-via-malicious
- Socket on the TanStack npm compromise (same pattern, different ecosystem): https://socket.dev/blog/tanstack-npm-packages-compromised-mini-shai-hulud-supply-chain-attack

The pattern in all three: a maintainer's CI or local token gets exfiltrated, the attacker publishes a malicious version, that version self-propagates and runs install-time payloads on every consumer. We are a high-value target for exactly this. We already do a few of the right things (trusted publishing for PyPI, no direct commits to `main`), but there is a long list of standard controls we have either not turned on or not verified.

What I know we have today (from a quick look at this repo):

- `pypi-publish.yaml` already uses PyPI trusted publishing (OIDC, no long-lived token). Good.
- `main` is protected against direct commits.
- No `SECURITY.md`, but `CODEOWNERS` (maybe not used well enough).
- Workflows pin third-party actions to version tags (e.g. `actions/checkout@v6`), **not full commit SHAs.**
- No top-level `permissions: contents: read` default on workflows, so they inherit the default token scope.

## Areas to audit and decide on

Grouping these so we can triage them in chunks. 

PyPI publishing:
- Confirm trusted publishing is set up for both `linkml` and `linkml-model` (and any other package we own, like copier).
- Require 2FA for every maintainer with PyPI upload rights. Audit who currently has rights and remove anyone dormant.
[...]

PR #3650, opened 2026-06-16
add uv audit test and cooldown, as well as dependebot cooldown.  to pass security audit, bumped several package dependencies.  Needs careful review.

sierra-moxon:
## Summary

related to https://github.com/linkml/linkml/issues/3623

1. uv audit — scans uv.lock for dependencies with known security vulnerabilities (CVEs). Fails the job if any are found.
2. uv sync --frozen with UV_MALWARE_CHECK=1 — when uv installs packages, it looks them up against OSV's malware (MAL) advisory database and blocks the install if a package is flagged as malicious.

The malware gate only blocks packages that already have an OSV advisory. There is a window between a malicious package being uploaded and an advisory being published, during which the gate will not catch it.

uv's https://docs.astral.sh/uv/concepts/resolution/#dependency-cooldowns (exclude-newer, added in uv 0.9.17) addresses this gap by refusing to resolve package versions newer than a chosen window (e.g. 7 days).  

<edit> per discussion with tech team, also add cooldowns to both uv and dependebot:
  1. pyproject.toml — added a [tool.uv] table with exclude-newer = "7 days".
  2. uv.lock — regenerated.
  3. 8 workflows — bumped UV_VERSION: "0.7.13" → "0.11.21" (required; cooldown needs uv ≥ 0.9.17)
  4. dependency-audit.yaml — added UV_EXCLUDE_NEWER: "7 days" to the env block
  5. .github/dependabot.yml — default 7 day cooldown
  6. updated several dependencies flagged for security concerns, except pytest:
       "tests/conftest.py:15 imports the private pytest API _diff_text, used at line 107-109 to render snapshot-mismatch diffs. pytest 9 removed that private function, so the whole suite fails to collect. Notably, the existing TODO at lines 104-105 already flags this as fragile and suggests moving to difflib."  But in an effort to keep this PR small enough to review, I would really rather not upgrade pytest and change to difflib in this same PR.  So instead I decided to skip this security fix explicity (for the short term, pytest update and difflib change coming in another PR), eg.: "run: uv audit --ignore GHSA-6w46-j5rx-g56g" 

## How was this tested?

ran the uv commands encoded in the action locally.  
workflow will run on draft PR. 
55 known CVEs in the current lock, so this action will fail today.  

- [n/a] I have added tests that prove my fix/feature works

For reviewer: We should ensure via CI/CD that uv bump is ok.


.github/workflows/dependency-audit.yaml
```diff
@@ -0,0 +1,67 @@
+name: Audit dependencies
+
+# Global environment variables applied to all steps
+env:
+  UV_PREVIEW: "1"          # Enables the preview uv audit and malware engines
+  UV_MALWARE_CHECK: "1"    # Automatically blocks malicious installs on sync/run
+  UV_EXCLUDE_NEWER: "7 days"  # Mirror the pyproject.toml cooldown for this job's resolution
```
pkalita-lbl: It's a little unclear to me what the purpose of setting `UV_EXCLUDE_NEWER` to the same value that's already set in `pyproject.toml` is here.

pkalita-lbl (commented):
I like the approach. I just have one comment, and I think Copilot's suggestions are worth considering as well.

.github/workflows/dependency-audit.yaml
sierra-moxon: removed.

sierra-moxon:
@amc-corey-cox - I think this is ready for merge when you are.

sierra-moxon:
@matentzn - bumping for merge?

matentzn (approved):
Note that relative-duration syntax needs uv ≥ 0.9.17 so developers need to upgrade if they have not already.

Issue #2714, opened 2025-05-19
gen_doc ignores `description` supplied to qualify a value of the `examples` in a schema.


proccaserra:
**Describe the bug**
A clear and concise description of what the bug is (including the error message that is printed, if any)
* part1: 
```linkml-lint --validate instrument.yaml``` throws ```Error: In classes > Instrument > examples > [0]: Additional properties are not allowed ('value_description' was unexpected) (valid-schema)```
to pass validation/linting, one needs to use the 'description' alias instead of the 'value_description' specified by the metadata schema.

* part2:
```gen-doc instrument.yaml -d ./instrument_doc``` generates the markdown documentation but in the section 'examples', only 'value' is rendered, 'description' is ignored/dropped

snippet of rendered markdown:

##Examples:
| Value |
| --- |
| centrifuge |
| microtome |

**Version of LinkML you are using**
1.9.2

**Please provide a schema (and if applicable, a data file) that replicates the issue**

```linkml schema:
id: test
name: instrument
prefixes:
  linkml: https://w3id.org/linkml/
  namespace: http://example.com

import:
  - linkml:types
default_range: string

classes:
 Instrument:
  name: Instrument
  description: "a manufactured device"
  alt_descriptions:
    alt_description_text: "another description"
    alt_description_source: "another source"
  slots:
    - manufacturer
  examples:
     - value: centrifuge
       description: "some description"
     - value: microtome
       description: "some other description"
slots:
  manufacturer:
    description: "the company responsible for producing the instrument"
    range: string
    required: true
````

Additional Observations:
 'alt_descriptions' are not rendered by gen-doc AND the linkml linter/validator does not seem to accept more than one value.

PR #3652, opened 2026-06-17
fix(docgen): render example descriptions in the Examples table (#2714)

gaoflow:
Fixes part of #2714.

### Problem

The docgen Examples table only rendered each example's `value`. The metamodel `Example` class has a `value_description` slot (exposed as `.description`), which the issue uses to qualify a value, but the template dropped it entirely, so the rendered docs lost that information.

### Fix

`packages/linkml/src/linkml/generators/docgen/common_metadata.md.jinja2`: add a `Description` column to the Examples table and render `x.description`, falling back to an empty cell when an example has no description (so missing descriptions don't render as `None`). This affects every element that renders examples (classes, slots, etc.), which is the intended behaviour.

### Test

`test_docgen_example_description` builds an inline schema with two examples (one with a description, one without) and asserts both rows render correctly. Without the template change the test fails; with it the full `test_docgen.py` module passes (44 passed, 1 skipped).

Scope: this addresses the gen-doc rendering half of #2714. The validation/alias question raised in the first half of the issue is independent and left untouched.

Disclosure: I used AI assistance (Claude) to help locate and draft this fix, under my direction and review.

### DIFF
--- a/packages/linkml/src/linkml/generators/docgen/common_metadata.md.jinja2
+++ b/packages/linkml/src/linkml/generators/docgen/common_metadata.md.jinja2
@@ -33,10 +33,10 @@
 {% if element.examples %}
 ## Examples
 
-| Value |
-| --- |
+| Value | Description |
+| --- | --- |
 {% for x in element.examples -%}
-| {{ x.value }} |
+| {{ x.value }} | {{ x.description if x.description is not none else '' }} |
 {% endfor %}
 {% endif -%}
 
--- a/tests/linkml/test_generators/test_docgen.py
+++ b/tests/linkml/test_generators/test_docgen.py
@@ -1730,3 +1730,46 @@ classes:
     # Each file must describe only its own element type
     assert_mdfile_does_not_contain(class_file, "# Subset:")
     assert_mdfile_does_not_contain(subset_file, "# Class:")
+
+
+def test_docgen_example_description(tmp_path):
+    """Regression for linkml/linkml#2714 -- gen-doc must render the ``description`` of
+    metamodel ``examples`` entries, not just their ``value``.
+
+    The Examples table previously had a single ``Value`` column and silently dropped
+    each example's ``description``.
+    """
+    schema_yaml = """\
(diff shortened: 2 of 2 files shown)


binggao1230:
Rebased this onto the latest main (head is now 521aaee7b8259981c3e856f85cac4030cdabf7d9). Local verification passed:

- uv run pytest tests/linkml/test_generators/test_docgen.py::test_docgen_example_description -q
- uv run pytest tests/linkml/test_generators/test_docgen.py -q
- uv run ruff check tests/linkml/test_generators/test_docgen.py
- uv run ruff format --check tests/linkml/test_generators/test_docgen.py

packages/linkml/src/linkml/generators/docgen/common_metadata.md.jinja2
```diff
@@ -33,10 +33,10 @@
 {% if element.examples %}
 ## Examples
 
-| Value |
-| --- |
+| Value | Description |
+| --- | --- |
 {% for x in element.examples -%}
-| {{ x.value }} |
+| {{ x.value }} | {{ x.description if x.description is not none else '' }} |
```
sujaypatil96, line 39: Thanks for the changes in this PR @gaoflow! Fully agree with the sentiment/philosophy here, but in a lot of linkml schemas that i've seen written for projects, i've rarely ever seen the description field being filled out for `examples`, causing it to be empty more often that not.

If you have a project where you are diligently populating the description for an example as well, then perhaps you can create a custom jinja template and modify the `common_metadata.md.jinja2` file's `## Examples` section to show a table with `Value` and `Description` columns for examples instead of just `Value`?

Whether or not this should be in the generic base template should probably be a bigger group decision. We should definitely make an issue for it and source some feedback from the community.

binggao1230:
Agreed the base-template placement is a group call. I opened linkml/linkml#3824 to source community feedback on whether the Description column should be default; happy to narrow this PR to an opt-in custom template instead if the discussion lands that way.

PR #3658, opened 2026-06-18
javagen: fix extra_slots support

### LINKED ISSUE #3657: Java generator may generate bogus `extraSlots` maps in derived classes
**Describe the bug**
The Java generator handles the `extra_slots.allowed` feature by generating, for any class that is configured to allow extra slots, an additional `Map<String,Object> extraSlots;` field intended to store the value of any extra slot (this is what the LinkML-Java runtime calls the “extension holder”).

However, that additional field only needs to be added _once_ in any class hierarchy. Once a parent class has it, all its subclasses automatically inherit it, and the field _must not_ be added again in a subclass, even if the subclass has an explicit (and redundant) `extra_slots.allowed` attribute.

Tip of the main branch as of 18 June 2026.

Minimal schema:

```yaml
id: https://example.org/redundant-extra-slots
name: redundant-extra-slots
prefixes:
imports:
  - linkml:types
default_range: string

classes:

  Foo:
    extra_slots:
      allowed: true

  Bar:
    is_a: Foo
    extra_slots:
      allowed: true
```

Note how the `Bar` class has a redundant `extra_slots.allowed: true` setting (redundant since the class is already set to allow extra slots, by virtue of being derived from `Foo` which already has a `extra_slots.allowed: true`).

From LinkML’s point of view the redundant `extra_slots.allowed` is harmless (it does not do anything useful, but it does not do any harm either), _but_ the Java generator would erroneously produce an `extraSlots` map in _both_ the `Foo` class and the `Bar` class, while only the `Foo` class needs it.

gouttegd:
## Summary

Fixes #3657

This PR primarily fixes #3657 by making sure that the Java generator does _not_ generate a redundant `extraSlots` map in a class that already inherits such a map from one of its ancestors.

Along the way, it includes two unrelated fixes for the Java generator:

(1) It fixes the problem that, when an `extraSlots` map field is present in a class, the field was previously not taken into account when (a) comparing instances of the class (with the `equals()` method) and (b) computing the hash for an instance (with the `hashCode()` method).

(2) It removes some code within `javagen` that needlessly duplicated functionalities that are provided by `SchemaView`.

## How was this tested?

Checked that existing tests pass, added a new test specifically for the core feature of the PR.

## Areas of uncertainty

None.

## Checklist

## AI Assistance

None.


tests/linkml/utils/fileutils.py
```diff
@@ -18,4 +19,4 @@ def assert_file_contains(filename, text, after=None) -> None:
                         found = True
             if after is not None and after in line:
                 is_after = True
-    assert found
+    assert found or invert
```
kevinschaper: When invert is true this will return true no matter what, I _think_ `found != invert` would make more sense here?

kevinschaper (changes requested):
Looks good, with (I think) one small bug

tests/linkml/utils/fileutils.py
gouttegd: > When invert is true this will return true no matter what

No, I think it’s the opposite: when `found` (not `invert`) is true, this will return true no matter what. The `or` operator does not evaluate its second operand when the first evaluates to `true`, so if `found` is true, then the entire expression evaluates to `true` regardless of the value of `invert`.

> I _think_ `found != invert` would make more sense here?

I agree.

packages/linkml/src/linkml/generators/javagen.py
```diff
@@ -441,6 +429,22 @@ def get_read_accessor_name(self, slot: SlotDefinition) -> str:
     def get_slot_actual_name(self, slot: SlotDefinition) -> str:
         return slot.alias if slot.alias and self.use_aliases else slot.name
 
+    def needs_extra_slots(self, klass: ClassDefinition) -> bool:
+        """Indicates whether a class requires handling of extra slots.
+
+        A Java class representing a LinkML class requires special code to
+        handle extra slots iff (1) the LinkML class is configured to allow
```
noelmcloughlin: nit `iff` -> `if`

packages/linkml/src/linkml/generators/javagen.py
```diff
@@ -441,6 +429,22 @@ def get_read_accessor_name(self, slot: SlotDefinition) -> str:
...
+        extra slots and (2) none of its parent classes are already
+        configured to allow extra slots.
+        """
+        if klass.extra_slots is None or not klass.extra_slots.allowed:
```
noelmcloughlin, line 444: Per https://linkml.io/linkml-model/latest/docs/extra_slots/  `allowed` is defined on applicable `ClassDefinition" but could be `null` (None) so would extra guards be needed?

```
>>> bob=None
>>> not bob
True
```
Searching for `extra_slots.range_expression` returns a hit in jsonschemagen with extra guards.
https://github.com/linkml/linkml/blob/65bc9bf35d70a936e9fbc4ad66be2e3a5b025b2c/packages/linkml/src/linkml/generators/jsonschemagen.py#L966-L971

packages/linkml/src/linkml/generators/javagen.py
```diff
@@ -441,6 +429,22 @@ def get_read_accessor_name(self, slot: SlotDefinition) -> str:
...
+            return False
+        for ancestor in self.schemaview.class_ancestors(klass.name, mixins=False, reflexive=False):
```
noelmcloughlin, line 446: Regarding "_and (2) none of its parent classes are already configured to allow extra slots_."

I understand that `extra_slots` has no inheritance so does it matter if parent classes allow?
Is `mixins=False` intended (dropping mixin-declared extra_slots)?

packages/linkml/src/linkml/generators/javagen.py
```diff
@@ -441,6 +429,22 @@ def get_read_accessor_name(self, slot: SlotDefinition) -> str:
...
+            ancestor_extra_slots = self.schemaview.get_class(ancestor).extra_slots
+            if ancestor_extra_slots is not None and ancestor_extra_slots.allowed:
```
noelmcloughlin, line 448: Similar issue to above, `allowed` may be unset.

noelmcloughlin (commented):
I just had time for a quick review and added some comments.
This jinja2 looks correct to me.  
I will be offline for over a week so I will assign core team as reviewers.

packages/linkml/src/linkml/generators/javagen.py
gouttegd: If `allowed` is `None` (instead of being explicitly set to `True` or `False`), then `not extra_slots.allowed` will return `True`, which is all we need to know here – it doesn’t matter whether this is because `allowed` is `None` or because `allowed` is `False`.

The difference matters for the JSONSchema generator because that generator needs to behave differently in the case where `allowed` is `None`: it needs to check whether some range expressions are allowed instead. But the Java generator currently does not support the `range_expression` case, so it only cares about whether extra slots are allowed generally or not.

packages/linkml/src/linkml/generators/javagen.py
gouttegd: > I understand that extra_slots has no inheritance so does it matter if parent classes allow?

(1) It’s unclear to me from the spec that `extra_slots` is not supposed to be inherited, and I would find very strange that it is not.

If I do this in a schema:

```yaml
classes:
  MyBaseClass:
    ...
    extra_slots:
      allowed: true

  MyFirstDerivedClass:
    is_a: MyBaseClass

  MySecondDerivedClass:
    is_a: MyFirstDerivedClass
```

I would _not_ expect to have to explicitly set `extra_slots.allowed: True` to all subclasses of `MyBaseClass`! I would expect that the value of the parent class applies, _unless_ a derived class overrides it.

I will open a ticket to get this point clarified.

(2) Regardless of whether `extra_slots` is _intended_ to be inherited or not, the way that feature is supported in Java, it can only be inherited.

Extra slots are supported in a class by making sure that class has a `Map<String,Object> extraSlots` field that can be used to store the extra slots (this is necessary since, contrary to Python, in Java you cannot just dynamically add fields to a class). And once a class has that `extraSlots` field, then all its subclasses necessarily have it too.

So it matters that a parent class has already allowed extra slots, because if that’s the case then the `Map<String,Object> extraSlots` field has already been added and we _must not_ add another one in the child class.

> Is `mixins=False` intended (dropping mixin-declared extra_slots)?

Yes. The use of `extra_slots` on a mixin class is even less specified than its use on a class, and I don’t think it actually makes any sense on a mixin. 

For example, what should happen if a class has one mixin that declares that it allows extra slots and another mixin that declares that it does not?

packages/linkml/src/linkml/generators/javagen.py
gouttegd: Same answer. :) We don’t care if `allowed` is `None` or if it is `False`, we just need to know whether it is `True`.

packages/linkml/src/linkml/generators/javagen.py
noelmcloughlin: Thanks for detailed clarification - my comment was too loose - it does make perfect sense to expect `extra_slots` to be inherited but glad you tracked with issue.  

[ClassDefinition](https://linkml.io/linkml-model/latest/docs/ClassDefinition/) defines many slots and presumably some are inheritable and some are not (for some reason).  For example the ClassDefintiion Linkml Source **induced** code has mixture of true/false inherited values.

For Mixins, I tend to agree - there may be DRY reasons to expect *Mixin -> extra_slots: true* to be attached to consuming classes - but should be "use case" driven - and this PR does not present a use case.

packages/linkml/src/linkml/generators/javagen/_visitor.jinja2
```diff
@@ -16,7 +16,7 @@ package {{ doc.package }};
  */
 public interface {{ doc.name }} {
     public void visit({{ gen.get_class_name(doc.visited_object) }} visited);
-{%- for child in gen.get_descendants(doc.visited_object) %}
+{%- for child in gen.get_descendants(doc.visited_object)|sort %}
```
noelmcloughlin, line 19: I was tracing back through `Javagen.py` to see how input is passed to see if extra jinja2 validation is needed on inputs (and logged some minor issues).  

**Optionally** you could add some guards - I have worked enough with Jina2 to know it does not handle bad input gracefully - but ideally input is validated in `javagen.py` as trust relationship starts there - templates should trust/expect good input is passed.

```python
{%- set descendants = gen.get_descendants(doc.visited_object) if doc.visited_object is defined and doc.visited_object else [] %}
{%- for child in (descendants or []) | sort(attribute='name') %}
    {%- set class_name = gen.get_class_name(child) %}
    {%- if class_name %}
    public void visit({{ class_name }} visited);
    {%- endif %}
{%- endfor %}
```

noelmcloughlin (commented):
I just had time for a quick review and added some comments.
This jinja2 looks correct to me.  
I will be offline for over a week so I will assign core team as reviewers.

packages/linkml/src/linkml/generators/javagen/_visitor.jinja2
gouttegd: > Optionally you could add some guards

The one thing we should guard against is the possibility that the argument to `--visitor` (the name of the class to visit) does not match a known class name in the schema.

But this is something I would much rather do in `javagen.py` rather than in the templates. Will do that know. :+1:

kevinschaper (approved):

Issue #3408, opened 2026-04-17
Re-add morph-kgc to test dependencies once pyoxigraph pin is resolved


matentzn:
morph-kgc was removed from the `tests` dependency group in #3407 because it pins `pyoxigraph < 0.4.0`, which conflicts with the new `pyoxigraph >= 0.5.6` requirement for deterministic RDF serialization. The YARRRML tests use `pytest.importorskip("morph_kgc")` so they skip gracefully, but those tests are now effectively dead cc @Ostrzyciel @lapkinvladimir 

## Context

- `packages/linkml/pyproject.toml` previously had `"morph-kgc >= 2.9.0; python_version >= '3.10'"` in the `tests` group
- morph-kgc pins `pyoxigraph < 0.4.0` upstream: https://github.com/morph-kgc/morph-kgc
- Once morph-kgc relaxes or drops that pin, we should bring back the dependency and re-enable the YARRRML test coverage

## Suggested approach

Periodically check morph-kgc releases for an updated pyoxigraph pin. When compatible, re-add `"morph-kgc >= <version>; python_version >= '3.10'"` to the test deps.

## Related

- #3407 — PR that introduced the pyoxigraph dependency and removed morph-kgc

melonora:
Just to note here, that morph_kgc depends on duckdb. Standard version being installed on python 3.14 is version 1.4.1 of duckdb which takes an excessive amount of time to build (20 minutes is not uncommon). Morph-kgc itself does not officially support python 3.14 yet which could be a reason why they did not push yet for one of the newer duckdb versions. I will most likely check early next week whether I can push morph-kgc to python 3.14 with a newer version of duckdb that hopefully will provide a better experience.

melonora:
I fixed the comment above regarding slow build differently in https://github.com/linkml/linkml/pull/3172. It was possible to add duckdb as a dev dependency with a lowerbound set to 1.5.2. This builds quickly in python 3.10-3.14.

PR #3664, opened 2026-06-22
test(yarrrmlgen): unit suite to 100% line coverage, plus prefix and blank-node fixes

nicholsn:
## Summary

Related to #3408

The YARRRML generator had almost no line coverage because its only tests
(`tests/linkml/test_generators/test_yarrrmlgen.py`) materialise RDF with the
optional `morph_kgc` dependency and are skipped whenever it is absent — the
default, and the situation tracked in #3408. This PR adds a **dependency-free
unit suite** (`test_yarrrmlgen_unit.py`) that drives `YarrrmlGenerator` directly
on small in-memory schemas and asserts on the structure of the emitted YARRRML
document, bringing `yarrrmlgen.py` to **100% line coverage**.

Writing those tests surfaced two output bugs, each fixed in its own commit:

1. **`fix(yarrrmlgen): resolve default_prefix to a declared CURIE prefix`** —
   when a schema omits `default_prefix`, LinkML synthesises it from the full
   schema IRI. The generator used that verbatim as a CURIE prefix, emitting
   malformed YARRRML such as `s: https://example.org/:$(id)` and
   `p: https://example.org/id`. Prefixes are now resolved before mappings are
   built, and a `default_prefix` that isn't a declared prefix is replaced with a
   real one (`ex`, else the first non-reserved declared prefix).

2. **`fix(yarrrmlgen): emit a blank node for classes without an identifier`** —
   a class with no identifier, key, or owning-parent id produced the subject
   `ex:Class/$(id)`, referencing an `id` field the class does not have. It now
   omits the subject, which is how YARRRML expresses a blank node — the correct
   representation for a resource with no identity.

The branch is organised as three atomic commits (the two fixes, then the test
suite), so each is self-contained and the test suite reflects the fixed behaviour.

## How was this tested?

- **New unit suite** (`test_yarrrmlgen_unit.py`): 48 mock-free tests that exercise
  the real generator and assert on the emitted document — source inference,
  iterators, subject templates, scalar/object predicate-objects, inlined-object
  joins, mixins, prefix handling, `tree_root` filtering, validation errors,
  serialisation, and the CLI. The two fixes are pinned via the public `as_dict()`
  path.
- **Coverage** — `uv run pytest tests/linkml/test_generators/test_yarrrmlgen_unit.py --cov --cov-report=term-missing`
  reports **100% line coverage** of `yarrrmlgen.py` (237/237 statements; the four
  remaining partial branches are unreachable defensive guards). Note: use bare
  `--cov` (the repo's `.coveragerc` scopes to the `linkml` package); scoping
  `--cov` to the submodule breaks metamodel loading under coverage.
- The existing `test_yarrrmlgen.py` is unaffected (it still skips without
  `morph_kgc`), and `ruff` (0.11.13) check + format pass.

## Areas of uncertainty

- **Blank-node subjects**: confirming that omitting `s` is the intended YARRRML
  representation for id-less classes. A pre-existing case — an inlined id-less
  child of an id-less parent — still emits a `{mapping: …}` reference with no join
  condition; this change does not make that worse, but reviewers may want to weigh
  in on the broader handling.
- **Prefix normalisation**: `_prefixes_with_defaults()` now also normalises
  `self.schema.default_prefix` and runs before the mapping loop so `get_uri`-derived
  predicates/types pick it up. For **imported** elements, `get_uri` resolves via the
  imported schema's own `default_prefix`, so subjects and predicates could diverge
  there (pre-existing) — an extra set of eyes welcome.
- **Enum `meaning` IRIs (intentionally not changed)**: enum-ranged slots are emitted
  as plain literals. Standard LinkML JSON data carries the enum *code* (e.g.
  `"active"`), so the literal is faithful; resolving it to a permissible value's
  `meaning` IRI would need a code→IRI lookup (an FnO function or a generated
  enum-table join) that isn't portably expressible in a single mapping. Flagged as a
  possible follow-up.

## Checklist

## AI Assistance


Ostrzyciel:
Hey, just noticed that I was nominated for CR due to CODEOWNERS... well, we are no longer using this generator, nor will be developing it further, so I will make a PR to remove me from CODEOWNERS.

In short, we have switched to developing [LinkML-Scala](https://github.com/NeverBlink-OSS/linkml-scala) as a more portable and robust solution. We have also given up on RML and YARRRML entirely.

packages/linkml/src/linkml/generators/yarrrmlgen.py
matentzn: `None | str` is now a bit more precise of an outcome for this function rather than `Any`...

packages/linkml/src/linkml/generators/yarrrmlgen.py
```diff
@@ -200,7 +214,10 @@ def _prefixes_with_defaults(self) -> dict[str, str]:
         if not has_user_prefix:
             px.setdefault("ex", "https://example.org/default#")
 
-        if not self.schema.default_prefix:
+        # Replace a missing default_prefix, or one that is not a usable CURIE
+        # prefix (e.g. the full schema IRI LinkML synthesises when it is omitted),
+        # with a real declared prefix.
+        if not self.schema.default_prefix or self.schema.default_prefix not in px:
             if "ex" in px:
                 self.schema.default_prefix = "ex"
```
matentzn: While in there: mutating `self.schema.default_prefix` is affects callers. I confirmed `gen.schema.default_prefix` goes from `'https://example.org/test/'` to `'ex'` across an `as_dict()` call, so anyone who hands you `sv.schema` gets their schema silently rewritten. It's pre-existing, but you've broadened when it fires — it used to need a falsy `default_prefix`, now any value that isn't a key in `px` triggers it. Documenting it in the docstring is an improvement on what was there; resolving into a local and leaving the input alone would be better.

matentzn (commented):
Schema with no declared prefixes beyond `linkml`, `default_prefix` are omitted so LinkML synthesises `https://example.org/test/` from the schema IRI. 

1. Before: malformed CURIEs. 
2. After: `ex: https://example.org/default#`, and every subject and predicate lands under `example.org/default#`.

That's better than malformed, but it throws away the schema's actual namespace — which is sitting right there in the synthesised `default_prefix` you're discarding. Register the synthesised IRI under `ex` than substitute the placeholder?

```python
# default_prefix is an IRI, not a declared prefix -> keep the namespace, name it
px["ex"] = str(self.schema.default_prefix)
self.schema.default_prefix = "ex"
```

packages/linkml/src/linkml/generators/yarrrmlgen.py
nicholsn: Done — _subject_template_for_class is now -> str | None

packages/linkml/src/linkml/generators/yarrrmlgen.py
nicholsn: Agreed, and thanks for actually running it. It resolves into a local now: _prefixes_with_defaults() returns (prefixes, default_prefix), and as_dict() passes the resolved prefix down to _subject_template_for_class / _po_list_for_class. Nothing writes to self.schema any more. Added a regression test that hands the generator a live SchemaDefinition and asserts default_prefix is untouched after as_dict().

nicholsn:
> Schema with no declared prefixes beyond `linkml`, `default_prefix` are omitted so LinkML synthesises `https://example.org/test/` from the schema IRI.
> 
> 1. Before: malformed CURIEs.
> 2. After: `ex: https://example.org/default#`, and every subject and predicate lands under `example.org/default#`.
> 
> That's better than malformed, but it throws away the schema's actual namespace — which is sitting right there in the synthesised `default_prefix` you're discarding. Register the synthesised IRI under `ex` than substitute the placeholder?
> 
> ```python
> # default_prefix is an IRI, not a declared prefix -> keep the namespace, name it
> px["ex"] = str(self.schema.default_prefix)
> self.schema.default_prefix = "ex"
> ```

Good catch — substituting the placeholder was the wrong trade. The synthesised IRI is now declared instead of discarded, so a schema with id: https://example.org/test and no default_prefix emits ex: https://example.org/test/ and s: ex:$(id) — real namespace, well-formed CURIE. Two edge cases fell out of it: if ex is already bound to a different namespace we bind ex1 rather than clobbering it, and if a declared prefix already points at the schema namespace we reuse that name instead of adding a duplicate. Both are tested.

One consequence worth your eyes: because schema.default_prefix is no longer rewritten, get_uri()-derived predicates and types still come out as full IRIs in this case (p: https://example.org/test/id) while the subject is a CURIE. They expand to the same thing, and the IRIs are exactly what main already emits — the only genuinely broken output was the subject. Happy to add a small shortener so all three positions render as ex:… if you'd prefer the uniformity.

nicholsn:
@matentzn Is this good? LMK if you want additional changes!

PR #3675, opened 2026-06-25
block not yet supported python versions

### LINKED ISSUE #3674: block installation of not-yet supported python versions
Current `pyproject.toml` gives a minimum supported Python version, but not a maximum. Since we tend to lag behind Python version releases, it makes sense to also specify the highest supported version.

Silvanoc:
## Summary

Fixes #3674

Block installation on upcoming, not-yet supported Python versions.

## Checklist

## AI Assistance

### DIFF
--- a/packages/linkml/pyproject.toml
+++ b/packages/linkml/pyproject.toml
@@ -35,7 +35,7 @@ classifiers = [
     "Programming Language :: Python :: 3.13",
     "Programming Language :: Python :: 3.14",
 ]
-requires-python = ">=3.10"
+requires-python = ">=3.10,<3.15"
 dynamic = ["version"]
 dependencies = [ # Specifier syntax: https://peps.python.org/pep-0631/
     "antlr4-python3-runtime >= 4.9.0, < 4.10",
--- a/packages/linkml_runtime/pyproject.toml
+++ b/packages/linkml_runtime/pyproject.toml
@@ -32,7 +32,7 @@ classifiers = [
     "Programming Language :: Python :: 3.13",
     "Programming Language :: Python :: 3.14",
 ]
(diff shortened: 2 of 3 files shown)


matentzn (changes requested):
I am personally against adding an upper Python cap - not severely (so can be swayed), but fairly strongly. 

For a foundational library like linkml, an upper bound on python_requires doesn't just affect linkml, it propagates to everything downstream, and resolvers respond by either silently backtracking to an ancient linkml or failing the whole environment with an error that misleadingly points at linkml. That's usually worse than the runtime failure it's meant to prevent.

Also, new Python minors almost always just work, so we'd be blocking versions that could've worked fine and cutting a release every cycle just to bump the number. Capping the lower bound is of course important; capping the upper bound for a library is widely considered an anti-pattern (see [Henry Schreiner's writeup and the scientific-Python packaging guidance](https://iscinumpy.dev/post/bound-version-constraints/)).

If the worry is untested versions, I'd rather test against new/prerelease Python in CI and ship an informed constraint only if something actually breaks.

Silvanoc:
> Also, new Python minors almost always just work

I have a different experience... that's why I came to the idea of adding the upper bound 😄 

> capping the upper bound for a library is widely considered an anti-pattern (see [Henry Schreiner's writeup and the scientific-Python packaging guidance](https://iscinumpy.dev/post/bound-version-constraints/)).

Even the TL;DR is too long for a topic I don't have real stakes on. I'll then simply drop it.

> If the worry is untested versions, I'd rather test against new/prerelease Python in CI and ship an informed constraint only if something actually breaks.

I've already added a reminder on my calendar for 2026.08.04, when the first Python v3.15 RC comes out 😉

PR #3678, opened 2026-06-26
oocodegen: Produce correct slot URI for locally defined attributes.

### LINKED ISSUE #3677: The object code generator produces wrong slot URI for local, class-specific attributes
**Describe the bug**
The `OOCodeGen` generator (used notably by the Java generator) may produce an incorrect value for the `slot_uri` field of the `OOField` class.

Tip of the main branch as of June 26th.

Consider first the following schema:

```yaml
id: https://example.org/slot-uri-on-attribute
name: slot-uri-on-attribute
prefixes:
imports:
  - linkml:types
default_range: string

classes:

  PrefixDeclaration:
    attributes:
      name:
        slot_uri: "http://www.w3.org/ns/shacl#prefix"
      prefix:
        range: uri
        slot_uri: "http://www.w3.org/ns/shacl#namespace"
```

The `PrefixDeclaration` class has two local, class-specific attributes that both have an explicitly assigned slot URI.

No problem here, the Java code generator (and thus the underlying object code generator) produces the expected output:

```java
@LinkURI("http://www.w3.org/ns/shacl#prefix")
private String name;

@LinkURI("http://www.w3.org/ns/shacl#namespace")
private URI prefix;
```

But now consider this second schema:

```
id: https://example.org/slot-uri-on-attribute-2
name: slot-uri-on-attribute-2
prefixes:
imports:
  - linkml:types
default_range: string

classes:

  PrefixDeclaration:
    attributes:
      name:
        slot_uri: "http://www.w3.org/ns/shacl#prefix"
      prefix:
        range: uri
        slot_uri: "http://www.w3.org/ns/shacl#namespace"

  AnotherClass:
    attributes:
      name:
```

This is the same schema as above, but this time with a second class that happens to also have a local, class-specific attribute named `name` (which is perfectly valid; the point of `attributes` is precisely that they are class-specific, so it’s perfectly fine for two classes to have identically named attributes).

But now the generated code is no longer correct:

```java
@LinkURI("https://example.org/slot-uri-on-attribute-2/name")
private String name;

@LinkURI("http://www.w3.org/ns/shacl#namespace")
private URI prefix;
```

The presence of the identically named attribute in a completely unrelated class somehow confuses the OOCodeGenerator, which now ignores the `slot_uri` slot.

**Cause**
The cause of the bug is on this line in the `oocodegen` module:

```python
slot_uri=sv.get_uri(slot.name, expand=True)
```

This calls the `SchemaView::get_uri` method with the slot _name_, forcing this method to perform a name-based lookup to find the slot. This is unnecessary since (1) at this point we already have a `SlotDefinition` object corresponding to the attribute (which has the correct `slot_uri` definition) and (2) the `SchemaView::get_uri` is perfectly happy to accept a `Element` (that is, a full definition) rather than only a `ElementName`.

gouttegd:
## Summary

Fixes #3677

When the object code generator produces the OOField object representing a field in a class (where the field is itself the representation of a LinkML slot or a LinkML attribute), it fills the `slot_uri` member by calling the `SchemaView::get_uri` method and passing it the slot/attribute's name (rather than the slot/attribute's definition).

This forces the SchemaView to look up for the actual definition from the specified name, which it may fail to do correctly if the name is not the name of a globally defined slot but of a locally defined attribute (which is expected; you cannot lookup a locally defined attribute by its name only).

The fix is to provide SchemaView directly with the correct SlotDefinition object (which the OOCodeGen already has), dispensing it from having to look it up.

## How was this tested?

Existing tests pass. A new test was added to test specifically the case that triggered this specific bug (two different classes that both have an identically named local attribute).

## Areas of uncertainty

None.

## Checklist

## AI Assistance

None.


kevinschaper (approved):
Nice catch @gouttegd!

PR #3683, opened 2026-06-26
Make EnumDefinitionImpl instances hashable

### LINKED ISSUE #3590: EnumDefinitionImpl instances are unhashable (TypeError)

SAY-5:
## Summary

Fixes #3590

`EnumDefinitionImpl` inherits `__eq__` from `jsonasobj2` (via `YAMLRoot`) but never defines `__hash__`, so Python sets `__hash__` to `None` and instances become unhashable. This breaks `_normalize_inlined()` in `yamlutils.py`, which adds enum-typed keys to a `set` for duplicate detection, with `TypeError: unhashable type`. The minimal reproducer from the issue (`hash(E("A"))`) also raises.

I added a `__hash__` that hashes on the permissible value text. Two equal enum values always share the same `_code.text`, so this stays consistent with the inherited equality and matches what `__str__` already returns.

## How was this tested?

Added a regression test in `tests/linkml_runtime/test_utils/test_enumerations.py` that checks the enum value is hashable, that equal values hash equally, and that it works as a set member and dict key. Confirmed the test fails on `main` (TypeError) and passes with this change. The existing tests in that module still pass.

## Areas of uncertainty

Hashing on `_code.text` (rather than including the class) is deliberate: cross-class enum values with the same code compare equal under the inherited `__eq__`, so folding in the class would break the equal-implies-equal-hash invariant.

## Checklist

## AI Assistance

### DIFF
--- a/packages/linkml_runtime/src/linkml_runtime/utils/enumerations.py
+++ b/packages/linkml_runtime/src/linkml_runtime/utils/enumerations.py
@@ -117,6 +117,16 @@ class EnumDefinitionImpl(YAMLRoot, metaclass=EnumDefinitionMeta):
         """The string representation of an enumerated value should be the code representing this value."""
         return self._code.text
 
+    def __hash__(self) -> int:
+        """Hash on the permissible value text.
+
+        ``YAMLRoot`` inherits ``__eq__`` from ``jsonasobj2`` without a matching
+        ``__hash__``, which leaves ``__hash__`` set to ``None`` and makes
+        instances unhashable.  Two equal enum values always share the same
+        permissible value text, so hashing on it stays consistent with equality.
+        """
+        return hash(self._code.text)
+
     def __repr__(self) -> str:
         rlist = [(f.name, getattr(self._code, f.name)) for f in fields(self._code)]
         return self.__class__.__name__ + "(" + ", ".join([f"{f[0]}={repr(f[1])}" for f in rlist if f[1]]) + ")"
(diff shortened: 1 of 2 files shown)


amc-corey-cox (approved):
This looks good to me @SAY-5 . Sorry for the run-around on the CI.

PR #3686, opened 2026-06-29
feat: new annotation-driven multi file json schema generator

### LINKED ISSUE #3685: Feature: multi-file JSON Schema generator (gen-multifile-json-schema)

I would like to add new `gen-multifile-json-schema` to allows controlled annotation-driven generation of multiple JSON schema files as output - the generator would inherit existing mechanics of JsonSchemaGen, but add the multi-file feature.   The Annotation mechanism is generic - the same tag scheme can later drive Pydantic, Rust, or GraphQL generators through the reserved prefix space.

### Why this feature

Currently `gen-json-schema` emits one JSON Schema file per schema. But many specifications are published as **several files**, one per topic, that point at each other with `$ref: "core.json#/$defs/Foo"`. 

Consumers who need to output multiple-files can do so with difficulty - wrapping `JsonSchemaGenerator`
themselves and rebuild the same plumbing that could be captured in a dedicated multi-file JSON Schema generator does that work once, driven entirely by annotations, reserving a prefix space so future generators to reuse.

### Proposed example A - generic: publish multi-file JSON Schema

A project wants to publish `core.json`, `controls.json`, and `interfaces.json`, with classes grouped by topic and one class as the root of each file (so consumers can `$ref: "core.json"` directly):

```yaml
subsets:
  control_subset:
    annotations:
      output_file: controls               # .json suffix is optional, generator can append

classes:
  Architecture:
    annotations:
      output_file: core
      jsonschema:is_root_of_file: true
  Control:
    in_subset:
      - control_subset                    # inherits `output_file` for DRY reasons

  Interface:
    annotations:
      output_file: interfaces
      jsonschema:is_root_of_file: true
```

One command emits all three files with the cross-file `$ref`s wired up. No custom post-processor, no routing tables in build scripts.

### Example B - adopting LinkML in downstream JSON Specification-first projects

Some downstream *Specification-first* projects ship a fixed set of `.json` schema files as their official artifact. Adopting LinkML means switching to one generated file, which breaks every consumer that depends on specification as multi-file names.  This generator can help:

1. Convert the existing JSON Schema layout into a LinkML schema.
2. Annotate each class with the `output_file` (and any `jsonschema:` tags) that reproduce the current layout.
3. Run the generator and diff against the published files until the output matches exactly.
4. Switch the build to generate from LinkML. Consumers see no change.

Afterwards, the adopting project gets LinkML toolchains for free, and adoption cost is lower.

### Why the tags are tiered

If every multi-file generator invented its own tags, one schema couldn't drive a Pydantic + JSON Schema + Rust release at once, and a flat namespace would fill up with target-specific terms. So the tags split in two:

- **Tier 1** - shared concepts (`output_file`, `inline_class`,  `additional_properties`).
- **Tier 2** - target-specific concepts under a `<target>:` prefix  (`jsonschema:` now; `pydantic:`/`rust:`/`graphql:` reserved).

This prefix syntax for keay follows existing LinkML practice and is accepted by linkml lint/validate and yamllint:
- https://github.com/linkml/linkml/blob/a9af9eca09999b4bec08705c14ff3590a743b9e8/packages/linkml/src/linkml/generators/sqlalchemygen.py#L122
- https://github.com/linkml/linkml/blob/a9af9eca09999b4bec08705c14ff3590a743b9e8/packages/linkml/src/linkml/transformers/relmodel_transformer.py#L309-L310

### Possible future changes

- Maybe the Tier 1 tags listed above could become real metamodel slots in future?
- Is a `<target>:` prefix the right convention for generator-scoped tags?

noelmcloughlin:
## Summary

Fixes #3685 

- Generate JSON Schema split into multiple target files via Annotations

## How was this tested?

Tested locally via `test_generators`

## Areas of uncertainty

I think this feature should be separate from incumbent `jsonschemagen.py` as it offers a model to other generators, and avoids introducing breaking changes into a widely used generator.

## Checklist

## AI Assistance
LLM used


kevinschaper (changes requested):
Hi @noelmcloughlin 

I'm a little nervous about this PR as the place where global vs namespaced annotations are introduced, would you be ok with cutting that from the docs, and tackling it in a PR that has more of a high level scope?

noelmcloughlin:
Yes, I'm bit nervous myself about this PR, feature is valuable, just the design needs iterations and I submitted hoping to be challenged to steer towards optimal design. Will revise and consider.

PR #3688, opened 2026-06-29
fix(jsonschemagen): keep $refs for inlined any_of class unions (#3684)

### LINKED ISSUE #3684: JSON Schema generator reduces class ranges in `any_of` unions to `type: string`
**Describe the bug**
When using `linkml:Any` as a slot range combined with the `any_of` metaslot to define a range union of classes, the JSON Schema generator incorrectly produces `{"type": "string"}` for each member of the union instead of generating a `{"$ref": "#/$defs/ClassName"}`.

**Crucially, this bug only manifests when the classes listed in the `any_of` definition contain an identifier slot (a slot where `identifier: true` is set).** If the classes in the union do not have an identifier slot, the generator correctly produces `$ref` references. It appears the logic blindly reduces the class to a string type rather than preserving the class reference entirely.

This behavior happens regardless of whether the parent slot is marked as `inlined: true`, `inlined: None` or `inlined: false`.

**Expected Behavior**
Given a slot definition like:
```yaml
primary_topic:
  description: A link to the Dataset, Data service or Catalog described in the record.
  slot_uri: foaf:primaryTopic
  range: Any
  required: true
  multivalued: false
  inlined: true
  any_of:
    - range: Catalogue
    - range: Dataset
```
the generated JSON Schema should look like:
```
"primary_topic": {
    "$ref": "#/$defs/Any",
    "anyOf": [
        {
            "$ref": "#/$defs/Catalogue"
        },
        {
            "$ref": "#/$defs/Dataset"
        },
    ],
    "description": "A link to the Dataset, Data service or Catalog described in the record."
                },
```
**Actual Behavior**
The generator currently produces:
```
"primary_topic": {
    "$ref": "#/$defs/Any",
    "anyOf": [
        {
            "type": "string"
        },
        {
            "type": "string"
        },
    ],
    "description": "A link to the Dataset, Data service or Catalog described in the record."
}
``` 
**Version of LinkML you are using:** LinkML v1.11.1
**Please provide a schema (and if applicable, a data file) that replicates the issue:**
``` 
id: https://example.org/test
prefixes:
  ex: https://example.org/
default_prefix: ex

classes:
  Any:
    name: Any
    description: This abstract class is needed to create the union of Dataset, DatasetSeries, Catalogue and DataService for the range of the slot [primary_topic](https://nfdi-de.github.io/chem-dcat-ap/elements/primary_topic/).
    class_uri: linkml:Any

  CatalogueRecord:
    class_uri: dcat:CatalogRecord
    description: See [DCAT-AP specs:Catalogue](https://semiceu.github.io/DCAT-AP/releases/3.0.0/#Catalogue)
    attributes:
      primary_topic:
        description: A link to the Dataset, Data service or Catalog described in the record.
        slot_uri: foaf:primaryTopic
        range: Any
        required: true
        multivalued: false
        inlined: true
        any_of:
          - range: Catalogue
          - range: Dataset

  # Class WITH identifier -> Triggers the bug
  Catalogue:
    class_uri: dcat:Catalog
    description: See [DCAT-AP specs:Catalogue](https://semiceu.github.io/DCAT-AP/releases/3.0.0/#Catalogue)
    slots:
      - id
      - title

  # Class WITH identifier -> Triggers the bug
  Dataset:
    class_uri: dcat:Dataset
    description: See [DCAT-AP specs:Catalogue](https://semiceu.github.io/DCAT-AP/releases/3.0.0/#Catalogue)
    slots:
      - id
      - title

slots:
      id:
        identifier: true
        range: uriorcurie
      title:
        range: string
``` 

**Fix Proposed by the EURIA AI**
The issue seems to be located in `linkml/generators/jsonschemagen.py` within the `get_subschema_for_slot` method and its interaction with `get_type_info_for_slot_subschema`.

**Root Cause:** When processing `any_of` lists, the code recursively calls `get_subschema_for_slot` on the anonymous slot expressions. Inside `get_type_info_for_slot_subschema`, there is logic that checks if a class range is inlined. If it is not inlined (which is the case for anonymous members of `any_of` as they don't inherit the parent's inlined flag), the code attempts to return the schema for the class's identifier slot instead of the class itself [jsonschemagen.py#L701-L703](https://github.com/linkml/linkml/blob/main/packages/linkml/src/linkml/generators/jsonschemagen.py#L701-L703):
```
else:
    id_slot = self.schemaview.get_identifier_slot(slot.range)
    return self.get_type_info_for_slot_subschema(id_slot)
```
This recursive call resolves the identifier slot's type. However, instead of preserving the full class structure, the resulting schema in the anyOf context often collapses to a generic `{"type": "string"}`. More importantly, for a union of classes, the semantic intent is to reference the class, not just its identifier.

**Solution:** Modify `get_subschema_for_slot` to detect when a union member's range is a class and force it to generate a `$ref` directly, bypassing the identifier reduction logic entirely. This ensures the output points to the full class definition (which correctly includes the identifier constraints) rather than a flattened primitive.

Replace the list comprehensions in the `bool_subschema` section ([jsonschemagen.py#L858-L873](https://github.com/linkml/linkml/blob/main/packages/linkml/src/linkml/generators/jsonschemagen.py#L858-L873)) with a helper function:
```
# Minimal helper: If range is a class, return $ref. Otherwise, recurse exactly as before.
        def _get_union_subschema(s):
            if s.range and s.range in self.schemaview.all_classes().keys():
                return JsonSchema.ref_for(s.range)
            # Preserve original behavior exactly:
            return self.get_subschema_for_slot(s, include_null=False)

        if slot.any_of is not None and len(slot.any_of) > 0:
            bool_subschema["anyOf"] = [_get_union_subschema(s) for s in slot.any_of]
            # Preserve original null handling exactly:
            if not slot.required and not prop.is_array and include_null:
                bool_subschema["anyOf"].append({"type": "null"})

        if slot.all_of is not None and len(slot.all_of) > 0:
            bool_subschema["allOf"] = [_get_union_subschema(s) for s in slot.all_of]

        if slot.exactly_one_of is not None and len(slot.exactly_one_of) > 0:
            bool_subschema["oneOf"] = [_get_union_subschema(s) for s in slot.exactly_one_of]

        if slot.none_of is not None and len(slot.none_of) > 0:
            bool_subschema["not"] = {
                "anyOf": [_get_union_subschema(s) for s in slot.none_of]
            }
```

CedricConday:
Fixes #3684

## What was wrong

When a slot uses `any_of` (or `all_of` / `exactly_one_of` / `none_of`) to express a union of classes, and the classes have an identifier slot, the JSON Schema generator produced `{"type": "string"}` for each union member instead of `{"$ref": "#/$defs/ClassName"}` — even when the parent slot was `inlined: true`.

As the issue notes, this only manifested when the union classes had an identifier slot; without one, the generator already produced `$ref`s.

### Root cause

The `any_of` members are anonymous slot expressions. They are rendered by recursing into `get_subschema_for_slot`, but they do **not** carry the parent slot's `inlined` setting. So `schemaview.is_inlined(member)` is `False`, and `get_type_info_for_slot_subschema` takes the non-inlined branch:

```python
else:
    id_slot = self.schemaview.get_identifier_slot(slot.range)
    return self.get_type_info_for_slot_subschema(id_slot)  # -> string
```

which reduces an identifier-bearing class to its identifier's (string) type.

## Fix

A small helper, `_subschema_for_boolean_member`, propagates the parent slot's `inlined` / `inlined_as_list` to class-ranged combinator members that don't set their own, before recursing. Identifier-bearing class members are then referenced with `$ref`.

This is deliberately narrow:
- Foreign-key (non-inlined) references to identifier classes still render as `{"type": "string"}`.
- Unions of classes **without** an identifier are unchanged (they already produced `$ref`s).

## Test

Adds `test_issue_3684_inlined_any_of_class_union_keeps_refs`, which builds the schema from the issue and asserts the `anyOf` members are `$ref`s to `Catalogue`/`Dataset` rather than string types.

Verified locally: the new test passes and the full `test_jsonschemagen.py` suite is green (48 passed, 261 subtests, 0 regressions).

---
*AI-assisted, human-reviewed — I'm an AI engineer; I find, fix, and test with AI, then review and verify before opening.*


StroemPhi:
While this also fixes the `linkml-validate` command to run without errors, which previously wasn't the case, it somehow doesn't work on a more complex schema, for which I need it. THe JSON generator, works fine now, but the `linkml-validate`still fails. See the [dcat_ap_plus.yaml](https://github.com/nfdi-de/dcat-ap-plus/blob/40-fix-the-underspecified-supportive-entites-issue/src/dcat_ap_plus/schema/dcat_ap_plus.yaml) in https://github.com/nfdi-de/dcat-ap-plus/pull/100, where the [CatalogueRecord-complete.yaml](https://github.com/nfdi-de/dcat-ap-plus/blob/40-fix-the-underspecified-supportive-entites-issue/tests/data/invalid/CatalogueRecord-complete.yaml) example had to be put into `/tests/data/invalid/` because it fails validation on the `primary_topic` slot with the cryptic error message:
```
[ERROR] [.\tests\data\valid\CatalogueRecord-001.yaml/0] {'id': 'https://example.org/dataset/env-monitoring-001'} 
is not valid under any of the given schemas in /primary_topic
```
That would either mean, that I made an error in dcat_ap_plus.yaml, which could totally be the case, or there is more to fix here. For reference, the [`primary_topic` slot definition from dcat_ap_plus.yaml](https://github.com/nfdi-de/dcat-ap-plus/blob/40-fix-the-underspecified-supportive-entites-issue/src/dcat_ap_plus/schema/dcat_ap_plus.yaml#L1074-L1086) is:
```
primary_topic:
        name: primary_topic
        description: A link to the Dataset, Data service or Catalog described in the record.
        slot_uri: foaf:primaryTopic
        range: Any
        required: true
        multivalued: false
        inlined: true
        any_of:
        - range: Catalogue
        - range: Dataset
        - range: DatasetSeries
        - range: DataService
```

CedricConday:
Thanks for testing this against a real schema — I reproduced your case with `dcat_ap_plus.yaml` and can confirm what's going on.

The union `$ref` generation from this PR is working: with `primary_topic` as `inlined: true`, the generated schema is `anyOf: [$ref Catalogue, $ref Dataset, $ref DatasetSeries, $ref DataService]`, and a **full** inlined object validates fine.

The `{'id': ...}` value fails because it's an *identifier-only reference*, and `inlined: true` means "embed the object." All four target classes have required non-id slots — Catalogue (`description`, `publisher`, `title`), Dataset (`description`, `title`, `was_generated_by`), DatasetSeries (`description`, `title`), DataService (`endpoint_URL`, `title`) — so an id-only value can't satisfy any of them. This isn't specific to the union: LinkML emits the same full-object `$ref` for any single-valued inlined class slot.

So it comes down to what you want `primary_topic` to hold:

- **A reference by identifier** → set `inlined: false` (or drop `inlined`) and give the identifier as a string, e.g. `primary_topic: https://example.org/dataset/env-monitoring-001`. That generates `anyOf: [{"type": "string"}, …]` and validates.
- **An embedded object** → keep `inlined: true` and include the required fields (`title`, `description`, …).

The `{'id': ...}` dict is the one form that validates under neither — a bare string is the reference form, a full object is the inline form.

If you'd specifically want an id-only dict to validate against inlined identifier-bearing classes, that'd be a separate enhancement (and a bit non-standard) — happy to open an issue for it if it's useful for your workflow.

StroemPhi:
> The `{'id': ...}` dict is the one form that validates under neither — a bare string is the reference form, a full object is the inline form.

You are right, that was it! I wasn't aware about that detail. Thank you @CedricConday for this really quick fix! 

@matentzn & @kevinschaper it would be really great if this PR could get merged as soon as all checks have passed, as we'll need the "any_of" range unions in other places of DCAT-AP+ as well.

CedricConday:
Glad that cleared it up, @StroemPhi — and thanks for running it against a real DCAT-AP+ schema; that's exactly the kind of test that surfaces the edge cases.

The branch is rebased on current `main` and green, so it's ready whenever @matentzn / @kevinschaper have a moment.

If having an id-only reference (the `{'id': ...}` form) validate against inlined class unions would help your DCAT-AP+ work, I'm happy to spec that as a separate enhancement issue — just say the word.

kevinschaper (approved):
Thanks @CedricConday! (and @StroemPhi for the extra check and vote of confidence)

CedricConday:
This one's approved but the branch has fallen behind main. Happy to rebase it if that helps get it in — just say the word.

kevinschaper:
Closes #3684 (thanks @StroemPhi)

CedricConday:
Brought the branch up to date with `main` — it was 34 commits behind, which is what was showing it as out-of-date.

No conflicts, and `tests/linkml/test_generators/test_jsonschemagen.py` is green: 52 passed, 2 skipped, 261 subtests passed.

Nothing else changed — the diff is the same as when you approved it. Ready to merge whenever suits.

CedricConday:
@linkml/core-team — this one has been waiting on a second review since @kevinschaper
requested one from the team on 12 August. He approved it back on 16 July, so I think that
team review is the only thing left.

Short version of the change: `jsonschemagen` was dropping `$ref`s when a class union was
inlined via `any_of`, so the generated schema silently lost the reference targets. It now
keeps them. Fixes #3684, and the five `test_issue_3684_*` regression tests cover it —
reverting `jsonschemagen.py` fails exactly those five and passes the other 157.

The branch is current with `main` again and CI is green. Happy to rebase, split it, or
adjust anything that would make it easier to review — and if someone other than core-team
should be looking at generator changes, point me at them and I will redirect.

Issue #3689, opened 2026-06-30
Adapt JSON-LD generator to use SchemaView


Silvanoc:
The JSON-LD generator is still using the deprecated SchemaLoader instead of SchemaView.

This is a meta-issue to gather individual baby-steps needed to accomplish the refactoring successfully.

Silvanoc:
This issue is closely related to #3797

PR #3694, opened 2026-06-30
fix(schemaview): resolve inlined through is_a and mixin chains

### LINKED ISSUE #3695: SchemaView.is_inlined() ignores inherited inlined=true from slot is_a chain
**Describe the bug**

`SchemaView.is_inlined()` incorrectly returns `False` for a slot that inherits `inlined: true` from a parent slot via `is_a`, when the child slot does not explicitly re-declare `inlined: true` itself.

According to the LinkML specification, slot properties are inherited through the `is_a` hierarchy. A slot should be considered inlined if any ancestor slot in its `is_a` chain declares `inlined: true`. The implementation only checks the slot's own `inlined` / `inlined_as_list` fields, ignoring the inherited value.

No error message is printed — the method silently returns the wrong value, causing downstream code (e.g. JSON schema generator, RDF loader) to treat an inlined slot as a non-inlined (by-reference) slot.

```yaml
id: https://example.org/inlined-is-a
name: inlined-is-a-example
prefixes:
  ex: https://example.org/

default_prefix: ex
default_range: string

classes:
  Address:
    attributes:
      street:
      city:

  Person:
    attributes:
      name:
      home_address:
        range: Address
        inlined: true        # parent slot declares inlined
        multivalued: false

  Employee:
    is_a: Person
    slot_usage:
      home_address:
        description: "overridden slot — inherits inlined=True from parent"
        # NOTE: inlined is NOT redeclared here
```

Minimal Python reproduction:

```python

sv = SchemaView("schema.yaml")

# Get the slot_usage-derived slot for Employee.home_address
# (or equivalently, a slot with is_a pointing to an inlined slot)
parent_slot = sv.all_slots()["home_address"]
print("parent inlined:", parent_slot.inlined)   # True

# A child slot that inherits via is_a but doesn't redeclare inlined
from linkml_runtime.linkml_model.meta import SlotDefinition
child_slot = SlotDefinition(name="home_address_child", is_a="home_address", range="Address")

print("is_inlined (expected True):", sv.is_inlined(child_slot))   # Returns False — BUG
```

Silvanoc:
## Summary

Fixes #3695

`SchemaView.is_inlined()` only checked `slot.inlined` and `slot.inlined_as_list` directly on the passed slot object, without traversing the slot's `is_a` chain. This meant a slot that inherited `inlined: true` from a parent slot via `is_a` was incorrectly reported as non-inlined.

The fix walks the slot's `is_a` chain through `all_slots()` to resolve the inherited value of `inlined` / `inlined_as_list` before making the determination, consistent with how `induced_slot()` already handles inherited slot properties. A `getattr` guard is also added so that `AnonymousSlotExpression` objects — which some callers pass despite the `SlotDefinition` type annotation — are handled safely.

This issue has appeared working on issue #3689: "Adapt JSON-LD generator to use SchemaView".

## How was this tested?

The fix is verified by the full existing test suite (`uv run pytest tests/`), which passes without regression. The bug had no existing test coverage because no test schema exercised the specific pattern of a slot inheriting `inlined` via `is_a` without redeclaring it. A dedicated test reproducing the issue has been added.

## Areas of uncertainty

- The `is_a` chain traversal uses `all_slots()`, which may be expensive for large schemas with deep slot hierarchies. An alternative would be to call `induced_slot()` directly, but that requires a class name which `is_inlined()` does not accept. Reviewers may want to assess whether the traversal cost is acceptable or whether `induced_slot()` should be used instead (requiring a signature change).
- The `getattr(slot, "is_a", None)` guard silently skips the chain traversal for `AnonymousSlotExpression` objects. Whether `is_inlined()` should instead reject non-`SlotDefinition` inputs with a `TypeError` is an open question.

## Checklist

## AI Assistance

### DIFF
--- a/packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
+++ b/packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
@@ -1948,13 +1948,33 @@ class SchemaView:
     def is_inlined(self, slot: SlotDefinition, imports: bool = True) -> bool:
         """Return true if slot is inferred or asserted inline.
 
+        A slot is inlined when:
+
+        * ``slot.inlined`` or ``slot.inlined_as_list`` is ``True`` (directly or
+          inherited through the slot's ``is_a`` chain), **or**
+        * the slot's range is a class with no identifier/key slot (which forces
+          inlining because there is no reference to use).
+
         :param slot:
         :param imports:
         :return:
         """
         slot_range = slot.range
         if slot_range in self.all_classes():
-            if slot.inlined or slot.inlined_as_list:
+            # Resolve inlined through the slot's is_a chain so that a slot that
+            # inherits inlined=True from a parent slot is correctly detected.
+            resolved_inlined = slot.inlined
+            resolved_inlined_as_list = slot.inlined_as_list
+            parent_name = getattr(slot, "is_a", None)
+            while (resolved_inlined is None and resolved_inlined_as_list is None) and parent_name:
+                parent = self.all_slots(imports=imports).get(parent_name)
+                if parent is None:
+                    break
+                resolved_inlined = parent.inlined
+                resolved_inlined_as_list = parent.inlined_as_list
+                parent_name = parent.is_a
+
+            if resolved_inlined or resolved_inlined_as_list:
                 return True
 
             id_slot = self.get_identifier_slot(slot_range, imports=imports)
--- a/packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
+++ b/packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
@@ -1951,7 +1951,7 @@ class SchemaView:
         A slot is inlined when:
 
         * ``slot.inlined`` or ``slot.inlined_as_list`` is ``True`` (directly or
-          inherited through the slot's ``is_a`` chain), **or**
+          inherited through the slot's ``is_a`` / ``mixins`` ancestry), **or**
         * the slot's range is a class with no identifier/key slot (which forces
           inlining because there is no reference to use).
 
@@ -1961,18 +1961,29 @@ class SchemaView:
         """
         slot_range = slot.range
         if slot_range in self.all_classes():
-            # Resolve inlined through the slot's is_a chain so that a slot that
-            # inherits inlined=True from a parent slot is correctly detected.
             resolved_inlined = slot.inlined
             resolved_inlined_as_list = slot.inlined_as_list
-            parent_name = getattr(slot, "is_a", None)
-            while (resolved_inlined is None and resolved_inlined_as_list is None) and parent_name:
-                parent = self.all_slots(imports=imports).get(parent_name)
-                if parent is None:
-                    break
-                resolved_inlined = parent.inlined
-                resolved_inlined_as_list = parent.inlined_as_list
-                parent_name = parent.is_a
+
+            slot_name = getattr(slot, "name", None)
+            if (resolved_inlined is None and resolved_inlined_as_list is None) and slot_name:
+                # walk all ancestors (reflexive=False to skip the slot itself,
+                # already checked above).
+                try:
+                    # slot_ancestors() is @lru_cache and handles both is_a and mixins;
+                    ancestors = self.slot_ancestors(slot_name, imports=imports, reflexive=False)
+                except ValueError:
+                    ancestors = []
+                if ancestors:
+                    # all_slots() is @lru_cache
+                    all_slots = self.all_slots(imports=imports)
+                    for anc_name in ancestors:
+                        anc = all_slots.get(anc_name)
+                        if anc is None:
+                            continue
+                        resolved_inlined = resolved_inlined or anc.inlined
+                        resolved_inlined_as_list = resolved_inlined_as_list or anc.inlined_as_list
+                        if resolved_inlined or resolved_inlined_as_list:
+                            break
 
             if resolved_inlined or resolved_inlined_as_list:
                 return True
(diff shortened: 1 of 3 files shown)


packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
```diff
@@ -1927,13 +1927,33 @@ def _get_string_type(self) -> TypeDefinition:
...
         :return:
         """
         slot_range = slot.range
         if slot_range in self.all_classes():
-            if slot.inlined or slot.inlined_as_list:
+            # Resolve inlined through the slot's is_a chain so that a slot that
+            # inherits inlined=True from a parent slot is correctly detected.
+            resolved_inlined = slot.inlined
+            resolved_inlined_as_list = slot.inlined_as_list
+            parent_name = getattr(slot, "is_a", None)
```
noelmcloughlin: Could you move `all_slots = self.all_slots(imports=imports)` here.
Its more performant and `all_slots()` is (probably) not `@lru_cache` but even if it is, there is no need to hit cache on each while loop iteration.

packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
```diff
@@ -1927,13 +1927,33 @@ def _get_string_type(self) -> TypeDefinition:
...
+            while (resolved_inlined is None and resolved_inlined_as_list is None) and parent_name:
+                parent = self.all_slots(imports=imports).get(parent_name)
```
noelmcloughlin: then `parent = all_slots.get(parent...`

noelmcloughlin (changes requested):
There appears to be more optimal fix. 

`induced_slot` will "_yield a dynamic SlotDefinition that has all properties materialized_" and already applies inheritance and overrides, so `inlined` inherited flag should  be present.

- `is_a` and `mixins` (cached):  https://github.com/linkml/linkml/blob/e94d2b8368ae15b06a0fa7323df7c18a8bcaccd2/packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py#L1061
- Leverage uses `is_a` and `mixins` and `SlotDefinition._inherited_slots` here: https://github.com/linkml/linkml/blob/e94d2b8368ae15b06a0fa7323df7c18a8bcaccd2/packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py#L1633

The while loop
- calls `self.all_slots` on every iteration = trivial fix
- walks `is_a` but not `mixins` so may fail to resolve some inlined
- Only focuses on `inlined` and `inlined_as_list` but the same bug probably exists on every other [SlotExpression](https://github.com/linkml/linkml/blob/e94d2b8368ae15b06a0fa7323df7c18a8bcaccd2/packages/linkml_runtime/src/linkml_runtime/linkml_model/meta.py#L2240) like `recommended`, `multivalued`, etc.
- duplicates pre existing (cached) traversal logic

Option A

- Reuse `slot_ancestors()` as per pseudo code example below (`inlined` only).
```
 resolved_inlined = slot.inlined
 if not (resolved inlined or resolved_inlined_as_list) and slot.is_a:
    try:
      ancestors = self.slot_ancestors(slot.is_a. imports=imports, reflexive=True)
   except:
     ancestors = [slot.is_a]               # not in schema
  all_slots = self.all_slots(imports=imports)
 for aname in ancestors:
    anc = all_slots.get(aname)
    if anc is None:
      continue
   resolved_inlined = resolved_inlined or anc.inlined

   if resolved_inlined or resolved_inlined_as_list:  
        break
 if resolved_inlined or resolved_inlined_as_list:  
        return true

return self.get_identifier_slot(slot_range, imports=imports) is None
```

Option B is generic helper that benefits everyone if you want.

Broadens your PR to cover `recommended`, `multivalued`, `inlined`, `inlined_as_list`, etc

```
def _resolve_inherited_metaslot(
    self, slot: SlotDefinition, metaslot_name: str, imports: bool = True
) -> Any:
    """Return the effective value of an inheritable metaslot, walking is_a + mixins."""
    v = getattr(slot, metaslot_name, None)
    if v is not None or not slot.is_a:
        return v
    try:
        ancestors = self.slot_ancestors(slot.is_a, imports=imports, reflexive=True)
    except ValueError:
        ancestors = [slot.is_a]
    all_s = self.all_slots(imports=imports)
    for anc_name in ancestors:
        anc = all_s.get(anc_name)
        if anc is None:
            continue
        av = getattr(anc, metaslot_name, None)
        if av is not None:
            return av
```
The implementation of `is_inlined()` after line 1943 becomes simpler 
```
  resolved_inlined = self._resolve_inherited_slot_metaslot(slot, "inlined", imports=imports)
  ..
  if resolved_inlined or resolved_inlined_as_list:
    return True

  id_slot = self.get_identifier_slot(slot_range, imports=imports)
  return id_slot is None
return False
```
Clearly more unit tests are needed but should include the mixin pattern, plus existing `is_a`
```
  child_via_mixin_no_redeclare:
    mixins:
       - inlined_parent_via_mixin       # inlined=true inherited
    range: ThingWithId
```

packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
Silvanoc: Correct, there's no need to get the same value (since `self` is not changing on each iteration) on each iteration. Will fix it.

Silvanoc:
@noelmcloughlin I've added a [new commit](https://github.com/linkml/linkml/pull/3694/changes/a4f4aec79482e26f4323b7dbc6649d8db79ef723) that should be addressing [your change request](https://github.com/linkml/linkml/pull/3694#pullrequestreview-4613042006), could you please review it again?

Silvanoc:
> Option A
> 
> * Reuse `slot_ancestors()` as per pseudo code example below (`inlined` only).
>
> [...] 
> 
> Option B is generic helper that benefits everyone if you want.
> 
> Broadens your PR to cover `recommended`, `multivalued`, `inlined`, `inlined_as_list`, etc
> 
> [...]

I've chosen Option A to keep this PR focused on the issue that it's fixing. Changing later on for Option B is always possible, but I consider it a separate concern.

noelmcloughlin (approved):
LGTM, thanks!!

PR #3696, opened 2026-06-30
feat(generators): add --no-generation-date for reproducible output

amc-corey-cox:
Adds a `--generation-date/--no-generation-date` flag (default **off**, so output is reproducible out of the box) that suppresses the load-time `generation_date` timestamp.

Addresses bullet 2 of #3516 (requested by @dalito for opt-in suppression, and @noelmcloughlin re: reproducible builds on #3518).

### Why a dedicated flag
The `generation_date` stamp is applied on every schema load and is serialized as a `linkml:generation_date` **data triple** in RDF/JSON-LD output, so it changes every run and defeats byte-stable generation. `--no-metadata` does *not* remove it — that flag only gates the `@context` comment block, not the data triple — so a dedicated control is needed.

### Approach
- Clear `schema.generation_date` at the base `Generator` level when `include_generation_date=False`, so it applies uniformly to every generator that serializes the schema.
- Forward the flag into the inner `JSONLDGenerator` that `gen-rdf` builds from `original_schema`.
- Guard the jsonld-context comment line so it doesn't emit a null when suppressed.

Scope is intentionally limited to `generation_date` (the per-run timestamp); `source_file_date`/`source_file_size` are left alone.

### Test
`test_generation_date_suppressed_by_default` asserts the triple is absent by default, present when explicitly requested, and that default output is byte-identical across runs.

---

### Addendum: default flipped to opt-in

The flag originally defaulted **on** (opt-out). @dalito and @noelmcloughlin both argued for the inverse — the date of generation is normally recoverable from version control, so paying for it with non-reproducible output is a bad trade. Flipped: the timestamp is now suppressed by default and `--generation-date` opts back in.

This makes the PR a behaviour change rather than a pure addition: generators no longer emit `generation_date` unless asked.

Two follow-ons from the flip:

- **`pythongen` and `namespacegen` gated their entire metadata banner on `generation_date` being truthy**, so suppressing the date also dropped the `# Auto generated from` and `# Schema:` lines. Only the date line is conditional now.
- **Added `test_output_is_byte_identical_across_processes`.** The original test ran twice in one interpreter, so it could only ever catch the timestamp — never `PYTHONHASHSEED`-dependent ordering, which is the other half of #3516. The new test runs three fresh interpreters with different seeds and compares bytes, and rejects empty output so a broken generator can't pass by producing nothing.

Verified against a 6-schema × 3-seed sweep of `gen-rdf` / `gen-shacl` / `gen-owl` / `gen-jsonld-context` in fresh processes: all byte-stable with the new default, and `gen-rdf` / `gen-jsonld-context` reproducibly *unstable* with `--generation-date`, confirming the timestamp was the remaining cause.


noelmcloughlin (commented):
LGTM - but one question

Should generation date not be suppressed by default, because "date-of-generation" is usually captured externally (i.e source revision control timestamps) so inserting timestamps feels unnecessary. On next iteration the generated outputs have no timestamp, but every generation afterwards is reproducible.

noelmcloughlin:
> Should generation date not be suppressed by default

See also @dalito  comment  https://github.com/linkml/linkml/issues/3516#issuecomment-4508510016

amc-corey-cox:
@noelmcloughlin @dalito — you were both right, and I've flipped it. `generation_date` is now suppressed by default; `--generation-date` opts back in (dc602a9aa).

I'd been hesitant to change the default since it's a behaviour change for every generator, but the argument that the date is already recoverable from version control settles it — it isn't worth paying for with non-reproducible output.

The flip also surfaced a latent bug worth flagging: `pythongen` and `namespacegen` gated their *entire* metadata banner on `generation_date` being truthy, so suppressing the date silently dropped the `# Auto generated from` and `# Schema:` lines too. Only the date line is conditional now.

I also strengthened the test. The original one ran twice in the same interpreter, so it could only ever catch the timestamp — never `PYTHONHASHSEED`-dependent ordering, which is the other half of #3516. `test_output_is_byte_identical_across_processes` now runs three fresh interpreters with different seeds and compares bytes.

noelmcloughlin (approved):
LGTM.

Aside, not blocking.

The only external impact is the lack of Linkml documentation?  File `utils/generator.py` provides a base class and common "micro-capabilities" for all child generators - its clearly used  by four generators updated in this PR.  But File [docs/generators/project-generator.rst](https://github.com/linkml/linkml/blob/main/docs/generators/project-generator.rst) - nearest equivalent to a "Base doc"  does not seem to provide expose what `utils/generator.py` offers to all child generators - i.e. `--no-generation-datebase`.  Am I missing something.  

Anyway, I suspect RST file format does not offer native support for "import commit-bits.rst" which could be inherited by all child generators that use the Base generator class.

packages/linkml/src/linkml/utils/generator.py
```diff
@@ -220,6 +225,12 @@ def __post_init__(self) -> None:
             if not self.schema.metamodel_version:
                 self.schema.metamodel_version = metamodel_version
 
+        # Drop the load-time generation_date stamp unless it was explicitly asked for.
+        # This is the metaslot that gets serialized as a data triple by the RDF/JSON-LD
+        # generators, so clearing it here covers every generator uniformly.
+        if not self.include_generation_date and self.schema is not None:
+            self.schema.generation_date = None
```
kevinschaper, line 242: Apologies for leaving the claude-ese as-is, but it captures the details: 

This mutates the caller's schema object rather than the generator's own copy.

On the SchemaView path, line 222 does `self.schema = self.schemaview.schema`, and `SchemaView` handed a `SchemaDefinition` keeps the reference instead of copying it. So `self.schema` and the caller's variable are the same object, and this assignment reaches back out of the generator and edits it.

Reproduced at `7d1b44c75`:

```python
sch = SchemaView("mini.yaml").schema
sch.generation_date = "2020-01-01T00:00:00"

ShaclGenerator(sch)          # just constructing it — no serialize() call

print(sch.generation_date)   # None
```

```
caller's generation_date: '2020-01-01T00:00:00' -> None
gen.schema is caller's object: True
```

Two patterns this breaks:

- **Reuse** — load once, run through several generators (normal, since parsing is the expensive part). The first generator constructed strips the field for every one after it.
- **Reading your own data** — anything inspecting `schema.generation_date` for a manifest, a log line, or a build comparison finds it gone after an unrelated generator is constructed.

No exception, no warning; the field just disappears.

Scope is narrower than it first looks: only the `uses_schemaloader = False` generators (SHACL, OWL, …) are affected. `_initialize_using_schemaloader` builds its own copy, so `PythonGenerator` comes back `False` on the identity check and is fine.

The intent is right — the output shouldn't carry the stamp. It's just applied one level too far out. Clearing it on a copy the generator owns (or restoring it once generation finishes) produces identical output and leaves the caller's schema intact. I didn't propose a specific patch because a blanket `deepcopy` here has a cost on large schemas that you're better placed to weigh than I am.

Worth a regression test either way — assert that `schema.generation_date` still holds its value after a generator is constructed from it.

kevinschaper (changes requested):
quick change request on changing a value in the schema that's passed into the generator

packages/linkml/src/linkml/utils/generator.py
amc-corey-cox: Fixed in 8fa3fb4eb. The suppression now deepcopies the schema before constructing the SchemaView when the caller hands us a SchemaDefinition (and only when there's actually a date to clear), so the generator owns what it mutates and gen.schema / schemaview.schema stay the same object internally. Your repro now leaves the caller's generation_date intact. Added the regression test you suggested in test_generator.py: constructs a ShaclGenerator from a SchemaDefinition with generation_date set, asserts the caller's value survives and the generator's own copy has it cleared. Thanks for the careful repro — the copy-on-write scoping (rather than a blanket deepcopy) means the cost only lands on the object-passing path where parsing was already skipped.

kevinschaper (approved):
🕺

PR #3703, opened 2026-07-01
fix(rdf): make pyoxigraph fallback deterministic for non-standard graphs (#3516)

amc-corey-cox:
Closes item 3 of #3516: `gen-owl` on the LinkML metamodel produces different output run-to-run.

## Root cause

The metamodel uses bare values like `status: testing` on the `status` slot (`range: uriorcurie`), which serialize to scheme-less/relative IRIs `<testing>` — 32 such triples (`bibo:status <testing>`/`<unstable>`). pyoxigraph rejects relative IRIs at parse time, so `canonicalize_rdf_graph()` fell through to its `SyntaxError` fallback, which did a plain `graph.serialize()` with **no blank-node canonicalization**. The metamodel OWL is dense with `owl:Restriction` blank nodes, which rdflib labels non-deterministically, so the fallback output varied across processes.

## Fix

Make the fallback itself deterministic instead of silently emitting unstable output:

- Canonicalize blank-node labels via `rdflib.compare.to_canonical_graph` (content-hash based, not run-local ids).
- Rebind the source graph's prefixes/base, which the canonical graph drops (otherwise rdflib's auto-generated `ns1:`/`ns2:` prefix naming is itself non-deterministic — the dominant leak).
- Sort lines for line-oriented formats (nt/nquads), where rdflib has no stable order.

Relative IRIs are preserved verbatim (not resolved against the base), so the underlying data issue stays visible rather than being silently rewritten into an absolute IRI. This changes no behaviour for standard graphs — only the pre-existing non-standard-RDF fallback path.

## Verification

- `gen-owl` on the metamodel is now byte-identical across 3 fresh processes.
- Two parametrized regression tests (subprocess + `PYTHONHASHSEED`, over turtle/nt) that fail on the pre-fix code and pass with the fix.
- Existing `test_rdf_canonicalize.py`, `test_gen_owl.py` (`--with-slow`), and the shacl/rdf/shex generator suites all pass.

## Out of scope

The metamodel data (`status: testing`) and the vendored `linkml_model/owl/meta.owl.ttl` are untouched — the source of truth is the separate `linkml/linkml-model` repo. This PR fixes the standing determinism hole in the fallback path itself.


matentzn (changes requested):
I am planting my foot here with "request changes" but only to mark that I want to review this, not because I saw anything wrong! Thanks @amc-corey-cox

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
```diff
@@ -70,6 +74,51 @@ class RDFCanonicalizationWarning(UserWarning):
...
+
+    Relative IRIs are preserved verbatim (not resolved against the base): the
+    goal is deterministic output, and silently rewriting ``<testing>`` into an
+    absolute IRI would mask what is really a data problem in the source graph.
+
+    :param graph: The rdflib Graph that pyoxigraph could not parse.
+    :param output_format: Target serialization format (e.g. ``"turtle"``, ``"nt"``).
+    :return: Deterministic string serialization of the graph.
+    """
+    canonical = to_canonical_graph(graph)
```
matentzn, line 109: just noting that the whole fix depends on the assumption that `rdflib.compare.to_canonical_graph` is deterministic :P for most of the LinkML internal stuff this will hold, so I think this is ok for now.

matentzn (approved):

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
amc-corey-cox: Yes, agreed. But if we can't count on that - well, that's a whole other issue.

Issue #3702, opened 2026-07-01
RDFC-1.0 canonicalization gives run-to-run stability but not change-locality (large diffs from small schema edits)


amc-corey-cox:
PR #3407 made RDF output (OWL/SHACL/RDF/ShEx + the rdflib dumper) reproducible across runs via pyoxigraph RDFC-1.0. That solved *run-to-run* stability. It does **not** solve *change-locality*: a small schema edit still produces a large artifact diff.

**Demonstration:** #3481 (a DEMO — DO NOT MERGE) adds a single optional string slot (`person_comment`) to the PersonInfo schema, with everything already in RDFC form. Result:

| File | Changed lines | Lines mentioning `person_comment` |
|---|---:|---:|
| `personinfo.owl.ttl` | 494 | 5 |
| `personinfo.shacl.ttl` | 632 | 2 |
| **Total** | **~1,126** | **~7** (~0.6% signal) |

The non-signal ~99% is blank-node **renumbering** (`_:c14n130 → _:c14n133`) cascading downstream of the insertion point — including in classes structurally unrelated to the change. This is inherent to RDFC-1.0: its labels are stable for a *fixed* graph but position-dependent, so any insertion shifts every label after it.

This matters for anyone version-controlling generated RDF (raised during #3407 review by @jdsika and @sneakers-the-rat; reproduced downstream in ASCS-eV/ontology-management-base#67 vs #68).

**This issue is for tracking/design only** — no solution is proposed here. RDFC-1.0 alone cannot deliver change-locality; a content-based stable blank-node labeling scheme (or skolemization) is likely needed. Related run-to-run leak-fixes (#3696, #3693, #3516) are separate and do not address this.

Issue #3531, opened 2026-05-19
Emit multivalued slots as `rdf:List` when `list_elements_ordered: true`


matentzn:
The pyoxigraph/RDFC-1.0 canonicalizer (#3407) makes the RDF dumper's output reproducible across runs, but it does **not** preserve the original order of `multivalued` slot values, even when `list_elements_ordered: true` is set on the slot. Items are emitted as flat triples and end up sorted lexicographically by the canonicalizer.

This is what @druimalban was asking about in [#1943 (comment)](https://github.com/linkml/linkml/issues/1943#issuecomment-4453426929) when he asked whether an RDF collection (linked-list semantics) is actually being produced. It is not. Issue #2069 is the same underlying problem from a year earlier.

## Reproducer

Schema:

```yaml
id: http://example.org/scratch
name: scratch
prefixes:
  ex: http://example.org/
  linkml: https://w3id.org/linkml/
default_prefix: ex
default_range: string
imports: [linkml:types]
classes:
  ColumnDesc:
    tree_root: true
    attributes:
      atom: { identifier: true, slot_uri: ex:atom }
      scope:
        multivalued: true
        list_elements_ordered: true   # <-- explicitly asks for ordering
        slot_uri: ex:scope
```

Data:

```yaml
atom: ex:sampled_data
scope: [latitude, longitude, easting, northing, depth, sampling_notes]
```

Output of `RDFLibDumper().dumps(...)`:

```turtle
ex:sampled_data ex:scope "depth" , "easting" , "latitude" , "longitude" , "northing" , "sampling_notes" ;
    a ex:ColumnDesc .
```

The values are alphabetised. No `rdf:first` / `rdf:rest`. `list_elements_ordered` has no effect on the output.

## Expected behaviour

When `list_elements_ordered: true`, emit an actual `rdf:List` so the order is part of the RDF semantics and survives any canonicalization / re-serialisation:

```turtle
ex:sampled_data ex:scope ( "latitude" "longitude" "easting" "northing" "depth" "sampling_notes" ) ;
    a ex:ColumnDesc .
```

When `list_elements_ordered` is unset/false, the current flat-triples-plus-canonical-sort behaviour is fine.

## Suggested approach

[...]

PR #3704, opened 2026-07-01
feat(rdf): opt-in content-hash blank-node labels for diff-stable RDF

amc-corey-cox:
Addresses #3702 (change-locality of canonicalized RDF).

RDFC-1.0 (#3407) made RDF output reproducible run-to-run, but its `_:c14n0, c14n1, …` labels are **ordinal positions in a global sort**, so a one-line schema edit renumbers every downstream blank node — a ~500–1100 line artifact diff that is ~85% pure renumbering ([#3481](https://github.com/linkml/linkml/pull/3481)).

The stable per-node hashes needed to fix this already exist *inside* RDFC-1.0, but pyoxigraph doesn't expose them and can't be monkey-patched (compiled PyO3). We filed [oxigraph#1788](https://github.com/oxigraph/oxigraph/issues/1788) asking for them upstream; until that lands, this implements the minimal thing on top of pyoxigraph's canonical output.

## What this does

Adds an **opt-in** `stable_blank_node_labels` (default off) that relabels blank nodes by a Merkle hash of their subtree content instead of the ordinal `c14n` label, so an unrelated edit leaves other labels byte-identical.

- Runtime: `_relabel_blank_nodes_by_hash` in `rdf_canonicalize.py`, plus a `stable_blank_node_labels` kwarg on `canonicalize_rdf_graph` and `RDFLibDumper.dumps`.
- Generators: `--stable-blank-node-labels/--no-stable-blank-node-labels` on `gen-owl` and `gen-shacl`.

## Measured effect

Demo pair (add one optional slot), `PYTHONHASHSEED=0`, reproducible across processes:

| Artifact | Churn (default `c14n`) | Churn (hash labels) |
|---|---:|---:|
| OWL | 539 | **19** |
| SHACL | 822 | **88** |

## Correctness

- Output verified **isomorphic** to the default (no triples dropped or merged) — only blank-node labels change.
- **Byte-reproducible** across processes.
- Blank nodes whose closure contains a **cycle** keep their `c14n` label (no infinite recursion); genuinely automorphic siblings stay distinct.

Tests cover all of the above plus `gen-shacl` and dumper integration.

## Scope / follow-ups

Default off, so nothing changes for existing users. Residual SHACL churn is understood — `rdf:List` spine cascade (ties to #3531/#3693) and automorphic-collision suffixes — and is exactly where RDFC-1.0's full n-degree (incoming-context) hashing matters, i.e. what oxigraph#1788 would provide. Wiring the remaining `gen-rdf`/`gen-shex`/`rdf_dumper` call sites is left as a follow-up.


matentzn (changes requested):
Same as before, planting my foot here - will review this carefully. I am glad its here.

tests/linkml_runtime/test_utils/test_rdf_canonicalize.py
```diff
@@ -202,3 +202,124 @@ def test_fallback_on_invalid_rdf():
...
+            g.add((bn, URIRef("http://example.com/q"), Literal(f"val_{letter}")))
+        print(canonicalize_rdf_graph(g, output_format="turtle", stable_blank_node_labels=True), end="")
+        """
+    )
+
+    def run(seed: str) -> str:
+        env = {"PYTHONHASHSEED": seed, "PATH": "/usr/bin:/bin"}
+        return subprocess.run(
+            [sys.executable, "-c", program], check=True, capture_output=True, text=True, env=env
+        ).stdout
```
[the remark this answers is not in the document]
amc-corey-cox: Fixed. Both subprocess determinism tests now inherit the full environment and override only PYTHONHASHSEED ({**os.environ, ...}) instead of building a minimal env dict. Applied to this test and its sibling test_sort_is_load_bearing for consistency.

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
```diff
@@ -284,9 +285,86 @@ def _is_safe_prefix_iri(iri: str) -> bool:
...
+    """
+    out_edges: dict[str, list[tuple[str, ox.Term]]] = {}
+    for t in triples:
+        if isinstance(t.subject, ox.BlankNode):
+            out_edges.setdefault(t.subject.value, []).append((str(t.predicate), t.object))
+
+    # Merkle hash per blank node; None marks a node that reaches a cycle.
+    memo: dict[str, str | None] = {}
+
+    def node_hash(bid: str, stack: frozenset[str]) -> str | None:
```
matentzn, line 318: ⚠️ AI feedback, but I checked the output of the run and from my cursory look I agree with it. When you address this, please run your own test to confirm the cycle issue.

`node_hash` (`rdf_canonicalize.py:318`) is recursive, one frame per `rdf:first`/`rdf:rest` cell, so depth == list length. I re-ran `_relabel_blank_nodes_by_hash` on a synthetic `sh:in`-style list: N=100 fine, N=1500 and N=5000 both `RecursionError`. A ~1500+ value enum (routine for ontology-derived enums) through `gen-shacl --stable-blank-node-labels` crashes instead of degrading — the exact large-artifact case this feature exists to help. It's opt-in and default-off so nobody's broken today, but this undercuts the headline use case and needs handling before merge. Fix mirrors your cycle handling: make the traversal iterative (post-order over the blank-node DAG), or catch the depth blowout and fall back to the `c14n` label for that node.

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
```diff
@@ -284,9 +285,86 @@ def _is_safe_prefix_iri(iri: str) -> bool:
...
+    # Group by hash to disambiguate genuine automorphic collisions.
+    by_hash: dict[str, list[str]] = {}
+    for bid, digest in memo.items():
+        if digest is not None:
+            by_hash.setdefault(digest, []).append(bid)
+
+    label_map: dict[str, str] = {}
+    for digest, bids in by_hash.items():
+        if len(bids) == 1:
+            label_map[bids[0]] = "b" + digest[:24]
```
matentzn, line 351: `by_hash` groups on the full sha256, but the emitted label is `b + digest[:24]`. Two nodes with different full digests sharing a 24-hex prefix each land in their own singleton group and both get the bare `b<prefix>` label — silently merging two distinct nodes. 96 bits, so astronomically unlikely, but the uniqueness guarantee should sit on the string you actually emit: group/uniquify on `digest[:24]`, not the full digest. Make airtight.

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
```diff
@@ -284,9 +285,86 @@ def _is_safe_prefix_iri(iri: str) -> bool:
...
+        else:
+            for k, bid in enumerate(sorted(bids)):
```
matentzn, line 353: if the behaviour of `sorted` changes the method will change, but ok. Not a big deal.

amc-corey-cox:
Revived: merged current main and re-verified (239 tests passing locally; change-locality re-confirmed on a one-slot demo edit: OWL churn 64→14 lines, SHACL 39→17).

One deliberate change since the original push: the label contract now matches what will land upstream. oxigraph#1824 (merged to oxigraph main, not yet released) labels blank nodes with the **bare hash** and disambiguates automorphic collisions with a bare counter suffix (`hash`, `hash1`, …). This PR previously used `b<hash>` with `_k` suffixes; it now emits the bare-hash shape, so when a pyoxigraph release ships `UnstableHashedIds` we can swap the Python post-pass for the real n-degree hashes and only the hash values change — the labeling contract stays put.

PR #3709, opened 2026-07-02
tests: diverting from model `main` only raises warning

### LINKED ISSUE #3708: Expecting model `HEAD` of `main` blocks PRs when `linkml_model` contains unreleased changes
**Describe the bug**

The test https://github.com/linkml/linkml/blob/e94d2b8368ae15b06a0fa7323df7c18a8bcaccd2/tests/linkml_runtime/test_linkml_model/test_linkml_files.py#L152 fails if https://github.com/linkml/linkml_model contains un-released changes.

But `linkml` should neither distribute an unreleased version of `linkml_model` nor test against it.

See [failing test](https://github.com/linkml/linkml/actions/runs/28574708293/job/84727821899?pr=2602).

Silvanoc:
## Summary

As of now tests fail if `linkml_model` `main` has unreleased changes. This PR only make tests fail if vendored copy of the model is not the latest release and also creates a PR comment if `linkml_model` `main` has unreleased changes.

The last commit of this PR also reverts the vendored `linkml_model` from the [current `main`](https://github.com/linkml/linkml-model/tree/main) (pulled by commit https://github.com/linkml/linkml/commit/00bb66b0bb899299eebee43f2b149a2024c178b1) to the [latest (pre)release (v1.11.0)](https://github.com/linkml/linkml-model/tree/v1.11.0), which is the content of commit https://github.com/linkml/linkml-model/commit/d028ff3a7e369a892da2bdfc69514766624a8d78. 

Fixes #3708

## How was this tested?

Test of this PR itself.

## Areas of uncertainty

Need to force pulling unreleased `linkml_model` changes is still unclear and must be confirmed in #3708

Solved:
**On pinning to prereleases:** `update_model` deliberately resolves the newest release *including* prereleases. Release-only pinning would stall linkml behind linkml-model's release cadence — we can't develop against model functionality we can't vendor, and we can't vendor it without a tag to pin to. Allowing rcs keeps the pin a named, published artifact rather than an arbitrary commit, while also keeping release-prep PRs unblocked, since those legitimately vendor rc content ahead of a stable release.

## Checklist

## AI Assistance


matentzn:
Thanks @Silvanoc, I support this, and sorry again for causing the mess this is cleaning up. Three things before merge:

1. `releases/latest` excludes prereleases, and the linkml-model rcs are marked as prereleases. The vendored files are historically bumped to rc content during release prep (5e33f583 vendored 1.11.0-rc2 while the latest stable release was still 1.10.0), so the hard test would block exactly those PRs, which led me to b6fdecac in the first place. Comparing against the newest release *including* prereleases (first entry of `/repos/linkml/linkml-model/releases`) fixes that. An alternative I'd also be happy with: record the linkml-model commit SHA at vendoring time (e.g. a `VENDORED_FROM` file next to the copies) and compare against raw URLs at that SHA, which is fully deterministic and independent of release semantics.
2. Unauthenticated `api.github.com` calls are limited to 60/h per IP, and shared Actions runner IPs will hit that, so `_get_latest_github_release_tag` will 403 flakily. Cache it (`functools.lru_cache`) and send `Authorization` from `GITHUB_TOKEN` when the env var is present.
3. The drift job posts a new PR comment on every push while drift exists, which gets noisy on active PRs. find-and-update an existing comment instead?

Silvanoc:
> 1. `releases/latest` excludes prereleases, and the linkml-model rcs are marked as prereleases. The vendored files are historically bumped to rc content during release prep ([5e33f58](https://github.com/linkml/linkml/commit/5e33f5832ba6d7b5ef2045acddcc125b64b1ceb8) vendored 1.11.0-rc2 while the latest stable release was still 1.10.0), so the hard test would block exactly those PRs, which led me to [b6fdeca](https://github.com/linkml/linkml/commit/b6fdecacdf3716a148a69df32c54e77a3de2f4c1) in the first place. Comparing against the newest release _including_ prereleases (first entry of `/repos/linkml/linkml-model/releases`) fixes that. An alternative I'd also be happy with: record the linkml-model commit SHA at vendoring time (e.g. a `VENDORED_FROM` file next to the copies) and compare against raw URLs at that SHA, which is fully deterministic and independent of release semantics.

I see.

1. Including prereleases/release-candidates shouldn't be a big deal.
2. Recording the SHA at vendoring time shouldn't be a big deal too, the [`Makefile`](https://github.com/linkml/linkml/blob/00bb66b0bb899299eebee43f2b149a2024c178b1/packages/linkml_runtime/Makefile) could easily take care of it. It'd also fix the point 2 in your comment, since no API call would be needed and `raw.githubusercontent.com` can be used instead (killing 2 birds with 1 stone).

I'd vote to go for 2, @matentzn which one do you vote for?

Silvanoc:
> 2\. Unauthenticated `api.github.com` calls are limited to 60/h per IP, and shared Actions runner IPs will hit that, so `_get_latest_github_release_tag` will 403 flakily. Cache it (`functools.lru_cache`) and send `Authorization` from `GITHUB_TOKEN` when the env var is present.

If we go for SHA pinning when vendoring (option 2 in [this comment](https://github.com/linkml/linkml/pull/3709#issuecomment-4866758228)), this issue gets resolved automatically.

Otherwise setting `GITHUB_TOKEN` on an envvar (like [here](https://github.com/linkml/linkml/pull/3709/changes#diff-71cabc4177e41ea8f15a89eb65398fa8187739f68a2762706d84885cd8b2ddc3R272)) is very easy and no caching is needed. How does `lru_cache` ensure that the cache is not outdated without making the calls anyway? Then caching doesn't make any sense.

Silvanoc:
> 3\. The drift job posts a new PR comment on every push while drift exists, which gets noisy on active PRs. find-and-update an existing comment instead?

Good point, will have a look at it.

matentzn:
> I'd vote to go for 2, @matentzn which one do you vote for?

I am fine with 2!

Silvanoc:
@matentzn this PR is now IMO ready for review. You can see that I've addressed the [PR comment topic](https://github.com/linkml/linkml/pull/3709#issuecomment-4866934846) looking at the multiple versions of the now [single warning comment](https://github.com/linkml/linkml/pull/3709#issuecomment-4868038258).

Silvanoc:
Now thinking on the daily developer workflow, I'm wondering if it really makes sense to have a warning on EVERY SINGLE PR that gets a pull whenever `linkml_model` has ongoing changes and is therefore out-of-sync with `linkml_runtime`. **That's too much noise!**

Are (pre)releases really being created manually? I haven't found any CI workflow for this purpose. If they are being created manually, then there is no chance to block it. The warning can be raised on the corresponding CI workflow and even block `.github/workflows/pypi-publish.yaml`, but it won't avoid the creation of the GitHub (pre)release including tag.

I've assigned this PR to me to avoid that @matentzn merge it before we have clarified this point.

dalito:
Pre-releases are only different from releases by the [tag format](https://packaging.python.org/en/latest/specifications/version-specifiers/#pre-releases). So there is no separate pre-release pipeline. The process of tagging is manual.

Silvanoc:
> Pre-releases are only different from releases by the [tag format](https://packaging.python.org/en/latest/specifications/version-specifiers/#pre-releases).

I'm aware of that. Therefore I'm trying to consequently write *(pre)release*.

> So there is no separate pre-release pipeline. The process of tagging is manual.

That's IMO not the best idea, since it's very error-prone. I would at least go for a script (which could take over the warning WRT `linkml_model` actuality).

matentzn:
@Silvanoc you let me know when you think its ready for re-review!

Silvanoc:
> @Silvanoc you let me know when you think its ready for re-review!

@matentzn I consider it ready for review

.github/workflows/main.yaml
```diff
@@ -231,3 +231,69 @@ jobs:
...
+
+          # Look for an existing comment that contains our marker
+          COMMENT_ID=$(gh api \
+            "/repos/${{ github.repository }}/issues/${PR_NUMBER}/comments" \
+            --paginate \
+            --jq ".[] | select(.body | contains(\"${MARKER}\")) | .id" \
+            | head -1)
+
+          if [ -n "$COMMENT_ID" ]; then
+            gh api --method PATCH \
```
matentzn: Are you sure this will work from forks, given that `GITHUB_TOKEN` is read-only? 

Maybe guard the comment step with `if: ... && github.event.pull_request.head.repo.full_name == github.repository`, or add `continue-on-error: true` or some such?

.github/workflows/main.yaml
```diff
@@ -231,3 +231,69 @@ jobs:
...
+        run: uv sync --all-groups
+
+      - name: Run upstream-main vendored-file checks
+        id: upstream_main_tests
+        run: |
+          uv run pytest tests/linkml_runtime/test_linkml_model/test_linkml_files.py \
+            --with-network --with-upstream-main -v
+        continue-on-error: true
+
+      - name: Warn via PR comment on drift (create or update)
```
matentzn: A downside is here that stale comments never clear. find-and-update only runs on failure, so once drift resolves the warning sticks on every PR that got it, still claiming drift. Add a success branch that PATCHes the comment to "resolved" or deletes it?

matentzn (commented):
Thanks a ton!

The design question I'd like to settle (you raised it yourself in https://github.com/linkml/linkml/pull/3709#issuecomment-4873937625, and I think you were right): drift against upstream `main` is a repo-level fact, not a property of any individual PR, and since main sits ahead of the last release almost all the time in steady state, this warns on effectively every open PR until the next release. @dalito confirmed tagging is manual so we can't gate on that. 

At the moment, but I might not have thought this through as well as you or @dalito, my preference would be to drop the per-PR comment and run `--with-upstream-main` on a schedule (we already have `metamodel-compat.yaml` doing something adjacent) against a single tracking issue that opens on drift and closes when resolved.

.github/workflows/main.yaml
Silvanoc: Good point, fixed on last commit.

.github/workflows/main.yaml
Silvanoc: Good point, fixed on last commit

Silvanoc:
> Thanks a ton!
> 
> The design question I'd like to settle (you raised it yourself in [#3709 (comment)](https://github.com/linkml/linkml/pull/3709#issuecomment-4873937625), and I think you were right): drift against upstream `main` is a repo-level fact, not a property of any individual PR, and since main sits ahead of the last release almost all the time in steady state, this warns on effectively every open PR until the next release. @dalito confirmed tagging is manual so we can't gate on that.
> 
> At the moment, but I might not have thought this through as well as you or @dalito, my preference would be to drop the per-PR comment and run `--with-upstream-main` on a schedule (we already have `metamodel-compat.yaml` doing something adjacent) against a single tracking issue that opens on drift and closes when resolved.

@matentzn I've modified it to run with the `metamodel-compat` tests, since they IMO belong together, and to act equivalently: open an issue, if needed, and update it, whenever something changes.

.github/workflows/metamodel-compat.yaml
```diff
@@ -113,3 +116,66 @@ jobs:
...
+          EXISTING_ISSUE=$(gh issue list --label "vendored-model-drift" --state open --json number --jq '.[0].number // empty')
+          echo "existing_issue=${EXISTING_ISSUE}" >> $GITHUB_OUTPUT
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Create or update issue on drift
+        if: always() && steps.drift_tests.outcome == 'failure'
+        run: |
+          ISSUE_TITLE="Vendored linkml-model files have drifted from upstream main"
+          ISSUE_BODY=$(cat <<'EOF'
```
matentzn: The heredoc is quoted - are you sure the test output makes it into the issue? 

`ISSUE_BODY=$(cat <<'EOF' ... EOF)` with the delimiter in single quotes disables afaik command substitution, so `$(tail -n 50 temp/drift_output.txt)` lands in the issue body *verbatim* as the literal string. 

Maybe I am wrong - its been a while since I wrote these :) Sorry.

.github/workflows/metamodel-compat.yaml
```diff
@@ -113,3 +116,66 @@ jobs:
...
+          EOF
+          )
+
+          if [ -n "${{ steps.check_drift_issue.outputs.existing_issue }}" ]; then
+            gh issue comment "${{ steps.check_drift_issue.outputs.existing_issue }}" --body "${ISSUE_BODY}"
+          else
+            gh issue create \
+              --title "${ISSUE_TITLE}" \
+              --body "${ISSUE_BODY}" \
+              --label "vendored-model-drift"
```
matentzn, line 191: Just to avoid this failing I created the label now: https://github.com/linkml/linkml/issues?q=sort%3Aupdated-desc+state%3Aopen+label%3Avendored-model-drift+

.github/workflows/metamodel-compat.yaml
```diff
@@ -113,3 +116,66 @@ jobs:
...
           delete-branch: true
           labels: metamodel-compat
+
+      - name: Check vendored files for drift against upstream main
+        id: drift_tests
+        if: always()
+        run: |
+          mkdir -p temp
+          uv run pytest tests/linkml_runtime/test_linkml_model/test_linkml_files.py \
+            --with-network --with-upstream-main -v 2>&1 | tee temp/drift_output.txt
```
matentzn, line 141: If the hard test ever fails on `main` or a transient `git fetch`/network blip occurs, you'll open a "drifted from upstream main" issue that's mislabeled or spurious?

tests/linkml_runtime/test_linkml_model/test_linkml_files.py
```diff
@@ -130,15 +132,57 @@ def test_fixed_meta_url():
...
+    assert UPSTREAM_SHA_FILE.exists(), (
+        f"{UPSTREAM_SHA_FILE} not found. Run 'make update_model' in packages/linkml_runtime/ to regenerate it."
+    )
+    assert len(open(UPSTREAM_SHA_FILE).read()) > 0, (
+        f"{UPSTREAM_SHA_FILE} is empty. Run 'make update_model' in packages/linkml_runtime/ to regenerate it."
+    )
+    return UPSTREAM_SHA_FILE.read_text().strip()
+
+
+def _git_show_file(sha: str, repo_path: str) -> str:
```
matentzn: `_git_show_file` seems to re-init a bare repo and fetches the commit tree once per parametrized file. Fine for 6 files.. but.. Maybe only fetch once is better?

matentzn (commented):
Nice @Silvanoc I think this is the right shape. 

Another round of smaller comments now.. You tell me where I am off base.

.github/workflows/metamodel-compat.yaml
matentzn: Not addressed yet.

.github/workflows/metamodel-compat.yaml
matentzn: Still an issue!

matentzn (dismissed):
@Silvanoc there are still a few unresolved conversations / feedback!

Thanks for your patience.

Silvanoc:
> @Silvanoc there are still a few unresolved conversations / feedback!
> 
> Thanks for your patience.

@matentzn I'm aware of it, but I'm trying to get progress on other areas. I've therefore marked it as `draft` so that it doesn't appear in the list of PRs pending merging.

.github/workflows/metamodel-compat.yaml
amc-corey-cox: Confirmed and fixed in 53d960aea — you were right, the quoted `<<'EOF'` disabled the substitution, so the body would have been filed with a literal `$(tail -n 50 temp/drift_output.txt)`.

Both bodies are now assembled in a file and passed with `--body-file`, which also keeps untrusted test output away from the shell (log lines containing backticks or `$(...)` are inserted verbatim rather than evaluated). Verified by extracting the actual run block and executing it against sample output containing `\`a\``, `$HOME` and `$(whoami)` — all pass through literally.

The same bug was already on main in the metamodel-compat issue body (`$(tail -n 100 temp/test_output.txt)`), so I fixed that one too rather than leave one of the pair broken.

.github/workflows/metamodel-compat.yaml
amc-corey-cox: Thanks — confirmed the `vendored-model-drift` label exists, so no change needed here.

.github/workflows/metamodel-compat.yaml
amc-corey-cox: Good catch — fixed in 53d960aea. The step ran the whole test file and treated any non-zero exit as drift, so a network blip, a collection error, or a hard `UPSTREAM_SHA` mismatch would all have filed an issue claiming we'd diverged from upstream main.

Two changes: the step now runs only the soft test (`-k test_vendored_files_match_upstream_main`), so a hard-check failure can't be misreported as drift; and issue creation keys off pytest exit code 1 (tests failed) via a `drifted` output rather than `outcome == 'failure'`. Exit 2/3/4 — interrupted, internal error, usage error — now fail the job loudly instead of filing something misleading. The close-on-resolved step keys off the same output.

tests/linkml_runtime/test_linkml_model/test_linkml_files.py
amc-corey-cox: Fixed in 53d960aea. It's actually 10 parametrised files rather than 6 (every `Source` × YAML/JSONLD), so that was 10 `git init --bare` + `git fetch` round trips per run.

Replaced with a session-scoped `upstream_file_reader` fixture that fetches once and does `git show` per lookup. Used a fixture rather than `lru_cache` so the `TemporaryDirectory` stays owned for the session instead of relying on cache lifetime. The 10 hard-check files now run in ~0.9s.

amc-corey-cox:
@Silvanoc - this has started to become somewhat load-bearing for other work. I took the liberty of resolving the threads by @matentzn
They were primarily mechanical in nature. I currently have this pinned to a commit but I'm releasing a linkml-model -rc soon and I'll remove the pin as soon as I do. I believe this will be ready to merge as soon as we have the linkml-model -rc released.

packages/linkml_runtime/Makefile
```diff
@@ -1,26 +1,43 @@
...
 update_model:
+	# Resolve the latest release or pre-release tag from the GitHub Releases API
+	$(eval LINKML_MODEL_TAG := $(shell \
+		curl -fsSL "https://api.github.com/repos/linkml/linkml-model/releases?per_page=1" \
+		| python3 -c "import sys, json; releases = json.load(sys.stdin); print(releases[0]['tag_name'])"))
+
+	# Resolve the tag to its commit SHA via the Git refs API
+	$(eval LINKML_MODEL_SHA := $(shell \
+		curl -fsSL "https://api.github.com/repos/linkml/linkml-model/git/refs/tags/$(LINKML_MODEL_TAG)" \
+		| python3 -c "import sys, json; d = json.load(sys.stdin); print(d['object']['sha'])"))
```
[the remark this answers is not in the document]
amc-corey-cox: Good catch — fixed in 58ebc0a29.

Switched to the commits endpoint, which dereferences the ref for us:

```
curl -fsSL ".../repos/linkml/linkml-model/commits/$(LINKML_MODEL_TAG)" \
  | python3 -c "import sys, json; print(json.load(sys.stdin)['sha'])"
```

This is correct for annotated and lightweight tags alike. It only worked before because GitHub's UI creates lightweight tags, so `object.sha` happened to be the commit — an annotated tag would have written a tag-object SHA that `git fetch <sha>` can't resolve. It's also one API call instead of two, which helps the rate-limiting concern raised earlier in this PR.

Verified against v1.11.1-rc1: resolves to `7657de64`, matching the committed `UPSTREAM_SHA`, so no re-vendoring was needed.

tests/linkml_runtime/test_linkml_model/test_linkml_files.py
```diff
@@ -130,15 +133,64 @@ def test_fixed_meta_url():
...
+
+    Returns the 40-character hex SHA written to ``UPSTREAM_SHA`` at vendoring time.
+    """
+    assert UPSTREAM_SHA_FILE.exists(), (
+        f"{UPSTREAM_SHA_FILE} not found. Run 'make update_model' in packages/linkml_runtime/ to regenerate it."
+    )
+    assert len(open(UPSTREAM_SHA_FILE).read()) > 0, (
+        f"{UPSTREAM_SHA_FILE} is empty. Run 'make update_model' in packages/linkml_runtime/ to regenerate it."
+    )
+    return UPSTREAM_SHA_FILE.read_text().strip()
```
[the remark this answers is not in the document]
amc-corey-cox: Fixed in 58ebc0a29 — reads once, validates `re.fullmatch(r"[0-9a-f]{40}", sha)`, no leaked handle.

The leak was real rather than theoretical: `ResourceWarning: unclosed file ... UPSTREAM_SHA` was showing up in our own test output. Confirmed it's gone, and the new validation produces a usable message:

```
AssertionError: .../UPSTREAM_SHA does not contain a 40-character hex commit SHA
(got 'not-a-sha'). Run 'make update_model' in packages/linkml_runtime/ to regenerate it.
```

One correction: the line 184 reference looks like stale numbering — there's no second `open()` in the file. The other reference is a `_get_upstream_sha()` call site, which is fine and picks up the fix transitively.

amc-corey-cox (approved):
This appears to be ready to go now.

Silvanoc:
@matentzn @amc-corey-cox thanks for taking over here! I am currently having some time off and therefore not very responsive.

Issue #2155, opened 2024-06-11
Inconsistent nullability in generated Pydantic model


candleindark:
**Describe the bug**
The nullability of a field in the generated Pydantic model corresponding to a LinkML class is inconsistent with the nullability of the the corresponding slot in the class.

**To reproduce**
1. Given the following LinkML schema
    `personinfo.yaml`
    ```yaml
    id: https://w3id.org/linkml/examples/personinfo
    name: personinfo
    prefixes:
      linkml: https://w3id.org/linkml/
      personinfo: https://w3id.org/linkml/examples/personinfo
    imports:
      - linkml:types
    default_range: string
    default_prefix: personinfo
    
    classes:
      Person:
        attributes:
          name:
    ```
2. and the following data instance
    `data.yaml`
    ```yaml
    name: null
    ```
3. Produces the following outcome when running `linkml-validate`
    ```sh
    ❯ linkml-validate -s personinfo.yaml data.yaml
    [ERROR] [data.yaml/0] None is not of type 'string' in /name
    ```
    The `name` slot is not nullable.
4. However, running `gen-pydantic` to generate a Pydantic model to represent `Person`.
    ```sh
    gen-pydantic --pydantic-version 2 personinfo.yaml > model.py
    ```
    We have the following outcome.
    `model.py`
    ```python
    from __future__ import annotations 
    from datetime import (
        datetime,
        date
    )
    from decimal import Decimal 
    from enum import Enum 
    import re
    import sys
    from typing import (
        Any,
        List,
        Literal,
        Dict,
        Optional,
        Union
    )
    from pydantic.version import VERSION  as PYDANTIC_VERSION 
    if int(PYDANTIC_VERSION[0])>=2:
        from pydantic import (
            BaseModel,
            ConfigDict,
            Field,
            field_validator
        )
    else:
        from pydantic import (
            BaseModel,
            Field,
            validator
        )
    
    metamodel_version = "None"
    version = "None"
    
    
    class ConfiguredBaseModel(BaseModel):
        model_config = ConfigDict(
            validate_assignment = True,
            validate_default = True,
            extra = "forbid",
            arbitrary_types_allowed = True,
            use_enum_values = True,
            strict = False,
        )
        pass
    
    
    class Person(ConfiguredBaseModel):
        name: Optional[str] = Field(None)
    
    
    # Model rebuild
    # see https://pydantic-docs.helpmanual.io/usage/models/#rebuilding-a-model
    Person.model_rebuild()
    ```
[...]

PR #3737, opened 2026-07-09
Mark non-required enum slots as nullable in JSON Schema output

### LINKED ISSUE #3736: Non-required enum slots emit non-nullable values in JSON Schema generator output
**Describe the bug**

JSON Schema output does not mark non-required enum fields as nullable, unlike other kinds of values like scalars or nested classes. This is related to #2740 but from the opposite direction. If a field is not marked as required, then it should consistently be marked as nullable.

1.11.1

Given the schema:

```yaml
id: https://example.org/nullable-enum-bug
name: nullable_enum_bug
prefixes:
default_range: string
imports:
  - linkml:types

enums:
  StatusEnum:
    permissible_values:
      active:
      inactive:

classes:
  Person:
    attributes:
      name:
        range: string
  
  Position:
    attributes:
      title:
        range: string
        required: false
      person:
        range: Person
        inlined: true
        required: false
      status:
        range: StatusEnum
        required: false
```

This JSON Schema is generated (quoting just a portion):

```json
"Position": {
    "description": "",
    "properties": {
        "person": {
            "anyOf": [
                {
                    "$ref": "#/$defs/Person"
                },
                {
                    "type": "null"
                }
            ]
        },
        "status": {
            "$ref": "#/$defs/StatusEnum"
        },
        "title": {
            "type": [
                "string",
                "null"
            ]
        }
    },
    "title": "Position",
    "type": "object"
},
```

* The optional string is typed `{ "type": ["string", "null"] }`
* The optional inlined class is typed `{ "anyOf": [ ...PersonDef..., { "type": "null" } ]}`
* The optional enum is emitted as a bare ref to `StatusEnum` def, rather than `{ "anyOf": [ ...StatusEnumDef..., { "type": "null" }]}`

kevinschaper:
Fixes #3736

## Problem

Optional slots with enum ranges were emitted as a bare `$ref` in JSON Schema output, so `{"status": null}` failed validation — inconsistent with optional type-ranged slots (`"type": ["string", "null"]`) and optional class-ranged slots (`"anyOf": [{"$ref": ...}, {"type": "null"}]`).

## Why null is correct here (spec basis)

- The instance specification states an assignment of a slot to `None` is equivalent to omitting the assignment, and the JSON mapping defines JSON `null` as the serialization of `None`.
- In the validation rules, the only check that fires on `None` is `Required`; the enum `Permissible` check only applies to actual permissible-value instances, so a range never constrains a null.
- This matches the position in linkml/linkml#2155 and https://github.com/orgs/linkml/discussions/1975 ("there is no semantic distinction between a null value and a missing value"), under which type and class ranges were already made nullable.

Users who want strict presence-only JSON Schema semantics retain the existing `include_null=False` escape hatch; the fix passes `required=slot.required or not include_null`, identical to the class-range path.

## Fix

In `get_subschema_for_slot`, the non-inlined reference branch (only reachable for enum ranges, since non-inlined class ranges resolve to their identifier's type) now forwards the slot's requiredness to `JsonSchema.ref_for`, producing `anyOf: [{"$ref": ...}, {"type": "null"}]` for optional single-valued enum slots. For multivalued slots, nullability applies to the array itself (via `array_of`), not the elements, matching type- and class-range behavior — pinned by an `items` assertion in the unit test. Metamodel and biolink JSON Schema snapshots regenerated accordingly.

## Tests

- **Unit**: added `enum_range` and `enum_range_multivalued` to `not_required.yaml` (the "all the ways things can be not required" input for `test_slot_not_required_nullability`). Also closed a silent hole in that test: a bare `$ref` property matched neither the `type` nor `anyOf` assertion branch and passed unchecked; it now fails explicitly.
- **Compliance**: new `test_optional_enum_slot_nullability` in `test_enum_compliance.py` covering present / absent / explicit-null / invalid values across all core frameworks. Getting this red required a harness fix: `check_data` stripped null-valued keys (`clean_null_terms`) before jsonschema/SHACL validation, so explicit-null cases silently degenerated into absent-key cases. A new `strip_nulls: bool = True` parameter preserves existing behavior for all other tests.
- Pre-fix, the only failure in the 13-framework × 4-case matrix was jsonschema × explicit-null; pydantic, python dataclasses, SHACL, and sqlite all already accept the null. Post-fix, everything is green: full jsonschemagen unit suite (67 passed + 261 subtests), jsonschemagen-marked compliance tests (771), full enum compliance file (754), and validator suites (140).

https://claude.ai/code/session_01Qnu2VdqDqDqAmWq3ovekqw

### DIFF
--- a/packages/linkml/src/linkml/generators/jsonschemagen.py
+++ b/packages/linkml/src/linkml/generators/jsonschemagen.py
@@ -836,7 +836,11 @@ class JsonSchemaGenerator(Generator, LifecycleMixin):
 
             else:
                 if reference is not None:
-                    prop = JsonSchema.ref_for(reference, required=slot.required or not include_null)
+                    # for multivalued slots, nullability applies to the array (via array_of
+                    # below), not to the individual elements
+                    prop = JsonSchema.ref_for(
+                        reference, required=slot.required or slot_is_multivalued or not include_null
+                    )
                 elif typ and fmt is None:
                     prop = JsonSchema({"type": typ})
                 elif typ:
--- a/packages/linkml/src/linkml/generators/jsonschemagen.py
+++ b/packages/linkml/src/linkml/generators/jsonschemagen.py
@@ -836,7 +836,7 @@ class JsonSchemaGenerator(Generator, LifecycleMixin):
 
             else:
                 if reference is not None:
-                    prop = JsonSchema.ref_for(reference)
+                    prop = JsonSchema.ref_for(reference, required=slot.required or not include_null)
                 elif typ and fmt is None:
                     prop = JsonSchema({"type": typ})
                 elif typ:
--- a/tests/linkml/test_biolink_model/__snapshots__/biolink.schema.json
+++ b/tests/linkml/test_biolink_model/__snapshots__/biolink.schema.json
@@ -3031,7 +3031,14 @@
                     ]
                 },
                 "object_direction_qualifier": {
-                    "$ref": "#/$defs/DirectionQualifierEnum",
+                    "anyOf": [
+                        {
+                            "$ref": "#/$defs/DirectionQualifierEnum"
+                        },
+                        {
+                            "type": "null"
+                        }
+                    ],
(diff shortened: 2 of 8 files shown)


amc-corey-cox (approved):
Looks good to me.

PR #3750, opened 2026-07-13
Fix 3749 (raw CURIEs for predicate URIs for schemas with subschemas)

### LINKED ISSUE #3749: RDFLibDumper silently emits raw CURIEs as predicate/type URIs for schemas with imported sub-schemas
**Describe the bug**

When dumping an object to RDF using `RDFLibDumper` with a schema that imports sub-schemas, predicate URIs and `rdf:type` objects belonging to the imported sub-schemas are emitted as raw, unexpanded CURIEs instead of full `https://…` URIs. No error or warning is raised — the output is silently wrong.

For example, using the `phenopackets` schema (which imports `phenopackets/base`, where the `base:` prefix is declared as `https://w3id.org/linkml/phenopackets/base/`), dumping an `OntologyClass` instance produces:
```turtle
<http://purl.obolibrary.org/obo/HP_1>  base:label       "test label" .
<http://purl.obolibrary.org/obo/HP_1>  rdf:type         base:OntologyClass .
```
instead of the correct:
```turtle
<http://purl.obolibrary.org/obo/HP_1>  <https://w3id.org/linkml/phenopackets/base/label>        "test label" .
<http://purl.obolibrary.org/obo/HP_1>  rdf:type         <https://w3id.org/linkml/phenopackets/base/OntologyClass> .
```

`URIRef("base:label")` is not a resolvable URI and will not be recognised correctly by any RDF processor or SPARQL endpoint.

**Version of LinkML you are using**: 1.11.0

The existing test schema and model in the repository are sufficient to reproduce the issue:
```python
from linkml_runtime.dumpers import rdflib_dumper
from rdflib import Graph

# phenopackets.yaml imports phenopackets/base.yaml, which declares:
#   prefixes:
#     base: https://w3id.org/linkml/phenopackets/base/
sv = SchemaView("tests/linkml_runtime/test_loaders_dumpers/input/phenopackets/phenopackets.yaml")

from tests.linkml_runtime.test_loaders_dumpers.models.phenopackets import OntologyClass
c = OntologyClass(id="HP:1", label="test label")
ttl = rdflib_dumper.dumps(c, sv)

g = Graph()
g.parse(data=ttl, format="ttl")
for s, p, o in g:
    print(s, p, o)
```

Output shows raw CURIEs:
```
http://purl.obolibrary.org/obo/HP_1  base:label       test label
http://purl.obolibrary.org/obo/HP_1  rdf:type         base:OntologyClass
```

---
**Root cause**
Two interacting issues:
 
1. [`SchemaView.namespaces()`](https://github.com/linkml/linkml/blob/main/packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py) is decorated with `@lru_cache(None)`. It builds its `Namespaces` object by iterating `self.schema_map`, which is populated lazily — only when the import closure is walked. If `namespaces()` is called before `imports_closure()` has run, the cache is seeded with an incomplete namespace map that is missing all prefixes declared in imported sub-schemas.
2. [`RDFLibDumper.as_rdf_graph()`](https://github.com/linkml/linkml/blob/main/packages/linkml_runtime/src/linkml_runtime/dumpers/rdflib_dumper.py) calls `schemaview.namespaces()` near the top of the method to bind prefixes to the RDF graph. This seeds the namespace cache before `inject_triples()` walks the schema and triggers full import resolution. Subsequent calls to `get_uri(slot, expand=True)` inside `inject_triples()` use the same incomplete cache, so CURIEs like `base:label` cannot be expanded and are silently returned as-is, producing malformed triples.

### LINKED ISSUE #3751: test_output_prefixes in test_rdflib_dumper.py has a hidden test-isolation bug and a brittle sdo: / schema1: assertion
**Describe the bug**

`test_output_prefixes` in
`tests/linkml_runtime/test_loaders_dumpers/test_rdflib_dumper.py` fails when
run in isolation (i.e. without `test_rdf_output` having run first in the same
session):

```
FAILED tests/linkml_runtime/test_loaders_dumpers/test_rdflib_dumper.py::test_output_prefixes
- assert 'prefix sdo:' in '@prefix personinfo: <https://w3id.org/linkml/examples/personinfo/> .\n
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-...tpersoninfo:phone "555-555-5555" .\n
_:c14n0 a personinfo:Container ;\n\tpersoninfo:persons ORCID:1234 , ORCID:4567 .\n'
```

Two bugs interact to produce this failure:

1. **Missing fixture dependency.** The test reads `OUT_429` from disk but
   declares no dependency on the `issue_429_graph` fixture that generates it.
   When `test_rdf_output` (which does use that fixture) happens to run first,
   `OUT_429` is freshly written to disk and the test passes. When
   `test_output_prefixes` runs first — or in isolation — the file is either
   absent or stale from a previous session.

2. **Brittle `sdo:` assertion.** The `semweb_context` prefix map binds
   `http://schema.org/` as `schema1:` rather than `sdo:`. A freshly generated
   graph therefore serialises to `schema1:Person`, not `sdo:Person`, causing
   the assertions `'prefix sdo:'` and `'sdo:Person'` to always fail on a
   fresh file.

The test was only passing in full-suite runs because `test_rdf_output` happened
to execute immediately before it and left a file on disk from an older run
(or a run with a different rdflib/semweb_context version) where `sdo:` was used.

Reproducible on current `main`.

No schema change is needed. Run the failing test in isolation:

```bash
uv run pytest tests/linkml_runtime/test_loaders_dumpers/test_rdflib_dumper.py::test_output_prefixes -v
```

The test fails because OUT_429 is either absent or stale. Then run the full
file and observe it passes only because test_rdf_output runs first and
regenerates the file:
uv run pytest tests/linkml_runtime/test_loaders_dumpers/test_rdflib_dumper.py -v

Silvanoc:
## Summary

Fixes #3749 
Fixes #3751

`RDFLibDumper.as_rdf_graph()` called `schemaview.namespaces()` before the
import closure had been walked. Because `namespaces()` is `@lru_cache`'d, the
resulting `Namespaces` object was missing all prefixes declared in imported
sub-schemas (e.g. the `base:` prefix from `phenopackets/base`). Subsequent URI
expansion inside `inject_triples()` therefore silently emitted raw CURIEs such
as `base:label` and `base:OntologyClass` as RDF predicates and type objects
instead of their full `https://…` equivalents.

The fix is a single `schemaview.imports_closure()` call at the top of
`as_rdf_graph()`, ensuring `schema_map` is fully populated before the namespace
cache is seeded.

## How was this tested?

Manual testing with code provided in issue description

Assertions were added to the existing `test_phenopackets` test in
`tests/linkml_runtime/test_loaders_dumpers/test_rdflib_dumper.py` to verify
that every predicate and `rdf:type` object in the dumped graph is a fully
expanded URI (contains `://`). The test fails on the unfixed code and passes
after the fix. The two changes were committed separately so the failing test
commit can be inspected and tested independently of the fix commit.

*Recommendation for review*:

1. Reproduce test failure running `uv run pytest tests/linkml_runtime/test_loaders_dumpers/test_rdflib_dumper.py::test_output_prefixes -v` on `main`.
2. Check out the 2nd branch commit (1. one fixing dependencies and 2. one fixing test). Previously failing test should now succeed.
3. Checkout the previous to last commit and observe following tests of `tests/linkml_runtime/test_loaders_dumpers/test_rdflib_dumper.py` failing: `test_phenopackets[prefix_map0]` and `test_phenopackets[prefix_map1]`
4. Checkout `HEAD` of the branch and see all tests succeeding.

## Areas of uncertainty

`SchemaView.namespaces()` has a docstring note that its output depends on
whether import-processing functions have been run beforehand. There may be other
call sites across the codebase that call `namespaces()` before
`imports_closure()` and are therefore affected by the same latent bug. Reviewers
may want to audit other callers of `namespaces()` for the same ordering issue.

## Checklist

## AI Assistance

### DIFF
--- a/packages/linkml_runtime/src/linkml_runtime/dumpers/rdflib_dumper.py
+++ b/packages/linkml_runtime/src/linkml_runtime/dumpers/rdflib_dumper.py
@@ -43,6 +43,7 @@ class RDFLibDumper(Dumper):
         :return:
         """
         g = Graph()
+        schemaview.imports_closure()  # ensure all imported sub-schemas are in schema_map before namespaces() caches
         if isinstance(prefix_map, Converter):
             # TODO replace with `prefix_map = prefix_map.bimap` after making minimum requirement on python 3.8
             prefix_map = {record.prefix: record.uri_prefix for record in prefix_map.records}
--- a/tests/linkml_runtime/test_loaders_dumpers/test_rdflib_dumper.py
+++ b/tests/linkml_runtime/test_loaders_dumpers/test_rdflib_dumper.py
@@ -420,6 +420,26 @@ def test_phenopackets(prefix_map):
             assert len(list(g.objects(resource_uri))) == 1
 
 
+def test_rdflib_phenopackets_roundtrip():
+    """Round-trip a phenopackets OntologyClass through dump then load.
+
+    Uses a fresh SchemaView for the load (cold namespace cache) to verify
+    that the loader also resolves sub-schema prefixes after imports_closure().
+    """
+    schema_path = str(INPUT_PATH / "phenopackets" / "phenopackets.yaml")
+    view = SchemaView(schema_path)
+    c = OntologyClass(id="HP:1", label="test label")
+    ttl = rdflib_dumper.dumps(c, view)
+    g = Graph()
+    g.parse(data=ttl, format="ttl")
+    view2 = SchemaView(schema_path)
+    objs = rdflib_loader.from_rdf_graph(g, target_class=OntologyClass, schemaview=view2)
+    assert len(objs) == 1
+    loaded = objs[0]
+    assert loaded.id == "HP:1"
+    assert loaded.label == "test label"
+
+
 def test_rdf_output(issue_429_graph):
     """Test RDF output for issue 429."""
     g = issue_429_graph
@@ -434,7 +454,7 @@ def test_rdf_output(issue_429_graph):
 
 
 def test_output_prefixes(issue_429_graph):
-    """Test output prefixes and key triples for issue 429.
+    """Test output namespace binding for issue 429.
 
     Assertions use the graph object and namespace manager directly rather than
     matching against the turtle serialisation string, so they are independent
@@ -450,15 +470,6 @@ def test_output_prefixes(issue_429_graph):
     # http://schema.org/ may be serialised as sdo: or schema1: depending on the
     # semweb_context version; check the URI rather than the label.
     assert "http://schema.org/" in bound_uris
-    # Both persons must be typed and have their properties present.
-    for orcid_id, full_name, age in [
-        ("1234", "Clark Kent", "32"),
-        ("4567", "Lois Lane", "33"),
-    ]:
-        subject = ORCID[orcid_id]
-        assert (subject, RDF.type, SDO.Person) in g
-        assert (subject, personinfo.age, Literal(age)) in g
-        assert (subject, personinfo.full_name, Literal(full_name)) in g
 
 
 def test_pydantic_model_dump():
--- a/tests/linkml_runtime/test_loaders_dumpers/test_rdflib_dumper.py
+++ b/tests/linkml_runtime/test_loaders_dumpers/test_rdflib_dumper.py
@@ -424,13 +424,19 @@ def test_rdf_output(issue_429_graph):
     assert (ORCID["4567"], personinfo.phone, Literal("555-555-5555")) in g
 
 
-def test_output_prefixes():
+def test_output_prefixes(issue_429_graph):
     """Test output prefixes for issue 429."""
-    with open(str(OUT_429), encoding="UTF-8") as file:
-        file_string = file.read()
-    prefixes = ["prefix ORCID:", "prefix personinfo:", "prefix sdo:", "sdo:Person", "personinfo:age", "ORCID:1234"]
-    for prefix in prefixes:
-        assert prefix in file_string
+    g = issue_429_graph
+    file_string = g.serialize(format="turtle")
+    # Serialisation-stable assertions: these prefixes are always present
+    # regardless of rdflib / semweb_context version.
+    for token in ["prefix ORCID:", "prefix personinfo:", "personinfo:age", "ORCID:1234"]:
+        assert token in file_string
+    # http://schema.org/ may be serialised as "sdo:" or "schema1:" depending on
+    # the semweb_context version; check at the graph level instead.
+    bound_uris = {str(ns) for _, ns in g.namespace_manager.namespaces()}
+    assert "http://schema.org/" in bound_uris
+    assert (ORCID["1234"], RDF.type, SDO.Person) in g
 
 
 def test_pydantic_model_dump():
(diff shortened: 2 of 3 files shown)


packages/linkml_runtime/src/linkml_runtime/dumpers/rdflib_dumper.py
```diff
@@ -43,6 +43,7 @@ def as_rdf_graph(
         :return:
         """
         g = Graph()
+        schemaview.imports_closure()  # ensure all imported sub-schemas are in schema_map before namespaces() caches
```
amc-corey-cox, line 46: Confirmed the fix locally — cold it dumps `<base:label>`, warmed it dumps `base:label`. It's also safer than it looks: `all_types()`/`all_enums()` downstream call `imports_closure` anyway, so this only reorders.

But the loader does the same thing from the same cold position (`loaders/rdflib_loader.py:67`) and isn't fixed. Dump with a warmed SchemaView, load back with a cold one:

```
MappingError: No pred for https://w3id.org/linkml/phenopackets/base/label
```

Same one-line fix in `from_rdf_graph`. Could you add it here, with a round-trip test? As it stands, I think we'd write RDF we can't read back.

amc-corey-cox, line 46: Your "areas of uncertainty" note looks right to me. `namespaces()` is `@lru_cache`'d but reads mutable `schema_map`, so whichever caller lands first freezes it — ~15 call sites share the latent bug.

I tried moving `imports_closure()` into `namespaces()` itself; it works and doesn't recurse. Too wide for this PR, but worth a follow-up issue — want me to file it?

tests/linkml_runtime/test_loaders_dumpers/test_rdflib_dumper.py
```diff
@@ -424,13 +433,32 @@ def test_rdf_output(issue_429_graph):
...
+    assert "http://schema.org/" in bound_uris
+    # Both persons must be typed and have their properties present.
+    for orcid_id, full_name, age in [
+        ("1234", "Clark Kent", "32"),
+        ("4567", "Lois Lane", "33"),
+    ]:
+        subject = ORCID[orcid_id]
+        assert (subject, RDF.type, SDO.Person) in g
+        assert (subject, personinfo.age, Literal(age)) in g
+        assert (subject, personinfo.full_name, Literal(full_name)) in g
```
amc-corey-cox: Good move off string matching, and checking the schema.org URI instead of the `sdo:`/`schema1:` label is the right call.

Nit: this now overlaps `test_rdf_output` — same fixture, nearly the same triples. The binding assertions above are the new value; these could go.

amc-corey-cox (approved):
Nice find, and the diagnosis is spot on — I reproduced the bug and confirmed the fix. The separated commits made that easy to verify, thanks for that.

One thing before this goes in: the loader has the same bug at loaders/rdflib_loader.py:67, and it isn't fixed here. That means we'd still write RDF we can't read back, so I'd rather the round trip land in one change than fix half of it now. Same one-line call in from_rdf_graph, plus a round-trip test.

packages/linkml_runtime/src/linkml_runtime/dumpers/rdflib_dumper.py
Silvanoc: I also considered the integration of `imports_closure()` into `namespaces()` and an AI review luckily raised a blocking issue (something with caches) why I finally discarded it. I never thought anybody would come up with that same idea and therefore haven't documented it 🫣

Silvanoc: Addressed

tests/linkml_runtime/test_loaders_dumpers/test_rdflib_dumper.py
Silvanoc: Addressed

PR #3753, opened 2026-07-13
fix(owlgen): call imports_closure() + namespaces.cache_clear() before URI expansion

### LINKED ISSUE #3752: OwlSchemaGenerator silently emits unexpanded CURIEs for types from imported sub-schemas (e.g. xsd:string instead of http://www.w3.org/2001/XMLSchema#string)
**Describe the bug**

When generating OWL from a schema that imports sub-schemas (e.g.
`https://w3id.org/linkml/types`), `SchemaView.get_uri(type, expand=True)` for
built-in types returns unexpanded CURIEs such as `xsd:string` instead of the
correct full URI `http://www.w3.org/2001/XMLSchema#string`. No error or warning
is raised — the output is silently wrong.

The root cause is a premature-namespace-cache bug: `SchemaView.namespaces()` is
`@lru_cache`'d. Inside `OwlSchemaGenerator.as_graph()`, `imports_closure()` is
called to walk the import tree, but `imports_closure()` itself internally calls
`namespaces()` (via `load_import()` → `map_import()`) mid-walk, before all
imported schemas have been added to `schema_map`. This seeds the namespace cache
with an incomplete `Namespaces` object that is missing prefixes declared only in
imported schemas (e.g. `xsd:` from `https://w3id.org/linkml/types`).

Subsequent calls to `get_uri(expand=True)` then silently return the unexpanded
CURIE. An existing workaround in `_type_uri()` (`if expanded.startswith("xsd:")`)
compensates for this in normal generation, but the root cause remains and surfaces
as a `ValueError` if `expand_curie()` is made strict.

```python
from linkml.generators.owlgen import OwlSchemaGenerator

schema_str = """
id: http://example.org/description-export
name: descriptionexport
imports:
default_range: string
default_prefix: ex
prefixes:
  ex: http://example.org/description-export/
classes:
  Person:
    slots:
      - name
slots:
  name:
    range: string
```

Silvanoc:
## Summary

Fixes #3752 

`OwlSchemaGenerator.as_graph()` triggered a `namespaces()` cache population
mid-walk through the import closure (via `imports_closure()` →
`load_import()` → `map_import()` → `namespaces()`), before all imported
sub-schemas had been added to `schema_map`. The resulting `Namespaces` object
was missing prefixes declared only in imported schemas (e.g. `xsd:` from
`https://w3id.org/linkml/types`). Subsequent calls to `get_uri(expand=True)`
on built-in types then silently returned unexpanded CURIEs such as `xsd:string`
instead of `http://www.w3.org/2001/XMLSchema#string`.

Fix: call `imports_closure()` explicitly at the start of `as_graph()` to
ensure `schema_map` is fully populated, then call `namespaces.cache_clear()`
to force a rebuild of the namespace cache from the complete `schema_map`.

Similar root-cause as that of #3750.

## How was this tested?

A new test `test_owlgen_xsd_prefixes_from_imported_schema_are_expanded` was
added to `tests/linkml/test_issues/test_linkml_issue_692.py`. It calls
`gen.serialize()` and then asserts that `get_uri(expand=True)` on the built-in
`string` type returns a fully expanded `http://...` URI rather than the raw
`xsd:string` CURIE. The test was committed separately before the fix so it can
be verified to fail on the test-only commit and pass on the fix commit.

## Areas of uncertainty

The same premature-namespace-cache pattern is likely present in other generators
that call `namespaces()` (or methods that call it transitively) before the import
closure is fully walked. A broader audit of all `namespaces()` call sites across
the generator codebase may be warranted.

Further issues might be created for this purpose.

## Checklist

## AI Assistance


packages/linkml/src/linkml/generators/owlgen.py
```diff
@@ -300,6 +300,8 @@ def as_graph(self) -> Graph:
             self.consolidate_cardinality_axioms = False
 
         sv = self.schemaview
+        sv.imports_closure()  # populate schema_map with all imported sub-schemas
```
amc-corey-cox, line 303: Nice diagnosis — I reproduced it and confirmed the fix works. But this is the third call site to get the same band-aid (the matching `rdflib_dumper` commits on this branch, plus #3778), and your own "areas of uncertainty" predicts more.

`imports_closure()` is what poisons the cache, so it can also clean up after itself. I tried this, with **no change to owlgen.py at all**:

```python
# in imports_closure(), after: closure = list(dict.fromkeys(closure).keys())
if len(self.schema_map) != _n_before:
    self.set_modified()
```

Your new test passes, and `test_schemaview.py` + `test_owlgen.py` are green (1320 passed, 6 skipped). I didn't run the full suite, so this is promising rather than proven — but worth trying before adding a fourth per-generator patch.

packages/linkml/src/linkml/generators/owlgen.py
```diff
@@ -300,6 +300,8 @@ def as_graph(self) -> Graph:
...
+        sv.namespaces.cache_clear()  # ensure namespace cache is rebuilt with the full schema_map
```
amc-corey-cox, line 304: `namespaces` is `lru_cache`'d on the *class*, so this clears the cache for every live `SchemaView` in the process, not just `sv`:

```
before: CacheInfo(currsize=2)   # two unrelated SchemaViews warmed
s1.namespaces.cache_clear()
after : CacheInfo(currsize=0)   # s2's entry gone too
```

Still correct (entries just rebuild), but wasteful in a long-lived process. `SchemaView` already has an instance-scoped idiom: `set_modified()` bumps `modifications`, which is part of `__key()`/`__hash__`, so that instance's cache entries are orphaned. Swapping in `sv.set_modified()` on your test-only commit passes the test.

amc-corey-cox, line 304: Follow-up opportunity: the `if expanded.startswith("xsd:")` workaround in `_type_uri()` (~line 1619) is commented `# TODO: fix upstream in schemaview` — this PR is arguably that fix. I deleted those three lines on top of your head commit and `test_owlgen.py` + this file stayed green (96 passed, 4 skipped) on Linux. The comment blames Windows `default_curi_maps`, so CI's Windows jobs are the real check. Optional, but it removes the layer that masked this bug in the first place.

Nit: `imports_closure()` here ignores `mergeimports`, which `as_graph()` otherwise threads through every `all_*()` call. Probably harmless, but if `mergeimports=False` means "don't touch imports", this walks (and fetches) them anyway.

tests/linkml/test_issues/test_linkml_issue_692.py
```diff
@@ -66,3 +66,42 @@ def test_owlgen_rdfs_profile():
 
     # and check rdfs:comment for property 'name' too
     assert (name_prop, RDFS.comment, None) in graph
+
+
+def test_owlgen_xsd_prefixes_from_imported_schema_are_expanded():
```
amc-corey-cox, line 71: This asserts internal `SchemaView` state and uses `serialize()` only as a trigger — and it picks the one case where output *doesn't* change. I diffed the generated TTL for the issue's schema before/after the fix: **byte-identical**, because `_type_uri()`'s `xsd:` workaround already patches that case.

The fix does change real output, though. `map_import()` only calls `namespaces()` for imports containing `:`, so the cache is poisoned at the first colon-import — anything loaded after it loses its prefixes. Put a local sub-schema *after* a colon-import and you get a garbage IRI:

```yaml
# main.yaml — sub.yaml declares `foo:` and a type with `uri: foo:Bar`
imports: [sub, https://w3id.org/linkml/types]
```

before: `<foo:Bar> a rdfs:Datatype .`
after:  `<http://example.org/foo/Bar> a rdfs:Datatype .`

That's the test I'd write — it asserts the defect a user sees and survives a refactor of *how* the cache gets invalidated.

Two nits: this file is scoped to issue 692 (convention here is one file per issue — `test_linkml_issue_3752.py`), and the docstring documents the fix's implementation, which goes stale if the fix moves. Describe the guarantee instead.

tests/linkml/test_issues/test_linkml_issue_692.py
```diff
@@ -66,3 +66,42 @@ def test_owlgen_rdfs_profile():
...
+        skip_vacuous_local_range_axioms=True,
+        consolidate_cardinality_axioms=True,
+    )
+    # Trigger the full generation pipeline so as_graph() applies the fix
+    gen.serialize()
+
+    sv = gen.schemaview
+    string_type = sv.get_type("string")
+    expanded = sv.get_uri(string_type, expand=True)
+    assert "://" in expanded, (
```
amc-corey-cox: The expected value is known exactly — assert `expanded == "http://www.w3.org/2001/XMLSchema#string"` rather than `"://" in expanded`.

amc-corey-cox (approved):
Nice catch on the root cause — I reproduced the bug and confirmed the two-line fix works, so this is a real improvement over the status quo.

My main suggestion is to fix it once in imports_closure() rather than per-generator (I sketched a patch that passes your test with no owlgen change at all), and to assert the generated output rather than internal SchemaView state — for the schema in the issue the TTL is byte-identical before and after.

Sorry for the lengthy comments. I investigated these with Claude Code and I didn't pare them down much. I hope that is helpful. If there are logic issues that's on me but I did vet this before submitting.

amc-corey-cox (changes requested):
Sorry, the previous review should have been request changes. You don't have to fix all of what I put there but please respond with pushback if you disagree.

packages/linkml/src/linkml/generators/owlgen.py
Silvanoc: I remember having tried to fix it also in `SchemaView` directly, seeing some promising results and then failing on the full-blown CI workflow (some cache cross-polution, if I remember right). I don't remember though if it was the same fix.

I cannot test it right now in the system I'm using to write this. But I'd prepared PR #3793 to enable testing CI workflows locally. @amc-corey-cox could you please give it a try to confirm/deny that it works?

amc-corey-cox: Ran the experiment — turned out I didn't need #3793, I just ran the three CI test jobs locally.

On current `main` + your test-only commit (65f41a6), with the `imports_closure()` change and **no change to `owlgen.py` at all**:

| job | result |
|---|---|
| `tests/linkml/` (`-n auto --dist loadgroup`) | 11682 passed, 0 failed |
| `tests/linkml/` **serial** | 11682 passed, 0 failed |
| `tests/linkml_runtime/` | 1838 passed, 0 failed |
| slow + biolink | 46 passed, 0 failed |

I ran it serially as well specifically to look for the cross-pollution you remembered — order-dependent breakage is exactly what `--dist loadgroup` can hide. Identical counts both ways, so I can't reproduce it.

My guess is your earlier attempt was a different patch. An unguarded `set_modified()` churns the cache on every call, which is the failure mode `relmodel_transformer.py:464` already warns about in a comment. The `len(schema_map)` guard is what keeps this one cheap — it only fires the first time, when imports actually got loaded:

```python
n_before = len(self.schema_map)        # top of imports_closure()
...
if len(self.schema_map) != n_before:   # after the closure dedup
    self.set_modified()
```

(There's precedent too — `merge_imports()` already calls `set_modified()` after mutating the schema for the same reason.)

**That said, I don't want to hold this PR for a refactor you can't currently test.** Your two lines are correct, your CI is green including Windows, and this has been sitting a while. Let's merge as-is. I'll open a follow-up to move the invalidation into `imports_closure()` and drop the band-aid from `owlgen`, `rdflib_dumper` and #3778 in one go — your "areas of uncertainty" section was right, and that's the place to act on it.

packages/linkml/src/linkml/generators/owlgen.py
amc-corey-cox: Withdrawing this as a change request — it folds into the follow-up.

The observation stands for the record (`namespaces` is `lru_cache`'d on the class, so `cache_clear()` clears every live `SchemaView`), but it's a minor waste rather than a correctness problem, and it disappears entirely once the invalidation moves into `imports_closure()`.

amc-corey-cox: Both of these are follow-up material, not blockers. Recording them so they don't get lost:

- the `xsd:` workaround in `_type_uri()` (`# TODO: fix upstream in schemaview`) looks removable once the root cause is fixed properly — Windows CI is the real check there
- `imports_closure()` here ignores `mergeimports`, which `as_graph()` otherwise threads through every `all_*()` call

I'll carry both into the follow-up issue.

tests/linkml/test_issues/test_linkml_issue_692.py
amc-corey-cox: Following up on my own comment: the output-level repro does exist, and import order is the part I hadn't pinned down. Because `todo` is a FILO stack, a local sub-schema listed *before* a colon-import is *processed* after it — so it loads into the already-poisoned cache:

```yaml
# main.yaml — sub.yaml declares `foo:` and a type with `uri: foo:Bar`
imports: [sub, https://w3id.org/linkml/types]
```

before: `<foo:Bar> a rdfs:Datatype .`
after:  `<http://example.org/foo/Bar> a rdfs:Datatype .`

The reverse order is byte-identical either way, which is probably why an internal-state assertion was the natural thing to reach for here — the obvious schema doesn't show the defect in its output.

Not asking you to rewrite the test for this PR. I'll fold the output-level case into the follow-up, along with the filename-convention nit.

tests/linkml/test_issues/test_linkml_issue_692.py
amc-corey-cox: I went ahead and did this one myself since it's mechanical and you mentioned you can't test right now — pushed 5d987fc to your branch, changing the assertion to:

```python
assert expanded == "http://www.w3.org/2001/XMLSchema#string"
```

Flagging it explicitly since it's a commit on your branch that you didn't write — revert it if you'd rather take it differently.

amc-corey-cox (approved):

Issue #3721, opened 2026-07-06
Enable diff-stable (content-hash) blank-node labels across LinkML's own RDF generators


amc-corey-cox:
Once #3704 lands the opt-in `stable_blank_node_labels` path, we should turn it on for LinkML's own RDF-producing generators so our generated artifacts get change-locality (small schema edit → small diff), not just run-to-run stability.

**Depends on #3704** — do not start until it has merged.

### Scope
- [ ] Wire `stable_blank_node_labels` into the call-sites #3704 left out: `gen-rdf`, `gen-shex`, and the non-rdflib `rdf_dumper`. (#3704 only wired `gen-owl`, `gen-shacl`, and `RDFLibDumper.dumps`.)
- [ ] Turn the option **on** for LinkML's own artifact generation.
- [ ] Land the resulting one-time re-serialization as a single, clearly-labeled commit so the churn isn't conflated with a real change.

### Notes
- Default stays **off** for downstream users (opt-in); this only changes what *we* emit. Downstream users get change-locality by opting in themselves.
- Residual SHACL churn (the `rdf:List` spine cascade and automorphic-collision suffixes) is expected and tracked separately — not a blocker here.

Relates to #3702, #3212.

jdsika:
It was suggested to add the diff-stability into its own library: https://github.com/ASCS-eV/diffable-rdf (first tests I am doing here as I needed the feature somewhere else)

PR #3754, opened 2026-07-13
feat(rdf): enable stable blank-node labels across gen-rdf and RDFDumper (#3721)

amc-corey-cox:
Implements #3721. Completes the opt-in `stable_blank_node_labels` wiring across the RDF entry points LinkML owns, so change-locality is available everywhere — not just `gen-owl`/`gen-shacl` (wired in #3704).

## Changes
- **`gen-rdf`** (`RDFGenerator`): new `stable_blank_node_labels` field + `--stable-blank-node-labels/--no-stable-blank-node-labels` CLI flag, threaded into `canonicalize_rdf_graph`. Mirrors `gen-owl`.
- **non-rdflib `RDFDumper.dumps`**: adds the `stable_blank_node_labels` kwarg, mirroring `RDFLibDumper.dumps`.

`gen-shex` is intentionally out of scope (ShEx-side determinism is tracked separately).

## Test
`test_stable_blank_node_labels` (gen-rdf on personinfo): default emits ordinal `c14n` labels; opt-in introduces content-hash labels and stays isomorphic. A residue of `c14n` remains for cyclic blank-node closures — #3704's documented cycle fallback (the metamodel-as-RDF is recursive).

## Notes
- **Stacked on #3704** — until that merges, this PR's diff includes #3704's changes; it reduces to just these three files once #3704 lands and this is rebased onto `main`.
- The "enable for our own generation" item in #3721 is **N/A in this repo** — there are no committed RDF *build products* here (the metamodel TTL is vendored from `linkml-model`; the `examples/` artifacts are tutorial content). The actual flip is a consumer-side (linkml-model / downstream) opt-in.
- `RDFDumper` (non-rdflib) is wired for API symmetry but is a **legacy path** — no internal callers or existing tests, and it produces degenerate output in a basic smoke test — so no dedicated test was added. Whether it should be deprecated rather than extended is worth a separate look.


amc-corey-cox:
Revived alongside #3704: merged the updated stack branch (which now includes current main), and the failing `test_stable_blank_node_labels` passes again — the failure was staleness against the moved base, not a logic bug. The test now also asserts the bare-hash label shape that matches oxigraph#1824's `UnstableHashedIds` contract (see the alignment note on #3704), and carries `@pytest.mark.network` for the offline test guard that landed in #3897. Full rdfgen/gen-rdf/dumper test surface passing locally (170 tests).

PR #3756, opened 2026-07-13
feat(javagen): add javabundle dataclass, serialize and render

noelmcloughlin:
## Summary

Fix for #3755

- Add `JavaBundle` dataclass in `javagen.py`
- Add `render(template_variant=None, extra_templates=None, visitors=None) -> JavaBundle` (No I/O)
- Rewrite `serialize()` internally to `render()` then loop-write
- Tests in `test_javagen.py` with asserts

## How was this tested?

Full test suite
Added new tests

## Checklist

- [x] My code follows the [contributor guidelines](https://linkml.io/linkml/maintainers/contributing.html)
- [x] I have added tests that prove my fix/feature works
- [x] Existing tests pass locally with my changes

## AI Assistance
LLM assistance, human review, and testing

PR #3758, opened 2026-07-13
feat(javagen): fix generator.serialize -> str contract (2 of 2)

### LINKED ISSUE #3755: rearchitect javagen for project generator interoperability
**Describe the bug**

Issue #2537 calls out that "_the project generator does not know about the pydantic and java generators_" and a reason is supplied in projectgen:
https://github.com/linkml/linkml/blob/668a0b1083da605946ba0b8435897051473725f4/packages/linkml/src/linkml/generators/projectgen.py#L52

Before considering whether including java is good thing, it seems sensible to align `javagen` architecture if practical  (respecting current java "one public class per file" rule - but also noting a future feature for single project tarball could add value = persistent identified release artifact).

The re-architecture should follow `rustgen` and (split-mode) `pydanticgen` - separate a pure `render()` method from a side-effecting `serialize()` - fixing the `Generator.serialize -> str` contract required by `projectgen`.

**Pattern seen**

- `PydanticGenerator.render()` -> `PydanticModule + serialize(rendered_module=None) -> str`
- `RustGenerator.render(mode='file'|'crate')` -> `FileResult | CrateResult + serialize() -> str` with centralized I/O
- `GolangGenerator` has same `render + serialize` split (single file use case)

1.11.1

noelmcloughlin:
## Summary

Fixes #3755

Aligning `javagen` architecture to other generator patterns for `projectgen.py` interoperability.
- update `serialize` signature
- return value delimited concatenation (parallels rustgen)
- update abstract `OOCodeGeneator.serialze` as needed
- delete dead mutation `self.directory = directory` in javagen.py
- Tests added (assert render is not re-invoked when `rendered=` is passed

## How was this tested?

Full tests + new tests

## Checklist

## AI Assistance
LLM assisted, human review, testing

### DIFF
--- a/packages/linkml/src/linkml/generators/javagen.py
+++ b/packages/linkml/src/linkml/generators/javagen.py
@@ -339,35 +339,54 @@ class JavaGenerator(OOCodeGenerator):
 
     def serialize(
         self,
-        directory: str,
+        directory: str | Path | None = None,
         template_variant: str | None = None,
         extra_templates: list[str] | None = None,
         visitors: list[str] | None = None,
+        rendered_module: JavaBundle | None = None,
         **kwargs,
-    ) -> None:
-        """Generate and write the Java code to files.
+    ) -> str:
+        """Generate the Java code and, if ``directory`` is given, write it to disk.
 
-        Thin wrapper around :meth:`render` that writes each rendered file in
-        the returned :class:`JavaBundle` to ``directory``.
+        Thin wrapper around :meth:`render` that additionally writes each
+        rendered file to ``directory`` and returns a single ``str``
+        containing every rendered file, each preceded by a ``// FILE: <name>``
+        header (mirrors `rustgen.RustGenerator.serialize`).
 
-        :param directory: The directory where to write the code files.
+        :param directory: Optional directory to write the code files to. If
+            ``None``, files are only rendered in memory and returned.
         :param template_variant: The name of the template variant to use, if any.
+            Ignored when ``rendered_module`` is provided.
         :param extra_templates: A list of additional templates from which to generate
-            additional code files. See :meth:`render` for details.
+            additional code files. See :meth:`render` for details. Ignored when
+            ``rendered_module`` is provided.
         :param visitors: A list of class names for which to generate a visitor
-            interface. See :meth:`render` for details.
+            interface. See :meth:`render` for details. Ignored when
+            ``rendered_module`` is provided.
+        :param rendered_module: Optional pre-computed :class:`JavaBundle` to
+            write instead of calling :meth:`render` afresh. Allows caller to
+            render once and inspect/write multiple times. When supplied,
+            ``template_variant``, ``extra_templates``, and ``visitors``
+            are ignored (the bundle is used as-is).
+        :return: All rendered files concatenated, each preceded by a
+            ``// FILE: <name>`` header.
         """
-        bundle = self.render(
-            template_variant=template_variant,
-            extra_templates=extra_templates,
-            visitors=visitors,
+        bundle = (
+            rendered_module
+            if rendered_module is not None
+            else self.render(
+                template_variant=template_variant,
+                extra_templates=extra_templates,
+                visitors=visitors,
+            )
         )
-        self.directory = directory
-        os.makedirs(directory, exist_ok=True)
-        for filename, code in bundle.files.items():
-            path = os.path.join(directory, filename)
-            with open(path, "w", encoding="UTF-8") as stream:
-                stream.write(code)
+        if directory is not None:
+            os.makedirs(directory, exist_ok=True)
+            for filename, code in bundle.files.items():
+                path = os.path.join(directory, filename)
+                with open(path, "w", encoding="UTF-8") as stream:
+                    stream.write(code)
+        return "\n\n".join(f"// FILE: {name}\n{code}" for name, code in bundle.files.items())
 
     # The following methods are intended to be used from within a code
     # template.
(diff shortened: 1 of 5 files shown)


packages/linkml/src/linkml/generators/javagen.py
```diff
@@ -101,6 +101,30 @@ def __post_init__(self):
...
+    CrateResult) and `pydanticgen.PydanticGenerator` (PydanticModule).
+    """
+
+    files: dict[str, str] = field(default_factory=dict)
+    """Rendered Java source, keyed by filename (e.g. ``"Address.java"``)."""
+
+    package: str = ""
+    """Java package name the rendered files belong to (informational)."""
+
+    documents: list[OODocument] = field(default_factory=list)
```
gouttegd, line 123: So far, `OODocument` (and the related classes such as `OOField`, etc.) could have been considered a relatively “internal” implementation detail (for as much as the notion even exists in Python) of the generators based on `oocodegen`, because they were only used internally by those generators without ever being exposed further.

This will no longer be the case after this change.

Not opposed to it, just something to be aware of for anyone working with `oocodegen`.

packages/linkml/src/linkml/generators/javagen.py
```diff
@@ -306,11 +333,60 @@ class and all its descendants will include a `accept(IFooVisitor)`
...
+        visitors: list[str] | None = None,
+        rendered_module: JavaBundle | None = None,
+        **kwargs,
+    ) -> str:
+        """Generate the Java code and, if ``directory`` is given, write it to disk.
+
+        Thin wrapper around :meth:`render` that additionally writes each
+        rendered file to ``directory`` and returns a single ``str``
+        containing every rendered file, each preceded by a ``// FILE: <name>``
+        header (mirrors `rustgen.RustGenerator.serialize`).
```
gouttegd, line 354: What is the rationale for returning “a single `str` containing every rendered file”?

OK, the Rust generator does it, but in the case of the Rust generator I believe it makes sense. For the Java generator, I don’t see the point. No Java tool (especially no Java compiler) could ever use such a string.

The only way to use such a string would be to split it on the `// FILE: <name>` headers and to dispatch the contents of each section into its own file – exactly what this method already does when provided with a `directory` parameter.

I can imagine that some callers would want to do the dispatching themselves, but in that case they might as well call `render()` and then work from the `JavaBundle` (where the contents of each file is already available in a separate string).

tests/linkml/test_generators/test_javagen.py
```diff
@@ -228,3 +230,120 @@ def test_refined_ranges(input_path):
...
+
+
+def test_render_template_variant(kitchen_sink_path, tmp_path):
+    """`render(template_variant=...)` is honoured (records variant here)."""
+    gen = JavaGenerator(kitchen_sink_path, package=PACKAGE)
+    bundle = gen.render(template_variant="records")
+
+    assert "Address.java" in bundle.files
+    assert "public record Address(String street, String city, BigDecimal altitude)" in bundle.files["Address.java"]
+    assert list(tmp_path.iterdir()) == []
```
gouttegd, line 260: Checking that no files are written in rendering mode has already been tested just above. There should be no need to test that again here.

tests/linkml/test_generators/test_javagen.py
```diff
@@ -228,3 +230,120 @@ def test_refined_ranges(input_path):
...
+    gen = JavaGenerator(kitchen_sink_path, package=PACKAGE, true_enums=True)
+    bundle = gen.render()
+
+    assert "CordialnessEnum.java" in bundle.files
+    assert "public enum CordialnessEnum" in bundle.files["CordialnessEnum.java"]
+    assert list(tmp_path.iterdir()) == []
+
+
+def test_serialize_delegates_to_render(kitchen_sink_path, tmp_path, monkeypatch):
+    """`serialize()` calls `render()` and writes its `files` to `directory`."""
```
gouttegd, line 287: As long as `serialize` writes what it is expected to write, whether it delegates its job to `render` should be irrelevant, and not something that has to be tested.

This kind of intrusive testing, trying to test not only the observable behaviour (the “contract”) of the method but also its inner workings, will only make it needlessly more difficult to change the implementation of `serialize` later on.

gouttegd (commented):
Looks good to me overall, with two main concerns:

(A) I am skeptical about the interest of having `serialize()` return the entire Java code as a single string – as I mention in the inline comments, this does not really make any sense for Java code (contrary to other generators).

(B) Some tests are too “intrusive” in that they test _how_ a method does its job instead of testing that it actually does its job.

packages/linkml/src/linkml/generators/javagen.py
noelmcloughlin: I agree and on reflection will remove as it is currently unnecessary - fixed in #3756

packages/linkml/src/linkml/generators/javagen.py
noelmcloughlin: Agree it should be dropped.  I was caught "ball watching" since existing contract promises to return "str" but implicitly returns "None" without any documentation or commentary - so I assumed a bug was frustrating the original authors intent to return a str - but your comments seem to confirm we can assume the signature was incorrect, and intent was "return None".

noelmcloughlin: Fixed in #3756

tests/linkml/test_generators/test_javagen.py
noelmcloughlin: Good catch.  - fixed in #3756

noelmcloughlin:
Closing / Consolidated into #3756

PR #3761, opened 2026-07-14
minor: fix excelgen serialize bug and projectgen duplicate call

### LINKED ISSUE #3759: ExcelGenerator.serialize returns None / projectgen calls twice
**Describe the bug**

`ExcelGenerator.serialize` is annotated `-> str` but has no return statement, explicitly implying `None`:

https://github.com/linkml/linkml/blob/668a0b1083da605946ba0b8435897051473725f4/packages/linkml/src/linkml/generators/excelgen.py#L170

Excel output is binary and cannot cleanly be a pure `render() -> str` without awkward solutions:
(a) returning `bytes` breaking `-> str`
(b) base64
(c) returning a Workbook object

Excel could be migrated to render a `ExcelBundle` as in memory representation.  See #3756 and #3758 

`projectgen.py` also calls `gen.serialize()` twice:

https://github.com/linkml/linkml/blob/668a0b1083da605946ba0b8435897051473725f4/packages/linkml/src/linkml/generators/projectgen.py#L140
https://github.com/linkml/linkml/blob/668a0b1083da605946ba0b8435897051473725f4/packages/linkml/src/linkml/generators/projectgen.py#L151

1.11.1

noelmcloughlin:
## Summary

Fixes #3759

- ExcelGenerator.serialize was annotated '-> str' but has no return statement, so implicitly returns None. Fix.

- projectgen called gen.serialize() twice for excel (once at  generic call site,  again in 'excel' branch). The second call was redundant since the first already wrote the .xlsx file. Fix with generic 'if gen_dump is None: continue' - any generator whose serialize writes its own files and returns None is now auto-supported without special-casing.

- Harden `test_projectgen` to assert excel/kitchen_sink.xlsx is actually produced, guarding the is-None skip

## How was this tested?

Full suite
adjusted one test to guard against regression

## Checklist

## AI Assistance
LLM assist, human review

### DIFF
--- a/packages/linkml/src/linkml/generators/excelgen.py
+++ b/packages/linkml/src/linkml/generators/excelgen.py
@@ -171,7 +171,7 @@ class ExcelGenerator(Generator):
 
         dv.add(f"{column_letter}2:{column_letter}1048576")
 
-    def serialize(self, **kwargs) -> str:
+    def serialize(self, **kwargs) -> None:
         sv = self.schemaview
         all_classes = sv.all_classes(imports=self.mergeimports)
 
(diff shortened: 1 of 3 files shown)


noelmcloughlin:
Superceded by https://github.com/linkml/linkml/pull/3799

Issue #3421, opened 2026-04-18
Update test suites to use 'raw.githubusercontent.com' url, not 'github.com'


noelmcloughlin:
CI failed for #3420 due to flaky network.
https://github.com/linkml/linkml/actions/runs/24606760280/job/71954218254?pr=3420

The root cause may be using `github.com` instead of ``raw.githubusercontent.com`.

For example `tests/linkml_runtime/test_utils/test_schemaview.py` has this:
```
CREATURE_SCHEMA_RAW_URL = "https://github.com/linkml/linkml/raw/refs/heads/main/tests/linkml_runtime/test_utils/input/mcc/creature_schema.yaml"
```
This would be more stable for raw content:
```
CREATURE_SCHEMA_RAW_URL = ("https://raw.githubusercontent.com/linkml/linkml/refs/heads/main/tests/linkml_runtime/test_utils/input/mcc/creature_schema.yaml")
```

noelmcloughlin:
And `HTTP Error 503: https://biolink.github.io/biolink-model/biolink-model.yaml`

amc-corey-cox:
Still live, and it cost us three CI cycles today on #3709 — recording the evidence since it's now blocking merges rather than just being noise.

Same constant, `CREATURE_SCHEMA_RAW_URL` at `test_schemaview.py:60`, still pointing at the redirecting `github.com/.../raw/...` form. When it trips, all six parametrisations of `test_creature_schema_entities_with_without_imports[*-creature_view_direct_url]` fail together with:

```
http.client.RemoteDisconnected: Remote end closed connection without response
```

Seen three times in a few hours: once locally, once on `test (ubuntu-latest, 3.13)`, once on `test (ubuntu-latest, 3.10)`. Passes on retry every time. Because it lands in the required `test` jobs, a merge-ready PR goes red on network luck alone.

The redirect hypothesis in the original report checks out:

```
302, 1 redirect, final 200   https://github.com/linkml/linkml/raw/refs/heads/main/...
200, 0 redirects             https://raw.githubusercontent.com/linkml/linkml/refs/heads/main/...
```

So every one of these fetches is an extra round trip through github.com, which is more aggressively rate-limited than the raw host.

Worth noting the swap makes the flake *less likely* rather than impossible — the test still does a live network fetch inside an assertion. If it keeps recurring after the URL change, the fixture is probably the thing to address rather than the host.

amc-corey-cox:
[...]

PR #3765, opened 2026-07-14
fix(openapigen): replace raw schema with schemaview provided one

### LINKED ISSUE #3764: Use of unresolved raw schema in OpenAPI generator
**Describe the bug**

The OpenAPI Generator is using the raw parse schema (which does not resolve imports) instead of the one provided by SchemaView, which resolves imports lazily.

**Version of LinkML you are using** 1.11.1

### LINKED ISSUE #3768: Missing OpenAPI schema for constrained types
**Describe the bug**

LinkML types that add constraints to standard types (like a `string` with a `pattern`) should be added as `components/schema` data schemas in OpenAPI. As of now only the base type without the constraint would be added.

**Version of LinkML you are using** `main`

None provided

Silvanoc:
## Summary

Fixes #3764
Fixes #3768 

`self.schema` returns the raw parsed schema without any imports resolution. This patch replaces it with the schema provided by the SchemaView which provides lazy resolution.

Replacing the wording `schema class` with `data schema` in the OpenAPI context. That specification uses the workding *data* (whereas *class*  is not used with that meaning at all) and provides the section `components/schema` to provide those *data's schemas*. Therefore the new naming.

## Checklist

## AI Assistance

### DIFF
--- a/packages/linkml/src/linkml/generators/openapigen.py
+++ b/packages/linkml/src/linkml/generators/openapigen.py
@@ -250,7 +250,9 @@ class OpenApiGenerator(Generator):
             if referenced_resource_name != class_name:
                 self._renaming[class_name] = referenced_resource_name
             referenced_class_names.add(class_name)
-            json_schema = JsonSchemaGenerator(self.schema, include_null=False, top_class=class_name).generate()
+            json_schema = JsonSchemaGenerator(
+                self.schemaview.schema, include_null=False, top_class=class_name
+            ).generate()
             json_schema_classes = json.loads(json_schema.to_json())["$defs"]
             class_schemas = class_schemas | json_schema_classes
         class_schemas = self._sanitize_schemas(class_schemas, referenced_class_names)
@@ -267,7 +269,8 @@ class OpenApiGenerator(Generator):
 
     def printout_template(self) -> str:
         """Return a generic OpenAPI template pre-filled with the first class of the schema."""
-        first_class = next(iter(self.schema.classes.keys()))
+        class_names = self.schemaview.all_classes().keys()
+        first_class = next(iter(class_names))
         if re.search(r"[ :\d]", first_class):
             first_class = f'"{first_class}"'
         return openapi_generic_template.format(schema_id=self.schema.id, schema_class=first_class)
--- a/packages/linkml/src/linkml/generators/openapigen.py
+++ b/packages/linkml/src/linkml/generators/openapigen.py
@@ -31,27 +31,27 @@ paths:
     get:
       responses:
         '200':
-          description: Endpoint example involving random schema class
+          description: Endpoint example involving random data schema
           content:
             application/json:
               schema:
                 # any broken reference will cause template instantiation to fail
                 # OpenAPI editors typically also report them
-                $ref: '#/components/schemas/{schema_class}'
+                $ref: '#/components/schemas/{data_schema}'
 components:
-  # any class schema provided here that is not used by at least
+  # any data schema provided here that is not used by at least
   # one endpoint will be eliminated from the template instantiation
   # OpenAPI editors typically also report them
   schemas:
     # this resource name can differ from the name in the LinkML schema
     # it must only match the corresponding endpoint `$ref` references
-    {schema_class}:
+    {data_schema}:
       type: object
       description: Resource schema to be generated from the LinkML data model.
       # schema ID mismatching with provided schema will cause template
       # instantiation to fail
-      x-linkml-schema: {schema_id}
-      x-linkml-source: {schema_class}
+      x-linkml-schema: {linkml_schema_id}
+      x-linkml-source: {data_schema}
 """
 
 
@@ -62,9 +62,9 @@ class OpenApiGenerator(Generator):
 
     The generator composes a user-provided OpenAPI template (containing the API header,
     paths/endpoints, and security schemes) with JSON Schema components generated from
-    the LinkML schema via :class:`.JsonSchemaGenerator`. Only classes referenced by the
-    template's endpoints (and their transitive dependencies) are included in the
-    ``components/schemas`` section.
+    the LinkML schema via :class:`.JsonSchemaGenerator`. Only data schemas referenced
+    by the template's endpoints (and their transitive dependencies) are included in
+    the ``components/schemas`` section.
     """
 
     generatorname = os.path.basename(__file__)
@@ -90,7 +90,7 @@ class OpenApiGenerator(Generator):
         repr=False,
     )
 
-    def _find_referenced_resources(self) -> set[str]:
+    def _find_referenced_schemas(self) -> set[str]:
         """Return the set of resource names referenced by the template's endpoints."""
         result = set()
         for endp_spec in self._template["paths"].values():
@@ -105,8 +105,8 @@ class OpenApiGenerator(Generator):
                         if "content" in response:
                             for content_spec in response["content"].values():
                                 if "$ref" in content_spec["schema"]:
-                                    class_name = content_spec["schema"]["$ref"].removeprefix("#/components/schemas/")
-                                    result.add(class_name)
+                                    resource_name = content_spec["schema"]["$ref"].removeprefix("#/components/schemas/")
+                                    result.add(resource_name)
         return result
 
     def _fix_openapi_spec(self, element: dict | list) -> dict | list | None:
@@ -143,7 +143,7 @@ class OpenApiGenerator(Generator):
 
     def _rename(self, element: dict | list) -> dict | list | None:
         """
-        If the resource names do not correspond the schema class names,
+        If the resource names do not correspond the data schema names,
         then some renaming is needed so that OpenAPI resource names
         are properly referenced throughout the whole OpenAPI file.
         """
@@ -156,9 +156,9 @@ class OpenApiGenerator(Generator):
                 if isinstance(value, dict | list):
                     value = self._rename(value)
                 elif isinstance(value, str) and value.startswith("#/components/schemas/"):
-                    class_name = value[len("#/components/schemas/") :]
-                    if class_name in self._renaming:
-                        value = value.replace(class_name, self._renaming[class_name])
+                    data_schema_name = value[len("#/components/schemas/") :]
+                    if data_schema_name in self._renaming:
+                        value = value.replace(data_schema_name, self._renaming[data_schema_name])
                 renamed_element[key] = value
         elif isinstance(element, list):
             renamed_element = []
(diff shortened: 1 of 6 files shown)


noelmcloughlin:
I see this is 1st priority. Will test this.

Silvanoc:
> I see this is 1st priority. Will test this.

@noelmcloughlin I recommend you to review at commit-level. In fact the relevant commit is [the 1st one](https://github.com/linkml/linkml/pull/3765/changes/622611c62721347ce046fa59f872fbc08db80386) (which is also a trivial one). **All other commits might be postponed, if you don't have time to review it all.**

Silvanoc:
@noelmcloughlin I've added a small commit replacing some missing `self.schema` with `self.schemaview.schema`.

I would also typically sync with `main` with a rebase, what breaks your local history (if you are doing the review locally). Would it be fine if I rebase onto `main`? If you are reviewing on the GitHub WebUI, then you shouldn't notice anything.

Silvanoc:
Working on follow-up PRs I've realized that the code in `serialize` is quite convoluted and hard to grasp (also some unfortunate variable names and no-op code are responsible for it). The review can therefore be a bit tough.

I've created a refactoring commit that should make it all much easier to grasp, but I'm not pushing it to avoid pulling @noelmcloughlin the carpet from under his feet in an ongoing review. If it's though desired, I can push it.

noelmcloughlin:
Yes, push the commit - I may review this evening or tomorrow morning - had no time yet.

Silvanoc:
> Yes, push the commit - I may review this evening or tomorrow morning - had no time yet.

Done. @noelmcloughlin the changes cut through the changes done by many of the other commits. Therefore a commit-based review doesn't make so much sense now as before.

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -84,32 +89,93 @@ class OpenApiGenerator(Generator):
     )
     # Mapping of OpenAPI version strings to validators from openapi-spec-validator.
     # Extend this dict when adding support for additional OpenAPI versions.
-    _openapi_validators: dict[str, type] = field(
+    _openapi_validators: dict[str, type[OaSpecValidator]] = field(
         default_factory=lambda: {"3.0.3": OpenAPIV30SpecValidator},
```
noelmcloughlin, line 93: Version 3.0.3 binds the solution to a minor semantic version (see also `docs/generators/openapi.rst`) - is this necessary or required by python validator library?  Does this cause tangible restriction on users?

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -31,27 +36,28 @@
     get:
       responses:
         '200':
-          description: Endpoint example involving random schema class
+          description: Endpoint example involving random data schema
```
noelmcloughlin, line 39: This string is changed in multiple places. The issue, PR, and documentation text does not explain what is `data schema / data_schema` and why its better than `schema_class / schema_class`.  The OAS is not helpful either - it mentions data types - but not data schemas.

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -31,27 +36,28 @@
...
+    # it creates a mapping between names in OpenAPI and LinkML
+    {data_schema}:
       type: object
       description: Resource schema to be generated from the LinkML data model.
       # schema ID mismatching with provided schema will cause template
       # instantiation to fail
-      x-linkml-schema: {schema_id}
-      x-linkml-source: {schema_class}
+      x-linkml-schema: {linkml_schema_id}
+      x-linkml-source: {data_schema}
```
noelmcloughlin, line 60: The assignment seems inconsistent.  I would have expected following:

- `x-linkml-schema = linkml_schema_id`
- `x-linkml-source = linkml_data_schema`       # instead of `data_schema`

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -84,32 +89,93 @@ class OpenApiGenerator(Generator):
...
     # Extend this dict when adding support for additional OpenAPI versions.
-    _openapi_validators: dict[str, type] = field(
+    _openapi_validators: dict[str, type[OaSpecValidator]] = field(
         default_factory=lambda: {"3.0.3": OpenAPIV30SpecValidator},
         init=False,
         repr=False,
     )
 
-    def _find_referenced_resources(self) -> set[str]:
+    def _validate_oa_template(self, oa_validator_class: type[OaSpecValidator], expected_version: str, format_name: str):
```
noelmcloughlin: Minor nit - the [glossary](https://learn.openapis.org/glossary.html) does not define `OA` so presumably `_oa_` refers to  `OAD` or `OAS`.  Maybe change to  `_validate_oad_template` as OpenAPI developers may be picky.

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -84,32 +89,93 @@ class OpenApiGenerator(Generator):
...
+        # Validate that the template declares the expected OpenAPI version
+        declared_version = self._template.get("openapi")
+        if declared_version != expected_version:
+            raise ValueError(
+                f"Template OpenAPI version is '{declared_version}', "
+                f"but format '{format_name}' requires version '{expected_version}'"
+            )
+        # Validate the input template against the OpenAPI specification.
+        # This also catches dangling $ref targets in endpoints.
+        openapi_validate(self._template, cls=oa_validator_class)
```
noelmcloughlin: Maybe `oa_` should be `oad_` - see previous comment above?

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -84,32 +89,93 @@ class OpenApiGenerator(Generator):
...
+        # Validation: every template schema must declare this LinkML schema.
+        if "components" in self._template and "schemas" in self._template["components"]:
+            for name, schema in self._template["components"]["schemas"].items():
+                if schema["x-linkml-schema"] != self.schemaview.schema.id:
```
noelmcloughlin, line 117: If `x-linkml-schema` is absent then python will throw error/exception.  Maybe a try/except block is warranted.

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -141,136 +207,166 @@ def _fix_openapi_spec(self, element: dict | list) -> dict | list | None:
...
-                self._find_references(class_schema, referenced_classes)
-        return class_schemas
+    def _find_schemas_line(self, template_text: str) -> int:
+        """Return the 0-indexed line number of the ``schemas`` key under ``components``."""
+        doc = yaml.compose(template_text)
+        if not isinstance(doc, MappingNode):
+            raise ValueError("OpenAPI template is not a YAML mapping")
+        components_node = None
+        for key, value in doc.value:
+            if isinstance(key, ScalarNode) and key.value == "components":
```
noelmcloughlin, line 279: Any need for `key.value.trim()` or is it already valid yaml?
Same question for `key.value == "schemas"` below.

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -284,12 +380,15 @@ def printout_template(self) -> str:
 def cli(yamlfile, template, **args):
     """Generate an OpenAPI v3.0.3 spec with resources modelled with LinkML.
```
noelmcloughlin, line 385: Just curious about v3.0.3 as mentioned in earlier comments - its very specific - why not `v3.0.4` or `v3.0.n`?

tests/linkml/test_generators/input/openapi/spec-fixed.openapi.yaml
```diff
@@ -0,0 +1,22 @@
+openapi: 3.0.3
```
noelmcloughlin, line 3: Does name mean template was broken so is now fixed (my first instinct) or template contains fixed content (my second instinct)?  Could you add a comment to top of the file so  users can understand context.  

Same comment on other two input files.

noelmcloughlin (changes requested):
The OpenAPI Specification is large so I spent some time reading documentation - a few chapters to become acquainted with the technical details.

Here is my review of the changeset - I have not reviewed with LLM.
The intent looks good and the test coverage appears to be sufficient.
I had various queries based on visual inspection of the code changeset.
Once PR is merged I hope to test the feature.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: It is not strictly needed, but I've pinned to that version due to two reasons:
- that's a version being used on a project I'm working on and I can therefore test in my project that it works
- none of the previous 3.0.x versions really got widely adopted as 3.0.3 did and therefore I doubt much users will feel restricted, we can lax that versioning on a later patch once we get better experience on 3.0.x (where `x =/= 3`.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: What I was sure about is the need to remove the term `class` to all the parts referring to the OpenAPI part, because that wording is not use in that context at all. Looking for some better wording, I noticed that they talk about *data*  and *data types*  and that the `components/schema` section refers to the schema of those data. That's why I thought that `data schema` would be the best fitting name. That wording change is probably only relevant for the PR, I can update the description accordingly.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: This is a bridge between the OpenAPI and the LinkML contexts. Since the wording differ between them, there is no wording that fits both. Since `{data_schema}` is the best fitting variable name in the OpenAPI `endpoints` , that's the name being used. I personally don't see a better fitting name.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: `_validate_oa_template` validates the OpenAPI Description template , therefore replacing to `_validate_oad_template`.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: Same as above, renaming to `_oad_`.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: No need, the template must be valid YAML or JSON (which is valid YAML).

That's exactly one of the intentions of this PR, that the given template MUST be valid OpenAPI (and therefore implicitly valid YAML/JSON) and therefore gets validated before the commented code gets executed.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: Same comment as above. Most widely used `3.0.x` version by far and only one tested on a real API.

tests/linkml/test_generators/input/openapi/spec-fixed.openapi.yaml
Silvanoc: Good point. Done.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: Good point. Done.

packages/linkml/src/linkml/generators/openapigen.py
noelmcloughlin: Okay, thanks for the rationale. Can be considered in future.

packages/linkml/src/linkml/generators/openapigen.py
noelmcloughlin: I can't find `data_schema` in OpenApi specification so I presumed it is a variable name we "own" and refers to the linkml schema.  If you think its not inconsistent thats fine.

packages/linkml/src/linkml/generators/openapigen.py
noelmcloughlin: Apologies for my loose suggestion - I realize `try/except` is wrong pattern in the generators (also `print` streams to stdout mixing diagnostics with generated output, and a KeyError).  I should have suggested this

```
for name, schema in self._template["components"]["schemas"].items():
    if "x-linkml-schema" not in schema:
        raise ValueError(f"Template data schema '{name}' is missing required 'x-linkml-schema'")
    if schema["x-linkml-schema"] != self.schemaview.schema.id:
        raise ValueError(
        ...
```

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: In the generation of an OAD template this variable is used to get the same name in an API endpoint referring to a `components/schema`. Both called `data_schema`, which is fine in the OpenAPI context. Then I use that very same name for the annotation referring to the LinkML class in order to avoid having renaming (OpenAPI data and LinkML class referring to the same object are called different). That is the moment where `data_schema` in the LinkML feels an alien. But it only signals that it's exactly the same name.

If OpenAPI and LinkML where to use different names for the same object, then it wouldn't be `data_schema` anymore.

I hope it clarifies it. IMO this shouldn't be a blocker for this fix. If you think that it should be changed, we can add an issue for it and address it later. I haven't had time to fully read #3871 and perhaps that the intention of that issue.

In any case I would expect you to resolve the conversation if you agree that it's not a blocker. That applies in general to all other open conversations. I only resolve them when I'm quite sure that they've been addressed.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: I didn't thought it through either, so nothing to apologize for. It should be fixed now.

packages/linkml/src/linkml/generators/openapigen.py
noelmcloughlin: Thanks for explaining

noelmcloughlin (approved):
LGTM, 

@matentzn please priortize as openapi is important feature

packages/linkml/src/linkml/generators/openapigen.py
noelmcloughlin: Hi, sorry no #3871 is not related to this comment.  This was a potential *nit* as I was just wondering why variable was not called `linkml_data_schema`  (instead of data_schema) to match `linkml_schema_id` - i.e. same `linkml_` prefix.  But its not important, just a review question.

Silvanoc:
@matentzn since I'm unsure if the additions of commit https://github.com/linkml/linkml/commit/ee91123527efe416f4f42daf49acb93515404086 are already applicable or not, I've assigned you this PR to avoid ignoring that rule just due to a misunderstanding.

noelmcloughlin:
I can merge but flaky tests seem to be causing failure (https://github.com/linkml/linkml/issues/3421).
 I don's see a "restart failed jobs" button.
 
 @Silvanoc could you do some edit and repush to see if jobs restart?

noelmcloughlin:
Nevermind - I see it now - rerunning failing jobs

PR #3769, opened 2026-07-15
ci: scope CVE audit to PRs that change dependencies

### LINKED ISSUE #3767: Make `uv audit` non-blocking in CI, or prevent it from spuriously failing unrelated PRs
The `dependency-audit.yaml` workflow runs `uv audit` as a _blocking_ step on every PR. When a new advisory is published for any dependency in `uv.lock`, all open PRs go red immediately, regardless of whether the PR touches dependencies at all. That's what's happening right now: PR #3766 (a one-line docs edit) is blocked by 6 newly-published CVEs in `click` and `pillow` that have nothing to do with it.

## Examples

Run [29372484892](https://github.com/linkml/linkml/actions/runs/29372484892/job/87218695887?pr=3766) on PR #3766 (update to `Community-Meetings.md`) failed at the "Audit lockfile for CVEs" step with:

```
Found 6 known vulnerabilities and no adverse project statuses in 256 packages

Vulnerabilities:

click 8.3.1 has 1 known vulnerability:
- PYSEC-2026-2132 (fixed in 8.3.3)

pillow 12.2.0 has 5 known vulnerabilities:
- PYSEC-2026-2253 through PYSEC-2026-2257 (all fixed in 12.3.0)

##[error]Process completed with exit code 1.
```

The step in `.github/workflows/dependency-audit.yaml` is:

```yaml
- name: Audit lockfile for CVEs
  run: uv audit --ignore GHSA-6w46-j5rx-g56g
```

No `continue-on-error: true`. No scoping to dependency-touching PRs. Result: every PR is a hostage to the upstream advisory database.

## Related

- #3766 — the docs PR currently blocked
- #3650 (PR) — introduced `dependency-audit.yaml`; already has one `--ignore` flag for GHSA-6w46-j5rx-g56g to work around a similar situation
- #3623 — broader supply-chain hardening context

amc-corey-cox:
## Problem

The `Audit lockfile for CVEs` step in `dependency-audit.yaml` reflects the state of the *upstream advisory database*, not the change under test. When a new advisory is published against an already-pinned package, **every open PR goes red at once** — and so does the next innocent merge to `main` — even when nothing touched dependencies. Right now 6 newly-published CVEs in `click` and `pillow` are failing unrelated PRs (e.g. #2602, #3766).

## Change

Gate the CVE audit on whether the change **actually altered dependencies**, diffed against each event's natural base:

- `pull_request` → the PR base
- `push` (main) → the commit before the push (`github.event.before`)
- `merge_group` → the queue base
- otherwise (`workflow_dispatch`, first/force push) → audit

A `pyproject.toml` change always counts as a real dependency change. `uv.lock` is regenerated non-deterministically (reordering, hashes, metadata), so a textual change there is only treated as real when the resolved `(name, version)` set actually differs — computed by a small stdlib-`tomllib` helper (`.github/scripts/uv_lock_deps_changed.py`). No third-party action added to the supply-chain workflow.

Implemented at the **step** level, not as an `on.pull_request.paths` filter: a path-filtered required workflow reports nothing on non-matching PRs and leaves the required check stuck *pending*. Here the job always runs and reports green; only the audit step is skipped.

Net effect: `main` can only go red from a merge that genuinely introduced a vulnerable dependency (which is a broken merge), never from an unrelated advisory. The malware `uv sync` anti-malware gate is unchanged and still hard-blocking on every event.

Closes #3767

### DIFF
--- a/.github/scripts/uv_lock_deps_changed.py
+++ b/.github/scripts/uv_lock_deps_changed.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+"""Decide whether the resolved ``uv.lock`` dependency set changed between refs.
+
+``uv.lock`` is regenerated non-deterministically: reordering, hashes, and
+metadata can differ between two lockfiles that resolve to exactly the same
+packages. ``uv audit`` only cares about the multiset of ``(name, version)``
+pairs, so this script compares that set between a base ref and ``HEAD`` and
+prints ``true`` only when it actually differs.
+
+Usage:
+    uv_lock_deps_changed.py <base_ref>
+
+Prints ``true`` when the resolved ``(name, version)`` set at ``HEAD`` differs
+from the one at ``<base_ref>`` (or when the lockfile is absent at either ref),
+otherwise ``false``.
+"""
+
+from __future__ import annotations
+
+import subprocess
+import sys
+import tomllib
+
+
+def package_set(ref: str) -> set[tuple[str, str | None]] | None:
+    """Return the ``{(name, version)}`` set from ``uv.lock`` at ``ref``.
+
+    Args:
+        ref: A git ref (SHA, branch, ``HEAD``) to read ``uv.lock`` from.
+
+    Returns:
+        The set of ``(name, version)`` tuples for every locked package, or
+        ``None`` if ``uv.lock`` does not exist at ``ref``.
+    """
+    result = subprocess.run(
+        ["git", "show", f"{ref}:uv.lock"],
+        capture_output=True,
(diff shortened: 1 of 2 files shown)


kevinschaper (approved):
Nice fix!

matentzn (approved):
What happened to adding `paths` to the action as a parameter, haha. I am ok with this but its a thorough and complex solution which catches many unnecessary corner cases compared to a super simple one where you simply sat: run the audit action ONLY when toml / uv.lock are touched.

Issue #3771, opened 2026-07-16
Several SchemaView generators still inherit uses_schemaloader = True


amc-corey-cox:
`rustgen`, `dbmlgen`, and `yarrrmlgen` are written against SchemaView — they build `self.schemaview` and use no visitor methods — but never set `uses_schemaloader = False`, so the base `__post_init__` takes the legacy `SchemaLoader` branch and runs a redundant `resolve()`. Beyond the wasted load, `self.schema` (set by the loader) can drift out of sync with `self.schemaview.schema`.

`yarrrmlgen` additionally overrides `__init__` rather than `__post_init__`, bypassing the dataclass init contract.

These should take the SchemaView init path they're actually written for.

Issue #3770, opened 2026-07-16
gen-sparql mutates the input schema in place


amc-corey-cox:
`SparqlGenerator.materialize_schema()` modifies the live `SchemaView`/`SchemaDefinition` instead of a copy — it moves attributes into `schema.slots` and then deletes `class.attributes` on the shared object (`sparqlgen.py:107-123`). Run standalone this is invisible, but under `gen-project` (or anything that reuses the `SchemaView`), a generator running after sparql receives a schema with its attributes already stripped.

The generators that transform the schema — docgen, openapigen, rdfgen, jsonschemagen — all deep-copy first; `gen-sparql` should do the same before materializing.

amc-corey-cox:
Digging into the fix, there's a second bug from the same code that changes what the fix should be.

Because `SparqlGenerator` sets `uses_schemaloader = False`, the base `__post_init__` **rebuilds `self.schemaview` from scratch** (`generator.py:211`) *after* `materialize_schema` has run:

```python
def __post_init__(self):
    self.schemaview = SchemaView(self.schema)   # build + materialize into THIS view
    materialize_schema(self.schemaview)
    super().__post_init__()                     # base does self.schemaview = SchemaView(schema, ...) again
```

So the behavior splits by input type:

- **Object input** — the rebuilt view happens to wrap the same mutated `SchemaDefinition`, so materialization survives *and* corrupts the caller's schema (the mutation reported here).
- **Path input** — the rebuilt view is a fresh load, so the materialization is thrown away entirely. Its one surviving effect — copying imported prefixes onto the schema — is lost, and the generated SPARQL references `rdf:`, `core:`, `prov:` prefixes that are never declared. That's silently invalid SPARQL today.

Repro on `kitchen_sink.yaml`: path-input and object-input outputs differ only in the `PREFIX` block, and the path-input output uses `rdf:`/`core:`/`prov:` in query bodies without declaring them.

So the fix isn't just "deep-copy first" — it needs to deep-copy the input *and* materialize into the view the base actually keeps (or drop the redundant early `SchemaView` + reorder), which resolves both the mutation and the missing-prefix breakage together.

Issue #3633, opened 2026-06-10
ValueError in DBMLGenerator when target class lacks id slot


noelmcloughlin:
**Describe the bug**

 `DBMLGenerator._generate_relationships()` raises `ValueError` for any reference whose target class has no identifier slot.

Note: reduced scope due to duplicate issue mentioned in comments below.

**Minimal repro**

```bash
mkdir -p /tmp/dbml-noid && cd /tmp/dbml-noid

cat > schema.yaml << 'EOF'
id: https://example.org/noid
name: noid
prefixes:
  linkml: https://w3id.org/linkml/
imports:
  - linkml:types
default_range: string
classes:
  Address:
    # No identifier slot.
    attributes:
      city:
      country:
  Person:
    attributes:
      id:
        identifier: true
      address:
        range: Address
EOF

uv run gen-dbml schema.yaml
# ValueError: Class Address has no identifier slot
```

**Version of LinkML you are using**
1.11.1

amc-corey-cox:
Heads up: #3778 fixes **bug 1** here. `DBMLGenerator` was inheriting `uses_schemaloader = True` and then rebuilding its own `SchemaView`, which dropped `base_dir` and resolved relative imports against the cwd. Setting `uses_schemaloader = False` lets the base build the view with the right base directory, and I added a regression test that runs `gen-dbml` from an unrelated cwd against a schema with relative imports.

**Bug 2** (`_generate_relationships` raising `ValueError` for references to identifier-less classes) is untouched in that PR — I left it for you since you're assigned. Happy to help there too if useful.

Separately, while fixing the init path I found a third issue: the generator computed induced slots but re-fetched the global slot via `get_slot()`, so per-class `slot_usage` (e.g. `required: true`) was silently dropped. That's fixed in the same PR.

noelmcloughlin:
@amc-corey-cox  I think I'm starting to understand better - `Schemaloader` is legacy and generators should move towards `Schemaview` aka `SV`. I was puzzled why `uses_schemaloader = True` is repeated in various child generators when super explicitly sets "True" - must be part of documenting #923 progress ..
https://github.com/linkml/linkml/blob/1288dbb6e9a6df8d1f04e250a1fb48af635bd530/packages/linkml/src/linkml/utils/generator.py#L100-L101

Seems  `uses_schemaloader = False` was direction needed - make more sense now.  I'll reduce scope of my Issue/PR as you suggest.

Btw, are some schema not "meta-model compliant"?"
[...]

PR #3778, opened 2026-07-17
Fix SchemaView initialization across sparql/rust/dbml/yarrrml generators

amc-corey-cox:
Fixes a cluster of related SchemaView-initialization bugs surfaced by the generators audit (#3770, #3771, and bug 1 of #3633).

## sparqlgen — mutates the input schema (#3770)
`materialize_schema()` rewrote the live `SchemaView` in place (moving attributes into slots, deleting `class.attributes`), so under `gen-project` any generator running after sparql received a schema with its attributes already stripped. It also ran *before* the base `__post_init__` rebuilt the view, so **path inputs silently lost the materialized prefixes** and produced SPARQL referencing undeclared `rdf:`/`core:`/`prov:` prefixes — invalid output, depending only on whether the caller passed a path or a pre-parsed schema.

Fix: deep-copy the whole view after base init, then materialize the copy. Copying the view rather than just the schema preserves import-resolution context.

## rust / dbml / yarrrml — wrong init path (#3771)
All three are written against `SchemaView` but never set `uses_schemaloader = False`, so the base took the legacy `SchemaLoader` branch and each generator then discarded and rebuilt the view. The rebuilt view dropped `base_dir`, so **dbmlgen resolved relative imports against the process cwd** instead of the schema directory — this is bug 1 of #3633. `yarrrmlgen` additionally overrode `__init__` instead of `__post_init__`, bypassing the dataclass contract.

Fix: set `uses_schemaloader = False` and let the base build the view; convert `yarrrmlgen` to a dataclass with `source` / `iterator_template` fields.

## dbmlgen — drops slot_usage overrides
Switching dbmlgen to the SchemaView path surfaced a latent bug: it computed induced slots but then re-fetched the *global* slot via `get_slot(slot_name.name)`, discarding per-class `slot_usage` (e.g. `required: true`). The old path masked this by baking slot_usage into mangled global slots. Fixed by using the induced slot directly.

## Tests
- sparqlgen: added regression tests that the caller's schema is not mutated and that path vs. pre-parsed-object inputs produce identical output.
- dbmlgen: added regression tests that `slot_usage: required` reaches the column and that relative imports resolve from the schema directory when run from an unrelated cwd.
- yarrrml suite (27 tests, requires `morph_kgc`) passes; projectgen integration passes.

Note: #3633 also reports a second, independent bug (`_generate_relationships` raising `ValueError` for references to identifier-less classes) that is **not** addressed here, so that issue is referenced rather than closed.


noelmcloughlin (approved):
LGTM (see also https://github.com/linkml/linkml/issues/3633#issuecomment-5006829937).

kevinschaper (approved):
looks good!

Issue #3780, opened 2026-07-20
Make java/golang package configurable via config.yaml


noelmcloughlin:
**What is your feature request?**

The Java Generator allows passing `package` name via CLI (see https://github.com/linkml/linkml/blob/main/docs/generators/java.rst)

```
gen-java --package org.biolink.model --output-directory org/biolink/model biolink-model.yaml
```

It would be nice to configure java package name via `config.yaml`
~~`schema-first` mechanism, i.e. annotations.java_package = "org.example.bob"~~

And generalize for Golang per #3781 PR comment

**How important is this feature to you?** Select from the options below:

• High

Issue #2537, opened 2025-02-10
The project generator does not know about the pydantic and java generators


dalito:
**Describe the bug**

Neither the pydantic nor the java generator can be run as part of the project generator `gen-project`. Both are just missing in [projectgen.py](https://github.com/linkml/linkml/blob/main/linkml/generators/projectgen.py).

**Version of LinkML you are using**

v1.8.6

PR #3781, opened 2026-07-20
feat(javagen): configure java/golang package name via config.yaml / gen-project integration

noelmcloughlin:
## Summary

JAVA - #3780
- Option to set Java `package` value from `config.yaml`
- Adds javagen support to `projectgen` per #2537
- Adds three missing Java keywords from https://docs.oracle.com/cd/E19316-01/819-3669/bnbuk/index.html 

GOLANG
- Add missing doc
- Generalize the solution (per https://github.com/linkml/linkml/pull/3781#pullrequestreview-4737743023)

Excel #3759
- `excelgen.py` - fix `serialize` return signature.
- `projectgen.py` - remove redundant excel serialization

## Full list of changes following PR review

Added

- --config-file/-C on gen-java and gen-golang, reading package from generator_args.{java,golang} - the same configuraion file style gen-project takes.
- gen-project now emits Java.
- Invalid package names are rejected instead of emitted into uncompilable output.
- Docs - created golang.rst, updated javagen.rst

Changed

- gen-golang uses --package / package= to match gen-java. --package-name / package_name= still work, deprecated.

Fixed
- gen-project ran the Excel generator twice.
- A directory set for one generator could be silently dropped from another's config.
- JAVA_KEYWORDS was missing true/false/null.
- Bad config files and generator_args values now give usage errors, not tracebacks.
- Go's derived fallback package is always a legal identifier.

## How was this tested?

Locally (slow, network)
added new  tests

## Checklist

## AI Assistance
LLM assisted

### DIFF
--- a/docs/generators/java.rst
+++ b/docs/generators/java.rst
@@ -102,6 +102,9 @@ The generated Java ``package`` statement is driven by the following precedence:
 2. ``generator_args.java.package`` set via ``--config-file``/``-C`` (see :ref:`java-configuration-file` below)
 3. Fallback default: ``example``
 
+The package name is validated wherever it comes from. An invalid name (``1bad.pkg``,
+``org.class.model``) is reported as an error.
+
 NOTE:
     The package name is *not* configurable via schema-level annotations: LinkML
     classes, slots, and enums live in a single global namespace regardless of
--- a/docs/generators/java.rst
+++ b/docs/generators/java.rst
@@ -94,6 +94,56 @@ expected to be serialised in the JSON or YAML serialisations; the
 ``--use-aliases`` option only affects the symbol used to represent the
 slot in the Java code.
 
+Package Configuration
+---------------------
+The generated Java ``package`` statement is driven by the following precedence:
+
+1. ``--package`` command-line option (or ``package=...`` when using ``JavaGenerator`` programmatically)
+2. ``generator_args.java.package`` set via ``--config-file``/``-C`` (see :ref:`java-configuration-file` below)
+3. Fallback default: ``example``
+
+NOTE:
+    The package name is *not* configurable via schema-level annotations: LinkML
+    classes, slots, and enums live in a single global namespace regardless of
+    which (sub-)schema declares them - the package is a global configuration for
+    the generator, not a property of the model.
+
+
+The package applies to every generated class and enum: ``gen-java`` emits one file per
+class into a single flat output directory, and cross-class references rely on all
+generated types sharing one package.
+
+.. _java-configuration-file:
+
+Configuration File
+-------------------
+
+As an alternative to the ``--package`` argument, ``gen-java`` accepts a
+``--config-file``/``-C`` YAML file -- the **same format** used by
+``gen-project``'s own ``--config-file`` (see :doc:`project-generator`), so a single
+project-wide ``config.yaml`` can be shared between ``gen-project`` and a
+standalone ``gen-java`` run. ``package`` lives under ``generator_args.java``,
+since that file is structured to configure every generator at once:
+
+.. code-block:: yaml
+
+    # config.yaml
+    generator_args:
+      java:
+        package: org.example.model
+
+``gen-java`` only ever reads ``generator_args.java.package`` out of this file --
+every other key (``directory``, ``excludes``, other generators' ``generator_args``
+entries, etc.) is ignored, so a full multi-generator project ``config.yaml`` can be
+passed as-is without modification.
+
+An explicit ``--package`` command-line option always overrides a value set via
+``--config-file``.
+
+The same configuration-file mechanism is supported by ``gen-golang`` (see
+:doc:`golang`), which reads ``generator_args.golang.package`` from the same file.
+
+
 Generating Visitor Patterns
 ---------------------------
 
--- a/packages/linkml/src/linkml/generators/excelgen.py
+++ b/packages/linkml/src/linkml/generators/excelgen.py
@@ -171,7 +171,7 @@ class ExcelGenerator(Generator):
 
         dv.add(f"{column_letter}2:{column_letter}1048576")
 
-    def serialize(self, **kwargs) -> str:
+    def serialize(self, **kwargs) -> None:
         sv = self.schemaview
         all_classes = sv.all_classes(imports=self.mergeimports)
 
--- a/packages/linkml/src/linkml/generators/golanggen/golanggen.py
+++ b/packages/linkml/src/linkml/generators/golanggen/golanggen.py
@@ -26,7 +26,7 @@ from linkml.generators.golanggen.template import (
 )
 from linkml.generators.oocodegen import OOCodeGenerator
 from linkml.utils.deprecation import deprecation_warning
-from linkml.utils.generator import read_generator_config_arg, shared_arguments
+from linkml.utils.generator import read_generator_config, shared_arguments
 from linkml_runtime.linkml_model.meta import ClassDefinition, EnumDefinition, SlotDefinition
 from linkml_runtime.utils.formatutils import camelcase, underscore
 from linkml_runtime.utils.schemaview import SchemaView
@@ -696,7 +696,7 @@ def cli(
         if package is None:
             package = package_name
     if package is None:
-        package = read_generator_config_arg(config_file, "golang", "package")
+        package = read_generator_config(config_file, "golang").get("package")
 
     if template_dir is not None:
         if not Path(template_dir).exists():
--- a/packages/linkml/src/linkml/generators/golanggen/golanggen.py
+++ b/packages/linkml/src/linkml/generators/golanggen/golanggen.py
@@ -25,7 +25,8 @@ from linkml.generators.golanggen.template import (
     Imports,
 )
 from linkml.generators.oocodegen import OOCodeGenerator
-from linkml.utils.generator import shared_arguments
+from linkml.utils.deprecation import deprecation_warning
+from linkml.utils.generator import read_generator_config_arg, shared_arguments
 from linkml_runtime.linkml_model.meta import ClassDefinition, EnumDefinition, SlotDefinition
 from linkml_runtime.utils.formatutils import camelcase, underscore
 from linkml_runtime.utils.schemaview import SchemaView
@@ -614,9 +615,25 @@ _TEMPLATE_NAMES = [
 
 @shared_arguments(GolangGenerator)
 @click.option(
-    "--package-name",
+    "--package",
+    "package",
     type=str,
     help="Override the Go package name (default: derived from schema name)",
+)
+@click.option(
+    "--package-name",
+    "package_name",
+    type=str,
+    help="DEPRECATED alias for --package",
+)
+@click.option(
+    "--config-file",
+    "-C",
+    type=click.File("rb"),
+    help="Path to a gen-project-style YAML config file setting "
+    "'generator_args: {golang: {package: ...}}'. An explicit --package always "
+    "takes precedence over the config file. If not given, a 'config.yaml' in the current "
+    "working directory is used automatically if present.",
 )
 @click.option(
     "--alphabetical-sort/--no-alphabetical-sort",
@@ -657,7 +674,9 @@ Available templates to override:
 @click.command(name="golang")
 def cli(
     yamlfile,
+    package: str | None = None,
     package_name: str | None = None,
+    config_file=None,
     alphabetical_sort: bool = False,
     nullable_primitives: bool = True,
     named_slot_types: bool = False,
@@ -673,13 +692,20 @@ def cli(
     - JSON tags for serialization
     - Struct embedding for inheritance
     """
+    if package_name is not None:
+        deprecation_warning("golanggen-package-name-option")
+        if package is None:
+            package = package_name
+    if package is None:
+        package = read_generator_config_arg(config_file, "golang", "package")
+
     if template_dir is not None:
         if not Path(template_dir).exists():
             raise FileNotFoundError(f"The template directory {template_dir} does not exist!")
 
     gen = GolangGenerator(
         yamlfile,
-        package_name=package_name,
+        package_name=package,
         alphabetical_sort=alphabetical_sort,
         nullable_primitives=nullable_primitives,
         named_slot_types=named_slot_types,
--- a/packages/linkml/src/linkml/generators/golanggen/golanggen.py
+++ b/packages/linkml/src/linkml/generators/golanggen/golanggen.py
@@ -7,8 +7,10 @@ Based on the PydanticGenerator architecture.
 import logging
 import os
 import re
+from collections.abc import Mapping
 from dataclasses import dataclass, field
 from pathlib import Path
+from typing import Any
 
 import click
 from jinja2 import ChoiceLoader, Environment, FileSystemLoader
@@ -24,8 +26,8 @@ from linkml.generators.golanggen.template import (
     Import,
     Imports,
 )
-from linkml.generators.oocodegen import OOCodeGenerator
-from linkml.utils.deprecation import deprecation_warning
+from linkml.generators.oocodegen import PACKAGE, OOCodeGenerator
+from linkml.utils.deprecation import deprecated_fields, deprecation_warning
 from linkml.utils.generator import read_generator_config, shared_arguments
 from linkml_runtime.linkml_model.meta import ClassDefinition, EnumDefinition, SlotDefinition
 from linkml_runtime.utils.formatutils import camelcase, underscore
@@ -71,7 +73,81 @@ GO_PRIMITIVE_TYPES = frozenset(
     }
 )
 
+# Fallback package when a schema name yields no usable identifier,
+# mirroring JavaGenerator's "example" default.
+DEFAULT_GO_PACKAGE = "example"
 
+# The 25 reserved words that may not be used as identifiers.
+# https://go.dev/ref/spec#Keywords
+GO_KEYWORDS = frozenset(
+    {
+        "break",
+        "case",
+        "chan",
+        "const",
+        "continue",
+        "default",
+        "defer",
+        "else",
+        "fallthrough",
+        "for",
+        "func",
+        "go",
+        "goto",
+        "if",
+        "import",
+        "interface",
+        "map",
+        "package",
+        "range",
+        "return",
+        "select",
+        "struct",
+        "switch",
+        "type",
+        "var",
+    }
+)
+
+
+def _is_valid_go_package(package_name: str) -> bool:
+    """Return True if ``package_name`` is a legal Go package identifier.
+
+    Unlike Java, a Go package clause (``package foo``) names a single
+    identifier, not a dotted path. It must start with a letter or
+    underscore, contain only letters/digits/underscores, and must not be a
+    reserved keyword. See specs for grammar/identifier rules:
+    https://go.dev/ref/spec#Package_clause
+    https://go.dev/ref/spec#Identifiers
+    """
+    if not package_name or not (package_name[0].isalpha() or package_name[0] == "_"):
+        return False
+    if any(not (ch.isalnum() or ch == "_") for ch in package_name):
+        return False
+    if package_name in GO_KEYWORDS:
+        return False
+    return True
+
+
+def _derive_go_package(schema_name: str) -> str:
+    """Derive a legal Go package identifier from a schema name.
+
+    Mirrors ``JavaGenerator``'s guarantee that the fallback package is always
+    valid: a schema name that sanitises to nothing (e.g. ``_private``) falls
+    back to :data:`DEFAULT_GO_PACKAGE`, and one that collides with a reserved
+    word is suffixed with ``_``, the same escape ``map_name`` applies to
+    keyword-named fields.
+    """
+    # Take the leading segment (e.g. "kitchen_sink" -> "kitchen")
+    stem = schema_name.split("_", 1)[0]
+    # Keep only characters legal in a Go package name (lowercase, alphanumeric, underscore)
+    candidate = re.sub(r"[^a-z0-9_]", "", stem.lower())
+    if candidate in GO_KEYWORDS:
+        candidate += "_"
+    return candidate if _is_valid_go_package(candidate) else DEFAULT_GO_PACKAGE
+
+
+@deprecated_fields({"package_name": "package"}, "golanggen-package-name-option")
 @dataclass
 class GolangGenerator(OOCodeGenerator):
     """
@@ -94,8 +170,13 @@ class GolangGenerator(OOCodeGenerator):
     file_extension = "go"
 
     # ObjectVars
-    package_name: str | None = None
-    """Override the package name. If None, derived from schema name."""
+    package: PACKAGE | None = None
+    """Override the Go package name. If unset, derived from the schema name.
+
+    Overrides :class:`~linkml.generators.oocodegen.OOCodeGenerator`'s ``"example"``
+    default so that an unset package can be told apart from an explicit one and
+    derived from the schema name instead. ``package_name`` is a deprecated alias.
+    """
 
     gen_slots: bool = True
     """Generate struct fields"""
@@ -167,6 +248,19 @@ class GolangGenerator(OOCodeGenerator):
 
     def __post_init__(self):
         super().__post_init__()
+        if self.package in (None, ""):
+            self.package = _derive_go_package(self.schemaview.schema.name)
+        else:
+            # guards against a non-string value, e.g. an unquoted number in config.yaml
+            self.package = str(self.package)
+        if not _is_valid_go_package(self.package):
+            raise ValueError(f"{self.package!r} is not a valid Go package name")
+
+    @classmethod
+    def validate_generator_args(cls, args: Mapping[str, Any]) -> None:
+        package = args.get("package")
+        if package not in (None, "") and not _is_valid_go_package(str(package)):
+            raise click.UsageError(f"{package!r} is not a valid Go package name")
 
     def default_value_for_type(self, typ: str) -> str:
         """
@@ -515,18 +609,6 @@ class GolangGenerator(OOCodeGenerator):
                 if sv.slot_children(slot_name):
                     self._parent_slot_names.add(slot_name)
 
-        # Determine package name
-        package_name = self.package_name
-        if package_name is None:
-            schema_name = sv.schema.name
-            # Extract package name from schema name (e.g., "kitchen_sink" -> "kitchen")
-            if "_" in schema_name:
-                package_name = schema_name[: schema_name.find("_")]
-            else:
-                package_name = schema_name
-            # Make it valid Go package name (lowercase, alphanumeric + underscore)
-            package_name = re.sub(r"[^a-z0-9_]", "", package_name.lower())
-
         # Generate enums
         enums = {}
         for enum_def in sv.all_enums().values():
@@ -566,7 +648,7 @@ class GolangGenerator(OOCodeGenerator):
             imports.sort()
 
         module = GolangModule(
-            package_name=package_name,
+            package_name=self.package,
             imports=imports,
             enums=enums,
             structs=structs,
@@ -697,14 +779,19 @@ def cli(
             package = package_name
     if package is None:
         package = read_generator_config(config_file, "golang").get("package")
+    GolangGenerator.validate_generator_args({"package": package})
 
     if template_dir is not None:
         if not Path(template_dir).exists():
             raise FileNotFoundError(f"The template directory {template_dir} does not exist!")
 
+    # The package was already validated above; a ValueError raised while actually
+    # constructing the generator here is therefore a genuine failure (e.g. an
+    # unloadable schema), not a bad package name, and is left to propagate with
+    # its own traceback rather than being caught and mistaken for one.
     gen = GolangGenerator(
         yamlfile,
-        package_name=package,
+        package=package,
         alphabetical_sort=alphabetical_sort,
         nullable_primitives=nullable_primitives,
         named_slot_types=named_slot_types,
--- a/packages/linkml/src/linkml/generators/golanggen/golanggen.py
+++ b/packages/linkml/src/linkml/generators/golanggen/golanggen.py
@@ -632,8 +632,7 @@ _TEMPLATE_NAMES = [
     type=click.File("rb"),
     help="Path to a gen-project-style YAML config file setting "
     "'generator_args: {golang: {package: ...}}'. An explicit --package always "
-    "takes precedence over the config file. If not given, a 'config.yaml' in the current "
-    "working directory is used automatically if present.",
+    "takes precedence over the config file.",
 )
 @click.option(
     "--alphabetical-sort/--no-alphabetical-sort",
--- a/packages/linkml/src/linkml/generators/javagen.py
+++ b/packages/linkml/src/linkml/generators/javagen.py
@@ -534,8 +534,8 @@ def cli(
     """Generate java classes to represent a LinkML model"""
     if package is None:
         package = read_generator_config(config_file, "java").get("package")
-        if package is not None and not _is_valid_java_package(str(package)):
-            logging.warning("Config value %r is not a valid Java package name; using it as-is.", package)
+    if package is not None and not _is_valid_java_package(str(package)):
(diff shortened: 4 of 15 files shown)


docs/generators/java.rst
```diff
@@ -32,6 +32,27 @@ Additional Notes
...
+
+.. code-block:: yaml
+
+        annotations:
+            java_package: org.example.model
+
+The package applies to every generated class and enum: ``gen-java`` emits one file per
+class into a single flat output directory, and cross-class references rely on all
+generated types sharing one package. If the annotation value is not a syntactically
+valid Java package name, a warning is logged but the value is still used.
```
gouttegd: I would like to see a small clarification that only the `java_package` annotation **from the top-level schema** is taken into any account; any `java_package` annotation in an _imported schema_ is ignored.

Consider the following scenario:

Alice creates and publishes the following schema `foo.yaml`:

```yaml
id: https://example.org/foo
name: foo
imports:
  - linkml:types
annotations:
  java_package: org.example.foo
classes:
  Thing:
    attributes:
      id:
        range: string
```

Then Bob creates its own schema `bar` in which it reuses (imports) Alice’s `foo` schema:

```yaml
id: https://example.org/bar
name: bar
imports:
  - linkml:types
  - foo # <- Importing schema foo.yaml
annotations:
  java_package: org.example.bar
classes:
  AnotherThing:
    ...
```

Bob could naively expect that, when running the Java generator over its `bar.yaml` schema, the `Thing` class (coming from the importing schema `foo`) would end up in the `org.example.foo` package while the `AnotherThing` class would end up in the `org.example.bar` package.

That is of course _not_ the case: both classes would end up in the `org.example.bar` package, as they should. I think the documentation should make that clear, to avoid any misunderstanding.

gouttegd (changes requested):
Overall this looks good to me.

The requested change is only a documentation issue, see the inline comment.

Apart from that, I have two “higher-level” comments (which I don’t consider blocking, just food for thought):

(A) For what it’s worth, I believe the Java package should _not_ be set in the schema. The same schema can be used to generate code for several different software projects (unless it’s a schema that is only ever intended to be used internally by one project), and each project will likely want to have its own namespace. That namespace should not be decided by the schema’s author, IMHO.

But as long as the `--package` option takes precedence (allowing whoever runs the code generator to specify its own namespace, notwithstanding what the schema says), I have no objection to this feature.

(B) I wonder if we should maybe coordinate with the generators for other package-scoped languages for better consistency. For example, the Go generator has a `--package-name` option (which is already different from the option name used by the Java generator, so inter-generator consistency is already missing here…). When that option is not specified, the Go generator derives the package name from the schema name.

docs/generators/java.rst
noelmcloughlin: I'm not Java developer so I reviewed [packages](https://dev.java/learn/packages) and few example java files.  

Just so I understand.

- By convention, the top-model (linkml importing/parent schema) should be authoritative owner of `package name`.

- If the top-level schema author and sub-schema author are same person then we expect a common `package name` (i.e. `org.example.foo` in github.com/foo/ organization) across all java files.

- If the sub-schema had **different author (i.e. `org.example.bar`** from github.com/bar/) then, by convention, we always override the foreign package name with native our name (`org.example.foo`).

This is normal design pattern - and should be forced by linkml?

noelmcloughlin:
> maybe coordinate with the generators for other package-scoped languages for better consistency. For example, the Go generator has a `--package-name` option

I will have a look, see what inventory looks like.

> the Java package should _not_ be set in the schema.

I'm can be easily convinced. Do you think `config.yaml` is better choice?  In theory everything in `config.yaml` could live as schema annotations, but that choice was never made.  So I am now wondering if this PR is muddying integration points too much - sure we can add `annotations.java_package` to linkml schema, but is that opening up a second configuration point that was previously rejected in favor of `config.yaml`?

docs/generators/java.rst
gouttegd: > This is normal design pattern

For LinkML, yes, this is a normal behaviour.

LinkML has no concept of namespace. All classes, slots, and enums in a LinkML schema live in a single, global, flat namespace – regardless of whether those classes, slots, and enums are declared in different sub-schemas.

The generated Java code should reflect that: all generated classes and enums should reside in a single Java package – not one package for each sub-schema. And then it is only logical that the top-level schema should be the one deciding what the package name is.

gouttegd:
> I'm can be easily convinced.

I was not trying to convince you to withdraw the feature! :) As long as the `java_package` annotation (if it is present) can be overridden by whoever runs the Java generator, I am completely fine with it.

And I am willing to admit that the feature does make sense for a schema that is strictly internal to a single project (i.e. a schema that nobody expects to be reused by anyone else to generate their own code).

docs/generators/java.rst
noelmcloughlin: Ok, thanks.  This implies `default_prefix` should be same - even if sub-schemas declare different `default_prefix` value - I do not recall seeing generator errors warning about different values on import but the merged model may flatten into parent's prefix value - but I have not given lot of thought.

noelmcloughlin:
I'm am leaning towards `config.yaml` since its global scope reinforces the notion package name must be global, unlike schema-level annotations potentially varying/absent across parent/sub-schemas.

On the other hand, I am researching possibility to generate java annotation type declarations (i.e. `java.lang.annotation`) which seems to require a tool to take `schema + data file + jinja2`, iterate a collection, and writes rendered files **AND/OR** schema-driven @interface support in gen-java (**would use annotations** `java_annotation_type: true`) but not sure use case is feasible.

I'll take some time to research more.

docs/generators/java.rst
gouttegd: > This implies `default_prefix` should be same - even if sub-schemas declare different `default_prefix` value

Not necessarily. When I say that “all classes, slots, and enums live in a single, global, flat namespaces”, I am talking about the classes/slots/enums’ _names_, not their _URIs_.

Consider the two following schemas:

`foo.yaml`:

```yaml
id: https://example.org/foo
name: foo
prefixes:
  foo: https://example.org/foo/
default_prefix: foo
imports:
  - linkml:types
annotations:
  java_package: org.example.foo
classes:
  Thing:
    attributes:
      id:
        range: string
```

`bar.yaml`:

```yaml
id: https://example.org/bar
name: bar
prefixes:
  bar: https://example.org/bar/
default_prefix: bar
imports:
  - linkml:types
  - foo
annotations:
  java_package: org.example.bar
classes:
  AnotherThing:
    attributes:
      ...
```

The `Thing` class and the `AnotherThing` classes have URIs in different namespaces (`https://example.org/foo/Thing` and `https://example.org/bar/AnotherThing`).

But the classes are _not_ identified by their URIs! They are identified by their _names_. (Formally, the “identifier slot” of a `Element` object is its `name`.)

This is why you cannot have two classes with the same name, even if they are declared in different sub-schemas and even if the sub-schemas have different default prefixes.

docs/generators/java.rst
```diff
@@ -32,6 +32,64 @@ Additional Notes
...
+
+``gen-java`` only ever reads ``generator_args.java.package`` out of this file --
+every other key (``directory``, ``excludes``, other generators' ``generator_args``
+entries, etc.) is ignored, so a full multi-generator project ``config.yaml`` can be
+passed as-is without modification.
+
+An explicit ``--package`` command-line option always overrides a value set via
+``--config-file``.
+
+If ``--config-file`` is not given, ``gen-java`` looks for a ``config.yaml`` in the
```
gouttegd: Unless I am mistaken, this is a different behaviour than the behaviour of the project generator, which does _not_ automatically fallback to a `config.yaml` file if a config file has not been explicitly specified with `--config-file`.

Importantly, this means that, after this change, `config.yaml` becomes _de facto_ a reserved filename in the top-level directory, since whenever the top-level directory contains such a file, it will _automatically_ be used by `linkml generate java` and `linkml generate golang` (and possibly by other `generate` subcommands later, if/when other generators adopt the same behaviour). Given how “generic” that filename is, I feel this warrants at least some discussion with the core devs. :)

packages/linkml/src/linkml/utils/generator.py
```diff
@@ -990,3 +991,41 @@ def decorator(f: Command) -> Command:
         return f
 
     return decorator
+
+
+DEFAULT_CONFIG_FILENAME = "config.yaml"
+
+
+def read_generator_config_arg(config_file: IO[bytes] | None, generator_name: str, arg: str):
```
gouttegd: This method will read the entire config file whenever it is called to get one particular configuration option.

This is fine _for now_ because there is currently one possible configuration option for each of the generators (Java, Go) that use this config file – so we know that this method will in fact only be called at once most once by any given generator.

But if/when the generators start needing to look up more than one configuration option (i.e. they start calling this method repeatedly), then the configuration file will be read entirely for every query.

I defer to the core devs to judge whether this inefficiency should be corrected or not. I personally think it should.

docs/generators/java.rst
noelmcloughlin: My mistake - I use `linkml-project-copier` heavily and it hardcodes `config.yaml` so I assumed it was linkml default: https://github.com/linkml/linkml-project-copier/blob/9654b2326e4ee08cc92351b48a688239f6512e41/template/config.public.mk.jinja#L17
```
## gen-project configuration file
LINKML_GENERATORS_CONFIG_YAML=config.yaml
```
That env file was introduced by me a few years ago as workaround for  `projectgen` many issues:

Meanwhile `projectgen.cli` docstring mentions `--config config.yaml` and click option is `--config-file` (inconsistent) so maybe maybe this idea of `config.yaml` being owned by LinkML became planted.  A better name would be `linkml-generators.yaml` (root of project) or even better`.linkml/generators.yaml` 

I will drop this behavior change.

noelmcloughlin: Reverted in commit https://github.com/linkml/linkml/pull/3781/changes/14be853fcf92ea415d4c424585060e641a48888b

packages/linkml/src/linkml/utils/generator.py
noelmcloughlin: Thanks for the review.
Good catch.
Fixed in commit https://github.com/linkml/linkml/pull/3781/changes/b7bd817995d4d9b9c7fc40695011158b29da7e58

docs/generators/java.rst
noelmcloughlin: Found pre-existing bug https://github.com/linkml/linkml/issues/3858 too

docs/generators/java.rst
```diff
@@ -32,6 +32,57 @@ Additional Notes
 
 The Java generator's default template uses Project Lombok's `@Data <https://projectlombok.org/features/Data>`__ annotation, which provides getters, setters, equals and hashcode functionality.
 
+Package Configuration
```
gouttegd, line 97: Nothing wrong here, but advance warning that I almost completely rewrote the entire `java.rst` document in #3822, so one of us (depending on which PR is merged first) will have some conflict resolution to do. ;)

tests/linkml/test_generators/test_javagen.py
```diff
@@ -298,3 +301,189 @@ def test_serialize_accepts_str_or_path_directory(kitchen_sink_path, tmp_path, as
...
+            ["--config-file", str(config_path), "--output-directory", str(out_dir), str(schema_path)],
+        )
+
+    assert result.exit_code == 0, result.output
+    assert_file_contains(out_dir / "Thing.java", "public class Thing", after="package 1bad.pkg")
+    assert any("not a valid Java package name" in record.message for record in caplog.records)
+
+
+def test_cli_config_file_missing_path_errors(tmp_path):
+    """A --config-file path that doesn't exist is a usage error (click.File's own behavior)."""
```
gouttegd: Not sure that test is needed. As the docstring says, this happens in Click. If a non-existing config file does not raise an error, then this is an issue with Click, not with LinkML.

gouttegd (commented):
I love the extensive testing, but the `test_cli_config_file_missing_path_errors` test (in `test_javagen.py`) is pushing things too far by testing what is actually a Click feature (and therefore testing it is the responsibility of Click’s developers, not ours).

tests/linkml/test_generators/test_javagen.py
noelmcloughlin: Yes too much (my `--config` paranoia #3858)

Fix folded into last commit.

gouttegd (approved):
Looks good to me.

matentzn:
@noelmcloughlin I will leave it like this because afaics you should be able to merge yourself now after a @gouttegd review.

gouttegd:
> you should be able to merge yourself now after a @gouttegd review.

Is that really the case? The PR touches more files than those for which I am code owner (e.g. `oocodegen.py`, `excelgen.py`, `deprecation.py`, etc.), so I would assume a review from the core team is still needed here, even after #3873. Or did I misunderstand how #3873 is supposed to work?

gouttegd (approved):

gouttegd:
GitHub seems to agree with my understanding: even with my review, a review from the core team is still needed (correctly, I believe).
Screenshot 2026-08-12 at 13 04 34

packages/linkml/src/linkml/generators/javagen.py
```diff
@@ -499,6 +530,10 @@ def cli(
     **args,
 ):
     """Generate java classes to represent a LinkML model"""
+    if package is None:
```
matentzn, line 546: So if I use ` --package 1bad.pkg` I silently produce broken Java and if I have the same in the config file I get a warning?

packages/linkml/src/linkml/generators/excelgen.py
```diff
@@ -171,7 +171,7 @@ def column_enum_validation(
 
         dv.add(f"{column_letter}2:{column_letter}1048576")
 
-    def serialize(self, **kwargs) -> str:
+    def serialize(self, **kwargs) -> None:
```
matentzn, line 174: Hm.. are you sure? 

The signature of the parent class returns str. Its not enforced by CI so you can do whatever you like, but this seems like problem in waiting..

`Generator.serialize`:

```
```

matentzn:
> GitHub seems to agree with my understanding: even with my review, a review from the core team is still needed (correctly, I believe).

Sorry jumped to conclusion based on the PR title `javagen`.

packages/linkml/src/linkml/generators/javagen.py
noelmcloughlin: Thanks for review - addressed in 4295dbd25a94dff5e28c8277924ed5197e6eee1b

packages/linkml/src/linkml/generators/excelgen.py
noelmcloughlin: It was intentional and already captured in comments at `projectgen.py`- because`excelgen.py` does actually return None (like `docgen.py`). But I can update the `generator.py` signature to match `oocodegen.py` pattern -> `str | None` - let me know if you have any concerns.

 See https://github.com/linkml/linkml/commit/4295dbd25a94dff5e28c8277924ed5197e6eee1b

packages/linkml/src/linkml/generators/javagen.py
```diff
@@ -493,6 +524,10 @@ def cli(
...
+        package = read_generator_config(config_file, "java").get("package")
+    if package is not None and not _is_valid_java_package(str(package)):
```
matentzn: AFAICS (correct me if I am wrong) `_is_valid_java_package` is only called in `cli()`. `gen-project` constructs `JavaGenerator(package=...)` directly, so it skips the check entirely. Move check into Move the check into `__post_init__`?

tests/linkml/test_scripts/__snapshots__/genshex/metashex.ttl
matentzn, line 1: Why is this file here?

packages/linkml/src/linkml/utils/deprecation.py
```diff
@@ -300,6 +300,16 @@ def warn(self, stack_level=3, **kwargs):
         recommendation="Use 'linkml validate schema.yaml' for schema validation without lint rules.",
         issue=3387,
     ),
+    Deprecation(
+        name="golanggen-package-name-option",
+        deprecated_in=SemVer.from_str("1.11.0"),
```
matentzn: 1.12?

packages/linkml/src/linkml/generators/golanggen/golanggen.py
matentzn, line 1: golang --package "1 bad-pkg"` still allows `package 1 bad-pkg` — no validation on the Go side at all, while Java now has it. Not a regression, but since @gouttegd's point (B) was specifically about consistency across the package-scoped generators, either give Go the same treatment or say why not.

packages/linkml/src/linkml/utils/generator.py
```diff
@@ -990,3 +992,55 @@ def decorator(f: Command) -> Command:
...
+    if not isinstance(value, dict):
+        raise click.UsageError(f"--config-file: expected a YAML mapping at {where}, found {type(value).__name__}")
+    return value
+
+
+def read_generator_config(config_file: IO[bytes] | None, generator_name: str) -> dict[str, Any]:
+    """Read one generator's settings out of a gen-project-style config file.
+
+    This is the configuration file that ``gen-project --config-file`` reads so a
+    project can keep a singlefile and each generator can have its own section.
```
matentzn:
```suggestion
    project can keep a single file and each generator can have its own section.
```

packages/linkml/src/linkml/generators/projectgen.py
```diff
@@ -49,12 +50,17 @@
...
+    # JavaGenerator writes one file per class directly into `directory`
+    # (a serialize()-time argument, not a constructor argument) and its
+    # serialize() always returns None; see SERIALIZE_ONLY_ARGS below.
+    "java": (JavaGenerator, "java/{name}.java", {"directory": "{parent}"}),
     "excel": (ExcelGenerator, "excel/{name}.xlsx", {"output": "{parent}/{name}.xlsx"}),
 }
 
+# Argument keys that must be passed to serialize() but NOT to the generator's
+# constructor (currently only JavaGenerator's `directory`).
+SERIALIZE_ONLY_ARGS = {"directory"}
```
matentzn: `SERIALIZE_ONLY_ARGS` is a module-level set applied to every generator in `GEN_MAP`. Today only java has `directory` so it's harmless, but a user writing `generator_args: {owl: {directory: ...}}` would have it silently dropped from the constructor. Can we move that as a per-entry flag in the `GEN_MAP` tuple next to the path format, not as a global name blocklist?

matentzn (changes requested):

tests/linkml/test_scripts/__snapshots__/genshex/metashex.ttl
noelmcloughlin: I'm not sure - it may have been a local test failure that I was dealing with (stale file).  Both #3870 (bump rdflib) and the isomorphic/diff PR got merged during this PR lifetime - so maybe text diff staleness is no longer a concern.

https://github.com/linkml/linkml/blob/26ebff6d41a1cd8b8d1888f32b85eab363a78735/tests/linkml/test_scripts/test_gen_shex.py#L32

I removed the file as `test_meta[arguments2-metashex.ttl] --with-network` passes on main.

packages/linkml/src/linkml/generators/javagen.py
noelmcloughlin: Fixed in 799f80855921738aeb23511dd8093e19b61448ef

tests/linkml/test_scripts/__snapshots__/genshex/metashex.ttl
noelmcloughlin: Fixed in 2108f186fbe2b53df651f9e0f024aa20a83e091b

packages/linkml/src/linkml/utils/deprecation.py
noelmcloughlin: Fixed -assumes PR is merged in time for 1.12 and included.

packages/linkml/src/linkml/generators/golanggen/golanggen.py
noelmcloughlin: Fixed in 799f80855921738aeb23511dd8093e19b61448ef

packages/linkml/src/linkml/generators/projectgen.py
noelmcloughlin: Fixed via refactoring in https://github.com/linkml/linkml/commit/799f80855921738aeb23511dd8093e19b61448ef

packages/linkml/src/linkml/generators/projectgen.py
```diff
@@ -128,7 +168,16 @@ def generate(schema_path: str, config: ProjectConfiguration = ProjectConfigurati
...
                     all_gen_args["output"] = all_gen_args["output"].format(name=name, parent=parent_dir)
 
-                gen = gen_cls(local_path, **all_gen_args)
+                # Some generators (e.g. JavaGenerator) accept an arg only in
+                # serialize(), not in their constructor; keep this generator's
+                # serialize-only keys out of the constructor call.
+                constructor_args = {k: v for k, v in all_gen_args.items() if k not in gen_config.serialize_only}
+                try:
+                    gen = gen_config.generator(local_path, **constructor_args)
+                except ValueError as e:
```
matentzn: This catch is too wide. Generator constructors load and resolve the schema, so they raise `ValueError` for reasons that have nothing to do with `generator_args`:

```
$ cat bad.yaml
justastring
$ gen-project -d out bad.yaml
Usage: gen-project [OPTIONS] YAMLFILE
Try 'gen-project --help' for help.
Error: graphql: File "bad.yaml", line 4, col 5: Unknown CURIE prefix: linkml
```

Options:

1. Give the package check its own exception (`class InvalidPackageName(ValueError)` next to the validator) and catch *that* in the three places. One line per catch, everything else keeps its traceback.
2. Validate `generator_args` up front — where the args are read, before any generator is constructed — and drop the try/except here and in `cli()` entirely. (I lean here)

packages/linkml/src/linkml/generators/javagen.py
```diff
@@ -517,18 +553,24 @@ def cli(
...
+            template_dir=template_dir,
+            template_file=template_file,
+            genmeta=genmeta,
+            gen_classvars=classvars,
+            gen_slots=slots,
+            true_enums=true_enums,
+            use_aliases=use_aliases,
+            **args,
+        )
+    except ValueError as e:
```
matentzn: Same problem as the one in `projectgen.py` — `JavaGenerator(...)` loads the schema, so this catches far more than the package check:

```
$ gen-java --output-directory out bad.yaml   # bad.yaml is the single line: justastring
Usage: gen-java [OPTIONS] YAMLFILE
Try 'gen-java --help' for help.
Error: Unexpected type justastring
```

The invocation is fine; the schema is broken. Suggesting `--help` sends the user in the wrong direction and the stack is gone.

Moving the validation into `__post_init__` was the right call and I don't want to undo it — the cheapest fix is to keep a `_is_valid_java_package(package)` call here in `cli()`, right after the `read_generator_config` line, raising `click.UsageError` for the flag/config case. Then `__post_init__` stays an uncaught library-level backstop for gen-project and programmatic use, and this try/except can go away.

packages/linkml/src/linkml/generators/golanggen/golanggen.py
```diff
@@ -673,19 +765,31 @@ def cli(
...
+        gen = GolangGenerator(
+            yamlfile,
+            package=package,
+            alphabetical_sort=alphabetical_sort,
+            nullable_primitives=nullable_primitives,
+            named_slot_types=named_slot_types,
+            template_dir=template_dir,
+            **args,
+        )
+    except ValueError as e:
```
matentzn: Ditto the javagen one — `GolangGenerator(...)` loads the schema, so this catches unrelated `ValueError`s and reports them as gen-golang usage errors. Whatever you pick for javagen, mirror it here.

packages/linkml/src/linkml/utils/deprecation.py
```diff
@@ -311,6 +311,16 @@ def warn(self, stack_level=3, **kwargs):
         recommendation="Remove the `materialize_patterns` option from generator calls and command lines.",
         issue=1557,
     ),
+    Deprecation(
+        name="golanggen-package-name-option",
+        deprecated_in=SemVer.from_str("1.12.0"),
```
matentzn, line 316: The message is CLI-phrased ("The `--package-name` option of gen-golang..."), but `deprecated_fields` now fires this same text for `GolangGenerator(package_name=...)` and for reading `.package_name` off an instance. Someone hitting it from Python gets told about a command-line flag they never used.

matentzn (changes requested):
Hey @noelmcloughlin thanks for the patience. I think thats the last one which I didnt see last time sorry; if you run out of steam here let me know, I dont want this to become a nit-athon.

packages/linkml/src/linkml/generators/projectgen.py
noelmcloughlin: Yes. validation moved early, before construction in all three places:

- `JavaGenerator.validate_generator_args({"package": package})`  (javagen.py line 548)
- `            gen_config.generator.validate_generator_args(all_gen_args)`  (projectgen line 158)
- `    GolangGenerator.validate_generator_args({"package": package})`    (golanggen/golanggen.py line 782)

So my tired mistake - wrapping contruction with try/catch - is no longer necessary. 

Fixed in 2108f186fbe2b53df651f9e0f024aa20a83e091b

packages/linkml/src/linkml/generators/javagen.py
noelmcloughlin: Fixed in 2108f186fbe2b53df651f9e0f024aa20a83e091b - and added unit test too be sure.

packages/linkml/src/linkml/generators/golanggen/golanggen.py
noelmcloughlin: Fixed in 2108f186fbe2b53df651f9e0f024aa20a83e091b

packages/linkml/src/linkml/utils/deprecation.py
noelmcloughlin: Good catch, done. Fixed in 2108f186fbe2b53df651f9e0f024aa20a83e091b

noelmcloughlin: Python user will see package_name/package named explicitly, not just the CLI flag spelling.

 pythongen.py has no `depreciation_name` arg inside `@deprecated_fields` so uses METADATA_FLAG

packages/linkml/src/linkml/generators/projectgen.py
```diff
@@ -94,61 +134,83 @@ class ProjectGenerator:
...
+                logger.info(f"Skipping {gen_name} as not in inclusion list: {config.includes}")
+                continue
+            if config.excludes is not None and gen_name in config.excludes:
+                logger.info(f"Skipping {gen_name} as it is in exclusion list")
+                continue
+            all_gen_args = {
+                **gen_config.default_args,
+                **config.generator_args.get(gen_name, {}),
+            }
+            gen_config.generator.validate_generator_args(all_gen_args)
```
matentzn: Nice.

One thing got lost in the refactor though. Last round the message named the generator:

```
Error: java: '1bad.pkg-name' is not a valid Java package name
```

and now it doesn't, because the prefix went out with the try/except:

```
$ gen-project --config-file cfg.yaml -d out kitchen_sink.yaml
Error: '1bad.pkg-name' is not a valid Java package name
```

For java you can just about infer it from "Java package name", but the generator name is the bit that tells you *which* `generator_args` section to go fix, and that stops being inferable the moment a second generator grows a validator. Catching `click.UsageError` specifically is narrow enough that it doesn't reintroduce the original problem:

```suggestion
            try:
                gen_config.generator.validate_generator_args(all_gen_args)
            except click.UsageError as e:
                raise click.UsageError(f"{gen_name}: {e.format_message()}") from e
```

Take it or leave it — if you're out of steam, merge as is and I'll open a follow-up.

packages/linkml/src/linkml/generators/projectgen.py
```diff
@@ -29,30 +30,69 @@
...
+    #    'rdf': GeneratorConfig(
+    #        RDFGenerator, 'rdf/{name}.ttl', {'context': '{parent}/../jsonld/{name}.context.jsonld'}
+    #    ),
+    "shex": GeneratorConfig(ShExGenerator, "shex/{name}.shex", {}),
+    "shacl": GeneratorConfig(ShaclGenerator, "shacl/{name}.shacl.ttl", {}),
+    "sqltable": GeneratorConfig(SQLTableGenerator, "sqlschema/{name}.sql", {}),
+    # JavaGenerator writes one file per class directly into `directory`
+    # (a serialize()-time argument, not a constructor argument) and its
+    # serialize() always returns None.
+    "java": GeneratorConfig(
```
matentzn, line 89: this means every `gen-project` run now emits a `java/` tree with `package example;`, which is exactly what #2537 asked for but is still a visible change in default output for everyone. Worth a line in the changelog when this lands.

And since this PR gives golang the same `--config-file` / `--package` treatment as java, is there a reason golang *isn't* in `GEN_MAP` too, or just scope?

matentzn (commented):
Two optional regression fix suggestions; say the word and I will approve merge.

packages/linkml/src/linkml/generators/projectgen.py
noelmcloughlin: I was half afraid - java in, golang out :-)

I think Java is sufficiently commodity enough language to get first class promotion but I think `linkml-project-copier` could exclude if needed.   What Changelog do you mean?

matentzn: > What Changelog do you mean

PR description.

> I was half afraid - java in, golang out :-)

OK! I trust that judgement :) You can resolve the comment!

matentzn (approved):

packages/linkml/src/linkml/generators/projectgen.py
noelmcloughlin: On this theme, LinkML does not have standard well known and "named" configuration for generators.

*Project Level*

I always misunderstood that `config.yaml` was the canonical project generator configuration file but learned its just what `linkml-project-cookiecutter / linkml-project-coper` used in absence of a linkml agreed filename or convention.  

> My mistake - I use `linkml-project-copier` heavily and it hardcodes `config.yaml` so I assumed it was linkml default: https://github.com/linkml/linkml-project-copier/blob/9654b2326e4ee08cc92351b48a688239f6512e41/template/config.public.mk.jinja#L17

I am now thinking a formal agreed name would be useful when `--config (file)` is not provided:

1. check for `$LINKML_GENERATORS_CONFIG_FILE` and use its sanitized value, else
2. check for `.linkml/generators.yaml`, else
3. fallback to existing code

The choice of `.linkml/` directory is bit strategic. By convention, its a well known convention - tools should support configuration inside a DOT_<toolname>/ directory.  Secondly, in my vision of "what is the best route to stealth adoption of linkml in projects that are resistant to the idea", I say "**LinkML is sidecar** and my preferred pattern is  `mkdir .linkml && cd .linkml` and run `linkml-project-copier` inside. 

So `.linkml/generators.yaml` fits nicely.   

*OS level*
Linkml could decide if we (a) support XDG standard, and (b) what filename we expose, although I do not see demand. I vaguely remember discussion on this in `linkml-project-cookercutter` issue years ago, but we focused instead on project level configuration solution (tech debt remains today).

https://specifications.freedesktop.org/basedir/latest/
_There is a single base directory relative to which user-specific configuration files should be written. This directory is defined by the environment variable $XDG_CONFIG_HOME._

matentzn: Is this somehow related to #3782?

noelmcloughlin: Maybe -  at first glance they seem to be discussing **OS level** configuration (i.e. global to my machine) which is related to my "OS Level" comment above which is secondary.

But I was mainly thinking about **project level** configuration. All Linkml has today is CLI parameter `--config` (and variants like `--config-file`, some deprecated maybe) so if that parameter is not passed, then all project config must be passed by CLL parameters (and there could be many).  Probably we need to socialize.

PR #3782, opened 2026-07-21
Global Config v2

### LINKED ISSUE #2331: A centralized config for LinkML

sneakers-the-rat:
## Summary

#2335 except redone over the monorepo structure. easier to just redo it than it was to rebase.

the description is the same as over there, just copy/pasting that here with one change - instead of the pattern being to instantiate `GlobalConfig` every time, which is expensive, or having some permanent module-level global, we instead have a `get_config()` function with a `reload` param so that we can get and store a globally cached version and have a predictable way to reload that if needed.

begin copy paste

---

Fix: https://github.com/linkml/linkml/issues/2331

alright here's a global config.

At the moement it's not hooked up into anything, but if we like this then i can continue the lovely work in https://github.com/linkml/linkml/pull/2323 and make it control log levels and whatnot as the config object suggests should be possible.

It should be all spelled out in the docs, but just for the good of the order here's the impl:

## Motivation

We want to be able to configure logging at first, but generally have a tidy way to specify behavior that might need to be customized in a per-system, per-session, and per-project way.

This is probably extremely overengineered, but i'm also laying groundwork for [`linkml build`](https://github.com/linkml/linkml/pull/2263) with a project-level config that can get sourced from one of potentially several places in several formats, so that's why. In the future i would anticipate additional global config stuff including whether we should try to resolve schema via url, location of cache directories, etc. so this sets us up for any of that.

## Implementation

Use `pydantic-settings`. 

Specifically, create a `GlobalConfig` class that sources from the following locations, in order of priority from high to low, where higher priority sources override lower priority

* in the arguments passed to the class constructor (not user configurable)
* in environment variables like ``export LINKML_LOG_DIR=~/``
* in a ``.env`` file in the working directory
* in a ``linkml_config.yaml`` file in the working directory
* in the ``tool.linkml.config`` table in a ``pyproject.toml`` file in the working directory
* in the global ``linkml_config.yaml`` file in the platform-specific data directory
  (use ``linkml config get config_file`` to find its location)
* the default values in the `GlobalConfig` model

There is an example of each kind of configuration in the docs in a tab group.

We can use nested models here, and imo we should do that to keep config reasonable instead of turning into a junk drawer, so I added an initial set of config options for handling stream and file logging to demo that. Descriptions of how to handle nested models are also in the docs.

I also added some cli commands to get and set global config values. values are printed as yaml so they can be re-parsed if needed (i did when writing tests)

```shell
$ linkml config get

user_dir: /Users/jonny/Library/Application Support/linkml
config_file: /Users/jonny/Library/Application Support/linkml/linkml_config.yaml
log:
  dir: /Users/jonny/Library/Logs/linkml
  file_name: linkml.log
  level: WARNING
  # ...
```

```shell
$ linkml config get log.level
WARNING
```

Setting values handles `None` to delete keys, nested models, and does model validation as well. We shouldn't be trying to set nonscalar/anything that doesn't have an unambiguous representation in a shell in the config, so hopefully that's as complex as we need it to be.

```shell
$ linkml config set log.level WARNING
Updated linkml config:
key: log.level
value: WARNING
config file: /Users/jonny/Library/Application Support/linkml/linkml_config.yaml
```

## Caveats

There's bound to be a bit of stickiness where people want to specify where the global config file lives. There's a sort of chicken and egg problem here, where we want the global config to be the lowest priority source of config, but then we should have read it and applied it before we reach the shell/cwd-based config options that might want to override it. That might need more consideration in the future, but for now i'm assuming that people will just use the local config options rather than trying to reposition the global config

There's also probably going to be some confusion like "what the heck i called `linkml config set <key> <value>`" and it didn't set my config! when people have some conflicting config file in the cwd. I didn't want to rig up a whole `linkml config set global <key> <value>` like eg. `pyenv` does at the moment, but that might be what we need to do if that starts happening. Tried to flag that clearly in the docs but we'll see.

## Re: Deps

I added two required deps:
- `pydantic-settings`
- `platformdirs`

`pydantic-settings` has one additional non-pydantic dep, `python-dotenv`, which is pretty light, and `platformdirs` has no additional deps and is extremely common (it's probably already in our dep tree somewhere).

imo it's worth it - since we're already installing pydantic, `pydantic-settings` doesn't add a lot more.

I also moved some docs deps that were hanging out in the `dev` group down into the `docs` group and added a package to show pydantic models cleanly (because the config models looked like total garbage without it)

## Tests

Added the basic tests that i thought we might want, but lmk if i missed something.

what ya think?

---

## AI Assistance

you can pry my keyboard from my cold dead hands

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -1,0 +1,149 @@
+# Configuration
+
+Global configuration in linkml uses [`pydantic-settings`](https://docs.pydantic.dev/latest/concepts/pydantic_settings/)
+
+
+## Sources
+
+Config values can be set (in order of priority from high to low, where higher
+priorities override lower priorities)
+
+* in the arguments passed to the class constructor (not user configurable)
+* in environment variables like `export LINKML_LOG_DIR=~/`
+* in a `.env` file in the working directory
+* in a `linkml_config.yaml` file in the working directory
+* in the `tool.linkml.config` table in a `pyproject.toml` file in the working directory
+* in the global `linkml_config.yaml` file in the platform-specific data directory
+  (use `linkml config get config_file` to find its location)
+* the default values in the {class}`~linkml.utils.config.GlobalConfig` model
+
+Parent directories are _not_ checked.
+
+## Keys
+
+### Prefix
+
+Keys for environment variables (i.e. set in a shell with e.g. `export` or in a `.env` file)
+are prefixed with `LINKML_` to not shadow other environment variables.
+Keys in `toml` or `yaml` files are not prefixed with `LINKML_` .
+
+### Nesting
+
+Keys for nested models are separated by a `__` double underscore in `.env`
+files or environment variables (eg. `LINKML_LOG__DIR`)
+
+Keys in `toml` or `yaml` files do not have a dunder separator because
+they can represent the nesting directly (see examples below)
+
+When setting values from the cli, keys for nested models are separated with a `.`.
+
+### Case
+
+Keys are case-insensitive, i.e. these are equivalent::
+
+    export LINKML_LOG__DIR=~/
+    export linkml_log__dir=~/
+
+## CLI
+
+Global settings can be gotten and set via the CLI, for example:
+
+### Getting values
+
+```{command-output} linkml config get
+---
+ellipsis: 5
+---
+```
+
+Index nested config models with `.`
+
+```{command-output} linkml config get log.level
+```
+
+### Setting values
+
+Values are set in the global `linkml_config.yaml` file with a similar syntax
+
+```{note}
+Values set via the CLI are set in the global `linkml_config.yaml`
+file, which is the lowest priority source of settings, and will thus be
+overridden by any local configs in `.env` files or otherwise.
+```
+
+```{command-output} linkml config set log.level WARNING
+```
+
+```{command-output} linkml config get log.level
+```
+
+
+## Examples
+
+
+`````{tab-set}
+````{tab-item} linkml_config.yaml
+```{code-block} yaml
+user_dir: ~/.config/linkml
+config_file: linkml_config.yaml
+log:
+  dir: /var/log/linkml
+  level_file: INFO
+  level_stream: WARNING
+  file_n: 5
+```
+````
+````{tab-item} env vars
+```{code-block} bash
+export LINKML_USER_DIR='~/.config/linkml'
+export LINKML_CONFIG_FILE='linkml_config.yaml'
+export LINKML_LOG__DIR='/var/log/linkml'
+export LINKML_LOG__LEVEL_FILE='INFO'
+export LINKML_LOG__LEVEL_STREAM='WARNING'
+export LINKML_LOG__FILE_N=5
+```
+````
+````{tab-item} .env file
+```{code-block} python
+LINKML_USER_DIR='~/.config/linkml'
+LINKML_CONFIG_FILE='linkml_config.yaml'
+LINKML_LOG__DIR='/var/log/linkml'
+LINKML_LOG__LEVEL_FILE='INFO'
+LINKML_LOG__LEVEL_STREAM='WARNING'
+LINKML_LOG__FILE_N=5
+```
+````
+````{tab-item} pyproject.toml
+```{code-block} toml
+[tool.linkml.config]
+user_dir = "~/.config/linkml"
+config_file = "linkml_config.yaml"
+
+[tool.linkml.config.log]
+dir = "/var/log/linkml"
+level_file = "INFO"
+level_stream = "WARNING"
+file_n = 5
+
+```
+````
+````{tab-item} cli
+```{code-block} bash
+linkml config set user_dir '~/.config/linkml'
+linkml config set config_file 'linkml_config.yaml'
+linkml config set log.dir '/var/log/linkml'
+linkml config set log.level_file 'info'
+linkml config set log.level_stream 'warning'
+linkml config set log.file_n 5
+```
+````
+`````
+
+
+## API
+
+```{eval-rst}
+.. automodule:: linkml.utils.config
+    :members:
+
+```
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -13,11 +13,18 @@ priorities override lower priorities)
 * in a `.env` file in the working directory
 * in a `linkml_config.yaml` file in the working directory
 * in the `tool.linkml.config` table in a `pyproject.toml` file in the working directory
-* in the global `linkml_config.yaml` file in the platform-specific data directory
+* in the global `linkml_config.yaml` file in the platform-specific data directory[^platformdirs]
   (use `linkml config get config_file` to find its location)
 * the default values in the {class}`~linkml.utils.config.GlobalConfig` model
 
 Parent directories are _not_ checked.
+
+[^platformdirs]: These directories come from [platformdirs](https://platformdirs.readthedocs.io/en/latest/index.html),
+  see [their documentation](https://platformdirs.readthedocs.io/en/latest/platforms.html) for more info.
+  The default configuration directories for linkml are:
+    - Linux: `~/.config/linkml/`
+    - Mac: `~/Library/Application Support/linkml/`
+    - Windows: `~\AppData\Local\linkml\linkml\`
 
 ## Keys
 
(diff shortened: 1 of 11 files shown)


docs/config/index.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Sources
+
+Config values can be set (in order of priority from high to low, where higher
+priorities override lower priorities)
+
+* in the arguments passed to the class constructor (not user configurable)
+* in environment variables like `export LINKML_LOG_DIR=~/`
+* in a `.env` file in the working directory
+* in a `linkml_config.yaml` file in the working directory
```
yarikoptic, line 14: I wonder if should be


```suggestion
* in a `.linkml.yaml` file in the working directory
```

or 

```suggestion
* in a `.linkml/config.yaml` file in the working directory
```

if we expect potentially anything else "useful" to come there.  And that could mirror the `user_dir: ~/.config/linkml` and potentially having user-wide `user_dir: ~/.config/linkml/config.yaml`?

docs/config/index.md
```diff
@@ -0,0 +1,149 @@
...
+* in the `tool.linkml.config` table in a `pyproject.toml` file in the working directory
+* in the global `linkml_config.yaml` file in the platform-specific data directory
+  (use `linkml config get config_file` to find its location)
```
yarikoptic, line 17: this is IMHO too ambigous, even having 


```suggestion
  e.g. `~/.config/linkml/linkml_config.yaml`  on Linux (use `linkml config get config_file` to find its location)
```
would be useful.

docs/config/index.md
```diff
@@ -0,0 +1,149 @@
...
+    export LINKML_LOG__DIR=~/
+    export linkml_log__dir=~/
+
+## CLI
+
+Global settings can be gotten and set via the CLI, for example:
+
+### Getting values
+
+```{command-output} linkml config get
```
yarikoptic, line 60: echoing `git config` - this should be "overall" thus potentially accounting for local config too. But there could be CLI option to specify at what level to request (global or current).

docs/config/index.md
```diff
@@ -0,0 +1,149 @@
...
+
+Values are set in the global `linkml_config.yaml` file with a similar syntax
+
+```{note}
+Values set via the CLI are set in the global `linkml_config.yaml`
+file, which is the lowest priority source of settings, and will thus be
+overridden by any local configs in `.env` files or otherwise.
+```
+
+```{command-output} linkml config set log.level WARNING
```
yarikoptic, line 81: again -- CLI option to instruct where to change might be handy

yarikoptic (commented):
initial comments on potential naming and overall handling of global-vs-local configs

docs/config/index.md
sneakers-the-rat: so three proposals here for the "unique name segment" of the linkml config when used in a project directory:

- `.config/linkml.yaml`
- `.linkml.yaml`
- `.linkml/config.yaml`

Several intersecting constraints/considerations for this name in particular: for the "user-level" or "systemwide" config, where one might want to set e.g. logging verbosity across projects, we should definitely use `platformdirs` - so that means that on different operating systems the root is different: 
- windows: `~\AppData\Local\linkml\`
- mac: `~/Library/Application Support/linkml/`
- linux: `~/.config/linkml/`

so i don't think we should use `.config/linkml.yaml` because that creates an inconsistency across OSes whether the in-repo directory mirrors the systemwide directory.

I personally find dotfiles a bit annoying because they are hidden by default in i think all OSes? but that's just a relatively weak style preference. similarly i would find it annoying to need to create a subdirectory with a single file in it on the off chance that there might be other files in it in the future (and that is similarly weak style preference). 

iirc i initially had it just at `linkml.yaml` but *i think* that got vetoed, and i understand why as it might be ambiguous or conflict with things people may have already named e.g. schemas, and if we shouldn't do `linkml.yaml` we probably shouldn't do `.linkml.yaml`.

I think i went with `linkml_config.yaml` because it's the most "linkml-like" name, which usually goes with names that are globally unique and unambiguous, rather than meaning implied by context (like, e.g. being `config.yaml` within a `linkml` directory, which is usually how i name things).

I would say my preferences are like 60% strong on having "a single uniquely named file even if the name is a bit cumbersome" but i would be amenable to changing this of course if outvoted.

docs/config/index.md
sneakers-the-rat: how about this? https://github.com/linkml/linkml/pull/3782/commits/052c10cb676246eca8f3604acb6d34d91b8b235d

Screenshot 2026-07-21 at 2 03 29 PM

Screenshot 2026-07-21 at 2 05 58 PM

docs/config/index.md
sneakers-the-rat: yeah this and below comment are a caveat I listed in PR description and also in the docs here: https://github.com/sneakers-the-rat/linkml/blob/052c10cb676246eca8f3604acb6d34d91b8b235d/docs/config/index.md?plain=1#L75-L79

I'll address both comments here since they are related.

`linkml config get` gets the currently effective value, wherever you're calling it - so this is already true:

> this should be "overall" thus potentially accounting for local config too.

`set` always sets in the global user-level config.

We could add flags to control where values are gotten from/set to, i had just not written that for YAGNI reasons, trying to get "any config to exist at all" before "writing all the bells and whistles" but if it's something we think we need on initial implementation i could write that. 

to do that, we would need to add directory traversal/project root finding which is always a bit touchy: if i call `linkml config set --project key value` from some subdirectory within my project, i should traverse up to find `linkml_config.yaml` rather than writing a new one in place. 

We're also using `pyyaml` rather than `ruamel.yaml` so setting from the CLI necessarily overwrites any document structure/comments/etc., which i would expect to be more frequently the case in project-level config than global userwide config.

but yeah anyway if we wanted to do that i think we should probably only target the project-level config and the global config, i'm a bit wary about editing someones `pyproject.toml` (and that would also require us to move `tomli-w` into runtime dependencies) and i think its sort of pointless to set env vars from the CLI since `export` is that already. So then we could mimic e.g. `npm`'s syntax and have the default be to set the local config by default, and then have a `-g/--global` flag set global user-level config.

for `get` i could add `-v/--verbose` or `-s/--show-sources` or something and show the value as it's set across the various sources, it would require a little bit of doing since pydantic-settings doesn't expose this afaik, but i could rig it up.

so ya lmk if we want those in this PR or can follow on with them, can do

docs/config/index.md
sneakers-the-rat: discussed in https://github.com/linkml/linkml/pull/3782#discussion_r3625941347

kevinschaper (approved):
Throwing an approval on, but obviously needs conflicts sorted to merge

sneakers-the-rat:
alright rebase'd and pass'd

PR #3786, opened 2026-07-22
Fix SchemaView prefix namespace initialization

### LINKED ISSUE #3785: `Generator._init_namespaces()` fails when a generator uses `uses_schemaloader=False` and is initialised with a file path
**Describe the bug**
`Generator._init_namespaces()` fails when a generator uses
`uses_schemaloader=False` and is initialised with a file path (the normal
supported use-case).

When a file path is passed, `Generator.__post_init__` constructs a
`SchemaView` via `load_schema_wrap()`, which parses the YAML and produces a
`SchemaDefinition` whose `.prefixes` is a `dict[str, Prefix]` — values are
`linkml_runtime.linkml_model.meta.Prefix` objects.  `_init_namespaces` then
takes the `isinstance(self.schema.prefixes, dict)` branch and assigns each
value directly to the `Namespaces` map, which raises
`ValueError: Not a valid URI` because it expects a plain URI string, not a
`Prefix` object.

The bug was introduced in commit `f04170e55` (released in v1.8.7), which
added the `dict` branch to handle pre-parsed `SchemaDefinition` objects whose
`.prefixes` had been set to plain `dict[str, str]` by the caller.  That test
case worked because the values were bare strings.  The same branch is also
reached in the normal file-path case, where the values are always `Prefix`
objects — and that was not accounted for.

The fix is to unwrap `value.prefix_reference` when the value carries that
attribute:

```python
self.namespaces[key] = value.prefix_reference if hasattr(value, "prefix_reference") else value
```
The bug affects any uses_schemaloader=False generator invoked with a file
path that has explicit prefix declarations.  It has gone largely unnoticed
because most such generators bypassed _init_namespaces by constructing their
own SchemaView before calling super().__post_init__() — a pattern that
PR #3778 is now fixing, which will expose this bug more broadly.

**Version of LinkML you are using:** Present since v1.8.7, still present on current main.

Any uses_schemaloader=False generator invoked with a file path whose schema
declares explicit prefixes will trigger this.  Minimal reproduction:
```python
from linkml.generators.shaclgen import ShaclGenerator

schema_str = """
id: https://example.org/test
name: test
prefixes:
  ex: https://example.org/test/
default_prefix: ex
imports:
  - linkml:types
default_range: string
classes:
  Foo:
    attributes:
      bar:
"""
# Raises ValueError: Not a valid URI: Prefix({'prefix_prefix': 'ex', ...})
gen = ShaclGenerator(schema_str)
gen.serialize()
```

Sanjays2402:
## Summary

Fixes #3785

Unwrap parsed `Prefix` objects before adding them to generator namespaces, while preserving support for caller-provided string prefix maps.

## How was this tested?

Added a SchemaView-backed generator regression using a schema file path. The test fails on the original code and passes with this fix; all generic generator tests and Ruff checks pass.

## Areas of uncertainty

None.

## Checklist

## AI Assistance


amc-corey-cox:
@Sanjays2402 thank you for the contribution. As @Silvanoc mentioned in the issue, we've decided to go a different direction for this in this PR: https://github.com/linkml/linkml/pull/3796

PR #3792, opened 2026-07-23
Require core team approval for LinkML PRs

### LINKED ISSUE #3777: Require core-team approval by default in CODEOWNERS unless an area owner approves
PR #3756 exposed a gap: a contributor with write access merged their own PR after receiving approval from another non-core-team member, with no core-team involvement. The CODEOWNERS file did not prevent this because **most of the codebase has no required reviewer at all** under the current opt-in model.

The fix is a default `* @linkml/core-team` rule at the top of `.github/CODEOWNERS`. GitHub CODEOWNERS precedence means the last matching rule wins, so the existing per-subsystem rules (pydanticgen, yarrrml) would continue to override the default for their paths, but every other path in the repo would require a core-team approval before merge. This is exactly the "core team must approve unless an area owner approves" behaviour we want.

This directly reverses the current file's stated design. The header comment reads:

> Opt-in model: no default (`*`) rule. Paths not listed here have no required CODEOWNER and are reviewed normally.

`docs/maintainers/codeowners.md` documents the same opt-in intent. Both the header and the doc would need updating if we switch models.

**Branch protection dependency:** CODEOWNERS only enforces required review if the branch protection rule "Require review from Code Owners" is enabled on `main`. If it is not currently on, adding the default rule alone does nothing. Someone with admin access needs to verify (and enable) that setting.

The opt-in rationale was to avoid creating a new approval gate for the whole codebase when CODEOWNERS was first introduced. That concern is now outweighed by the demonstrated gap: write-access holders can and will merge without core-team sign-off, not out of bad faith, but because nothing stops them. The default rule closes that gap.

The open question before acting: do we commit to the default-rule model, or address the immediate gap differently (e.g. stricter branch protection without touching CODEOWNERS)? If we go with the default rule, the contributor hierarchy docs (`docs/maintainers/contributor-hierarchy.md`) also reference the opt-in model in passing and should be reviewed for consistency.

## Examples

Proposed change to `.github/CODEOWNERS`:

```diff
-# Opt-in model: no default (`*`) rule. Paths not listed here have no required
-# CODEOWNER and are reviewed normally. Rules only appear where someone has
-# explicitly volunteered to steward an area.
+# Default model: all paths require core-team approval unless a more specific
+# CODEOWNER rule applies. Later rules override earlier ones (last match wins).

+# --- Default: require core-team approval everywhere ---
+*  @linkml/core-team
+
 # --- CODEOWNERS and governance docs ---
 /.github/CODEOWNERS                         @linkml/core-team
 ...

 # --- Per-subsystem ownership (opt-in) ---
 # pydanticgen (overrides the default for these paths):
 /packages/linkml/src/linkml/generators/pydanticgen/  @sneakers-the-rat @kevinschaper
 ...
```

Precedence resolution for a pydanticgen PR: GitHub evaluates all matching rules and uses the last one. The default `*` matches everything; the pydanticgen rule also matches and appears later, so `@sneakers-the-rat @kevinschaper` is the required reviewer. For a PR touching `packages/linkml/src/linkml/generators/javagen/`, only `*` matches, so `@linkml/core-team` is required.

## Related

- #3140 — tracking issue for the CODEOWNERS rollout; the opt-in model was the agreed starting point there
- #3756 — the PR where the gap surfaced; @gouttegd flagged the merge as inconsistent with contributing policy

matentzn:
## Summary

Fixes #3777.

This PR updates codeowners contract to require core team approval for PRs _unless_ approved by a codeowner. Right now anyone with write access can merge _anything_. That was not the intention.

## How was this tested?

Mostly documentation changes, but need to remember to activate branch 

## Checklist

- [X] Existing tests pass locally with my changes
- [x] AFTER MERGING ENSURE THAT BRANCH PROTECTION RULES ARE UPDATED

## AI Assistance

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,11 +1,13 @@
 # LinkML CODEOWNERS — DRAFT (see linkml/linkml#3140)
 #
-# Opt-in model: no default (`*`) rule. Paths not listed here have no required
-# CODEOWNER and are reviewed normally. Rules only appear where someone has
-# explicitly volunteered to steward an area.
+# Default model: all paths require core-team approval unless a more specific
+# CODEOWNER rule applies. Later rules override earlier ones (last match wins),
+# so per-subsystem rules below take precedence over the default for their paths.
 #
 # Process for adding rules: docs/maintainers/codeowners.md
-# Later rules override earlier ones for matched paths.
+
+# --- Default: require core-team approval everywhere ---
+*  @linkml/core-team
 
 # --- CODEOWNERS and governance docs ---
 /.github/CODEOWNERS                         @linkml/core-team
@@ -13,7 +15,7 @@
 /docs/maintainers/codeowners.md             @linkml/core-team
 /docs/maintainers/generator-governance.md   @linkml/core-team
 
-# --- Per-subsystem ownership (opt-in) ---
+# --- Per-subsystem ownership (overrides the default for these paths) ---
 
 # javagen
 /docs/generators/java.rst                            @gouttegd
--- a/docs/maintainers/codeowners.md
+++ b/docs/maintainers/codeowners.md
@@ -9,23 +9,33 @@ For the underlying GitHub mechanics, see the upstream
 [CODEOWNERS documentation](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners).
 This page only covers how LinkML uses it.
 
-## The LinkML model: opt-in stewardship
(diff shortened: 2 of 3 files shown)


melonora (approved):
I don't see any objections to this. Maybe the only thing is that if a person of the core team does not have much familiarity and is reviewing that in that case even 2 people of the core team would be required, but this is more a social rule.

amc-corey-cox (approved):
Yes, I agree with this.

matentzn:
I have also updated the branch protection rules now, we should be back to restricted.

PR #3796, opened 2026-07-23
Warn hybrid schemaloader-schemaview use

### LINKED ISSUE #3795: Detect and warn on hybrid SchemaLoader/SchemaView namespace access in generators
**Describe the bug**

`self.namespaces` is a SchemaLoader-era artifact populated by `_init_namespaces`
and consumed by visitor-pattern generators (`uses_schemaloader=True`).

SchemaView-based generators (`uses_schemaloader=False` like `LinkmlGenerator`,
`OwlSchemaGenerator`) should use `self.schemaview.namespaces()` for URI resolution
instead.  Two problems arise when they access `self.namespaces`:

1. **Hybrid-design anti-pattern** — the generator mixes SchemaLoader and SchemaView
   APIs, making future refactoring harder.

2. **Silent data corruption** (issue #3785) — `_init_namespaces()` runs unconditionally
   on the SchemaView path.  It copies values from `self.schema.prefixes` into the
   `Namespaces` map, but on the SchemaView path those values are `Prefix` model
   objects (not URI strings).  `Namespaces.__setitem__` calls `str()` on them,
   which yields the `__repr__` output — e.g.
   `"Prefix(prefix_prefix='pav', prefix_reference='http://purl.org/pav/')"` —
   instead of the expected URI `"http://purl.org/pav/"`.

The only actual hybrid use in the codebase is `OwlSchemaGenerator._range_uri()`,
which calls `self.namespaces.uri_for(typ.uri)` instead of
`self.schemaview.get_uri(typ, expand=True)`.

Kitchen sink schema: `tests/linkml/test_generators/input/kitchen_sink.yaml`.

To reproduce on the *unfixed* code (main):
```python
uv run python -c "
from linkml.generators.owlgen import OwlSchemaGenerator

g = OwlSchemaGenerator('tests/linkml/test_generators/input/kitchen_sink.yaml')
assert not g.uses_schemaloader

print('Expected: Namespace(\"http://purl.org/pav/\")')
print(f'Got: {g.namespaces[\"pav\"]}')
"
```

Silvanoc:
## Summary

Fixes #3795

Detect and warn on hybrid `SchemaLoader`/`SchemaView` namespace access in generators. `self.namespaces` is a `SchemaLoader`-era artifact; generators using `uses_schemaloader=False` (`SchemaView` path) should use `self.schemaview.namespaces()`. Rename internal field to `self._namespaces`, add property that emits `DeprecationWarning` and forwards to `self.schemaview.namespaces()` on `SchemaView` path, skip `_init_namespaces()` entirely on that path to prevent silent data corruption (#3785). Fix `OwlSchemaGenerator._range_uri()` — the only actual hybrid user — to call `self.schemaview.get_uri()` instead.

## How was this tested?
- New test `test_schemaview_generator_namespaces_access_warns` in `tests/linkml/test_generators/test_linkmlgen.py` verifies `DeprecationWarning` is emitted and forwarded `Namespaces` object is functional.
- `make test-linkml` passes.

## Areas of uncertainty
None — changes are tightly scoped to the hybrid namespace access pattern.

## Checklist
- [x] My code follows the contributor guidelines (https://linkml.io/linkml/maintainers/contributing.html)

## AI Assistance
If you used AI tools while preparing this PR, you are still the author and responsible for understanding, verifying, and defending your submission. Please engage with reviewers personally rather than through your agent during feedback and revisions. See our AI Covenant (AI_COVENANT.md) for details.


packages/linkml/src/linkml/utils/generator.py
```diff
@@ -155,8 +156,39 @@ class Generator(metaclass=abc.ABCMeta):
     """Path to output file. Note all generators may not implement this
     uniformly, see https://github.com/linkml/linkml/issues/923"""
 
-    namespaces: Namespaces | None = None
-    """All prefix expansions used"""
+    _namespaces: Namespaces | None = None
```
amc-corey-cox: Renaming the field renames the constructor kwarg: `LinkmlGenerator(path, namespaces=ns)` now raises `TypeError`. That call works on main. Is it worth keeping a `namespaces=` alias?

packages/linkml/src/linkml/utils/generator.py
```diff
@@ -155,8 +156,39 @@ class Generator(metaclass=abc.ABCMeta):
...
+        pre-built :class:`~linkml_runtime.utils.namespaces.Namespaces` object
+        populated by SchemaLoader.
+
+        On the SchemaView path (``uses_schemaloader=False``) accessing this
+        property is a sign of a hybrid design anti-pattern.  A deprecation
+        warning is emitted and the call is transparently forwarded to
+        ``self.schemaview.namespaces()`` so that existing callers continue to
+        work while being nudged towards the correct API.
+        """
+        if not self.uses_schemaloader and self.schemaview is not None:
```
amc-corey-cox, line 185: This forwards unconditionally, so an explicitly injected namespace map is silently dropped on the SchemaView path — on main it's honored. Returning `self._namespaces` first when it's not None would keep injection working.

packages/linkml/src/linkml/utils/generator.py
```diff
@@ -155,8 +156,39 @@ class Generator(metaclass=abc.ABCMeta):
...
+            warnings.warn(
```
amc-corey-cox, line 186: `DeprecationWarning` is filtered by default outside `__main__`, so as a guard this mostly won't be seen. `UserWarning` would be.

packages/linkml/src/linkml/utils/generator.py
```diff
@@ -269,18 +301,24 @@ def _initialize_using_schemaloader(self, schema: Union[str, TextIO, SchemaDefini
             self.schema_defaults = loader.schema_defaults
 
     def _init_namespaces(self):
-        if self.namespaces is None:
-            self.namespaces = Namespaces()
+        # SchemaView-based generators do not need a pre-built namespace map;
+        # they use self.schemaview.namespaces() directly.  Populating
+        # self._namespaces on the SchemaView path would silently produce a
+        # broken map (Prefix objects instead of URI strings) and is unnecessary.
+        if not self.uses_schemaloader:
```
amc-corey-cox: After this return the rest of the method is unreachable — the SchemaLoader path already sets `_namespaces` at lines 266/295 before `_init_namespaces()` runs. Might be cleaner to delete the method than leave the buggy branches inside it.

tests/linkml/test_generators/test_linkmlgen.py
```diff
@@ -26,6 +27,27 @@ def test_linkmlgen_prefixes():
     assert "equipment_schema" in parsed["prefixes"]
 
 
+def test_schemaview_generator_namespaces_access_warns(kitchen_sink_path):
```
amc-corey-cox, line 31: #3786 adds a small regression test that passes unchanged against this branch. Feel free to import it here if you find it useful.

tests/linkml/test_generators/test_linkmlgen.py
```diff
@@ -26,6 +27,27 @@ def test_linkmlgen_prefixes():
...
+    assert not gen.uses_schemaloader
+
+    with pytest.warns(DeprecationWarning, match="self.namespaces.*SchemaLoader-era"):
+        ns = gen.namespaces
+
+    # The returned value must still be a functional Namespaces object
+    # forwarded from self.schemaview.namespaces() so that existing callers
+    # continue to work while being migrated.
+    assert ns is not None
+    assert ns is gen.schemaview.namespaces()
```
amc-corey-cox: `is` relies on SchemaView caching the `Namespaces` object; `==` would be less brittle.

amc-corey-cox (changes requested):
@Silvanoc - I noticed a few things that could stand some small tweaks before this goes in. Push back if any of this is too nit-picky. Thanks!

packages/linkml/src/linkml/utils/generator.py
Silvanoc: Good point. Are we missing a test for such a use-case that would have made the issue visible to me?

I can have a look at it next week.

packages/linkml/src/linkml/utils/generator.py
Silvanoc: Good point. I wasn't aware of this issue. I'll address it.

packages/linkml/src/linkml/utils/generator.py
Silvanoc: The issue is that I coupled the hybrid use warning with the returned value and they should be independent.

Meaning:
1. if hybrid use, only raise warning
2. return namespaces
   1. if injected namespaces, return it
   2. if not injected namespaces, calculate namespaces over right Schema-Loader/View path

tests/linkml/test_generators/test_linkmlgen.py
Silvanoc: I'll import it next week.

tests/linkml/test_generators/test_linkmlgen.py
Silvanoc: Ok, will fix it.

packages/linkml/src/linkml/utils/generator.py
Silvanoc: Should be resolved now.

packages/linkml/src/linkml/utils/generator.py
Silvanoc: Should be resolved now.

packages/linkml/src/linkml/utils/generator.py
Silvanoc: Should be resolved now.

tests/linkml/test_generators/test_linkmlgen.py
Silvanoc: Should be resolved now.

tests/linkml/test_generators/test_linkmlgen.py
Silvanoc: Should be resolved now.

packages/linkml/src/linkml/utils/generator.py
Silvanoc: This was a tricky one... Commit https://github.com/linkml/linkml/pull/3796/changes/586260c941193af196df1852e0d781982a3316be should address it. The comments are quite verbose, possibly too much. I've put also part of the documentation into the commit message. @amc-corey-cox how do you see it? Too verbose? Fine? More explanation needed?

Silvanoc:
@amc-corey-cox I think I addressed all the requested changes, could you give me some feedback? If you are aware of it, but have other priorities right now, I can be patient and come back to you in 1 or 2 weeks.

packages/linkml/src/linkml/utils/generator.py
amc-corey-cox: Confirmed — the kwarg is preserved on both paths and covered by the new tests. The comments are on the verbose side, but the dataclass/property default trick is subtle enough that I'd rather have them than not. Fine as-is.

packages/linkml/src/linkml/utils/generator.py
amc-corey-cox: Confirmed fixed — an injected map is now returned first, covered by test_namespaces_constructor_kwarg_injection_schemaview.

packages/linkml/src/linkml/utils/generator.py
amc-corey-cox: Confirmed — now a UserWarning.

packages/linkml/src/linkml/utils/generator.py
amc-corey-cox: Confirmed — _init_namespaces is gone entirely. Thanks!

tests/linkml/test_generators/test_linkmlgen.py
amc-corey-cox: Test imported as test_schema_view_prefix_namespaces — thanks.

tests/linkml/test_generators/test_linkmlgen.py
amc-corey-cox: Confirmed — == now. Thanks.

packages/linkml/src/linkml/utils/generator.py
```diff
@@ -162,8 +162,67 @@ class Generator(metaclass=abc.ABCMeta):
...
+
+        On the SchemaView path (``uses_schemaloader=False``) accessing this
+        property is a sign of a hybrid design anti-pattern. A deprecation
+        warning is emitted and the call is transparently forwarded to
+        ``self.schemaview.namespaces()`` so that existing callers continue to
+        work while being nudged towards the correct API.
+        """
+        # Emit a warning when a SchemaView-based generator reads self.namespaces.
+        if not self.uses_schemaloader and self.schemaview is not None:
+            warnings.warn(
```
amc-corey-cox, line 192: Sorry I missed this on my first pass. It appears this warning now fires on every `gen-json-schema` run: `JsonSchemaGenerator` is `uses_schemaloader=False` but reads `self.namespaces` in its own `__post_init__` guard (jsonschemagen.py:475), so constructing the generator trips it — the jsonschemagen test suite emits 455 of them. Changing that guard to check `self.schemaview is None` would keep the warning targeted at actual user code.

amc-corey-cox (approved):
@Silvanoc - sorry I missed one on my first pass. I think this should it ready to go. Otherwise this looks really good. We can merge without getting this change but I dislike merging with spurious warnings so I'd appreciate if if you can get this in.

packages/linkml/src/linkml/utils/generator.py
Silvanoc: Well, there's no need for that guard in [`jsonschemagen.py`:475](https://github.com/linkml/linkml/blob/b4c3261fddcad14f0221ba980f6e84f9d88eb873/packages/linkml/src/linkml/generators/jsonschemagen.py#L475), anymore. If `super().__post_init__()` the line above the guard has succeeded, then [`self.schemaview` has been created](https://github.com/linkml/linkml/blob/b4c3261fddcad14f0221ba980f6e84f9d88eb873/packages/linkml/src/linkml/utils/generator.py#L227), no need to check it. It isn't true, if [`uses_schemaloader = True`](https://github.com/linkml/linkml/blob/b4c3261fddcad14f0221ba980f6e84f9d88eb873/packages/linkml/src/linkml/generators/jsonschemagen.py#L409), but that would break many more things. So I'm simply removing that unnecessary guard.

Silvanoc: @amc-corey-cox would you mind looking at this fix and, if you agree with the solution, directly merge it?

Issue #1532, opened 2023-07-13
Validation error for datetime/date data in YAML


rly:
**Describe the bug**
When I try to validate a YAML file with datetime data against a schema with a datetime or date attribute, I get these errors:
```
✗ datetime.datetime(2007, 3, 1, 13, 0, tzinfo=datetime.timezone(datetime.timedelta(days=-1, seconds=68400))) is not of type 'string' in $.a_datetime
✗ datetime.date(2007, 3, 1) is not of type 'string' in $.a_date
```

I think this is because PyYAML auto-parses datetimes as `datetime.datetime` and `datetime.date` objects. 

Is there a way to use YAML to store datetime data and validate it? 

When YAML is passed to the validator, should it check against `datetime.datetime` and `datetime.date` objects instead of `string`?

See https://stackoverflow.com/questions/34667108/ignore-dates-and-times-while-parsing-yaml for a potential fix.

**To reproduce**
Create `test_datetime.yaml`:
```yaml
id: https://w3id.org/test
name: test
prefixes:
  linkml: https://w3id.org/linkml/
imports:
  - linkml:types

classes:

  FileRoot:
    tree_root: true
    attributes:
      a_datetime:
        range: datetime
      a_date:
        range: date
```

Create `test_datetime_data.yaml`:
```yaml
a_datetime: 2007-03-01T13:00:00-05:00
a_date: 2007-03-01
```

Run: `linkml-validate -s test_datetime.yaml test_datetime_data.yaml`

**About your computer (if applicable, please complete the following information):**
 - OS: macOS
 - Python 3.10

I have these relevant packages installed:
```
linkml                    1.5.6
linkml-dataops            0.1.0
linkml-runtime            1.5.4
PyYAML                    6.0
```

rly:
I just figured out that I can get around this error by putting quotes around the datetime/date values:
```yaml
a_datetime: "2007-03-01T13:00:00-05:00"
a_date: "2007-03-01"
```

That works and it looks like it does validate the string as a datetime or date string. Maybe this YAML gotcha should be documented somewhere? Happy to close this.

ialarmedalien:
I've been caught out by the date conversion when dealing with yaml in the past, so I vote yes. There don't seem to be any other common datatypes that get converted into objects, which makes it seem a bit random when it happens with date/times.

matentzn:
[...]

PR #3802, opened 2026-07-24
Clarify that "YAML" means "the JSON subset of YAML".

gouttegd:
## Summary

~~Fixes~~ Documents why #1532 won’t be fixed.

This PR updates the documentation about loading/dumping from/to YAML to clarify that YAML-specific features MUST NOT be used in the YAML serialisation of LinkML-defined data.

As mentioned in the discussion of #1532, this is a common trap when working with date-typed slots, when the LinkML validator will (seemingly inexplicably, and very counter-intuitively) reject YAML-formatted data in which date-typed slots are serialized directly as untagged date scalars rather than strings, because the validator expects date-typed slots to have string values.

I personally believe this behaviour is faulty and should be fixed, but since it seems to actually be the _expected_ behaviour (as per the discussion in #1532), this should at least be documented.

## How was this tested?

This is a documentation change only.

## Areas of uncertainty

None.

## Checklist

- ~~[ ] I have added tests that prove my fix/feature works~~
- ~~[ ] Existing tests pass locally with my changes~~

## AI Assistance

None.


docs/data/conversion.rst
```diff
@@ -87,6 +87,24 @@ Loading from and dumping to YAML
 The native YAML representation for LinkML is essentially identical to
 JSON.
 
+Importantly, this means that YAML-specific features, such as the
+automatic typing of untagged nodes into non-numeric scalars (e.g.
```
matentzn, line 91: > non-numeric

Just with a cursory look, its not just non-numeric:

```yaml
version: 1.10   # -> 1.1, a float. Fails a string-typed slot, and silently eats the trailing zero
label: yes      # -> True. Fails a string-typed slot
```

Both of these look like they will fail the same way the date does. Maybe ensure we should not better say "untagged scalars get YAML's implicit typing, which is not JSON's"  (without exceptions)?

docs/data/conversion.rst
```diff
@@ -87,6 +87,24 @@ Loading from and dumping to YAML
...
+dates) or, conversely, the explicit typing of nodes through tags, are
+not expected to be used and can lead to undefined behaviours.
```
matentzn, line 93: Its _undocumented_ but a bit conservative. The way I see it `a_date: !!str 2007-03-01` would work ok.  The less conservative framing is that the python tag most precisely reflect the models slot range.

matentzn (commented):
I personally would prefer to keep #1532 open - @cmungall should pass over it once when he is gifting us "the spec"

docs/data/conversion.rst
gouttegd: Not sure what your point is?

The fact is that the current “spec” does not say anything about how data is expected to be serialised in YAML, apart that it is “essentially identical to JSON”. Therefore, what happens if you use a YAML feature that has no equivalent in JSON (such as the use of an untagged non-String scalar) is completely **unspecified** and can lead to **undefined behaviour**.

The fact that untagged scalars “get YAML’s implicit typing” which will then later result in the LinkML loader failing to load the data (because the implicit typing is not what the loader expected) is the behaviour that happens when you use the LinkML-Py runtime, but it’s merely one undefined behaviour among all the undefined behaviours that could happen. Maybe another implementation would have a different behaviour – we are in unspecified territory, so _anything_ can happen.

All we need to do here (apart from writing a real spec for YAML serialisation) is to clearly state that YAML specific features are not to be used. No need to explain why (especially to give a reason that is specific to the LinkML-Py runtime).

docs/data/conversion.rst
gouttegd: > The way I see it a_date: !!str 2007-03-01 would work ok.

You don’t know about that. Maybe it would work with LinkML-Py, maybe not, but even if it did it would be accidental. An undefined behaviour is still an undefined behaviour even if it happens to do the thing you would expect.

docs/data/conversion.rst
matentzn: Argh, example was bad, and yes, the loader failure is a linkml-py behaviour, not the spec, and the doc shouldn't explain the mechanism in terms of one runtime.  Same goes for the `!!! str` - I should have looked at the wider context of the change rather than just commented on the bubble and my half-a**ed understanding :P All good. The boolean case (label: yes in JSON would be interpreted as a boolen, right?) could have been an example, but no need, now,

docs/data/conversion.rst
matentzn: Point taken, see answer to other comment.

matentzn (approved):

Ostrzyciel (approved):

Ostrzyciel:
I support this! Could we also put this into the spec? As I understand this is only the docs of the Python implementation.

Issue #3760, opened 2026-07-14
metamodel status slot has wrong `slot_uri` (`bibo:status`) and mixed semantics — term maturity (`vs:term_status` strings) vs. document lifecycle (`bibo:DocumentStatus` URIs) need separate treatment


Silvanoc:
**Describe the bug**

I'm working on adding to `expand_curie()` strict input validation: either a CURIE (that gets expanded) or a URI (which is left untouched). And such validation has raised this issue.

The metamodel's `status` slot has conflicting semantics, an incorrect `slot_uri`, and an incorrect `range` for its actual usage.

What the metamodel declares:
```yaml
status:
  domain: element
  range: uriorcurie
  slot_uri: bibo:status
  examples:
    - value: "bibo:draft"
```

**Problem 1** - wrong `slot_uri`

`bibo:status` (<http://purl.org/ontology/bibo/status>) is an OWL object property
whose `range` is `bibo:DocumentStatus` - a class of named individuals representing
document publication lifecycle states (`accepted`, `draft`, `published`, `unpublished`,
…).  Its individuals live at http://purl.org/ontology/bibo/status/<name>, so even
the example value `bibo:draft` is wrong: the correct URI would require a separate
prefix such as `bibo_status: http://purl.org/ontology/bibo/status/`.

**Problem 2** - wrong `range` for actual usage

The values actually used throughout `meta.yaml` and `array.yaml` are the bare
strings `testing` (30+ occurrences, like in [meta.yaml](https://github.com/linkml/linkml-model/blob/986c7881bae48055d3dbd9106205bc86c5215df4/linkml_model/model/schema/meta.yaml#L862) and [array.yaml](https://github.com/linkml/linkml-model/blob/986c7881bae48055d3dbd9106205bc86c5215df4/linkml_model/model/schema/array.yaml#L13)) and `unstable` (2 occurrences, like in [meta.yaml](https://github.com/linkml/linkml-model/blob/986c7881bae48055d3dbd9106205bc86c5215df4/tests/input/meta.yaml#L1373)).  These are
not URIs or CURIEs - they are plain-string values probably from the W3C SemWeb Vocab
Status vocabulary (`vs:term_status`,
<http://www.w3.org/2003/06/sw-vocab-status/ns#term_status>), which explicitly
defines its values as literals: `"stable"`, `"testing"`, `"unstable"`, `"archaic"`.

**Problem 3** - mixed semantics: terms vs. documents

[...]

PR #3804, opened 2026-07-24
fix owl canonicalization hang: add timeout (rdf graph with high symmetric blank-node structure)

### LINKED ISSUE #3803: rdf_canonicalize.py rdflib fallback hangs in certain scenarios
**Describe the bug**

During PR #3641 preparation, `test_metamodel[OwlSchemaGenerator-.owl-serialize_kwargs0]` never finished (after 20 mins) - apparently hanging instead of pass or fail.

**Root cause**

OwlSchemaGenerator.serialize() tries to canonicalize its output for reproducibility. It first tries a fast method (pyoxigraph), and falls back to a slower one (rdflib) if the fast one can't handle the graph. See https://github.com/linkml/linkml/pull/3703

 `pyoxigraph` rejected the metamodel's graph because of one bad value - the status slot mentioned in https://github.com/linkml/linkml/issues/3760 which is declared as a URI but is actually set to a bare word producing an invalid relative IRI. **This triggers a slow rdflib fallback that actually hangs.**

Timing indicates sensitivity: the full graph (5658 triples, 838 blank nodes) never finishes in 20+ minutes. Removing just 33 triples (5 owl:unionOf list structures added by the newer "covering axiom" feature) and it finishes in 16 seconds. So a small increase in graph size pushed the slow fallback's runtime from "slow" to "never returns." 

**NOTE: A status slot's bad value is what triggers the fallback in the first place. Fixing that would avoid the slow path entirely for this schema. See discussion in #3760 and linkml/linkml-model#263 - out of scope**

**Two issues**

(1) The RDF slow fallback has no time limit. It's shared code used by several generators, so any schema that trips the same conditions can hang the same way.  `canonicalize_rdf_graph()` tries pyoxigraph's fast RDFC-1.0 canonicalization first; when pyoxigraph rejects a graph (relative IRIs, literal predicates - intentional fallback triggers), it calls `rdflib.compare.to_canonical_graph()` instead. That fallback algorithm has no bound and can blow up on graphs with symmetric blank-node structure:
-- a reduced metamodel OWL graph (covering-axiom owl:unionOf lists disabled, 5625 triples) canonicalizes via this fallback in 16.2s;
-- the same graph with covering axioms enabled (default, 5658 triples - only 33 more) does not complete in over 20 minutes. 
This utility is shared by OwlSchemaGenerator, RDFGenerator, SHACLGenerator, ShExGenerator, rdflib_dumper, rdf_dumper - any user schema that trips the same pyoxigraph-rejection condition is exposed to the same hang, not just this metamodel.

(2) `compare_rdf()` test-comparison helper is separately unbounded (and its near-duplicate `tests/linkml_runtime/support/compare_rdf.py`) calls `rdflib.compare.to_isomorphic()` and `graph_diff()` directly - same unbounded algorithm family as first bug, but a completely separate call path .  The code even has a standing comments acknowledging this pain point has been noted but never fixed:

https://github.com/linkml/linkml/blob/8e443f15faa4673acc3345923ea84199f73e7f50/tests/linkml/utils/compare_rdf.py#L86

https://github.com/linkml/linkml/blob/8e443f15faa4673acc3345923ea84199f73e7f50/tests/linkml/test_biolink_model/test_biolink_model.py#L37

## Impact
Blocked a OWL generator test (hangs instead of failing fast), and is a risk for any schema/generator.

1.11.1

noelmcloughlin:
## Summary

Fixes #3803

- `rdf_canonicalize.py` timeout hardening

- `compare_rdf.py` ×2 refactor - the target test no longer hangs and passes reliably (5/5 runs, ~34s each - it hits the 30s timeout and falls back to non-canonical output every time, but `compare_rdf`'s comparison still succeeds)

## How was this tested?

- new tests in `test_rdf_canonicalize.py`

## Areas of uncertainty

This PR treats `meta.yaml has malformed status value` as out of scope. The trade-off is the OWL/RDF/TTL metamodel generators stay ~30s slower for this one schema until #3760 is resolved upstream.

## Checklist

## AI Assistance
LLM assisted

### DIFF
--- a/packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
+++ b/packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
@@ -34,15 +34,80 @@ with stable blank node labels and sorted triples.
    is valid Turtle (PN_LOCAL_ESC), but rdflib's notation3 parser rejects
    it because it conflicts with the statement-terminator dot.  We
    post-process the output to expand such CURIEs to full ``<IRI>`` form.
+
+6. **Pathological blank-node structure**: ``rdflib.compare.to_canonical_graph``
+   (used for the fallback in limitation 2) has no complexity bound and can
+   take an unbounded amount of time on graphs with highly symmetric blank-node
+   structure (e.g. many structurally-similar ``owl:Restriction``/``rdf:List``
+   nodes). To guarantee ``canonicalize_rdf_graph`` always returns, this
+   canonicalization step runs in a subprocess with a timeout
+   (``fallback_timeout``, default :data:`DEFAULT_FALLBACK_TIMEOUT_SECONDS`).
+   If it does not finish in time, the subprocess is killed and the graph is
+   serialized directly (no blank-node canonicalization at all), with a
+   warning. Output is still valid RDF, just not guaranteed deterministic
+   across runs.
 """
 
 import io
+import multiprocessing
 import re
 import warnings
 
 import pyoxigraph as ox
 import rdflib
 from rdflib.compare import to_canonical_graph
+
+#: Default time budget, in seconds, for the rdflib fallback canonicalization
+#: (see limitation 6 above) before giving up and serializing non-canonically.
+DEFAULT_FALLBACK_TIMEOUT_SECONDS = 30.0
+
+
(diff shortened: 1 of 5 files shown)


noelmcloughlin:
Flaky test failure (https://github.com/linkml/linkml/issues/3421)

PR #3807, opened 2026-07-27
Improve configurability openapigen

Silvanoc:
## Summary

Add minor features and improvements to the OpenAPI Generator:
- Option to keep sub-schemas declared in the template, but not referenced anywhere in the API.
- Option to inline enums instead of creating separate objects for them.
- Detect dangling references.
- Avoid elements renaming due to canonalization.

## How was this tested?

Specific tests have been added.

## Areas of uncertainty

None, no known user yet of this generator.

## Checklist

## AI Assistance


Silvanoc:
As [discussed](https://obo-communitygroup.slack.com/archives/C06SMUB1NK1/p1785321862838509?thread_ts=1785312677.571759&cid=C06SMUB1NK1) with @noelmcloughlin, annotations in the model can be an alternative approach to ensure that classes, types that are not transitive dependencies are nevertheless shown in the OpenAPI `components/schema` section.

The biggest difference is that the `--keep-unreferenced` provides that configurability at OpenAPI-template-level without touching the model. The annotations approach somehow shifts an OpenAPI concern to the model.

noelmcloughlin:
I will review. Background is that I manually created [openapi.yaml](https://github.com/noelmcloughlin/ai-linkmo/blob/main/lib/api/openapi.yaml) to reflect key classes (endpoints) in [ai-atlas-nexus](https://github.com/IBM/ai-atlas-nexus/tree/main/src/ai_atlas_nexus/ai_risk_ontology/schema) and discussed CLI use case with [Inge](https://github.com/ingelise) (_she presented to Linkml community in May and expressed interest in CLI_).

OpenAPI generator seemed perfect. More generally I see many use cases (particularly in my employment/industry).

So happy to make time to familiarize myself with the feature. I see [docs](https://github.com/linkml/linkml/blob/main/docs/generators/openapi.rst) mention **OpenAPI template** so that looks best starting point.  Probably an example exists in `tests/linkml/test_generators/` ?

Silvanoc:
> I will review.

But please start with the bug fix #3765, which is release-relevant!

> Background is that I manually created [openapi.yaml](https://github.com/noelmcloughlin/ai-linkmo/blob/main/lib/api/openapi.yaml) to reflect key classes (endpoints) in [ai-atlas-nexus](https://github.com/IBM/ai-atlas-nexus/tree/main/src/ai_atlas_nexus/ai_risk_ontology/schema) and discussed CLI use case with [Inge](https://github.com/ingelise) (_she presented to Linkml community in May and expressed interest in CLI_).

What I see in your OpenAPI looks completely different from what I'm trying to accomplish. We declare almost all schemas in the `components/schema` section (which provides in the end the data model) and the `paths` section has almost no schemas and only references to the `components/schema` section. As of now the OpenAPI Generator only touches the `components/schema` section and leaves the rest untouched.

> So happy to make time to familiarize myself with the feature. I see [docs](https://github.com/linkml/linkml/blob/main/docs/generators/openapi.rst) mention **OpenAPI template** so that looks best starting point. Probably an example exists in `tests/linkml/test_generators/` ?

You will find examples in the different added tests. I'd recommend you to look at commit level to have the context.

noelmcloughlin:
Is this ready for review?  I can make some time.

Silvanoc:
> Is this ready for review? I can make some time.

@noelmcloughlin now it's rebased on current `main` and tests are passing, so it's ready for review.

noelmcloughlin:
Questions before review - as not covered in documentation.

1. What is the benefit (`--keep-unreferenced / -k`) of keeping template-declared schemas even when no endpoint references them?  Is this some kind of flexibility - maybe today I am not referencing those schema but tomorrow I might?

2. Should `--inline-enums / -e` not be default `true` - if the feature is worth introducing, and maybe OpenApi specification does not understand enum concept, then maybe we always want to perform the translation?  And if that is the case, then is the option useful, since `--inline-enums false` would always generate invalid JSON components/schemas?

3. Do dangling references result from deficiencies in LinkML schema translation (for example  #3940)?

4. And `preserve_names=True)` keeps original LinkML class names (e.g. lowercase class name).  I see `jsongen.py` defaults to False` but you have identified that `openapigen.py` works best preserving names.  I have not used `jsongen.py` enough to remember why that generator does not required preserving names by default.
```
class JsonSchema(dict):
    OPTIONAL_IDENTIFIER_SUFFIX = "__identifier_optional"
    PRESERVE_NAMES: bool = False

   ....

    preserve_names: bool = False
    """If true, preserve LinkML element names in JSON Schema output (e.g., for $defs, properties, $ref targets)."""
```

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -243,32 +249,53 @@ def _rename(self, name_map: dict[str, str], element: dict | list) -> dict | list
...
         to OpenAPI 3.0.3 compat, and apply any OpenAPI↔LinkML name renames.
         """
-        referenced_schemas = endpoint_ref_linkml_names.copy()
+
+        referenced_schemas = req_linkml_names.copy()
         for openapi_schema in openapi_schemas.values():
-            self._find_references(openapi_schema, referenced_schemas)
-        while set(openapi_schemas.keys()).difference(referenced_schemas):
+            referenced_schemas = self._find_references(openapi_schema, referenced_schemas)
+        if not self.keep_unreferenced:
```
noelmcloughlin: Will removing the `while` loop and this update not be regression?  A single pass increases risk of loss.

Could you add unit test as seems only Foo is generated in the resulting template?  
Also add classes with spaces (i.e. `Hoo Bar`, `foo bar`, `), or any other corner case you think of - the unit tests and kitchen sink schema may need more missing/presence examples to catch leaks. 
```
id: https://example.org/foobar
name: foobar
prefixes:
imports:
  - linkml:types
default_range: string

classes:
  Foo:
    description: the only class reachable from the endpoint
    attributes:
      name:
        range: string

  Baz:
    description: unrelated class, not reachable from Foo, but present in the schema
    attributes:
      bar_ref:
        range: Bar

  Bar:
    description: should be pruned - only referenced by the also-unreachable Baz
    attributes:
      value:
        range: string
```

Silvanoc:
> 1. What is the benefit (`--keep-unreferenced / -k`) of keeping template-declared schemas even when no endpoint references them?  Is this some kind of flexibility - maybe today I am not referencing those schema but tomorrow I might?

I knew I had documented it somewhere... but clearly [not in the best place](https://github.com/linkml/linkml/pull/3807/changes#diff-3584f14f8af5168b9bfc19b41a53f4e896dca43cecd407a27843d15e164873f3R199).

In the project where we are using it we define schemas that are not needed directly for the API, but needed for validation of some documents "hidden" in the payload (e.g. because it's a compressed file).

Some documentation is needed. The question is where... in the help it's probably too much. The commit message is probably not enough. Probably in the `.rst` documentation, right?

Silvanoc:
> 3. Do dangling references result from deficiencies in LinkML schema translation (for example  [Relational metaslots are under-specified, unenforced, gen-owl drops #3940](https://github.com/linkml/linkml/issues/3940))?

I have to admit that I forgot which issue that I faced raised that need...

I suppose that dangling references in the template should be possible to be detected validating the template before instantiation (since we require the template to declare placeholders for all the schemas being used) before it gets resolved/instantiated.

Dangling references inside of the LinkML schema should be detected by LinkML itself. At least I'd expect the JSON-Schema generation to detect it.

So perhaps the final validation is superfluous.

Silvanoc:
> 2. Should `--inline-enums / -e` not be default `true` - if the feature is worth introducing, and maybe OpenApi specification does not understand enum concept, then maybe we always want to perform the translation?  And if that is the case, then is the option useful, since `--inline-enums false` would always generate invalid JSON components/schemas?

I didn't have any profound thoughts about it. I suppose that I simply took the result before flag introduction as the default. But I can't really tell now.

OpenAPI does understand enum concept, but with the idiom `any_of` enumerating the options. But in general it's neither a native concept, nor a common thing to do.

I'll change that default.

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -322,9 +349,10 @@ def serialize(self, template_file: str = "", **kwargs) -> str:
         endpoint_ref_openapi_names = self._find_referenced_schemas()  # OpenAPI names referenced by endpoints
         openapi_schemas = self._template["components"]["schemas"]  # schemas provided by the OpenAPI template
         # collect the LinkML names referenced by endpoints (seed for sanitizing below)
-        endpoint_ref_linkml_names: set[str] = {
-            openapi_schemas[n]["x-linkml-source"] for n in endpoint_ref_openapi_names
-        }
+        if self.keep_unreferenced:
```
noelmcloughlin, line 375: As with condition above, will this conditional  `if self.keep_unreferenced` ("_keep schemas listed in the template even if not referenced by any endpoint_") not disable pruning?  As above can you add unit tests that targets absence more directly in `test_openapigen.py` and enrich kitchen sink schema if needed.

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -243,32 +249,53 @@ def _rename(self, name_map: dict[str, str], element: dict | list) -> dict | list
...
         for openapi_schema in openapi_schemas.values():
             openapi_schema.pop("title", None)
         openapi_schemas = cast(dict, self._fix_openapi_spec(openapi_schemas))
         if name_map:
             openapi_schemas = cast(dict, self._rename(name_map, openapi_schemas))
+        if self.inline_enums:
+            openapi_schemas = self._inline_enum_schemas(openapi_schemas)
         return openapi_schemas
 
+    def _inline_enum_schemas(self, data_schemas: dict) -> dict:
```
noelmcloughlin, line 296: Add unit test to cover types and update code to ensure `types` (see comment earier) are not mistaken as enums?  Then going further, extend unit tests with `inline_enums=True` and inverse `inline_enums=False`

```
id: https://example.org/foobarenum
name: foobarenum
prefixes:
default_range: string

types:
  FixedType:
    description: a LinkML type (not an enum) constrained to a single value
    base: str
    uri: xsd:string
    equals_string: "fixed-value"

classes:
  Foo:
    attributes:
      name:
        range: string
```

packages/linkml/src/linkml/generators/openapigen.py
noelmcloughlin, line 213: Linkml types becomes enums too.

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -12,6 +12,8 @@
 from openapi_spec_validator import OpenAPIV30SpecValidator
 from openapi_spec_validator import validate as openapi_validate
 from openapi_spec_validator.validation.validators import SpecValidator as OaSpecValidator
+from referencing import Registry, Resource
```
noelmcloughlin: `referencing` is not in https://github.com/linkml/linkml/blob/main/packages/linkml/pyproject.toml, but in my IDE I see it in `uv.lock` so maybe pulled in as transient dependency, so ok.

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -353,6 +381,38 @@ def serialize(self, template_file: str = "", **kwargs) -> str:
         indented_schemas = textwrap.indent(schemas_yaml, "    ")
         result = text_before_schemas + "  schemas:\n" + indented_schemas
 
+        # Check for dangling $ref references using the referencing library
+        result_obj = yaml.safe_load(result)
+        schemas = result_obj.get("components", {}).get("schemas", {})
+        registry = Registry().with_resources(
```
noelmcloughlin: I understand the `openapigen.py` is just emitting flat schema entries `#/components/schemas/<name>` so you could just simplify this:
 
- delete lines 15-16, lines 387-393, lines 408-412
- add simple check instead.
```
schema_names = set(result_obj.get("components", {}).get("schemas", {}).keys())
...
dangling = [
    ref
    for ref in all_refs
    if not ref.startswith("#/components/schemas/")
    or ref.removeprefix("#/components/schemas/") not in schema_names
]
```
- see if unit tests pass

noelmcloughlin (changes requested):
Sorry for delay, just spent some time on this review with comments, mainly concerning unit tests (major) and few other minor nit (minor).

noelmcloughlin:
> > 3. Do dangling references result from deficiencies in LinkML schema translation (for example  [Relational metaslots are under-specified, unenforced, gen-owl drops #3940](https://github.com/linkml/linkml/issues/3940))?
> 
> I have to admit that I forgot which issue that I faced raised that need...
> 
> I suppose that dangling references in the template should be possible to be detected validating the template before instantiation (since we require the template to declare placeholders for all the schemas being used) before it gets resolved/instantiated.
> 
> Dangling references inside of the LinkML schema should be detected by LinkML itself. At least I'd expect the JSON-Schema generation to detect it.
> 
> So perhaps the final validation is superfluous.

No need to change - I'm not saying its wrong or over-engineered. Users want generators to work, and not be blocked on principle that Linkml should have captured the danglers.

noelmcloughlin:
> > 1. What is the benefit (`--keep-unreferenced / -k`) of keeping template-declared schemas even when no endpoint references them?  Is this some kind of flexibility - maybe today I am not referencing those schema but tomorrow I might?
> 
> I knew I had documented it somewhere... but clearly [not in the best place](https://github.com/linkml/linkml/pull/3807/changes#diff-3584f14f8af5168b9bfc19b41a53f4e896dca43cecd407a27843d15e164873f3R199).
> 
> In the project where we are using it we define schemas that are not needed directly for the API, but needed for validation of some documents "hidden" in the payload (e.g. because it's a compressed file).
> 
> Some documentation is needed. The question is where... in the help it's probably too much. The commit message is probably not enough. Probably in the `.rst` documentation, right?

You could check a few other `docs/generators/*.rst` to see how others are handling.  I have mostly experience of enriching java RST file (with Damian) - I am not sure which generator documentation is exemplar example otherwise.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: Two different topics in this comment:

---

**while loop**

The original code was not needed. I was thinking on transitive dependencies, therefore I used the `while` loop to follow all the indirections. But since **`_find_references` returns a complete list of direct and transitive dependencies**, the `for` loop 2 lines down suffice. It takes every single schema and if it's not in the list, it can be directly removed.

---

**tests**

`keep_unreferenced` is only for resolving/instantiating/materializing schemas provided in the `components/schemas` part of the OpenAPI template and keeping them. Defaultwise a schema under `components/schemas` that is not directly or transitively needed by the endpoints gets removed.

I have the feeling that your expectation isn't aligned with mine and therefore with my implementation. That would mean **poor documentation**, not an error on your side 😄 

The tests use [this OpenAPI schema](https://github.com/linkml/linkml/pull/3807/changes#diff-48adebc4dfc1982d5c7b5d701d84b1c3887bb58bea5febdc64cd6b530429a6ed) to validate it. In which Person is referenced, but not OpaqueEvent. All other `kitchen_sink` elements are irrelevant.

Does it clarify it a bit?

IMO `kitchen_sink` provides enough corner cases (like names with spaces), I only need to make use of them in the tests. I'll have a look at that.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: Adding it explicitly is probably meaningful. I'll do it.

Silvanoc:
> > > 3. Do dangling references result from deficiencies in LinkML schema translation (for example  [Relational metaslots are under-specified, unenforced, gen-owl drops #3940](https://github.com/linkml/linkml/issues/3940))?
> > 
> > 
> > I have to admit that I forgot which issue that I faced raised that need...
> > I suppose that dangling references in the template should be possible to be detected validating the template before instantiation (since we require the template to declare placeholders for all the schemas being used) before it gets resolved/instantiated.
> > Dangling references inside of the LinkML schema should be detected by LinkML itself. At least I'd expect the JSON-Schema generation to detect it.
> > So perhaps the final validation is superfluous.
> 
> No need to change - I'm not saying its wrong or over-engineered. Users want generators to work, and not be blocked on principle that Linkml should have captured the danglers.

Found it! There is [a test](https://github.com/linkml/linkml/pull/3807/changes#diff-7e60318aa42a67fe1ed0b1564f7b7128b22d7578692de0bf9dad2278b7383f7fR41) with the kind of dangling reference that I had in mind (others are also possible): a schema placeholder referring to a LinkML element that does not exist.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: I don't understand what you mean.

noelmcloughlin: Oops - sorry - I was jumping around from code at Line 276 and forgot this and just have partial notes:

- In lines 167-168 above `type` / `equals_string` sets the key is "const"
```
        if type_def.equals_string is not None:
            schema["const"] = str(type_def.equals_string)
```
Later in `_santize_schemas` this method is called over the schemas:
```
        openapi_schemas = cast(dict, self._fix_openapi_spec(openapi_schemas))
```
And lines 199-201 below (in this method) the "const" key changes to "enum" key:
```
199  for key, value in element.items():
200      if key == "const":
201          fixed_element["enum"] = [value]
```
However I got distracted by work and picked up at line 276 - but then decided adding unit tests and flexing ths code with `inline_enums=True/False` would be easier.

packages/linkml/src/linkml/generators/openapigen.py
noelmcloughlin: Yes, please bear with me as its steep learning curve with (a) OpenAPI and (b) your generator (and JSON domain).   If you could review this minimal example - I don't think this is valid, so I wanted to question if removing the old code which iteratively scanned/pruned the invalid entries this example shows.

Schema:
```
# /tmp/foobar.yaml
id: https://example.org/foobar
name: foobar
classes:
  Foo: {}
  Baz:
    attributes:
      drinks:
        range: Bar
  Bar: {}
```
Template (I assume its valid) - one and only one endpoint references `Foo` class.
```
# /tmp/foobar-tmpl.yaml
openapi: 3.0.3
info:
  title: t
  version: '1.0.0'
paths:
  /foo:
    get:
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Foo'
components:
  schemas:
    Foo:
      type: object
      x-linkml-schema: https://example.org/foobar
      x-linkml-source: Foo
```
Reproducers

1. Output includes Foo/Baz/Bar
```bash
python -m linkml.generators.openapigen /tmp/foobar.yaml -t /tmp/foobar-tmpl.yaml -k
````

2. Output includes Foo/Bar
```bash
python -m linkml.generators.openapigen /tmp/foobar.yaml -t /tmp/foobar-tmpl.yaml
```

packages/linkml/src/linkml/generators/openapigen.py
noelmcloughlin: See https://github.com/linkml/linkml/pull/3807/changes#r3909107266 above with bit more detail.

packages/linkml/src/linkml/generators/openapigen.py
noelmcloughlin: Regarding documentation, its always poor :-)

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: `referencing` is a transitive dependency of the [pre-existing `jsonschema` dependency](https://github.com/linkml/linkml/blob/1643ff83c120a1668eb89beb7394c627282e8376/packages/linkml/pyproject.toml#L48) and of the newly introduced `openapi-schema-validator`. So no need to add it.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: The `_sanitize_schemas` code in main was way too complex and the addition of the `keep_unreferenced` flag was making it worse. I've changed it strongly and it now it should work as expected.

Additionally, I've documented the use of the newly added flags: https://github.com/linkml/linkml/blob/improve-configurability-openapigen/docs/generators/openapi.rst#advanced-usage

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: Applied.

Silvanoc:
> 4. And `preserve_names=True)` keeps original LinkML class names (e.g. lowercase class name).  I see `jsongen.py` defaults to False`but you have identified that`openapigen.py`works best preserving names.  I have not used`jsongen.py` enough to remember why that generator does not required preserving names by default.
> 
> ```
> class JsonSchema(dict):
>     OPTIONAL_IDENTIFIER_SUFFIX = "__identifier_optional"
>     PRESERVE_NAMES: bool = False
> 
>    ....
> 
>     preserve_names: bool = False
>     """If true, preserve LinkML element names in JSON Schema output (e.g., for $defs, properties, $ref targets)."""
> ```

One of the commits (not adding a link because they change on rebasing) adds a test to reproduce the issue. The resolution (next commit) does use the JsonSchema preserve_names flag.

Adding features I add the tests to the same commit adding the feature.

Fixing bugs I first add a commit that helps reproducing the bug and then another commit fixing it. So you can checkout the bug-reproducing commit, see it failing. Then checkout the fix and see it succeeding.

Silvanoc:
> > 2. Should `--inline-enums / -e` not be default `true` - if the feature is worth introducing, and maybe OpenApi specification does not understand enum concept, then maybe we always want to perform the translation?  And if that is the case, then is the option useful, since `--inline-enums false` would always generate invalid JSON components/schemas?
> 
> I didn't have any profound thoughts about it. I suppose that I simply took the result before flag introduction as the default. But I can't really tell now.
> 
> OpenAPI does understand enum concept, but with the idiom `any_of` enumerating the options. But in general it's neither a native concept, nor a common thing to do.
> 
> I'll change that default.

I think that is more common to keep enums as separate schemas in `components/schemas` so that they can referenced multiple times, than inlining. I'm therefore not changing the default.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: I've also added tests for transitive dependencies. Please check.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: Should be addressed by the last commits. Please check.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: Should be addressed by now with the last commits. Please check.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: Should be addressed by now with the last commits. Please check.

Silvanoc:
@matentzn I wonder why this PR is requiring an approval of the *core-team*, since according [CODEOWNERS.md](https://github.com/linkml/linkml/blob/main/.github/CODEOWNERS) the approval of @noelmcloughlin or mine would suffice. Perhaps the CODEOWNERS.md doesn't work as documented in that file itself?

packages/linkml/src/linkml/generators/openapigen.py
noelmcloughlin: Nice fix, clean refactor.

packages/linkml/src/linkml/generators/openapigen.py
noelmcloughlin: Yes, just checked.

1. Prune+drop covered by `test_unreferenced_chain_pruned_by_default` ( `assert "Foo")
2. Prune+keep covered by `test_unreferenced_chain_pruned_by_default` asserts all three presetn - `Foo, Foo Bar, Baz Qux` (because Foo Bar is template-declared in the fixture , so seed pulls `Baz Qux` transitively).

Could you add a third test to close this off please?

assert that -k does not pull in classes that are neither template-declared nor transitively required from a declared schema - prove -k stays scoped to the template rather than reverting to "dump everything.  **See my reproducer above**,  running `-k` would now only return `Foo` so its just a final regression edge to coveer.

Naive users (like me) who start using the generator are prone to mistakes so this collection of tests should capture everything regarding pruning.

packages/linkml/src/linkml/generators/openapigen.py
noelmcloughlin: Yes, checked.
Covered by `test_inline_enums_does_not_inline_types` using teh "schema_type_and_enums" fixture. And  `test_inline_enums_disabled_keeps_types_and_enums_separate` covers `inline_enums false".

Another issue - The inlining_enums logic still breaks when an endpoint references an enum directly. Method `_inline_enum_schemas` deletes from components/schemas without checking whether something outside `elem_schemas` references by name.  Minimal reproducer.
```
# /tmp/schema2.yaml
id: https://example.org/g1
name: g1
default_range: string
enums:
  FixedEnum:
    permissible_values:
      FOO: {}
      BAR: {}
```

```
# /tmp/schema2-tmpl.yaml
openapi: 3.0.3
info:
  title: t
  version: '1.0.0'
paths:
  /fixed-enum:
    get:
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/FixedEnum'
components:
  schemas:
    FixedEnum:
      type: object
      x-linkml-schema: https://example.org/g1
      x-linkml-source: FixedEnum
```

`_inline_enum_schemas` deletes entry from components/schemas without checking whether something outside `elem_schemas` points at it.

```
 python -m linkml.generators.openapigen /tmp/schema2.yaml -t /tmp/schema2-tmpl.yaml -e 

ValueError: Dangling $ref in generated OpenAPI spec: #/components/schemas/FixedEnum
```

noelmcloughlin: Also can you update logic so renaming a type in the template does not bypass the `all_types()` guard and add unit test like this reproducer:
```
# /tmp/renaming/yaml
id: https://example.org/g1
name: g1
default_range: string
types:
  FixedType:
    base: str
    uri: xsd:string
    equals_string: "fixed-value"
classes:
  Foo:
    attributes:
      name:
        range: string
```

```
# /tmp/renaming-tmpl.yaml
openapi: 3.0.3
info:
  title: t
  version: '1.0.0'
paths:
  /fixed:
    get:
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Fixed'
components:
  schemas:
    Fixed:
      type: object
      x-linkml-schema: https://example.org/g1
      x-linkml-source: FixedType
```

The above work fine without `-k` or without renaming.  So just this one fails:
```
python -m linkml.generators.openapigen /tmp/renaming.yaml -t /tmp/renaming-tmpl.yaml -e 
<frozen runpy>:128: RuntimeWarning: 'linkml.generators.openapigen' found in sys.modules after import of package 'linkml.generators', but prior to execution of 'linkml.generators.openapigen'; this may result in unpredictable behaviour
ValueError: Dangling $ref in generated OpenAPI spec: #/components/schemas/Fixed
```

docs/generators/openapi.rst
```diff
@@ -119,6 +119,31 @@ generated by :doc:`JsonSchemaGenerator <json-schema>`:
...
+
+Inlining Enums
+^^^^^^^^^^^^^^
+
+By default enumerations are declared as separate schemas in the
+``components/schemas`` of the resulting OAD (OpenAPI Description) document.
+
+Using the *inline_enums* option (object constructor boolean argument
+``inline_enums``, CLI flag ``--inline-enums``) it's possible to change
+that behavior so that inlined into the using ``components/schemas``
```
noelmcloughlin, line 144: The sentence feels incomplete, something missing - 

"_t's possible to change that behavior so that inlined into the using_`

tests/linkml/test_generators/test_openapigen.py
```diff
@@ -147,3 +162,211 @@ def test_renaming(input_path, kitchen_sink_path):
...
+    """
+    head_path = str(input_path("openapi/spec-comments.openapi.yaml"))
+    result = OpenApiGenerator(kitchen_sink_path).serialize(head_path)
+    # comments are dropped by a YAML round-trip but preserved by text handling
+    assert "# top-level comment must survive round-trip" in result
+    assert "# this endpoint comment must survive" in result
+    # original quoting style is preserved (round-trip would normalise this)
+    assert "version: '1.0.0'" in result
+    # the untouched template prefix is emitted byte-for-byte
+    template_text = Path(head_path).read_text()
```
noelmcloughlin, line 182: Not your issue - but could line 43 `open(head_path).read()` be updated to use this `Path(head_path).read` pattern.

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -353,8 +404,34 @@ def serialize(self, template_file: str = "", **kwargs) -> str:
...
+                    refs.append(ref)
+                for value in obj.values():
+                    _collect_refs(value, refs)
+            elif isinstance(obj, list):
+                for item in obj:
+                    _collect_refs(item, refs)
+
+        all_refs: list[str] = []
+        _collect_refs(result_obj, all_refs)
+        dangling = [
```
noelmcloughlin, line 424: The dangling check is too strict for valid schema and rejects all the reusable OpenApi template sections names.  The easy fix would be to only check check  refs that start with `#/components/schemas/` and skip the rest (or resolve each against its own section -which is more robust but could be done in future PR).

I am not sure if this is regression against main or always existed. 
```
# /tmp/schema3.yaml
id: https://example.org/g1
name: g1
default_range: string
classes:
  Foo:
    attributes:
      name:
        range: string
```

```
# /tmp/schema3-tmpl.yaml
openapi: 3.0.3
info:
  title: t
  version: '1.0.0'
paths:
  /foo:
    get:
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Foo'
        '404':
          $ref: '#/components/responses/NotFound'
components:
  responses:
    NotFound:             # <------- this
      description: not found
  schemas:
    Foo:
      type: object
      x-linkml-schema: https://example.org/g1
      x-linkml-source: Foo
```

```
python -m linkml.generators.openapigen /tmp/schema3.yaml -t /tmp/schema3-tmpl.yaml -e 

ValueError: Dangling $ref in generated OpenAPI spec: #/components/responses/NotFound
```

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -243,31 +270,54 @@ def _rename(self, name_map: dict[str, str], element: dict | list) -> dict | list
...
+        }
+        if not enum_schemas:
+            return data_schemas
+
+        def _replace_refs(obj):
+            if isinstance(obj, dict):
+                if "$ref" in obj:
+                    ref_name = obj["$ref"].split("/")[-1]
+                    if ref_name in enum_schemas:
+                        return enum_schemas[ref_name]
```
noelmcloughlin: When an enum is referenced twice, the `_replace_refs` logic returns same `enum_schemas[ref_name]` for each repeated reference and the ouput serializes with YAML anchors.

Note : _duplicate references seem to be a modeling pattern not covered extensively in kitchen sink perhaps, because I raised separate issue about two slots sharing same inverse value, and then found other relational slots not covered by specification docs - my point is I presume only shared `enums` is a problem in OpenAPI template but maybe you can think of other slotDefinition members to worry besides this `enum` use case _

Could you check what fix is appropriate and add unit tests - since I presume the anchors are not wanted.

```
# /tmp/schema4.yaml
# enum referenced from two classes
id: https://example.org/g1
name: g1
default_range: string
enums:
  FixedEnum:
    permissible_values:
      FOO: {}
      BAR: {}
classes:
  Foo:
    attributes:
      color:
        range: FixedEnum
  Bar:
    attributes:
      color:
        range: FixedEnum
```

```
# /tmp/schema4-tmpl.yaml
# endpoint for both classes referencing FixedEnum in schema
openapi: 3.0.3
info:
  title: t
  version: '1.0.0'
paths:
  /foo:
    get:
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Foo'
  /bar:
    get:
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Bar'
components:
  schemas:
    Foo:
      type: object
      x-linkml-schema: https://example.org/g1
      x-linkml-source: Foo
    Bar:
      type: object
      x-linkml-schema: https://example.org/g1
      x-linkml-source: Bar
```

```
python -m linkml.generators.openapigen /tmp/schema4.yaml -t /tmp/schema4-tmpl.yaml -e 
```

**Output is corrupted**
```
# /tmp/schema4-tmpl.yaml
# endpoint for both classes referencing FixedEnum in schema
openapi: 3.0.3
info:
  title: t
  version: '1.0.0'
paths:
  /foo:
    get:
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Foo'
  /bar:
    get:
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Bar'
components:
  schemas:
    Foo:
      type: object
      additionalProperties: true
      description: ''
      properties:
        color: &id001
          type: string
          description: ''
          enum:
          - FOO
          - BAR
    Bar:
      type: object
      additionalProperties: true
      description: ''
      properties:
        color: *id001
```

packages/linkml/src/linkml/generators/openapigen.py
```diff
@@ -243,31 +270,54 @@ def _rename(self, name_map: dict[str, str], element: dict | list) -> dict | list
...
+                return {k: _replace_refs(v) for k, v in obj.items()}
```
noelmcloughlin, line 346: The `-e` is dropping slot-level `description` and replace by the template "" value.

BUT seem my comment below - on reflection, may be beyond your  PR scope - your choice.

```
# /tmp/schema5.yaml

id: https://example.org/g1
name: g1
default_range: string
enums:
  FixedEnum:
    permissible_values:
      FOO: {}
      BAR: {}
classes:
  Foo:
    description: Foo class, I am
    attributes:
      color:
        description: the color of foo
        range: FixedEnum
```

```
# /tmp/schema5-tmpl.yaml
#
openapi: 3.0.3
info:
  title: t
  version: '1.0.0'
paths:
  /foo:
    get:
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Foo'
components:
  schemas:
    Foo:
      type: object
      x-linkml-schema: https://example.org/g1
      x-linkml-source: Foo
```


```
python -m linkml.generators.openapigen /tmp/schema5.yaml -t /tmp/schema5-tmpl.yaml -e 
<frozen runpy>:128: RuntimeWarning: 'linkml.generators.openapigen' found in sys.modules after import of package 'linkml.generators', but prior to execution of 'linkml.generators.openapigen'; this may result in unpredictable behaviour

# /tmp/schema5-tmpl.yaml
#
openapi: 3.0.3
info:
  title: t
  version: '1.0.0'
paths:
  /foo:
    get:
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Foo'
components:
  schemas:
    Foo:
      type: object
      additionalProperties: true
      description: 'Foo class, I am'
      properties:
        color:
          type: string
          description: ''             # Dropped here
          enum:
          - FOO
          - BAR
```

noelmcloughlin: I started going further finding more issues (if schema declares a slot and template overrides who wins / what about `slot_usage`, etc) but a rabbit hole is revealed -> outside this PR scope.

I will do analysis and create separate issue.

noelmcloughlin (changes requested):
I have added my review and tried to include as many reproducers as needed to speed your analysis up.  If something is out of scope, say so, and we can track via new issue.

packages/linkml/src/linkml/generators/openapigen.py
noelmcloughlin: Leave this out of scope. I will raise Issue (todo) and PR (https://github.com/linkml/linkml/pull/3961)

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: Good catch.

Regression test and fix added (two separate commits). Please check.

tests/linkml/test_generators/test_openapigen.py
Silvanoc: Fixed

docs/generators/openapi.rst
Silvanoc: Fixed

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: No, the anchors are not desired. I don't know if they are valid, but I personally consider that they would damage the readability of the OAD. Regression and fix added. Please check.

Silvanoc:
> I have added my review and tried to include as many reproducers as needed to speed your analysis up. If something is out of scope, say so, and we can track via new issue.

@noelmcloughlin you've raised a couple of issues that appear due to the newly introduced features, therefore I consider the corresponding regression tests and fixes part of this PR. Thus I've tried to address them all here. Thanks for the intensive testing!

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: The last added test should address this issue. Please check.

packages/linkml/src/linkml/generators/openapigen.py
noelmcloughlin: Hi @Silvanoc  doing first pass - I realize that #3961 cannot fix this issue.

Could you address in this PR - unless you object and template `description: ""` implies intent to eclipse the linkml schema slot description value which I think is not logical intent.

noelmcloughlin (changes requested):
Just added a comment - https://github.com/linkml/linkml/pull/3807#discussion_r3950730024

noelmcloughlin:
I forgot to log this issue - https://github.com/linkml/linkml/issues/3970 - its trivial on `main` branch - outside PR scope.

Also this on `main` branch: https://github.com/linkml/linkml/issues/3960

fyi

packages/linkml/src/linkml/generators/openapigen.py
noelmcloughlin: Some kind of future `--include-anchors` CLI arg?   If nobody screams ...

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: Should be fixed now. Please check.

packages/linkml/src/linkml/generators/openapigen.py
Silvanoc: I would wait for someone needing it before implementing it...

Silvanoc:
> I forgot to log this issue - #3970 - its trivial on `main` branch - outside PR scope.
> 
> Also this on `main` branch: #3960
> 
> fyi

@noelmcloughlin would you agree on addressing these bugs after this PR?

noelmcloughlin (approved):
LGTM, thanks @Silvanoc

Silvanoc:
@matentzn I don't understand why, but this PR is requiring someone from the core-team to approve.

The ["self-contained area" CODEOWNERS rule for the openapigen](https://github.com/linkml/linkml/blob/f43da63a37d04aac858322b5c2dabd25544c8cd3/.github/CODEOWNERS#L42) should in theory allow @noelmcloughlin and me to review each other's contributions and merge them. Why isn't working here? I can't identify any file out of the openapigen area...

Silvanoc:
> @matentzn I don't understand why, but this PR is requiring someone from the core-team to approve.
> 
> The ["self-contained area" CODEOWNERS rule for the openapigen](https://github.com/linkml/linkml/blob/f43da63a37d04aac858322b5c2dabd25544c8cd3/.github/CODEOWNERS#L42) should in theory allow @noelmcloughlin and me to review each other's contributions and merge them. Why isn't working here? I can't identify any file out of the openapigen area...

The schemas that are in `tests/linkml/test_generators/input/`, but outside of `tests/linkml/test_generators/input/openapi/` are the issue! I think that it make sense to keep them outside of the openapi directory, so we will wait for the approval of the core-team.

Silvanoc:
@noelmcloughlin by putting schemas for the openapigen tests in `tests/linkml/test_generators/input/` (instead of using its subdirectory `openapi`) we are blocking ourselves from having a "self-contained area". Although the schemas that we are adding for the OpenAPI generator could be used for other tests, they aren't as of now.

Therefore I'm considering seriously moving those schemas to `tests/linkml/test_generators/input/openapi/`, at least until someone else wants to use them an "promotes" them to the generator-independent area. Would you agree?

noelmcloughlin:
Yes - do that for the present justification you gave. Saves everyone time :-)

PR #3818, opened 2026-07-29
pydanticgen: YAMLRoot-style input coercion for SimpleDict with `--coerce-input` flag

### LINKED ISSUE #1862: Improve pydantic support for inlining as CompactDict SimpleDict

kevinschaper:
## Summary

This adds opt-in SimpleDict support to pydanticgen using `mode="before"` field validators 

Fixes #1862

One decision in here that I'm not so confident about here is that the includes.py as an additional file rather than inlined into the generated pydantic - I guess it's transitively imported and a downstream user doesn't necessarily need to care, but the name name is inflexible etc. 

## AI Assistance

This is split off from an experiment with Claude Fable (from its first week of availability) to fully convert LinkML to use Pydantic classes and swap out Pythongen classes. I didn't have any intention to unload that all on anyone to review, but I do think there are bite sized chunks that can be pulled out, and I guess that makes this the amuse bouche.


packages/linkml/src/linkml/generators/pydanticgen/pydanticgen.py
```diff
@@ -239,6 +241,15 @@ class PydanticGenerator(OOCodeGenerator, LifecycleMixin):
     the default templates will be used.
     """
     extra_fields: Literal["allow", "forbid", "ignore"] = "forbid"
+    coerce_input: bool = False
```
sneakers-the-rat, line 244: name is not very specific. input coercion has a specific meaning in pydantic

packages/linkml/src/linkml/generators/pydanticgen/pydanticgen.py
```diff
@@ -638,18 +649,43 @@ def generate_slot(self, slot: SlotDefinition, cls: ClassDefinition) -> SlotResul
                 collection_key = None
             if slot.inlined is False or collection_key is None or slot.inlined_as_list is True:
                 result.attribute.range = f"list[{result.attribute.range}]"
+                if self.coerce_input:
+                    result.attribute.coerce_to_list = True
+                    if slot.range in self.schemaview.all_classes() and self.schemaview.is_inlined(slot):
+                        inlined_key = self.generate_inlined_list_key_name(slot)
+                        if inlined_key is not None:
+                            result.attribute.inlined_list_key = inlined_key
+                            helper = InlinedListCoercion().render(self._template_environment())
```
sneakers-the-rat, line 658: super janky and not how the template system works

sneakers-the-rat: https://linkml.io/linkml/generators/pydantic.html#linkml.generators.pydanticgen.template.PydanticTemplateModel.render

sneakers-the-rat: https://linkml.io/linkml/generators/pydantic.html#templates

packages/linkml/src/linkml/generators/pydanticgen/pydanticgen.py
```diff
@@ -638,18 +649,43 @@ def generate_slot(self, slot: SlotDefinition, cls: ClassDefinition) -> SlotResul
...
                 else:
                     result.attribute.range = f"dict[{collection_key}, {result.attribute.range}]"
+                    coerce_keyed = True
+                if coerce_keyed and self.coerce_input:
+                    key_name = self.generate_collection_key_name(slot_ranges)
+                    if key_name is not None:
+                        result.attribute.keyed_dict_key = key_name
+                        if len(slot_ranges) == 1:
+                            result.attribute.keyed_dict_value = self.generate_collection_value_name(slot)
+                        helper = KeyedCollectionCoercion().render(self._template_environment())
```
sneakers-the-rat, line 684: yeah the whole point of the template system is to design out the need for a ton of imperative rendering code like this. the models hold the values to render and drive the rendering. the generator constructs the models.

packages/linkml/src/linkml/generators/pydanticgen/pydanticgen.py
```diff
@@ -879,6 +915,72 @@ def generate_collection_key(
...
+                continue
+            identifier_slot = self.schemaview.get_identifier_slot(slot_range, use_key=True)
+            if identifier_slot is not None:
+                alias = identifier_slot.alias if identifier_slot.alias else identifier_slot.name
+                key_names.add(make_valid_python_identifier(underscore(alias)))
+        if len(key_names) == 1:
+            return key_names.pop()
+        return None
+
+    def generate_collection_value_name(self, slot_def: SlotDefinition) -> str | None:
```
sneakers-the-rat, line 940: surely these three methods are done elsewhere as well and not unique to the pydantic generator.

packages/linkml/src/linkml/generators/pydanticgen/pydanticgen.py
```diff
@@ -1137,6 +1239,7 @@ def render(self) -> PydanticModule:
 
         base_model = PydanticBaseModel(
             extra_fields=self.extra_fields,
+            coerce_numbers_to_str=self.coerce_input,
```
sneakers-the-rat, line 1242: why would this be implicitly coupled

packages/linkml/src/linkml/generators/pydanticgen/template.py
```diff
@@ -210,6 +215,28 @@ class PydanticAttribute(PydanticTemplateModel):
...
+    """
+    When True (multivalued slots rendered as lists), a ``mode="before"``
+    validator is generated that wraps a single value into a singleton list.
+    """
+    inlined_list_key: str | None = None
+    """
+    For multivalued slots inlined as a list of objects: the identifier/key
+    slot (or first required slot) of the range class, allowing dict-keyed
+    input forms to normalize to a list with keys injected.
+    """
```
sneakers-the-rat, line 239: this is also janky, i see this is used to make deduplicated methods. there shouldn't be any need for this. we can use annotated types so that these behave as proper types - like i don't really understand what "simple dicts" are, they don't seem very simple to me, but whatever they are should probably work like `SimpleDict[T]` s.t.

```python
SimpleDict = Annotated[T, BeforeValidator(...)]
```

sneakers-the-rat: ah i see i have already said exactly this https://github.com/linkml/linkml/issues/1862#issuecomment-1971960288

packages/linkml/src/linkml/generators/pydanticgen/template.py
```diff
@@ -245,6 +272,46 @@ class PydanticValidator(PydanticAttribute):
     template: ClassVar[str] = "validator.py.jinja"
 
 
+class KeyedCollectionCoercion(PydanticTemplateModel):
```
sneakers-the-rat, line 275: yeah not how template system works. fable fumbled pretty much 100% of the design here.

packages/linkml/src/linkml/generators/pydanticgen/template.py
```diff
@@ -280,7 +347,24 @@ def _validators(self) -> dict[str, PydanticValidator] | None:
...
+            }
+        )
+        validators.update(
+            {
+                f"{k}__list": PydanticListValidator(**v.model_dump())
+                for k, v in self.attributes.items()
+                if v.coerce_to_list
+            }
+        )
+        return validators
```
sneakers-the-rat, line 367: this is what should happen in the gennerator. totally unclear why um we are... recreating models that we just created? but different?

packages/linkml/src/linkml/generators/pydanticgen/templates/base_model.py.jinja
```diff
@@ -8,6 +8,9 @@ class {{ name }}(BaseModel):
         arbitrary_types_allowed = True,
         use_enum_values = True,
         strict = {{ strict }},
+{% if coerce_numbers_to_str %}
+        coerce_numbers_to_str = True,
+{% endif %}
```
sneakers-the-rat, line 13: see the above `strict` declaration

packages/linkml/src/linkml/generators/pydanticgen/templates/coerce_inlined_list.py.jinja
```diff
@@ -0,0 +1,35 @@
+{% raw %}
+def _coerce_inlined_list(value: Any, key_name: str) -> Any:
+    """
+    Normalize the input forms of a multivalued slot inlined as a list of
+    objects: a dict keyed by ``key_name`` (the identifier/key slot of the
+    range class, or its first required slot) becomes a list with the keys
+    injected, and a single object is wrapped into a singleton list. Mirrors
+    the normalization YAMLRoot performs for the equivalent dataclass models.
```
sneakers-the-rat, line 8: link to the docs. this is not detail that should be injected in every downstream module

packages/linkml/src/linkml/generators/pydanticgen/templates/coerce_keyed_collection.py.jinja
```diff
@@ -0,0 +1,70 @@
+{% raw %}
+def _coerce_keyed_collection(value: Any, key_name: str, value_name: Optional[str] = None) -> Any:
```
sneakers-the-rat, line 2: surely this code exists elsewhere. if it doesn't it should. then we just inject the source of that function instead of creating these raw templates which aren't statically inspectable and create an unnecessary point of divergence. the python models should do whatever this is identically to how the pydantic models do it. that's only possible with shared code.

sneakers-the-rat (changes requested):
i'll return with more concrete revisions but fable has given us yet another example of how LLMs fail to understand even basic design systems that aren't rich in the training set and reliably produce hyperverbose, technically-procedurally-correct-but-bafflingly-convoluted code.

if this is an expected input data type, i'm not sure why it would be optional to support it, but i also don't really understand the need for the underlying thing that's being done here. this should take the form of a type like `SimpleDict[T]` that is generated exactly once and everything else just wraps the inner type with that annotation. 50 lines not 500. more work to review and fix llm code than to just write it correctly.

the template system isn't the problem here - the last human being to learn it described it as "you really cooked here"

edit: there are an infinite number of ways that any given abstract problem can be instantiated in code. design systems try and reduce that number into a small grammar of possible approaches that can be understood by a standing body of contributors, so that shared theories and understanding about how things work can be built. i am still trying to wrap my mind around this as one of many examples of how LLMs break this basic socio-technological covenant in flagrantly broken ways. this may be one way to do the problem, but it is as if it was created by an alien trying to guess at what locally idiomatic code might look like, even though I know you as a person understand it better. the several new points of code divergence and the use of the templating system - which tries to reduce procedural spaghetti - to introduce a few *new kinds* of procedural spaghetti are astonishing.

PR #3819, opened 2026-07-30
Fix 1813 pythongen any of union

### LINKED ISSUE #1521: "any_of" misbehaving
**Describe the bug**
Using "any_of" to get a union of ranges doesn't seem to work at all

**To reproduce**
1. Create schema
<details>
  <summary>schema.yaml</summary>

``` yaml
# schema.yaml
id: http://example.org/diag-anyof
name: diag-anyof
prefixes:
imports:
  - linkml:types
default_range: string

classes:
  ParentRange:
  ChildRange1:
    is_a: ParentRange
  ChildRange2:
    is_a: ParentRange
  BaseClass:
    slots:
      - some_slot
  ChildClass1:
    is_a: BaseClass
    slot_usage:
      some_slot:
        any_of:
          - range: ChildRange1
          - range: ChildRange2
  ChildClass2:
    is_a: BaseClass
    slots:
      - other_slot

slots:
  some_slot:
    range: ParentRange
  other_slot:
    any_of:
      - range: ChildRange1
      - range: ChildRange2
```
</details>

2. Generate Python classes: `gen-python ./schema.yaml > ./schema.py`

3. Observe resulting Python module:

<details>
  <summary>schema.py</summary>

``` python
# Auto generated from schema2.yaml by pythongen.py version: 0.9.0
# Generation date: 2023-07-04T17:36:13
# Schema: diag-anyof
#
# id: http://example.org/diag-anyof
# description:
# license: https://creativecommons.org/publicdomain/zero/1.0/

import dataclasses
import re
from jsonasobj2 import JsonObj, as_dict
from typing import Optional, List, Union, Dict, ClassVar, Any
from dataclasses import dataclass
from linkml_runtime.linkml_model.meta import EnumDefinition, PermissibleValue, PvFormulaOptions

from linkml_runtime.utils.slot import Slot
from linkml_runtime.utils.metamodelcore import empty_list, empty_dict, bnode
from linkml_runtime.utils.yamlutils import YAMLRoot, extended_str, extended_float, extended_int
from linkml_runtime.utils.dataclass_extensions_376 import dataclasses_init_fn_with_kwargs
from linkml_runtime.utils.formatutils import camelcase, underscore, sfx
from linkml_runtime.utils.curienamespace import CurieNamespace
from linkml_runtime.linkml_model.types import String

metamodel_version = "1.7.0"
version = None

# Overwrite dataclasses _init_fn to add **kwargs in __init__
dataclasses._init_fn = dataclasses_init_fn_with_kwargs

# Namespaces
LINKML = CurieNamespace('linkml', 'https://w3id.org/linkml/')
DEFAULT_ = CurieNamespace('', 'http://example.org/diag-anyof/')


# Types

# Class references



class ParentRange(YAMLRoot):

    class_class_uri: ClassVar[URIRef] = URIRef("http://example.org/diag-anyof/ParentRange")
    class_class_curie: ClassVar[str] = None
    class_name: ClassVar[str] = "ParentRange"
    class_model_uri: ClassVar[URIRef] = URIRef("http://example.org/diag-anyof/ParentRange")


class ChildRange1(ParentRange):

    class_class_uri: ClassVar[URIRef] = URIRef("http://example.org/diag-anyof/ChildRange1")
    class_class_curie: ClassVar[str] = None
    class_name: ClassVar[str] = "ChildRange1"
    class_model_uri: ClassVar[URIRef] = URIRef("http://example.org/diag-anyof/ChildRange1")


class ChildRange2(ParentRange):

    class_class_uri: ClassVar[URIRef] = URIRef("http://example.org/diag-anyof/ChildRange2")
    class_class_curie: ClassVar[str] = None
    class_name: ClassVar[str] = "ChildRange2"
    class_model_uri: ClassVar[URIRef] = URIRef("http://example.org/diag-anyof/ChildRange2")


@dataclass
class BaseClass(YAMLRoot):

    class_class_uri: ClassVar[URIRef] = URIRef("http://example.org/diag-anyof/BaseClass")
    class_class_curie: ClassVar[str] = None
    class_name: ClassVar[str] = "BaseClass"
    class_model_uri: ClassVar[URIRef] = URIRef("http://example.org/diag-anyof/BaseClass")

    some_slot: Optional[Union[dict, ParentRange]] = None

        if self.some_slot is not None and not isinstance(self.some_slot, ParentRange):
            self.some_slot = ParentRange()

@dataclass
class ChildClass1(BaseClass):

    class_class_uri: ClassVar[URIRef] = URIRef("http://example.org/diag-anyof/ChildClass1")
    class_class_curie: ClassVar[str] = None
    class_name: ClassVar[str] = "ChildClass1"
    class_model_uri: ClassVar[URIRef] = URIRef("http://example.org/diag-anyof/ChildClass1")

    some_slot: Optional[Union[dict, ParentRange]] = None

        if self.some_slot is not None and not isinstance(self.some_slot, ParentRange):
            self.some_slot = ParentRange()

@dataclass
class ChildClass2(BaseClass):

    class_class_uri: ClassVar[URIRef] = URIRef("http://example.org/diag-anyof/ChildClass2")
    class_class_curie: ClassVar[str] = None
    class_name: ClassVar[str] = "ChildClass2"
    class_model_uri: ClassVar[URIRef] = URIRef("http://example.org/diag-anyof/ChildClass2")

    other_slot: Optional[str] = None

        if self.other_slot is not None and not isinstance(self.other_slot, str):
            self.other_slot = str(self.other_slot)

# Enumerations


# Slots
class slots:
    pass

slots.some_slot = Slot(uri=DEFAULT_.some_slot, name="some_slot", curie=DEFAULT_.curie('some_slot'),
                   model_uri=DEFAULT_.some_slot, domain=None, range=Optional[Union[dict, ParentRange]])

slots.other_slot = Slot(uri=DEFAULT_.other_slot, name="other_slot", curie=DEFAULT_.curie('other_slot'),
                   model_uri=DEFAULT_.other_slot, domain=None, range=Optional[str])

slots.ChildClass1_some_slot = Slot(uri=DEFAULT_.some_slot, name="ChildClass1_some_slot", curie=DEFAULT_.curie('some_slot'),
                   model_uri=DEFAULT_.ChildClass1_some_slot, domain=ChildClass1, range=Optional[Union[dict, ParentRange]])
```
</details>

4. Generate JSON-Schema: `gen-json-schema ./schema.yaml > ./schema.schema.json`

3. Observe resulting JSON-Schema:

<details>
  <summary>schema.schema.json</summary>

``` json
{
    "$defs": {
        "BaseClass": {
            "description": "",
            "properties": {
                "some_slot": {
                    "$ref": "#/$defs/ParentRange"
                }
            },
            "title": "BaseClass",
            "type": "object"
        },
        "ChildClass1": {
            "description": "",
            "properties": {
                "some_slot": {
                    "$ref": "#/$defs/ParentRange",
                    "anyOf": [
                        {
                            "$ref": "#/$defs/ChildRange1"
                        },
                        {
                            "$ref": "#/$defs/ChildRange2"
                        }
                    ]
                }
            },
            "title": "ChildClass1",
            "type": "object"
        },
        "ChildClass2": {
            "description": "",
            "properties": {
                "other_slot": {
                    "anyOf": [
                        {
                            "$ref": "#/$defs/ChildRange1"
                        },
                        {
                            "$ref": "#/$defs/ChildRange2"
                        }
                    ],
                    "type": "string"
                },
                "some_slot": {
                    "$ref": "#/$defs/ParentRange"
                }
            },
            "title": "ChildClass2",
            "type": "object"
        },
        "ChildRange1": {
            "description": "",
            "title": "ChildRange1",
            "type": "object"
        },
        "ChildRange2": {
            "description": "",
            "title": "ChildRange2",
            "type": "object"
        },
        "ParentRange": {
            "description": "",
            "title": "ParentRange",
            "type": "object"
        }
    },
    "$id": "http://example.org/diag-anyof",
    "$schema": "http://json-schema.org/draft-07/schema#",
    "additionalProperties": true,
    "metamodel_version": "1.7.0",
    "title": "diag-anyof",
    "type": "object",
    "version": null
}
```
</details>

6. Create small inspecting program

<details>
  <summary>inspect-schema.py</summary>

``` python
#! /usr/bin/env python
#  inspect-schema.py

schema_file = "./schema.yaml"
print("Schema: " + schema_file)

view :SchemaView = SchemaView(schema_file)
view.imports_closure()

for class_name in view.all_classes():
    print("Class: " + class_name)
    linkml_class = view.get_class(class_name)
    for slot in view.class_slots(class_name):
        print(" Slot: " + slot)
        induced_slot = view.induced_slot(slot, class_name)
        print("    Range: {}".format(induced_slot.range))
        print("    Any of:")
        for item_any in induced_slot.any_of:
            print("      " + item_any.range)
```
</details>

7. Observe result:

<details>
  <summary>inspection output</summary>

``` shell
Schema: ./schema.yaml
Class: ParentRange
Class: ChildRange1
Class: ChildRange2
Class: BaseClass
  Slot: some_slot
    Range: ParentRange
    Any of:
Class: ChildClass1
  Slot: some_slot
    Range: ParentRange
    Any of:
      ChildRange1
      ChildRange2
Class: ChildClass2
  Slot: other_slot
    Range: string
    Any of:
      ChildRange1
      ChildRange2
  Slot: some_slot
    Range: ParentRange
Any of:
```
</details>

### Expected behavior
In the resulting Python classes (point 3 above) I was expecting
1. `ChildClass1.some_slot` to be of type `Optional[Union[dict, ChildRange1, ChildRange2]]`, since it's only limiting the range, not expanding it
2. `ChildClass2.other_slot` to be of type `Optional[Union[dict, ChildRange1, ChildRange2]]`, the resulting `Optional[str]` is completely wrong

In the resulting JSON-Schema (point 5 above) I was expecting
1. `"$defs".ChildClass1.properties` to be `{"some_slot": {"anyOf": [ {"$ref": "#/$defs/ChildRange1"},{"$ref": "#/$defs/ChildRange2"}]}}`, or any equivalent schema resulting from [factoring](https://json-schema.org/understanding-json-schema/reference/combining.html#factoring-schemas).
2. `"$defs".ChildClass1.properties` to be `{"other_slot": {"anyOf": [ {"$ref": "#/$defs/ChildRange1"},{"$ref": "#/$defs/ChildRange2"}]},{"some_slot": {"$ref": "#/$defs/ParentRange"}}}`, or any equivalent schema resulting from [factoring](https://json-schema.org/understanding-json-schema/reference/combining.html#factoring-schemas).

In the result of the inspection (point 7 above) I was expecting
1. `[ChildRange1, ChildRange2]` to be listed as the range of `ChildClass1.some_slot`
2. `[ChildRange1, ChildRange2]` to be listed as the range of `ChildClass2.other_slot` and `ParentRange` as the range of `ChildClass2.some_slot`

### About your computer (if applicable, please complete the following information):**
 - OS: MacOS Ventura
 - Version LinkML: 1.5.6

**Additional context**

### LINKED ISSUE #1813: Python Generator does not support keyword `any_of`
**Describe the bug**
Range of attributes can be defined by key word `range` or by key word `any_of`. The latter allows to define union of ranges, e.g. union of boolean and float. However current implementation of python generator does not support `any_of`. Instead, all attributes, whose data type is defined as `any_of` are mapped to python variables of type string. 

**To reproduce**
1. Take the following yaml file

```yaml
id: https://example.com/any_of
name: AnyOf
prefixes:
  ex: https://example.com/
imports:
  - linkml:types
default_range:  string

classes:
  AnyOf:
    attributes:
      example:
        any_of:
         - range: boolean
         - range: float
```

2. Run python generator. Resulting python class looks as follows

```python
# Auto generated from test.yaml by pythongen.py version: 0.0.1
# Generation date: 2024-01-05T15:20:26
# Schema: cryptography
#
# id: https://example.com/any_of
# description:
# license: https://creativecommons.org/publicdomain/zero/1.0/

import dataclasses
import re
from jsonasobj2 import JsonObj, as_dict
from typing import Optional, List, Union, Dict, ClassVar, Any
from dataclasses import dataclass
from linkml_runtime.linkml_model.meta import EnumDefinition, PermissibleValue, PvFormulaOptions

from linkml_runtime.utils.slot import Slot
from linkml_runtime.utils.metamodelcore import empty_list, empty_dict, bnode
from linkml_runtime.utils.yamlutils import YAMLRoot, extended_str, extended_float, extended_int
from linkml_runtime.utils.dataclass_extensions_376 import dataclasses_init_fn_with_kwargs
from linkml_runtime.utils.formatutils import camelcase, underscore, sfx
from linkml_runtime.utils.curienamespace import CurieNamespace
from linkml_runtime.linkml_model.types import String

metamodel_version = "1.7.0"
version = None

# Overwrite dataclasses _init_fn to add **kwargs in __init__
dataclasses._init_fn = dataclasses_init_fn_with_kwargs

# Namespaces
EX = CurieNamespace('ex', 'https://example.com/')
LINKML = CurieNamespace('linkml', 'https://w3id.org/linkml/')
DEFAULT_ = CurieNamespace('', 'https://example.com/any_of/')


# Types

# Class references



@dataclass
class AnyOf(YAMLRoot):

    class_class_uri: ClassVar[URIRef] = EX["any_of/AnyOf"]
    class_class_curie: ClassVar[str] = "ex:any_of/AnyOf"
    class_name: ClassVar[str] = "AnyOf"
    class_model_uri: ClassVar[URIRef] = URIRef("https://example.com/any_of/AnyOf")

    example: Optional[str] = None

        if self.example is not None and not isinstance(self.example, str):
            self.example = str(self.example)

# Enumerations


# Slots
class slots:
    pass

slots.anyOf__example = Slot(uri=DEFAULT_.example, name="anyOf__example", curie=DEFAULT_.curie('example'),
                   model_uri=DEFAULT_.anyOf__example, domain=None, range=Optional[str])
```

Data type of class' attribute `example` is of type `Optional[str]`

**Expected behavior**

Data type of class' attribute `example` should be of type `Optional[Union[bool, float]] `. Constructor and slot definition must be update, too.

HendrikBorgelt:
## Summary

Fixes #1813, Fixes #1521. Related to #1483.

This PR fixes issue #1813 ("Python Generator does not support keyword `any_of`") and issue #1521 ("`any_of` misbehaving"), and makes partial progress on #1483.

**The problem:** When a LinkML schema says a field can be "one of several types or classes" (using the `any_of` keyword), the Python code generator (`pythongen.py`) completely ignored that and just gave the field a generic, wrong type. This caused two concrete problems:
1. The generated Python type hint was wrong (e.g. a field that should be `Union[Dataset, Catalogue]` just came out as `Any` or a plain string).
2. When actually loading data into these generated classes, there was no logic to figure out which of the possible types a piece of data should become. This meant the data stayed an untyped, generic object instead of a real class instance — which then made exporting that data to RDF/Turtle crash with `ValueError: Class JsonObj not found in schema`, because the export code had no idea what class it was looking at.

**The fix, in short:** the Python generator now builds a real `Union[TypeA, TypeB, ...]` type hint for `any_of` fields, and generates code that tries each possible type in turn and picks whichever one actually fits the data. The same fix is also applied to `exactly_one_of` (a closely related keyword meaning "must match exactly one of these types, not zero and not several").

### What's fixed, in a bit more detail

- **Type hints:** a field using `any_of: [{range: A}, {range: B}]` now gets the Python type hint `Union[A, B]` instead of collapsing to something generic.
- **Actually building the right object:** when data is loaded, the generated code now tries constructing type `A` first, and if that fails, tries `B`, and so on, using whichever one succeeds first. This is a deliberate, predictable rule (whichever type is listed first in the schema wins if the data could match more than one) — not a guess.
- **`exactly_one_of` support:** for fields that must match *exactly one* possible type, the generator now tries every option and checks that precisely one of them fits, raising a clear error if none fit or if more than one fits (which would mean the data itself is ambiguous/invalid).
- **Two smaller bugs found and fixed along the way:**
  - When a field only had one possible type listed, the generated code had a small Python syntax slip (a missing comma) that made it crash immediately with `TypeError: 'function' object is not iterable`.
  - There was a mismatch between the code that decides "should this type be fully embedded, or just referenced by its ID" for the type hint, versus the code that decides the same thing for figuring out which Python imports are needed — meaning sometimes the generated file could ask for an import that wasn't actually needed. Both places now share one function so they always agree.

### What's not covered by this PR

- `all_of` and `none_of` are not touched — they mean something fundamentally different (an object can't literally *be* two unrelated classes at once, and "none of these types" doesn't have a resulting type at all), so they need a different, separate approach.
- Fields that are lists (multivalued) of `any_of`/`exactly_one_of` types are not covered — that's a bigger, separate piece of work. A test is included that locks in today's (unchanged) behavior for these, so a future change touching this is a deliberate choice, not an accident.
- This PR only touches the Python dataclass generator (`pythongen.py`). It does not touch the JSON Schema generator, the Pydantic generator (which already handles `any_of` correctly), or the `linkml_runtime` package.

### Functions changed, briefly

All changes are in `packages/linkml/src/linkml/generators/pythongen.py`.

- **`_resolve_branch_inlined`** *(new)* — decides, for one possible type in an `any_of`/`exactly_one_of` list, whether it should be treated as a full embedded object or just referenced by its ID. Used by both the type-hint code and the import-collecting code, so they can't disagree.
- **`_boolean_combinator_branch_slots`** *(new)* — takes a field with `any_of`/`exactly_one_of` and turns each listed option into something the rest of the generator can treat like an ordinary single-type field, so existing logic can be reused instead of duplicated.
- **`_flatten_boolean_combinator_range`** *(new, small helper)* — cleans up the generated type-hint text so it doesn't end up with type hints nested inside type hints.
- **`class_reference_type`** *(modified)* — builds the Python type hint for a field. It now checks for `any_of`/`exactly_one_of` first and builds a proper `Union[...]` from all the options; unchanged for every ordinary field.
- **`gen_postinit`** *(modified)* — generates the code that runs right after an object is created to fix up/validate its fields. It now recognizes `any_of`/`exactly_one_of` fields and hands them off to the new logic below, before the code path that used to skip them entirely.
- **`_boolean_combinator_coercion_expr`** *(new)* — for one possible type, decides exactly how to try building it from the raw data (as an empty object, from a plain value, or from a nested dictionary of fields).
- **`_gen_boolean_combinator_postinit`** *(new)* — writes the actual generated code block that tries each possible type in order (for `any_of`) or checks that exactly one fits (for `exactly_one_of`), including the error messages when nothing fits or too many things fit.
- **`add_slot_range` / `add_one_range`** *(modified, inside `gen_import_list`)* — figures out which Python `import` statements the generated file needs. Updated so it also looks at `any_of`/`exactly_one_of` options, not just the field's main type, and reuses `_resolve_branch_inlined` so it doesn't disagree with the type-hint code.

## How was this tested?

- Added a new set of unit tests to `tests/linkml/test_generators/test_pythongen.py` covering: correct type hints, correct type-picking (including deliberately ambiguous and deeply-nested cases), the "no match" and "too many matches" error cases for `exactly_one_of`, and a test that pins today's unchanged behavior for list-of-`any_of` fields.
- Flipped the existing `test_slot_any_of` compliance test (in `tests/linkml/test_compliance/test_boolean_slot_compliance.py`) from marked-as-not-implemented to a real, passing check for the Python dataclass generator.
- Ran the full `any_of`/`exactly_one_of`/boolean-expression compliance test suite: 707 passed, 0 failed.
- Regenerated two existing test snapshots (the LinkML metamodel's own generated Python, and a generic test-fixture schema's generated Python) since the fix legitimately changes their output — both were reviewed by hand to confirm the diffs are limited to exactly the fields using `any_of`.
- Manually confirmed the original real-world crash (`ValueError: Class JsonObj not found in schema` when exporting to RDF) no longer happens with this fix.
- Also ran the full test suite (`uv run pytest`) unscoped: 3 failures + 1 teardown error remain, all pre-existing and unrelated to this change — 3 in `test_rdf_canonicalize.py` (hardcode a Unix-only subprocess `PATH`, breaking on Windows) and 1 in `test_pydanticgen.py::test_template_noblack` (a Windows file-locking issue in a shared test fixture's teardown). Confirmed via targeted reruns that neither touches `pythongen.py`, `test_pythongen.py`, or `test_boolean_slot_compliance.py`.

## Areas of uncertainty

- **Tie-break rule for ambiguous data:** when a value could structurally match more than one candidate type in an `any_of` list, this PR resolves it by trying candidates in the exact order they're declared in the schema and taking the first success. This matches the resolution style already used elsewhere in the generator, but reviewers may want to double check this is the right default rather than, say, erroring out on ambiguity.
- **`exactly_one_of` always builds every candidate:** unlike `any_of` (which stops at the first match), `exactly_one_of` must try every candidate to detect if more than one matches — meaning for deeply nested schemas, every candidate's nested structure gets fully constructed on every attempt, not just until the first success. This is a real but (I believe) acceptable cost given what `exactly_one_of` is supposed to guarantee.
- **Scope boundary with `all_of`/`none_of`:** I've deliberately left these two combinators alone (see "What's not covered" above) since they don't have a clean type-union equivalent. Would appreciate a second opinion on whether that boundary is drawn in the right place.

## Checklist

## AI Assistance

**AI tool used:** Claude Code, under my direction. I reviewed the generated code and tests line by line, ran and verified the test suite myself, and take full responsibility for everything in this PR.


packages/linkml/src/linkml/generators/pythongen.py
```diff
@@ -937,6 +1042,113 @@ def gen_constructor(self, cls: ClassDefinition) -> str | None:
...
+        dedup_names = list(dict.fromkeys(branch_names))  # de-dup, preserve declared order -- this *is* the tie-break
+        tuple_str = ", ".join(dedup_names) + ("," if len(dedup_names) == 1 else "")
+        union_label = f"Union[{', '.join(dedup_names)}]" if len(dedup_names) > 1 else dedup_names[0]
+
+        guard = (
+            f"if not isinstance(self.{aliased_slot_name}, ({tuple_str})):"
+            if slot.required
+            else f"if self.{aliased_slot_name} is not None and not isinstance(self.{aliased_slot_name}, ({tuple_str})):"
+        )
+        lambdas = ", ".join(
```
amc-corey-cox, line 1117: **Regression: `exactly_one_of` now rejects input that `main` coerces fine.**

Schema below is the shape used by this repo's own `test_slot_exactly_one_of`
(`range: integer` plus two `minimum_value`/`maximum_value` branches):

| input | `main` | this branch |
|---|---|---|
| `s1=5` | `5` | `5` |
| `s1="5"` | `5` | `ValueError: Value '5' … ambiguously matches 2 candidate types … ['int', 'int']` |

Generated code:

```python
_matches = []
for _coerce in (lambda: int(value), lambda: int(value)):   # <- same candidate, twice
```

Branches that don't declare their own `range` fall back to `slot.range`, so both
resolve to `int`. Line 1096 dedups them for the `isinstance` guard and the error
message, but this join still emits one lambda per original branch — and
`exactly_one_of` counts successes instead of stopping at the first, so two
identical candidates always look ambiguous.

Coercing `"5"` → `5` is exactly what `__post_init__` exists for, and `main` does
it. The compliance test survives only because it passes pre-typed ints, which the
`isinstance` guard short-circuits before this loop is ever reached.

Deduping the lambdas alongside `dedup_names` fixes it. Falling back to the plain
single-range path when every branch resolves to the same range may be better
still — constraint-only `any_of`/`exactly_one_of` is a value disjunction, not a
type union, so there's no union for this machinery to build.

(`any_of` has the same duplication but no bug — first match wins. It just emits
redundant lambdas and a confusing "candidate types int" message on total failure.)

tests/linkml/test_generators/test_pythongen.py
```diff
@@ -405,3 +405,456 @@ def test_gen_references_cycle_safety_raises_value_error(monkeypatch):
...
+      s1:
+        range: Any
+        multivalued: true
+        any_of:
+          - range: D
+          - range: integer
+"""
+    code = PythonGenerator(schema).serialize()
+    class_c = code[code.index("class C(") : code.index("class D(")]
+    assert "__post_init__" not in class_c
```
amc-corey-cox, line 697: This assertion can't fail. `D` is generated before `C`, so the slice runs backwards
and `class_c` is always empty:

```
index of "class C(": 2047    index of "class D(": 1468   ->   slice length: 0
```

`"__post_init__" not in ""` is unconditionally true, so the test passes regardless
of what pythongen emits. It reads as coverage in the suite while pinning nothing.

The intent is satisfied by accident — `C` really does have no `__post_init__` — so
asserting on the annotation instead would both document the boundary and hold it:

```python
s1: Optional[Union[Union[dict, Any], list[Union[dict, Any]]]] = empty_list()
```

That breaks loudly whenever multivalued `any_of` support lands, which is what you
want from a scope-pinning test.

tests/linkml/test_compliance/test_boolean_slot_compliance.py
```diff
@@ -142,7 +142,7 @@ def test_slot_any_of(framework, data_name, value, is_valid, use_any_type, use_de
         core_elements=["any_of", "range"],
     )
     expected_behavior = ValidationBehavior.IMPLEMENTS
-    if framework in [PYTHON_DATACLASSES, SQL_DDL_SQLITE, PANDERA_POLARS_CLASS]:
+    if framework in [SQL_DDL_SQLITE, PANDERA_POLARS_CLASS]:
```
amc-corey-cox, line 145: Worth adding a regression test for the actual reported symptom — `exclude_rdf=True`
below means nothing in the suite covers it. Same schema (`any_of` over two
identifier-less classes), `main` vs this branch:

```
main:  item type: JsonObj  ->  ValueError: Class JsonObj not found in schema
this:  item type: Dataset  ->  ex:C1 a ex:Container ; ex:item _:c14n0 .
                               _:c14n0 a ex:Dataset ; ex:title "T" .
```

So the fix is real and #1521 is closed by it — there's just nothing that would
catch a regression. A few lines in `test_pythongen.py` would pin it:

```python
c = py_module.Container(cid="ex:C1", item={"title": "T"})
assert "ex:Dataset" in rdflib_dumper.dumps(c, schemaview=SchemaView(schema))
```

The flip to `IMPLEMENTS` itself looks right for what this test checks.

packages/linkml/src/linkml/generators/pythongen.py
```diff
@@ -937,6 +1042,113 @@ def gen_constructor(self, cls: ClassDefinition) -> str | None:
...
+            lambdas += ","
+
+        rlines.append(guard)
+        rlines.append(f"\tvalue = self.{aliased_slot_name}")
+        if not require_exactly_one:
+            rlines.append(f"\tfor _coerce in ({lambdas}):")
+            rlines.append("\t\ttry:")
+            rlines.append(f"\t\t\tself.{aliased_slot_name} = _coerce()")
+            rlines.append("\t\t\tbreak")
+            rlines.append("\t\texcept Exception:")
```
amc-corey-cox: `except Exception` catches the failures this is meant to catch — and also the ones
it isn't. Using the schema from this PR's own `test_any_of_self_referential_tree`,
a 400-deep tree reports:

```
ValueError: None of the candidate types Union[Leaf, Node] could be constructed from …
```

The real fault is `RecursionError`; the user is told their data doesn't match the
schema. Same for an `AttributeError` from a genuine bug in a nested candidate's
constructor.

Legitimate non-matches only ever raise two types — `int('abc')` → `ValueError`,
`D(**as_dict(5))` / `D(**as_dict('abc'))` → `TypeError`, `E('nope')` → `ValueError`
— so `except (ValueError, TypeError)` keeps every intended path and stops hiding
real faults. Applies to the `exactly_one_of` arm at line 1134 too.

amc-corey-cox (changes requested):
Verified this does fix #1521 — same schema on main leaves the value a `JsonObj` and blows up on RDF export with `Class JsonObj not found in schema`, while on this branch it becomes a real `Dataset` and serializes correctly, and the branch-to-synthetic-slot approach reads well.

One regression to sort before merge: `exactly_one_of` branches that don't declare their own `range` collapse into duplicate candidates, so `s1="5"` now raises "ambiguously matches 2 candidate types ['int', 'int']" where main coerces it to `5` — details plus a few test notes inline.

amc-corey-cox:
Meant to fold this into the review — answers to your three "areas of uncertainty":

- **Tie-break by declaration order** — good default. I checked the identifier-union
  case specifically and the arbitrary pick isn't observable in output; both dumpers
  resolve ranges from SchemaView rather than the value's Python type.
- **`exactly_one_of` constructing every candidate** — fine by me, it's inherent to
  what the keyword has to guarantee. Not worth optimizing.
- **Leaving `all_of`/`none_of` alone** — right boundary. Neither has a union
  equivalent to build.

packages/linkml/src/linkml/generators/pythongen.py
HendrikBorgelt: Good catch, thanks for the precise repro. Fixed by deduping the (branch_slot, branch_name) pairs before building the lambdas, not just for the isinstance guard/label — verified s1="5" on your exact schema now coerces to 5 again. Kept it going through the boolean-combinator path rather than falling back to the plain single-range coercion, since that's a bigger structural change — open to revisiting if you'd still prefer that direction.

tests/linkml/test_generators/test_pythongen.py
HendrikBorgelt: I fixed the slice direction and added an assertion on the actual generated annotation string (Optional[Union[Union[dict, Any], list[Union[dict, Any]]]]) so it can't pass trivially again.

tests/linkml/test_compliance/test_boolean_slot_compliance.py
HendrikBorgelt: Added test_any_of_rdf_export_regression_1521 in test_pythongen.py — builds a Container with an any_of slot over two identifier-less classes and dumps it via rdflib_dumper, pinning the actual #1521 crash directly.

packages/linkml/src/linkml/generators/pythongen.py
HendrikBorgelt: Narrowed both arms (any_of and exactly_one_of) to except (ValueError, TypeError). Had to regenerate two metamodel snapshot files too, since this changed pythongen's own output for ArrayExpression.maximum_number_dimensions.

HendrikBorgelt:
@amc-corey-cox , all requested changes should be implemented.

amc-corey-cox (approved):
This looks good to me now. @HendrikBorgelt - thanks for the extra effort in getting this across!

PR #3821, opened 2026-07-30
Emit sh:node for native class names

### LINKED ISSUE #3820: `--use-native-names` in ShaclGenerator emits sh:class with the shape's own URI instead of sh:node
---
name: Bug report
about: Create a report to help us improve
title: '`--use-native-names` in ShaclGenerator emits sh:class with the shape's own URI instead of sh:node'
labels: bug
assignees: ''

---

**Describe the bug**
When `ShaclGenerator` is used with `use_class_uri_names=False` (`--use-native-names`), property shapes whose range is a class emit `sh:class <native-shape-uri>` instead of `sh:class <class_uri>`. Real data is always typed with the class's actual `class_uri`, never with the native shape name, so this constraint can never be satisfied by valid data. The correct constraint here is `sh:node <native-shape-uri>` (validate the value against the shape), not `sh:class` (check its rdf:type).

v1.11.1

```yaml
id: http://example.org/test
name: range_class_test
prefixes:
  ex: http://example.org/
default_prefix: ex

imports:
  - linkml:types

classes:
  Target:
    class_uri: ex:ExternalTarget
    slots:
      - label

  Container:
    slots:
      - has_target

slots:
  label:
    range: string
  has_target:
    range: Target

### PR DESCRIPTION
**Pull request by HendrikBorgelt**
## Summary

Fixes #3820

**The problem:** LinkML's SHACL generator has an option `--use-class-uri-names/--use-native-names` that lets you choose whether generated SHACL shapes get named after a class's `class_uri` (the default) or after the plain LinkML class name ("native names"). In native-names mode, a property whose range is a class still generated `sh:class <native-shape-name>` — the same kind of constraint as default mode, just pointed at the wrong URI.

`sh:class` means "the value must have this exact RDF type." In native-names mode, the URI it was pointing at is just the shape's own identifier (e.g. `ex:Target`), which real data is never actually typed as — real data is typed with the class's true `class_uri` (e.g. `ex:ExternalTarget`). So the constraint could never be satisfied by valid data, and SHACL validation would reject correct instances.

**The fix, in short:** in native-names mode, the generator now emits `sh:node <native-shape-name>` instead of `sh:class`. `sh:node` means "the value must conform to this shape" — which is the right check when the URI in question is a shape identifier rather than a real RDF class. Default mode (`use_class_uri_names=True`) is unchanged, since there the URI already is the real `class_uri` and `sh:class` is correct.

**Example, before and after** (schema: class `Target` has `class_uri: ex:ExternalTarget`; `Container.has_target` ranges over `Target`; generated with `--use-native-names`):

Before (bug):
```
ex:Target a sh:NodeShape ;
    sh:targetClass ex:ExternalTarget .

_:propshape sh:class ex:Target ;   # data is never typed ex:Target, only ex:ExternalTarget
    sh:path ex:has_target .
```

After (fix):
```
_:propshape sh:node ex:Target ;    # correctly validates the value against the Target shape
    sh:path ex:has_target .
```

## How was this tested?

- Added a new unit test, `test_shacl_property_class_constraint_modes` (in `tests/linkml/test_generators/test_shaclgen_naming.py`), covering both modes explicitly: default mode must emit `sh:class` with the real `class_uri` and must never emit `sh:node`; native-names mode must emit `sh:node` with the native shape name and must never emit `sh:class` (nor accidentally leak the `class_uri` into `sh:node`).
- Ran the full targeted SHACL test surface: `test_shaclgen.py`, `test_shaclgen_naming.py`, `test_gen_shacl.py` — 60/60 pass.
- Ran the broader `-m shaclgen` compliance sweep — 542/542 pass, 0 failures.
- Ran the entire unscoped test suite (`uv run pytest`) twice, before and after merging in the latest `main`: identical result both times — 3 failures + 1 teardown error, all pre-existing and unrelated to this change (a hardcoded Unix-only subprocess `PATH` in `test_rdf_canonicalize.py`, and a Windows file-locking issue in a shared test fixture's teardown in `test_pydanticgen.py::test_template_noblack`). Neither touches `shaclgen.py` or any SHACL test file.

## Areas of uncertainty

- This only changes the property-shape-level constraint (`_add_class`, used when a slot's range is a class). I didn't touch any other place that might reference shape names vs. class URIs in native-names mode — reviewers familiar with the rest of the generator may want to confirm there isn't a second, similar spot I missed.
- I'm the original author of the `--use-native-names` option itself, so I have context on the intent but would appreciate a second pair of eyes confirming `sh:node` is the right SHACL construct here rather than, say, `sh:property`-level shape composition.

## Checklist

## AI Assistance

**AI tool used:** Claude Code, under my direction. I reviewed the generated code and test line by line, ran and verified the test suite myself, and take full responsibility for everything in this PR.


amc-corey-cox:
Reviewed and verified the fix — `sh:node` is the right construct here, and `_add_class` is the only place a native shape name gets emitted, so there was no second spot to catch.

I pushed one small commit to your branch (4439ad93f): with `--use-native-names --suffix X` the `sh:node` reference kept the unsuffixed URI, pointing at a shape that is never emitted. SHACL treats that as a no-op rather than an error, so invalid data came back `Conforms: True`. Added a parametrized test asserting every `sh:node` object is an emitted `sh:NodeShape`. Revert if you disagree.

Two follow-ups I do not think should block this: `sh:node` drops the `rdf:type` check that `sh:class` gave you, and self-referential ranges now produce recursive shape references (pySHACL warns and backs out). Happy to open an issue.

amc-corey-cox (approved):
This looks good to me.

PR #3832, opened 2026-08-03
Resolve structured patterns when computing induced slots and types

### LINKED ISSUE #1557: Materialize structured patterns in all relevant generators

pkalita-lbl:
## Summary

Fixes #1557 

These changes approach the problem slightly differently than what was outlined in the issue, but they accomplish the same goal. The original idea (and what was originally implemented in this branch) was to add a `materialize_patterns` option to a number of generator classes and CLIs. 

As pointed out in an [initial review](https://github.com/linkml/linkml/pull/3832#discussion_r3724230584), it's hard to imagine circumstances when -- looking at induced slots and types -- you _wouldn't_ want to see a materialized pattern there. After all, the docstring of `SchemaView.induced_slot()` even says "all properties materialized": https://github.com/linkml/linkml/blob/1e69b5b363ae438c65460da3b65b84083691f208/packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py#L1663-L1664

These changes update `SchemaView.induced_slot()` and `SchemaView.induced_type()` to do the pattern materialization. Therefore any generator (or other client of `SchemaView`) will automatically get the right behavior when induced slots and types are used. When it comes to slots there is a bit of trickiness to ensure that the right schema file's `settings` are consulted when there are levels of slot inheritance, class inheritance, slot usage, etc. Hopefully this is explained well enough in inline code comments.

To keep things consistent the existing `materialize_patterns` option on `JsonSchemaGenerator` and `LinkmlGenerator` have been deprecated. The `SchemaView` method that made modifications to the underlying schema has also been deprecated.

## How was this tested?

* Enable a previously-skipped test function in `test_pattern_compliance.py`
* Add new tests around new functionality

## Areas of uncertainty

None

## Checklist

## AI Assistance

If you used AI tools while preparing this PR, you are still the author and responsible for understanding, verifying, and defending your submission. Please engage with reviewers personally rather than through your agent during feedback and revisions. See our [AI Covenant](AI_COVENANT.md) for details. ✅


packages/linkml/src/linkml/generators/owlgen.py
```diff
@@ -993,6 +1000,21 @@ def add_constraints(
...
+        """Translate a materialized structured pattern to XML Schema regex semantics."""
+        pattern = element.pattern
+        structured_pattern = element.structured_pattern
+        if pattern is None or structured_pattern is None or not self.materialize_patterns:
+            return pattern
+        if structured_pattern.partial_match:
+            return f".*({pattern}).*"
+        if pattern.startswith("^(?:") and pattern.endswith(")$"):
+            return pattern[4:-2]
+        return pattern
```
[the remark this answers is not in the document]
pkalita-lbl: Seems reasonable

packages/linkml/src/linkml/generators/pydanticgen/pydanticgen.py
```diff
@@ -230,6 +230,11 @@ class PydanticGenerator(OOCodeGenerator, LifecycleMixin):
     If True, optional multivalued slots default to ``[]``; if False, they default to ``None``.
     """
 
+    materialize_patterns: bool = True
```
sneakers-the-rat: When would we want this to be `False`? It seems like it would be incorrect to not materialize the patterns if declared, so maybe we don't need the option at all

sneakers-the-rat (approved):
Only reviewed pydanticgen part, but very minimal and just making use of a schemaview method (which I did not review, but assume is correct).

One minor thing about the parameter since we already have a ton of params, and when something is unambiguously correct I don't think we need an option to turn that off, but very mild aesthetics/simplicity opinion, feel free to ignore.

packages/linkml/src/linkml/generators/pydanticgen/pydanticgen.py
sneakers-the-rat: I guess more generally it seems like the schema view should unconditionally materialize patterns as part of inducing a class/slot. Its already footgunny enough to have concrete/induced definitions, we should try and keep that number to two rather than having "partially induced classes and slots where you have to remember to call some other method to fully induce them"

pkalita-lbl: You raise a good point. I was being conservative with this PR and just propagating existing patterns from `LinkmlGenerator` and `JsonSchemaGenerator` into other generators. I'll try out the more general solution of materializing structured patterns as part of `induced_slot`/`induced_type` and see how that feels before going forward with this PR.

sneakers-the-rat: Sure sure, don't mean to blow up this PR, at your leisure/etc.

pkalita-lbl: I think the general solution of materializing patterns as part of the induced definitions was the right move. I think that will make things less error-prone in the future. Changes pushed up and description updated.

packages/linkml/src/linkml/generators/linkmlgen.py
```diff
@@ -103,7 +109,7 @@ def serialize(self, output: str = None, **kwargs) -> str:
     "--materialize-patterns/--no-materialize-patterns",
     default=True,
```
amc-corey-cox: Was leaving this at `True` deliberate? Since `__post_init__` warns on any non-`None` value, I think it means every `gen-linkml` invocation emits the deprecation once 1.12.0 lands, including for users who never touched the flag. `gen-json-schema`'s equivalent moved to `default=None`.

The `if materialize_patterns is None: materialize_patterns = materialize` fallback in `cli()` below is also unreachable while this default is `True`, which made me wonder whether `None` was the intent here too.

packages/linkml/src/linkml/generators/owlgen.py
```diff
@@ -993,6 +996,29 @@ def add_constraints(
...
+        if pattern is None and isinstance(element, SlotDefinition | TypeDefinition):
+            pattern = self.schemaview.resolve_pattern(element)
+        if pattern is None or structured_pattern is None:
+            return pattern
+        inner = pattern
+        if inner.startswith("^(?:") and inner.endswith(")$"):
+            inner = inner[4:-2]
+        if inner.startswith("^") and inner.endswith("$"):
+            inner = inner[1:-1]
+        if structured_pattern.partial_match:
```
amc-corey-cox: A type that asserts `pattern: "^EXPLICIT$"` comes out as `.*(EXPLICIT).*` — the whole-string intent gets inverted into a substring match:

```yaml
types:
  ParentType:
    structured_pattern: {syntax: "{word}", interpolated: true, partial_match: true}
  ChildType:
    typeof: ParentType
    pattern: "^EXPLICIT$"
```

```
$ gen-owl --no-type-objects typ.yaml
xsd:pattern ".*(EXPLICIT).*"
```

I think this is because `as_graph` now passes `induced_type(...)` into `add_type`, and on an induced definition `pattern` and `structured_pattern` can come from different ancestors — so `partial_match` gets read off the parent while the pattern itself came from the child.

Would it work to key the partial/whole decision off the `^(?:...)$` wrapper this method already strips a few lines up, rather than off `structured_pattern`? On an induced definition the resolved pattern is self-describing, so `structured_pattern` shouldn't need consulting at all.

Narrow trigger — needs `--no-type-objects` plus this exact inheritance shape.

tests/linkml/test_compliance/test_pattern_compliance.py
```diff
@@ -114,13 +114,13 @@ def test_pattern(framework, range, schema_name, pattern, data_name, value):
         (True, False, "abc", False),
         (True, False, "abc def ghi", False),
         (True, True, "abc def ghi", True),
+        (True, True, "123 abc def", True),
```
amc-corey-cox: Worth pinning the search-vs-match semantics while you're in here?

`test_pattern` above builds its oracle as `is_valid = bool(re.match(pattern, str(value)))` (line 84), but after this PR pydantic and pandera both match unanchored, and JSON Schema/SHACL always did.

None of the ten current `test_pattern` cases tell the two apart, so the suite passes either way — something like `("partial_match", "string", r"ab", "prefix_ab", "xab")` would. I checked, and switching that oracle to `re.search` doesn't change the expected result for any existing case.

As it stands I don't think anything in the suite would catch the `search` behavior being reverted to `match`.

packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
```diff
@@ -1711,29 +1741,59 @@ def induced_slot(
...
+        # only the final induced slot could combine them from different layers. For
+        # example, a child slot_usage's explicit `pattern` could otherwise be
+        # overwritten by a parent's inherited `structured_pattern`.
+        induced_slot.pattern = self.resolve_pattern(induced_slot)
         if not slot_comes_from_attribute:
             slot_anc_names = self.slot_ancestors(slot_name, reflexive=True)
             # inheritable slot: first propagate from ancestors
             for anc_sn in reversed(slot_anc_names):
                 anc_slot = self.get_slot(anc_sn, attributes=False)
+                if anc_slot is None:
```
amc-corey-cox: What made this guard necessary? Before this change a slot ancestor that `get_slot` couldn't resolve would have raised; now it's skipped and the induced slot comes out quietly under-constrained, which feels at odds with the fail-fast convention.

If nothing in the test corpus actually hits it, might be worth dropping. If something does, I'd guess that schema is the real bug.

amc-corey-cox (approved):

PR #3835, opened 2026-08-04
fix(rustgen): exclude type-designator slot from tagged variant structs

elasticdotventures:
## Summary

Builds on #3661 (included here as the first commit so this branch is
independently buildable/testable — happy to rebase down to just the
second commit once #3661 merges).

Once #3661 fixes tagged `OrSubtype` enum generation (`#[serde(tag = "...")]`
for classes with a `designates_type` slot), a further bug surfaces: the
type-designator slot is *also* emitted as a regular field on each
concrete variant struct (via `generate_class`'s `induced_attrs`, and
`gen_poly_trait`'s polymorphic trait accessor/impls).

Since the tag now lives at the enum level, a variant struct carrying the
same-named field collides with it. Serializing produces a JSON object
with a **literal duplicate key**:

```json
{"stereotype_kind":"Role","stereotype_kind":"Role","name":"..."}
```

which `serde_json` correctly rejects on deserialize (`"duplicate field"`).
This reliably reproduces for any `is_a` hierarchy of sibling subtypes
that differ only in the discriminant — exactly the shape a UFO-style
ontology stereotype taxonomy has (`Kind`/`SubKind`/`Role`/`Relator`/`Mode`,
all differing only by a type tag), which is how I found it: porting an
existing hand-written Rust enum of that shape to a LinkML schema.

## Fix

Exclude the type-designator slot from a class's own struct fields (and
from the polymorphic trait's accessor/impls) whenever that class is a
concrete descendant of some ancestor that generates a tagged
`OrSubtype` enum. The field stays on the *ancestor's own* struct — it's
only stripped from classes that appear as tagged variants of someone
else's enum, since only those actually collide with the enum-level tag.

Two call sites needed the same exclusion:
- `generate_class`'s `induced_attrs` (the struct's own serialized fields)
- `gen_poly_trait`'s `attribs` (the polymorphic trait property + impl
  generation, which otherwise emits `&self.<field>` for a field that
  no longer exists on the struct once the first fix is applied alone)

## Test plan

- [x] New `is_a` hierarchy with sibling subtypes sharing an attribute set
      generates, compiles, and round-trips correctly through
      `serde_json` (tagged, not duplicate-keyed)
- [x] Verified against this exact branch (cherry-picked on top of #3661's
      commit, rebased onto current `main`) — `cargo test` 4/4 passing,
      including a case that specifically exercises `Role`/`Mode` (variants
      that would previously deserialize as the wrong variant under
      `#[serde(untagged)]`, or fail to serialize at all once tagged with
      this bug present)
- [x] `linkml validate` on the reproduction schema: no issues

Happy to add a repo-native test fixture under `tests/linkml/test_generators/input/` if that's preferred over the ad-hoc schema I used locally — let me know the preferred location/format and I'll fold it in.

### DIFF
--- a/packages/linkml/src/linkml/generators/rustgen/rustgen.py
+++ b/packages/linkml/src/linkml/generators/rustgen/rustgen.py
@@ -574,7 +574,7 @@ class RustGenerator(Generator, LifecycleMixin):
             return RustStructOrSubtypeEnum(
                 enum_name=get_name(cls) + "OrSubtype",
                 struct_names=[get_name(self.schemaview.get_class(d)) for d in descendants],
-                type_designator_name=get_name(td) if td else None,
+                type_designator_field=get_name(td) if td else None,
                 as_key_value=get_key_or_identifier_slot(cls, self.schemaview) is not None,
                 type_designators=td_mapping,
                 key_property_type=key_type,
--- a/packages/linkml/src/linkml/generators/rustgen/rustgen.py
+++ b/packages/linkml/src/linkml/generators/rustgen/rustgen.py
@@ -500,6 +500,22 @@ class RustGenerator(Generator, LifecycleMixin):
         """
         cls = self.before_generate_class(cls, self.schemaview)
         induced_attrs = [self.schemaview.induced_slot(sn, cls.name) for sn in self.schemaview.class_slots(cls.name)]
+
+        # Exclude the type-designator slot from a class's own struct fields
+        # when this class is a concrete variant of some tagged ancestor
+        # OrSubtype enum. The tag is emitted at the enum level
+        # (#[serde(tag = "...")]); if the variant struct ALSO carries the
+        # same-named field, serde emits a duplicate JSON key on serialize
+        # and rejects it on deserialize ("duplicate field"). The field stays
+        # on the ancestor's own struct (it isn't wrapped as anyone's variant).
+        for ancestor_name in self.schemaview.class_ancestors(cls.name, reflexive=False):
+            if not has_real_subtypes(self.schemaview, ancestor_name):
+                continue
+            td = self.schemaview.get_type_designator_slot(ancestor_name)
+            if td is None:
+                continue
+            induced_attrs = [a for a in induced_attrs if a.name != td.name]
+
         induced_attrs = self.before_generate_slots(induced_attrs, self.schemaview)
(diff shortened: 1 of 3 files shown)


kervel:
Hello!

I agree with the problem but not with the fix for the problem here. Removing the type designator from the generated class fails when you just want to use the generated class directly without the orsubtype trait. A type designator is still a real slot so users of the generated structs would expect it to be there.

The proper fix is i think that the *OrSubtype enums should not have their own serialisation path: on serialize a OrSubtype enum would just call the serialize method of the inner value and return that. on deserialize the current code would remain.

Something like this:

```rust
#[cfg(feature = "serde")]
impl Serialize for {{ enum_name }} {
    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
        match self {
{%- for t in struct_names %}
            {{ enum_name }}::{{ t }}(inner) => inner.serialize(serializer),
{%- endfor %}
        }
    }
}
```
For deserialisation, the enum deserialize method must make sure it still passes the type designator to the actual struct (example code)

```rust
#[cfg(feature = "serde")]
impl<'de> Deserialize<'de> for {{ enum_name }} {
    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
        use serde::de::Error as _;
        let value = serde_value::Value::deserialize(deserializer)?;
        let tag = match &value {
            serde_value::Value::Map(map) => map
                .get(&serde_value::Value::String("{{ type_designator_field }}".to_string()))
                .and_then(|v| match v {
                    serde_value::Value::String(s) => Some(s.clone()),
                    _ => None,
                }),
            _ => None,
        };
        let tag = tag.ok_or_else(|| D::Error::missing_field("{{ type_designator_field }}"))?;
        match tag.as_str() {
            // one arm per variant, all accepted designator values as `|` patterns:
            "MedicalEvent" => MedicalEvent::deserialize(
                serde_value::ValueDeserializer::<D::Error>::new(value),
            ).map({{ enum_name }}::MedicalEvent),
            other => Err(D::Error::unknown_variant(other, &[/* ... */])),
        }
    }
}
```

kevinschaper (changes requested):
Putting this into an actual request changes state based on @kervel's comment

PR #3839, opened 2026-08-04
fix: preserve uncompactable URI type designators

### LINKED ISSUE #3838: `uriorcurie` type designator becomes `"None"` when the class URI cannot be compacted
**Describe the bug**

When a generated Python class has a `uriorcurie` type-designating slot and its `class_uri` cannot be converted to a CURIE, LinkML sets the designator to the literal string `"None"`.

This happens both when the concrete class is constructed directly and when data is loaded directly into it.

In the below example, `schema_type` should instead be `https://unregistered.example/Association`. The generated `class_class_uri` contains that full URI, so it should be used when `class_class_curie` is unavailable.

Note: I found this while working on [#3836](https://github.com/linkml/linkml/issues/3836). Fixing that issue allows full-URI polymorphic loading to proceed far enough to expose this bug.

Initially reproduced with:

```text
linkml==1.11.1
linkml-runtime==1.11.1
pydantic==2.13.4
rdflib==7.6.0
Python 3.12
```

Also reproduced on `linkml/linkml` `main` at commit:

```text
e5d97c45a6b84e0cb5fbb9665a965c4b0c70194f
```

Save this as `schema-none-issue.yaml`:

```yaml
id: https://example.org/designator-test
name: designator_test

prefixes:
  ex: https://example.org/

default_prefix: ex

imports:
  - linkml:types

slots:
  schema_type:
    range: uriorcurie
    designates_type: true

classes:
  Association:
    class_uri: https://unregistered.example/Association
    slots:
      - schema_type
```

Save this as `reproduce-none-issue.py` beside the schema:

```python
#!/usr/bin/env python3
# /// script
# requires-python = "==3.12.*"
# dependencies = [
#   "linkml==1.11.1",
#   "linkml-runtime==1.11.1",
#   "pydantic==2.13.4",
#   "rdflib==7.6.0",
# ]
# ///

from pathlib import Path

from linkml.generators import PythonGenerator
from linkml_runtime.loaders import json_loader

EXPECTED = "https://unregistered.example/Association"

schema_path = Path(__file__).with_name("schema-none-issue.yaml")
model = PythonGenerator(str(schema_path)).compile_module()

constructed = model.Association()
loaded = json_loader.load(
    {"schema_type": EXPECTED},
    target_class=model.Association,
)


assert constructed.schema_type == EXPECTED
assert loaded.schema_type == EXPECTED
```

Run with:
`uv run reproduce-none-issue.py`

leej3:
## Summary

Fixes #3838 

Generated Python models currently serialize a uriorcurie type designator as "None" when the class URI cannot be converted to a CURIE. This change falls back to the full class_class_uri when class_class_curie is unavailable.

## How was this tested?

- Added a regression test for #3838 covering direct construction and concrete-class loading.
- Python generator and issue regression tests: 14 passed.
- Type-designator compliance suite: 180 passed, 15 skipped.
- Ruff linting and formatting checks passed.

## Areas of uncertainty

No known unknowns.

## Checklist


kevinschaper (approved):
thanks @leej3!

PR #3842, opened 2026-08-05
javagen: Allow generating code for all schemas under a given directory

### LINKED ISSUE #3829: javagen: Propose and support Maven-like convention for use of LinkML within a Java project
A new option to the Java generator to trigger a behaviour (described below) that would make integrating LinkML within a Maven-managed Java project easier. When that option is _not_ used, the generator behaves as it currently does.

**Additional context**
Java projects generally tend to favour a “convention-over-configuration” approach, where many things don’t need to be explicitly configured in a project-level configuration file because the project is expected to follow some well-established (and well-supported by existing Java tools) conventions.

For example, in any given project, all source files are expected to be at a predictable location like `src/NATURE/TYPE/PACKAGE/`, where

* _NATURE_ is typically either `main` (main source code) or `test` (source code for test units);
* _TYPE_ is the type of source files (typically `java` for Java code, but if other languages are involved in the project they will have their own directories);
* _PACKAGE_ is a hierarchy of directories that represent the Java package in which the code belongs to.

For example, a _Foo_ project that uses both Java source files and ANTLR4 source files, and whose code should be in a `org.example.foo` package, could look like this:

```
src/
src/main
src/main/java
src/main/java/org/
src/main/java/org/example
src/main/java/org/example/foo
src/main/java/org/example/foo/Main.java
src/main/antlr4
src/main/antlr4/org/
src/main/antlr4/org/example
src/main/antlr4/org/example/foo
src/main/antlr4/org/example/foo/grammar.g4
```

As long as that convention is followed, neither the Java compiler nor the ANTLR4 grammar generator need to be explicitly told where to find their source files and what to do with them.

I would like to propose a similar convention for the use of LinkML within a Java project, where

* all LinkML-related files are expected to be under `src/main/linkml` (if they are used for the “main” code) or `src/test/linkml` (if they are used as part of the test suite);
* that directory will itself contain two directories:
    * a `schemas` directory containing the LinkML schemas, under a directory hierarchy that represents the package for the generated code;
    * a `templates` directory containing any custom templates (for cases where the default templates provided by LinkML-Py are not enough)

So for example, if the same _Foo_ project as above starts using LinkML schema(s) to generate parts of its code:
* the schema(s) would be expected to be under `src/main/linkml/schemas/org/example/foo/`;
* custom templates (if used) would be expected to be under `src/main/linkml/templates`;
* the generated code would be expected to be written under `src/main/java/org/example/foo` with the package name set to `org.example.foo`.

This is of course already doable “manually” by calling `linkml generate java` with the appropriate options:

```sh
$ for schema in src/main/linkml/schemas/org/example/foo/*.yaml ; do
    linkml generate java --package org.example.foo \
                         --template-dir src/main/linkml/templates \
                         --output-directory src/main/java/org/example/foo \
                         $schema
done
```

but (i) this is quite cumbersome and (ii) does not allow discovering the package(s) (unless you do some more complicated shell scripting).

What I would like to propose here is a single option that would allow to simply do:

```sh
$ linkml generate java --maven-linkml-dir src/main/linkml \
                       --output-directory src/main/java
```

When the `--maven-linkml-dir <DIR>` option (open to suggestions for a better name) is used, the generator automatically takes care of (i) finding all the schemas under `<DIR>/schemas`, and for each schema (2) inferring the package name `PKG` from the directory hierarchy (unless overridden by an explicit `--package` option), (iii) writing the generated code under `<OUTPUT_DIR>/<PKG>`.

gouttegd:
## Summary

Fixes #3829

This PR amends the CLI for the Java generator so that it can accept either a single schema file (as before) or a _directory_ as its positional argument. If the argument is a file, the existing behaviour is unchanged.

If the positional argument is a directory, then the generator will automatically

* find all schemas under that directory;
* infer the package name for a given schema from the directory hierarchy under which the schema is found;
* reproduce that directory hierarchy under the output directory.

For example, if the command is called as follows:

```console
$ linkml generate java --output-directory src/main/java \
                       src/main/linkml/schemas
```

and there is a LinkML schema under `src/main/linkml/schemas/org/example/foo/my-schema.yaml`, then the package name for that schema will be inferred to be `org.example.foo` and the generated code will be written under `src/main/java/org/example/foo`.

This allows for easy integration of LinkML schemas and LinkML-derived code in a typical Maven-managed Java project, assuming the convention proposed in #3829 is followed.

## How was this tested?

Ad-hoc testing on all my Java projects using LinkML (SSSOM-Java, KGCL-Java, LinkML-Java), plus a new test added to LinkML’s own test suite.

## Areas of uncertainty

Not really.

## Checklist

## AI Assistance

Well, I tried to ask an agent to generate the test for the new feature and to add it to the existing set of Javagen tests. It told me that there was _no_ existing tests for the Java generator, and therefore spent a few minutes trying to make up its own tests before I closed the session and wrote the test myself. So much for “AI assistance”…


kevinschaper (changes requested):
The one thing about this that I think could use a little bit of a tweak is the necessary package structure, or lack of a helpful error without the right directory structure. 

If I run this against a schema I have nearby sitting in a normal python-ish location, the message I get doesn't really explain what's complaining about. 

```
uv run gen-java ../monarch-app/backend/src/monarch_py/datamodels --output-directory /tmp/monarch-java
ValueError: '.' is not a valid Java package name
```

It turns out that specifying a package fails the same way: 
```
uv run gen-java ../monarch-app/backend/src/monarch_py/datamodels --output-directory /tmp/monarch-java --package org.monarchinitiative
ValueError: '.' is not a valid Java package name
```

packages/linkml/src/linkml/generators/javagen.py
```diff
@@ -212,6 +213,32 @@ def get_template(self, name: str, fallback: str = "class", variant: str | None =
...
+    :param importmap: The import map to use, if any.
+    :return: A list of paths to root schemas.
+    """
+
+    # This is woefully inefficient as we have to load each schema
+    # through SchemaView, but this is necessary to ensure that we
+    # use the very same import resolution logic as the generator.
+    imported_ids: dict[str, int] = {}
+    schema_ids: dict[Path, str] = {}
+    for schema_path in schema_directory.glob("**/*.yaml"):
```
kevinschaper: Three issues with this glob, in rough order of how likely someone is to hit them.

**A non-schema YAML anywhere under the tree kills the whole run.** Every match is fed straight to `SchemaView`. Dropping an ordinary config file into the tree gives:

```
TypeError: SchemaDefinition.__init__() got an unexpected keyword argument 'server'
```

exit 1, and no indication of which file was responsible. The motivating use case points at a source tree, so a stray `.yaml` is a matter of time. Skipping unparseable files with a warning — or at minimum naming the offending path — would make this much easier to diagnose.

**`.yml` is silently skipped.** The pattern matches `*.yaml` only. I added a valid schema as `qux.yml`; nothing was generated for it, exit 0, no warning. Silent under-generation is the worse of the two failure modes here, since the build looks like it succeeded.

**The result order is filesystem order, not sorted.** `Path.glob` walks in directory order, so the sequence varies by machine and checkout. That matters when two root schemas under the same package both define a class of the same name: both write the same `.java` file, last one wins, exit 0, no warning. I reproduced this with two schemas each declaring `Shared` — the output contained only one of them, and which one depends on directory order. A `sorted()` here wouldn't fix the collision, but it would at least make the outcome reproducible; detecting the collision and failing would be better still.

packages/linkml/src/linkml/generators/javagen.py
```diff
@@ -562,6 +583,28 @@ def cli(
...
+        # Generate code for all root schemas under the specified directory,
+        # inferring the package name from the directory hierarchy
+        for schema_path in _find_root_schemas(yamlfile, importmap):
+            package_dir = schema_path.relative_to(yamlfile).parent
+            output_dir = Path(output_directory) / package_dir
+            package_name = package_dir.as_posix().replace("/", ".")
+            JavaGenerator(
+                schema_path,
+                importmap=importmap,
+                package=package_name,
```
kevinschaper: Directory mode derives the package name from the path and never consults `--package`, which makes the flag both useless and misleading here.

**Flat directories can't work at all.** Point at a directory whose schemas sit at the top level — no package subdirectories — and `package_dir` is `.`, so:

```console
$ gen-java ./datamodels --output-directory /tmp/out
ValueError: '.' is not a valid Java package name
```

That's a realistic first attempt; I hit it immediately on a real project whose schemas live in one flat directory. Failing is correct, but the message never mentions that directory mode infers the package from the subdirectory path, so it's hard to act on.

**`--package` doesn't rescue it, and is silently discarded when things do work.** Both verified on this branch:

```console
$ gen-java ./flat --package org.monarchinitiative --output-directory /tmp/out
ValueError: '.' is not a valid Java package name      # explicit package ignored

$ gen-java ./tree --package com.override.me --output-directory /tmp/out2
$ head -1 /tmp/out2/org/monarchinitiative/model/Node.java
package org.monarchinitiative.model;                  # com.override.me dropped, exit 0
```

There's also an inconsistency at lines 567-569: `--package` is read from the config file and validated before this branch runs, so an *invalid* `--package` is rejected in a mode that ignores it, while a *valid* one is silently thrown away.

Wiring the flag in would turn the flat-directory case from a hard failure into a supported one. Something like: flat directory plus `--package` generates into that package; a package tree with no `--package` infers from the path as it does today; and a tree plus `--package` either treats it as a prefix or is rejected explicitly. I don't have a strong preference among those — you may well have intended the tree to be the sole source of package names, which is defensible. But accepting the flag and discarding it isn't a good outcome in any of those readings.

gouttegd:
@kevinschaper I think in that situation, erroring out is the only thing to do, but I agree that the error message should be more explicit.

kevinschaper (approved):
looks good!

noelmcloughlin (commented):
Lgtm

gouttegd:
@noelmcloughlin Can you _approve_ then instead of merely commenting? :) Your explicit approval is required for merging as you are a code owner on the Java generator.

noelmcloughlin (approved):
Lgtm, apologies, my finger hit wrong button on phone

PR #3843, opened 2026-08-05
fix: jsonschemagen, include URI form of uriorcurie

### LINKED ISSUE #3697: `uriorcurie`-typed type designator slot only accepts CURIE values
**Describe the bug**
When a type designator slot (`designates_type: true`) has a range set to `uriorcurie`, the LinkML validator will only accept CURIEs, _not_ URIs.

Tip of the main branch as of June 30th (b8fe8b7da89cf1332aece90bdccaa0a4243932dc).

Consider the following schema:

```yaml
id: https://example.org/td-as-uriorcurie
name: td-as-uriorcurie
description: >-
  A schema illustrating a uriorcurie-typed type designator slot.

prefixes:
  orcid: https://orcid.org/
  tdu: https://example.org/td-as-uriorcurie/
imports:
  - linkml:types
default_prefix: tdu
default_range: string

slots:

  type:
    description: The type designator.
    designates_type: true
    range: uriorcurie

  name:
    description: The name of the Foo.

classes:

  Root:
    description: The container class at the root of the tree.
    attributes:
      foos:
        description: All the foos in this object.
        range: Foo
        multivalued: true
    tree_root: true

  Foo:
    description: The base Foo class.
    slots:
      - type
      - name

  Bar:
    description: A foo that is more precisely a bar.
    is_a: Foo
    attributes:
      length:
        description: The length of the bar.
        range: integer
```

The `Foo` class has a `uriorcurie`-typed type designator slot.

I would expect both the following data files to be found valid:

(A) Type designator values as CURIEs (using the `tdu` namespace declared in the schema):

```
foos:
  - type: tdu:Foo
    name: A simple foo.
  - type: tdu:Bar
    name: An extended foo.
    length: 123
```

(B) Type designator values as URIs:

```
foos:
  - type: https://example.org/td-as-uriorcurie/Foo
    name: A simple foo.
  - type: https://example.org/td-as-uriorcurie/Bar
    name: An extended foo.
    length: 123
```

However `linkml validate` will only accept **A**. Trying to validate **B** will yield

```sh
$ linkml validate -s td-as-uriorcurie.yaml data-with-uris.yaml
[ERROR] [data-with-uris.yaml/0] {'type': 'https://example.org/td-as-uriorcurie/Foo', 'name': 'A simple foo.'} is not valid under any of the given schemas in /foos/0
[ERROR] [data-with-uris.yaml/0] {'type': 'https://example.org/td-as-uriorcurie/Bar', 'name': 'An extended foo.', 'length': 123} is not valid under any of the given schemas in /foos/1
```

If I change the range of the type designator slot to `uri`, then the validator will only accept **B** and reject **A** (as expected). Conversely, if I set the range to `curie`, then the validator will only accept **A** and reject **B** (as expected).

From the (laconic) definition of the `uriorcurie` type (“a URI or a CURIE”), I would expect to be able to use indifferently a URI or a CURIE for a `uriorcurie`-typed slot.

And from the following bit in the [documentation about type designators](https://linkml.io/linkml/schemas/type-designators.html):

> [The range] could also be `uri`, `curie`, or `uriorcurie`, in which case it should match value of the `class_uri` field normalized to the appropriate form.

I interpret “normalized to the _appropriate_ form” as follows:

* if the range is `uri`, the value should match the value of the `class_uri` field normalized to the URI form (OK);
* if the range is `curie`, the value should match the value of the `class_uri` field normalized to the CURIE form (OK);
* if the range is `uriorcurie`, the value should match the value of the `class_uri` field normalized to **either the URI or the CURIE form** (KO, the validator only accepts the CURIE form).

leej3:
# Summary

Fixes #3697

Allows uriorcurie-typed type designators to use either the compact CURIE or expanded URI form of the class’s effective class_uri.

The JSON Schema generator now includes both forms in the generated enum. The regression test uses distinct assigned and schema-native namespaces to verify that the values are derived from class_uri.

## How was this tested?

- Focused tests: 266 passed, 22 skipped
- Ruff check/format: passed
- Documentation build: passed

## Checklist


matentzn:
@leej3 needs deconflicting, now assigned to a reviewer.

kevinschaper (approved):
very nice @leej3!

PR #3854, opened 2026-08-06
Stop dropping inherited metaslot values of 0 and ""

### LINKED ISSUE #3845: Inherited metaslots cannot be overridden to false, and a value of 0 is dropped
When a slot inherits from another slot through `is_a` or `mixins`, `SchemaView.induced_slot` copies inherited values only when the value is non-zero, rather than when it is set:

```python
# linkml_runtime/utils/schemaview.py
for metaslot_name in SlotDefinition._inherited_slots:
    if getattr(anc_slot, metaslot_name, None):
        setattr(induced_slot, metaslot_name, copy(getattr(anc_slot, metaslot_name)))
```

This has two consequences. The second one is the more serious.

**A child cannot turn an inherited boolean off.** `False` is skipped, so the parent's `True` survives and the child's explicit value is discarded:

```yaml
slots:
  p_req_true:   {abstract: true, required: true}
  c_opt_out:    {is_a: p_req_true, required: false}
  p_multi_true: {abstract: true, multivalued: true}
  c_single:     {is_a: p_multi_true, multivalued: false}
```

Induced `c_opt_out.required` is `True`, not `False`. Induced `c_single.multivalued` is `True`, not `False`. Writing `required: false` to opt out of a parent's requirement does nothing, with no warning.

**A parent value of 0 is not inherited at all.** `minimum_value`, `maximum_value`, `minimum_cardinality`, `maximum_cardinality` and `exact_cardinality` are all inherited and all accept 0:

```yaml
slots:
  p_minval_zero: {abstract: true, range: integer, minimum_value: 0}
  c_minval:      {is_a: p_minval_zero}
```

Induced `c_minval.minimum_value` is `None`, not `0`. Validating `{"c_minval": -7}` against the owning class is **accepted**, so a negative value passes a constraint that says the minimum is zero. The constraint disappears rather than failing, so a schema can silently lose validation.

The same pattern was found and fixed in the SHACL generator in https://github.com/linkml/linkml/pull/3473, which changes four `if s.some_cardinality:` tests to `is not None`. That pull request has been green and unreviewed since 2026-05-02. `pydanticgen/array.py` tests array dimension cardinalities the same way in six places, though I have not worked out whether those are reachable as bugs.

There is a precedent for the fix in this repository already. `is_inlined` on main has been rewritten to walk `slot_ancestors` and it guards on `is None` rather than on truth, so the intended semantics are not in dispute. What is missing is applying them at the root, in `induced_slot`, rather than one metaslot at a time.

Changing the test to `is not None` fixes all three cases above. Against `test_schemaview.py` it produced no change in results: 1220 passed and the same 6 pre-existing failures both with and without the change. The generator test suites have not been run, and that is where a behavior change is most likely to surface, so this needs a full run before it goes in.

turbomam:
## Summary

Fixes 
- https://github.com/linkml/linkml/issues/3845

`induced_slot` only propagated an inherited metaslot when its value was truthy, so a parent's `minimum_value: 0` never reached the child. The constraint disappeared instead of failing:

```yaml
slots:
  p: {range: integer, minimum_value: 0}
  c: {is_a: p}          # induced minimum_value was None, so -7 validated
```

The specification's rules for merging a slot with its parent take the ancestor's value whenever the more specific slot has none, whatever that value is. See the Combine Slots algorithm:
https://linkml.io/linkml-model/latest/docs/specification/04derived-schemas/#algorithm-combine-slots

Of the 31 inherited metaslots, 19 are affected. The other 12 are boolean and keep their existing behavior.

## Three things this PR decides, and I would like them confirmed

1. **Boolean metaslots are now pinned to the specification.** A child's `required: false` still does not clear a parent's `required: true`, per the `v1 OR v2` row of that table. That was previously accidental, a side effect of the truthiness test. This makes it explicit and adds a test for it, so it can no longer change by accident. If the project would rather not tie Python's behavior to the specification here, I will drop that test.
2. **An empty list or dict counts as unset.** `equals_string_in` is backed by `empty_list()`, so an unset child holds `[]` rather than `None`. `slot_ancestors(reflexive=True)` includes the slot itself and is walked last, so without this the child's `[]` overwrote the parent's value. Consequence: a child still cannot deliberately clear an inherited `equals_string_in` by setting it to `[]`. That limitation is not new, but this PR makes it deliberate.
3. **A test here now depends on linkml-model.** Adding or removing a boolean inherited metaslot upstream will fail CI in this repository until a list in `schemaview.py` is updated. That is a cost for linkml-model contributors and I would rather flag it than have someone find it the hard way.

The list itself is private (`_BOOLEAN_INHERITED_METASLOTS`), so it is not a new public name to support. Why it is written out rather than computed is documented at the constant.

## Testing

Compliance test across the core frameworks, unit tests for the falsy and boolean cases, and a regression test named for the issue.

To disprove it: revert the change in `induced_slot` and the compliance test fails on pydantic, jsonschema and shacl, all of which silently accept `-7`. The boolean cases pass either way, which is the point of testing them.

## Not in scope

- The loop over `slot_ancestors` in `induced_slot` overwrites rather than combining, so `minimum_value: 10` on a grandparent under `minimum_value: 5` on a parent resolves to 5 rather than to the larger of the two. That loosens constraints in both directions. Separate issue if you agree it is a gap.
- The same class of bug in shaclgen is fixed and waiting for review in https://github.com/linkml/linkml/pull/3473.

## Checklist

## AI Assistance

Drafted with AI assistance and reviewed by me. I am the author and will engage with review myself.


packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
```diff
@@ -1717,8 +1743,15 @@ def induced_slot(
...
-                        setattr(induced_slot, metaslot_name, copy(getattr(anc_slot, metaslot_name)))
+                    anc_value = getattr(anc_slot, metaslot_name, None)
+                    if anc_value is None:
+                        continue
+                    if metaslot_name in BOOLEAN_INHERITED_METASLOTS:
+                        # Boolean metaslots combine with OR, so only a True propagates.
+                        if anc_value:
+                            setattr(induced_slot, metaslot_name, True)
+                    else:
+                        setattr(induced_slot, metaslot_name, copy(anc_value))
```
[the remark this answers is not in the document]
turbomam: Confirmed and fixed in 96ddd96af.

You are right about the mechanism. `slot_ancestors(reflexive=True)` puts the slot itself last in the reversed walk, so the child's default `[]` was the final write and it clobbered the parent's value. Reproduced before fixing: parent `equals_string_in: ["a", "b"]` with an unset child induced `[]` instead of `["a", "b"]`.

`equals_string_in` is the only multivalued metaslot among the 31 inherited ones, and it is the only one whose default is a container rather than `None`, so it was the whole of the exposure.

The guard now skips empty lists and dicts while still inheriting `0` and `""`:

```python
if isinstance(anc_value, (list, dict)) and not anc_value:
    continue
```

This restores the pre-existing behavior for that case rather than changing it. A child still cannot deliberately clear an inherited `equals_string_in` by setting it to `[]`, because the data model gives us no way to tell an explicit empty list from the default one. That limitation is not new and I did not try to fix it here.

tests/linkml_runtime/test_utils/test_schemaview.py
```diff
@@ -3511,3 +3512,77 @@ def test_induced_slot_domain_of_no_duplicates() -> None:
...
+    """
+    schema = SchemaDefinition(id="test", name="test")
+    view = SchemaView(schema)
+    view.add_slot(SlotDefinition("parent", **{metaslot_name: value}))
+    view.add_slot(SlotDefinition("child", is_a="parent"))
+    view.add_class(ClassDefinition("C", slots=["child"]))
+
+    assert getattr(view.induced_slot("child", "C"), metaslot_name) == value
+
+
```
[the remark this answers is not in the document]
turbomam: Added in 96ddd96af as `test_induced_slot_inherits_multivalued_metaslot`, parametrized over three cases:

- parent `["a", "b"]`, child unset, expect `["a", "b"]` (the case that was broken)
- parent unset, child `["x"]`, expect `["x"]`
- parent `["a"]`, child `["x"]`, expect `["x"]`

Verified it earns its place: with the empty-collection guard removed, the first case fails and the other two pass.

Worth recording that the four full suites were green over the broken commit (runtime 1833, compliance 9186, generators 1399, remaining linkml 1075), so nothing existing covered list-valued inherited metaslots. Your catch, not the test suite's.

matentzn:
@turbomam can you first get QC to pass, then I will assign someone for review.

packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
```diff
@@ -1717,8 +1752,20 @@ def induced_slot(
             for anc_sn in reversed(slot_anc_names):
                 anc_slot = self.get_slot(anc_sn, attributes=False)
                 for metaslot_name in SlotDefinition._inherited_slots:  # noqa: SLF001
-                    if getattr(anc_slot, metaslot_name, None):
-                        setattr(induced_slot, metaslot_name, copy(getattr(anc_slot, metaslot_name)))
+                    anc_value = getattr(anc_slot, metaslot_name, None)
+                    if anc_value is None:
+                        continue
+                    if isinstance(anc_value, list | dict) and not anc_value:
```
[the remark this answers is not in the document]
turbomam: This one is not correct, and the repository's own linter is what required this form.

`isinstance()` has accepted PEP 604 unions since Python 3.10, which is this package's floor (`requires-python = ">=3.10"` in both `packages/linkml/pyproject.toml` and `packages/linkml_runtime/pyproject.toml`). Checked on the oldest supported version:

```
$ uv run --python 3.10 python -c "import sys; print(sys.version.split()[0]); print(isinstance([], list | dict))"
3.10.20
True
```

The `TypeError: isinstance() argument 2 cannot be a union` you are describing comes from `typing.Union[...]`, which is a different object from the `types.UnionType` produced by `list | dict`.

This line was originally written as `isinstance(anc_value, (list, dict))`. CI failed it under `ruff` rule UP038, "Use `X | Y` in `isinstance` call instead of `(X, Y)`", and the change to the union form was made to satisfy that rule. Reverting to a tuple would put the check back in a failing state.

The line is also covered: `test_induced_slot_inherits_multivalued_metaslot` exercises it directly and passes across the matrix.

turbomam:
Thanks @matentzn I do think this is ready for review now

turbomam:
Thanks to @Ostrzyciel and @niegrzybkowski for helpful comments about changing the state of boolean metaslots during inheritance

- https://github.com/linkml/linkml/issues/3845#issuecomment-5201687145
- https://github.com/linkml/linkml/issues/3845#issuecomment-5201858104

packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
```diff
@@ -66,6 +66,43 @@
...
+#:
+#: This list is written out rather than computed, and that is a known weak point rather
+#: than a preference. Computing it from the metamodel at import time is circular,
+#: because ``linkml_runtime.utils.introspection`` imports ``SchemaView`` from this
+#: module. Computing it from the ``SlotDefinition`` type annotations does work, but
+#: ``typing.get_type_hints`` has no precedent anywhere in this package. The tidiest
+#: fix is probably for the Python generator to emit the list next to
+#: ``_inherited_slots``, the same way that list is already produced. Until then
+#: ``test_boolean_inherited_metaslots_matches_metamodel`` is what catches it if this
+#: list and the metamodel stop agreeing.
```
kevinschaper, line 88: This half of the block is doing a different job from the half above it, and I don't think it should live here permanently.

Lines 69–78 are durable: what the constant is, the `v1 OR v2` rule, the spec link, and the fact that a test keeps it in sync. That's what a future reader needs.

This part is rationale addressed to a reviewer — "a known weak point rather than a preference", "`typing.get_type_hints` has no precedent anywhere in this package", "the tidiest fix is probably for the Python generator to emit the list". It's a good PR description and it'll be stale narrative to whoever opens this file next year.

The one piece with lasting value is the circular-import constraint, because it stops someone "improving" the constant into an import cycle. That's a sentence:

```python
#: Not computed from the metamodel: `linkml_runtime.utils.introspection` imports
#: SchemaView from this module, so deriving it at import time would be circular.
```

The pythongen proposal belongs in an issue, where it can be acted on, rather than in a comment where it can only be read.

packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
```diff
@@ -1717,8 +1754,20 @@ def induced_slot(
             for anc_sn in reversed(slot_anc_names):
                 anc_slot = self.get_slot(anc_sn, attributes=False)
                 for metaslot_name in SlotDefinition._inherited_slots:  # noqa: SLF001
-                    if getattr(anc_slot, metaslot_name, None):
-                        setattr(induced_slot, metaslot_name, copy(getattr(anc_slot, metaslot_name)))
+                    anc_value = getattr(anc_slot, metaslot_name, None)
+                    if anc_value is None:
+                        continue
+                    if isinstance(anc_value, list | dict) and not anc_value:
```
kevinschaper, line 1760: Agree with this and with the reasoning in the description — `empty_list()` means `[]` genuinely carries no information, so treating it as set would clobber the ancestor.

The consequence you flag (a child can't deliberately clear an inherited `equals_string_in` by setting `[]`) is worth its own issue rather than a fix here: distinguishing "unset" from "explicitly empty" needs something the metamodel doesn't currently express, so it isn't resolvable inside this PR.

tests/linkml_runtime/test_utils/test_schemaview.py
```diff
@@ -3511,3 +3512,113 @@ def test_induced_slot_domain_of_no_duplicates() -> None:
...
+def test_boolean_inherited_metaslots_matches_metamodel() -> None:
+    """_BOOLEAN_INHERITED_METASLOTS must list exactly the inherited metaslots ranged on boolean.
+
+    Guards against drift: if the metamodel gains or loses a boolean inherited metaslot,
+    the OR-combination branch in induced_slot has to learn about it.
+    """
+    metamodel = package_schemaview("linkml_runtime.linkml_model.meta")
+    induced = [metamodel.induced_slot(name, "slot_definition") for name in metamodel.class_slots("slot_definition")]
+    expected = {slot.name for slot in induced if slot.inherited and slot.range == "boolean"}
+    assert expected == _BOOLEAN_INHERITED_METASLOTS
```
kevinschaper, line 3526: This is the right shape for the guard — computing `expected` from the metamodel rather than restating the literal is what makes it actually catch drift.

One ask, tied to your question 3: a bare set-equality failure will show a linkml-model contributor two sets and no instruction. Worth giving it a message that names the constant and the file, e.g.

```python
assert expected == _BOOLEAN_INHERITED_METASLOTS, (
    "Boolean inherited metaslots changed upstream. Update "
    "_BOOLEAN_INHERITED_METASLOTS in linkml_runtime/utils/schemaview.py: "
    f"added={expected - _BOOLEAN_INHERITED_METASLOTS} "
    f"removed={_BOOLEAN_INHERITED_METASLOTS - expected}"
)
```

That turns the cost you flagged from "CI broke, why" into "CI broke, here's the edit".

tests/linkml_runtime/test_utils/test_schemaview.py
```diff
@@ -3511,3 +3512,113 @@ def test_induced_slot_domain_of_no_duplicates() -> None:
...
+        ("multivalued", True, False, True),
+        ("multivalued", None, True, True),
+    ],
+)
+def test_induced_slot_boolean_metaslots_combine_with_or(
+    metaslot_name: str, parent_value: bool | None, child_value: bool | None, expected: bool
+) -> None:
+    """Boolean metaslots combine with OR, so a child cannot clear an inherited True.
+
+    This asserts the specification behavior so it cannot change by accident. Switching
```
kevinschaper, line 3572: Same instinct as the comment block in `schemaview.py`, less urgently: the paragraph explaining what *would* happen if someone switched the loop to `is not None` is arguing a counterfactual to a reviewer.

"Asserts the spec's OR behaviour so it can't change silently" plus the Combine Slots link carries the whole point, and survives better once the alternative implementation nobody wrote is no longer on anyone's mind.

kevinschaper (changes requested):
I left agnatically generated comments about tidying up your agnatically generated comments - mostly just claude doing its thing of putting too much issue context into the source.

PR #3855, opened 2026-08-06
fix(schemaview): keep track of which schema requested each import

### LINKED ISSUE #3499: SchemaView.imports_closure() resolves relative imports against root schema directory

noelmcloughlin:
## Summary

Fixes #3499 and unblocks https://github.com/IBM/ai-atlas-nexus/pull/274

This PR helps unblock imports and w3id on boarding:

- `imports_closure` queue now tracks provenance as discussed in #3499
- New _load_closure_import decides how to load each schema via fall-back policy (careful discrimination)
- _resolved_importmap() helper, previously inlined in load_import, now shared by both call sites

The second commit handles URL-keyed schemas. A absolute URL import has `/` in its key, so its relative imports took the normalization branch, `os.path.normpath()` collapsed the `//` causing: `file://…/middle` became `file:/…/middle`, which fails as `ValueError: Unknown CURIE prefix: file`. URLs are now excluded from that branch and located by the same importer-based resolution.

Added a commit to fix Windows test failure.

## How was this tested?

added tests for #3499

## Checklist


kevinschaper (approved):
nice improvement! thanks @noelmcloughlin!

PR #3857, opened 2026-08-07
fix: sync vendored meta.yaml with upstream linkml-model

amc-corey-cox:
`test_vendored_files_match_upstream` is currently failing on **every PR in the repo, including `main`** — nothing to do with the PRs themselves. Upstream `linkml-model` added three prefixes and our vendored copy went stale.

```yaml
dcterms: http://purl.org/dc/terms/
rdfs: http://www.w3.org/2000/01/rdf-schema#
rdf: http://www.w3.org/1999/02/22-rdf-syntax-ns#
```

`main`'s last green run simply predates the drift; I confirmed by diffing `origin/main`'s vendored file against upstream directly. Since generators resolve this file locally instead of fetching it, drift means generated output silently diverges from upstream — which is exactly what that test exists to catch.

## Why the snapshots move

The three prefixes change how the metamodel artifacts render CURIEs, so the derived snapshots have to be regenerated:

| File | Churn |
| --- | --- |
| `meta.yaml` | +3 (prefixes only) |
| `meta.json` | +12 |
| `meta.shex` | 118 |
| `meta.ttl` | 220 |

187 insertions / 166 deletions.

**`meta.owl` is deliberately untouched.** Its test compares RDF semantically rather than byte-for-byte and was already passing. Regenerating it produces ~2166 lines of pure serialization churn with no semantic change, so including it would have quadrupled the diff for nothing. Worth knowing if you regenerate locally and see it come up dirty.

No schema semantics change here — prefix additions and their downstream rendering only.

## Verification

- `test_vendored_files_match_upstream` — 199 passed
- Slow suite (`-m "slow and not kroki" --with-slow --with-biolink`) — 46 passed
- `tests/linkml_runtime` — 1819 passed
- `tests/linkml` — 11613 passed

## Sequencing

Two things worth flagging:

- **#3714** also touches `meta.yaml` (it uncomments the `linkml:validation` import) and several of the same snapshots. It does **not** add these prefixes, so it won't fix the drift, but it will conflict with this. Whichever lands second needs a merge.
- **#3853** (the dependabot sweep) is blocked on this — its only remaining failure is this same drift test. Once this merges it picks the fix up via a main merge.

### DIFF
--- a/packages/linkml_runtime/src/linkml_runtime/linkml_model/model/schema/meta.yaml
+++ b/packages/linkml_runtime/src/linkml_runtime/linkml_model/model/schema/meta.yaml
@@ -50,6 +50,9 @@ prefixes:
   qudt: http://qudt.org/schema/qudt/
   cdisc: http://rdf.cdisc.org/mms#
   SIO: http://semanticscience.org/resource/SIO_
+  dcterms: http://purl.org/dc/terms/
+  rdfs: http://www.w3.org/2000/01/rdf-schema#
+  rdf: http://www.w3.org/1999/02/22-rdf-syntax-ns#
 
 default_prefix: linkml
 default_range: string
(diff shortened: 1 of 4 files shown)


matentzn (approved):
Purely cosmetic: dc1 regression.. Corey will make follow up issue.

PR #3860, opened 2026-08-07
fix(projectgen): validate yaml config properly and merge generator_args with cli args

### LINKED ISSUE #3858: gen-project: AttributeError on a malformed config AND includes/excludes as string silently skips generators AND minor bugs
**Describe the bug**

During PR #3781 testing various bugs were found during testing::

- `gen-project` skips validation on the YAML it is handed. If `--config-file` or `-A/--generator-arguments` parses as valid YAML but are unexpected mappings, the value is stored and blows up later on attribute access. The user gets unhandled `AttributeError` traceback,  instead of helpful output like this ...
```bash
Error: --config-file: expected a YAML mapping at 'generator_args', found str
```

-   `generate()` does `gen_name in config.excludes.` With a list that's membership (good); with a string it's substring matching (bad here), and jsonld is a substring of jsonldcontext. So excluding one generator silently excludes another

- Validation is CLI-only.  ProjectGenerator.generate() is public and used programmatically and still dies on bad ProjectCofiguration:

`AttributeError**: 'str' object has no attribute 'get'`

- Arg -A silently replaces `--config-file`'s `generator_args` rather than merging. Undocumented behavior.

1.11.1

```bash
cat > /tmp/myschema.yaml <<'EOF'
id: https://example.org/test
name: test
prefixes:
imports:
  - linkml:types
default_range: string
classes:
  Thing:
    slots:
      - name
slots:
  name:
EOF
```

Test: top level isn't a mapping:
```bash
printf 'justastring\n' > /tmp/config.yaml
uv run gen-project --config-file /tmp/config.yaml -d out /tmp/myschema.yaml

  File ".../linkml/generators/projectgen.py", line 233, in cli
    for k, v in yaml.safe_load(config_file).items():
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'items'
```

Test: --config-file `generator_args` isn't a mapping:
```bash
printf 'generator_args: java.package=org.example.model\n' > /tmp/config.yaml
uv run gen-project --config-file /tmp/config.yaml -d out /tmp/myschema.yaml

  File ".../linkml/generators/projectgen.py", line 122, in generate
    **config.generator_args.get(gen_name, {}),
      ^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'get'
```

Test: `-A` is not a mapping
```bash
uv run gen-project -A 'notamapping' -d out /tmp/myschema.yaml

AttributeError: 'str' object has no attribute 'get'
```
.. crashing deep inside `generate()` with no cause signal

Test: excludes/includes as a string silently generates the wrong artifacts
```bash
printf 'excludes: jsonldcontext\n' > /tmp/config.yaml
gen-project --config-file /tmp/config.yaml -d out /tmp/schema.yaml  # exit 0

ls out/jsonld/     # empty/gone - loop matched 'jsonld' generator to 'jsonldcontext' string)
```

noelmcloughlin:
## Summary

Fixes #3858

- Various bugs found in projectgen configuration management
- **Allow merging configuration supplied by CLI (`-A` wins) and configuration file (`-C config.yaml`)**
- fix test gap: empty config file causing crashes was inadvertently  fixed (missing test case)

Starts addressing https://github.com/linkml/linkml/issues/3972 following PR comment.

## How was this tested?

expanded tests

## Areas of uncertainty

- merging generator_args with cli args (-A) is new behavior but old behavior was silent skip / undocumented

## Checklist

## AI Assistance
LLM assisted

### DIFF
--- a/packages/linkml/src/linkml/generators/projectgen.py
+++ b/packages/linkml/src/linkml/generators/projectgen.py
@@ -110,6 +110,45 @@ def get_local_imports(schema_path: Path, dir: Path):
     return all_imports
 
 
+def _config_mapping(value: Any, where: str, source: str) -> ARG_DICT:
+    """Check that one level of a configuration is a mapping, and hand it back.
+
+    An empty value (``generator_args:`` with nothing under it) reads as None saying
+    "nothing configured here", so it returns as empty mapping. Anything else that is
+    not a mapping is a mistake in the configuration, and says so where it was given
+    rather than failing later.
+
+    :param value: Whatever was found at this point in the configuration.
+    :param where: Where that was, for the error message, e.g. ``"'generator_args'"``.
+    :param source: The option it came from, e.g. ``"--config-file"``.
+    :return: The mapping, empty if nothing was configured.
+    :raises click.UsageError: if the value is neither a mapping nor empty.
+    """
+    if value is None:
+        return {}
+    if not isinstance(value, dict):
+        raise click.UsageError(f"{source}: expected a YAML mapping at {where}, found {type(value).__name__}")
+    return value
+
+
+def _generator_args(value: Any, source: str, prefix: str = "") -> dict[GENERATOR_NAME, ARG_DICT]:
+    """Check a block of per-generator arguments: generator name to that generator's settings.
+
+    The block and each generator's settings within it both have to be mappings, since
+    the settings are unpacked into the generator's constructor.
+
+    :param value: The block as read from YAML.
+    :param source: The option it came from, e.g. ``"--config-file"``.
+    :param prefix: Key path leading to the block, e.g. ``"generator_args."`` for a config
+        file, or empty when the block is the whole value (as with ``-A``).
+    :return: The validated block, empty if nothing was configured.
+    :raises click.UsageError: if the block, or any generator's settings, isn't a mapping.
+    """
+    where = f"'{prefix.rstrip('.')}'" if prefix else "the top level"
+    args = _config_mapping(value, where, source)
+    return {name: _config_mapping(section, f"'{prefix}{name}'", source) for name, section in args.items()}
+
+
 @dataclass
 class ProjectConfiguration:
     """
@@ -295,7 +334,12 @@ def cli(
     """
     project_config = ProjectConfiguration()
     if config_file is not None:
-        for k, v in yaml.safe_load(config_file).items():
+        config_data = _config_mapping(yaml.safe_load(config_file), "the top level", "--config-file")
+        if "generator_args" in config_data:
+            config_data["generator_args"] = _generator_args(
+                config_data["generator_args"], "--config-file", "generator_args."
+            )
+        for k, v in config_data.items():
             setattr(project_config, k, v)
     if exclude:
         project_config.excludes = list(exclude)
@@ -303,9 +347,10 @@ def cli(
         project_config.includes = list(include)
     if generator_arguments is not None:
         try:
-            project_config.generator_args = yaml.safe_load(generator_arguments)
-        except Exception:
-            raise Exception("Argument must be a valid YAML blob")
+            arguments = yaml.safe_load(generator_arguments)
+        except yaml.YAMLError as e:
+            raise click.UsageError(f"--generator-arguments must be a valid YAML blob: {e}") from e
+        project_config.generator_args = _generator_args(arguments, "--generator-arguments")
         logger.info(f"generator args: {project_config.generator_args}")
     if dir is not None:
         project_config.directory = dir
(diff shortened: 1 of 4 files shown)


packages/linkml/src/linkml/generators/projectgen.py
```diff
@@ -70,6 +70,135 @@ def get_local_imports(schema_path: Path, dir: Path):
...
+
+    :param value: YAML text, or an open file to read it from.
+    :param source: Where it came from, e.g. ``"--config-file"``.
+    :return: Whatever the YAML held.
+    :raises ValueError: if the YAML is malformed.
+    """
+    try:
+        return yaml.safe_load(value)
+    except yaml.YAMLError as e:
+        raise ValueError(f"{source} is not valid YAML: {e}") from e
```
kevinschaper, line 84: This validation now exists twice. #3781 landed `_config_mapping` / `read_generator_config` in `linkml/utils/generator.py` on 2026-09-01 — after this branch's base — and `gen-java` / `gen-golang` already use it against the *same* `config.yaml`. This PR adds a second implementation here.

They have already drifted on the error text for the same malformed file:

```
gen-java     ->  Error: --config-file: could not parse as YAML: ...
gen-project  ->  Error: --config-file is not valid YAML: ...
```

and after the (mandatory) rebase, `generate()` would raise `ValueError` from this path while main's shared helper raises `click.UsageError` for the same class of problem — two exception types one caller has to catch.

Since you have to rebase anyway, worth folding this into the shared helper rather than carrying a parallel implementation. If `projectgen` needs something the shared one doesn't do, extending it there means `gen-java` and `gen-golang` get the same fix.

Not a criticism of the code — #3781 simply landed underneath you.

packages/linkml/src/linkml/generators/projectgen.py
```diff
@@ -70,6 +70,135 @@ def get_local_imports(schema_path: Path, dir: Path):
...
+    :param value: Whatever was found at this point in the configuration.
+    :param where: Where that was, for the error message, e.g. ``"'excludes'"``.
+    :param source: The option it came from, e.g. ``"--config-file"``.
+    :return: The list of names, empty if nothing was configured.
+    :raises ValueError: if the value is not a list of names, nor empty.
+    """
+    if value is None:
+        return []
+    if not isinstance(value, list):
+        raise ValueError(f"{source}: expected a YAML list of generator names at {where}, found {type(value).__name__}")
```
kevinschaper, line 145: The docstring explains the reasoning well, and the bug is real. I'd argue coercion resolves it better than rejection, though.

The failure mode you document is substring matching:

```
bare 'jsonldcontext'       -> skips ['jsonld', 'jsonldcontext']   # the bug
coerced ['jsonldcontext']  -> skips ['jsonldcontext']             # correct
bare 'shex'                -> skips ['shex']                      # already correct today
```

Coercing `str` -> `[str]` fixes the `jsonld`/`jsonldcontext` collision *properly*, rather than outlawing the syntax it happened to surface in. And it matters because the bare form works correctly today for every generator name that isn't a prefix of another — which is all of them except that one pair. I verified `excludes: shex` on main: exit 0, shex correctly skipped. Rejecting it turns a working config into `exit 2` with no deprecation period.

```suggestion
    if value is None:
        return []
    if isinstance(value, str):
        # A bare string is accepted as a one-item list. Treating it as a list
        # matters: ``gen_name in "jsonldcontext"`` is a substring test, so a
        # bare string would wrongly exclude ``jsonld`` as well.
        return [value]
    if not isinstance(value, list):
        raise ValueError(f"{source}: expected a YAML list of generator names at {where}, found {type(value).__name__}")
```

That flips the `excludes-string` / `includes-string` test params from "errors" to "excludes exactly one", which is the behaviour worth pinning.

If you'd rather keep the rejection — defensible, the form is undocumented and fail-fast is the house style — then it needs a changelog entry, because what a user sees on upgrade is a hard error on a config that worked yesterday.

kevinschaper (changes requested):
I added a few claude generated inline comments

packages/linkml/src/linkml/generators/projectgen.py
noelmcloughlin: Fixed by new commit. I forgot about this PR and started work on Issue #3972 for better cross-generator solution and proposed a 1st PR https://github.com/linkml/linkml/issues/3972 - so I borrowed code from there to address this concern in a forward-compatible way - we leave javagen/golangto be addressed in the other PR.

packages/linkml/src/linkml/generators/projectgen.py
noelmcloughlin: Added coercing commit.

kevinschaper (approved):
thanks @noelmcloughlin!

Issue #3879, opened 2026-08-11
Notebook test suite is order-dependent: shared venv + runtime uv pip install cells let one notebook mask another's failures


kevinschaper:
## Problem

The notebook test suite (`tests/linkml/test_notebooks/test_notebooks.py`) produces results that depend on the order notebooks happen to run in, rather than on whether each notebook is actually correct. A genuinely broken notebook can pass, and an intermittent, PR-unrelated failure can appear on unrelated changes.

## Mechanism

Three properties combine to make this happen:

1. **Shared, mutating venv.** All notebooks execute serially against a single Python environment. Several notebooks contain runtime cells that install packages into that environment, e.g.:
   ```
   !uv pip install -q --disable-pip-version-check yamlmagic
   %reload_ext yamlmagic
   ```
   Once any notebook installs such a package, it persists in `site-packages` for every notebook that runs afterward in the same job (each notebook gets a fresh kernel, but the venv is shared).

2. **Nondeterministic execution order.** The test is parametrized over `os.listdir(NBBASEDIR)`, which returns entries in filesystem order — *not* sorted, and not guaranteed stable across fresh checkouts or runners.

3. **Cross-notebook dependency via side effects.** Because installs leak across notebooks, a notebook that fails to install a dependency itself can still succeed *if* an earlier notebook already installed it — and fail if it happens to run first.

## Concrete evidence

- `notebooks/examples.ipynb` contains a malformed install cell — `!uv pip -q --disable-pip-version-check install yamlmagic` — where the flag precedes the `install` subcommand. `uv` rejects this (`error: unexpected argument '--disable-pip-version-check' found`), so the install never runs and the following `%reload_ext yamlmagic` raises `ModuleNotFoundError: No module named 'yamlmagic'`.
- Despite this, the cell "passes" whenever a correctly-formed sibling notebook (`DistributedModels.ipynb`, `Type-Designators.ipynb`, `enumerations.ipynb`) runs earlier in the same job and installs `yamlmagic` into the shared venv first.
- This surfaced as an intermittent failure in **Notebook Tests (ubuntu-latest, 3.13)** — e.g. on #3873, a CODEOWNERS-only change with no relation to notebooks — where the `os.listdir` order placed `examples.ipynb` before any sibling that installs `yamlmagic`.
[...]

PR #3885, opened 2026-08-12
fix(ci): eliminate two intermittent test failures

amc-corey-cox:
Two unrelated causes have been reddening required checks on PRs that have nothing to do with either. Both are small, both are test-infrastructure only, and they're together because each was blocking the other's PR from going green.

## 1. `examples.ipynb` install cell

The flags precede the subcommand:

```
!uv pip -q --disable-pip-version-check install yamlmagic     # rejected by uv
!uv pip install -q --disable-pip-version-check yamlmagic     # the three sibling notebooks
```

uv errors with `unexpected argument '--disable-pip-version-check'`, yamlmagic is never installed, and the next line's `%reload_ext yamlmagic` raises `ModuleNotFoundError`.

Notebooks share one venv, so this only surfaces when `examples.ipynb` runs before a sibling that installs yamlmagic correctly. Order comes from `os.listdir` — filesystem order, unstable across runners. So: coin flip.

The uncomfortable half — **`examples.ipynb` has been broken all along and passing most of the time**, on a dependency it never installed itself.

## 2. Creature fixtures fetching through a redirect

`RemoteDisconnected: Remote end closed connection without response`, four times today, each taking out all six parametrisations of `test_creature_schema_entities_with_without_imports` at once.

Both remote fixtures fetched via `github.com/.../raw/...`, which 302s to `raw.githubusercontent.com`:

```
302, 1 redirect, final 200   https://github.com/linkml/linkml/raw/refs/heads/main/...
200, 0 redirects             https://raw.githubusercontent.com/linkml/linkml/refs/heads/main/...
```

Two URLs move — `CREATURE_SCHEMA_RAW_URL` (used by `creature_view_direct_url`) and the `mcc` prefix in `creature_schema_remote.yaml` (the fetch base for `creature_view_remote`'s import).

**The `github.com/.../tree/...` URLs elsewhere under `mcc/` are deliberately untouched.** Those are schema identifiers and CURIE bases, not fetch targets — rewriting them would change schema identity without fixing anything.

## Verification

Notebook fix, against a venv with yamlmagic absent so nothing could mask it:
- before — fails with `unexpected argument` then `ModuleNotFoundError`
- after — passes, and yamlmagic is importable afterwards, installed by the cell

URL fix:
- both new URLs resolve `200` with zero redirects
- `test_schemaview.py` — 1226 passed
- `tests/linkml_runtime` — 1829 passed

Notebook suite from a clean venv: 8 passed, 1 failed. That failure is `SQL-examples.ipynb` (`OperationalError: no such table: Person`), which reproduces identically with and without these changes and passes in CI — pre-existing and unrelated.

## Scope

Neither issue is fully closed by this:

- **#3879** — the shared mutating venv plus unstable `os.listdir` ordering is untouched. Note that sorting the order would make runs deterministic but would *not* fix masking; it'd make `examples.ipynb` consistently masked instead of intermittently. Having each notebook install its own dependencies, as here, is what breaks the coupling.
- **#3421** — also covers a `biolink.github.io` 503 on a different host, which this does not address.

### DIFF
--- a/notebooks/examples.ipynb
+++ b/notebooks/examples.ipynb
@@ -18,7 +18,7 @@
     }
    },
    "source": [
-    "!uv pip -q --disable-pip-version-check install yamlmagic\n",
+    "!uv pip install -q --disable-pip-version-check yamlmagic\n",
     "%reload_ext yamlmagic"
    ],
    "outputs": [],
--- a/tests/linkml_runtime/test_utils/input/mcc/creature_schema_remote.yaml
+++ b/tests/linkml_runtime/test_utils/input/mcc/creature_schema_remote.yaml
@@ -5,6 +5,6 @@ description: |
   Schema for testing remote imports via a prefix. This produces yields exactly the same schema
   as creature_schema.yaml, but pulls in the files via remote import.
 prefixes:
-  mcc: https://github.com/linkml/linkml/raw/refs/heads/main/tests/linkml_runtime/test_utils/input/mcc/
+  mcc: https://raw.githubusercontent.com/linkml/linkml/refs/heads/main/tests/linkml_runtime/test_utils/input/mcc/
 imports:
   - mcc:creature_schema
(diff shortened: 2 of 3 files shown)


kevinschaper (approved):
Good fixes for this specific notebook, this doesn't address the whole weird situation of notebook testing, but I agree that this is the right way to quickly unblock us.

Issue #3583, opened 2026-05-29
Notebook tests mutate the shared test venv via '!uv pip install --upgrade'


kevinschaper:
## Summary

`tests/linkml/test_notebooks/test_notebooks.py::test_redo_notebook` executes the notebooks in `notebooks/` **in-process, in the shared CI/test venv** (via `nbconvert`'s `ExecutePreprocessor`). Several notebooks contain shell-install cells, e.g.:

- `notebooks/ShExPrimerModel.ipynb`: `!uv pip install linkml --upgrade --disable-pip-version-check -q`
- `notebooks/context_issue.ipynb`: `!uv pip install … prefixcommons rdflib rdflib-jsonld jsonasobj`
- `notebooks/DistributedModels.ipynb`: `!uv pip install -q … ..` and `yamlmagic`
- `notebooks/examples.ipynb`, `enumerations.ipynb`, `Type-Designators.ipynb`, `SQL-examples.ipynb`: `yamlmagic` / `names`

Running these **mutates the venv the rest of the suite shares**, ignoring `uv.lock`. The `--upgrade` in `ShExPrimerModel.ipynb` is the worst offender: it pulls the *latest* version of linkml's whole dependency tree.

## Impact (already observed)

This caused #3581: `--upgrade` pulled **idna 3.17** over the locked 3.11; a later partial `uv` restore left a mismatched idna install (3.11 `core.py` + 3.17 `uts46data.py`), breaking `test_rewrite_rules` with `ImportError: cannot import name 'uts46data'`. Only Linux was affected because notebook tests are skipped on Windows. The #3581 lock bump papers over this specific case, but **any future dependency release that changes module structure can reintroduce the same class of failure.**

## Proposed fixes (pick one)

1. **Strip install cells before execution** (recommended): have the notebook test runner remove `!pip` / `!uv pip` / `%pip` / `%uv` shell-install cells before `ExecutePreprocessor.preprocess(...)`. The CI env already has everything installed, so these cells add nothing in-test.
2. **Isolate the kernel**: execute each notebook in a dedicated throwaway environment so installs can't touch the shared venv.
3. **Remove/soften install cells in the notebooks themselves** (at minimum drop `--upgrade`), and audit all `notebooks/*.ipynb` install cells.

## Repro / evidence

Instrumented CI (a teardown hook tracking `idna/uts46data.py` size) logged:
```
[...]

PR #3897, opened 2026-08-14
test: serve test network fetches from local files instead of retrying them

amc-corey-cox:
> [!NOTE]
> **Second revision.** After the serve-from-disk rewrite described below, @sneakers-the-rat's review pushed this further into a full redesign of how the suite relates to the network. What's actually in the PR now:
>
> - **Three test categories:** unmarked tests are *blocked* from the network at every layer (urllib, requests, raw sockets); `network`-marked tests run against local stubs (live under `--with-network`, which no longer skips anything); a new `upstream` marker replaces `upstream_main` for tests whose assertions are about the outside world — those run live-only, weekly.
> - **Per-PR CI is fully offline** — no job passes a network flag. The weekly `metamodel-compat` workflow is where reality gets checked: upstream tests, plus a live run of the `network`-marked tests to catch stub drift.
> - **The stub map mirrors the published w3id `.htaccess`**, so a URL that 404s live also fails here.
> - **`VENDORED_MANIFEST`**: `make update_model` records checksums; an offline per-PR test catches vendored files edited without re-vendoring, replacing the per-PR role of the upstream comparison that moved weekly.
> - **Notebook kernels are covered again** — the guard is injected via `sitecustomize`, superseding "What this does not change" below.
>
> Follow-ups filed: #3953, #3954, #3955. The description below is the serve-from-disk revision, kept for history; the original retry design is collapsed at the bottom.

---

Root cause for #3421.

**This PR changed direction after review.** It started as a retry wrapper around transient network failures. @kevinschaper showed the `urllib` half never actually took effect, and @sneakers-the-rat pushed back on the whole approach — make the tests touch the network *less*, not retry harder. Both were right, and the measurement below is what convinced me. The original description is preserved at the bottom.

## The actual bug

```
uv run pytest tests/linkml/ --with-slow --with-biolink -m "slow and not kroki" -n auto
```

That's the `Slow Tests` job, where every reported flake landed. It **never passes `--with-network`** — and I measured it making **34 outbound requests**.

The flakes weren't network tests failing. They were a job that declares itself offline reaching the network anyway.

## Where the traffic came from

Instrumenting `OpenerDirector.open` and `HTTPAdapter.send` across the real CI commands, the whole surface is four test functions:

| Test | Fetches | Local counterpart |
| --- | --- | --- |
| `test_base/test_metamodel.py::test_metamodel[RDFGenerator]` | `w3id.org/linkml/*.context.jsonld` | vendored in `linkml_runtime` |
| `test_scripts/test_gen_rdf.py::test_metamodel` (×3 params) | same | same |
| `test_schemaview.py::test_creature_schema_entities_with_without_imports` | `raw.githubusercontent.com/.../creature_schema.yaml` | in `tests/` |
| `test_schemaview.py::test_imports_remote_url_with_imports` | `raw.githubusercontent.com/.../kitchen_sink.yaml` | in `tests/` |

None are *about* the network. The metamodel ones fetch `@context` URLs as a side effect of rdflib parsing generated JSON-LD — nobody wrote "go fetch w3id". The SchemaView ones test URL-based import resolution.

**Every URL they fetch is already on disk.** So this isn't mocking in the invent-a-response sense; it's pointing at files we ship.

## What this does

`tests/offline_network.py` serves mapped URLs from local files and raises `UnexpectedNetworkAccess` on anything else, with a message naming the URL and the two ways forward. The map is derived from `LOCAL_PATH_FOR`, not hand-listed, so new model files are covered automatically.

Tests that genuinely need the network use the `@pytest.mark.network` opt-in that already exists in the repo.

**Verified offline**, not assumed: with external sockets hard-blocked at `socket.connect`, the full slow suite passes `46 passed, 29 skipped, 3 xfailed` — identical to before.

On timing, don't expect a speedup here: it cut the slow suite from 454s to 101s *on my machine*, where the network was the bottleneck, but CI job times are unchanged (~3m before and after). GitHub's runners reach w3id fast enough that latency was never the problem — flakiness was.

## Two things worth flagging

**It fixes a blind spot.** The GitHub URLs point at `refs/heads/main`, so a branch that edited `creature_schema.yaml` was tested against *main's* copy of it. They now resolve to the working tree.

**`test_url_for_format` was hitting the network unmarked** — gated only on `requests_cache` being importable, not on `@pytest.mark.network`. Now marked. That accounts for 42 tests moving from passed to skipped on the default runtime path; they still run in CI, which passes `--with-network`.

## What this does not change

Divergence checking. `test_vendored_files_match_upstream` / `_main` are separately marked `network` + `upstream_main` and driven by `metamodel-compat.yaml`, which opens a tracking issue on drift. Untouched — and worth saying explicitly, because if that system *didn't* exist, mapping these URLs to local files would have quietly destroyed the only drift signal we had.

## Not covered

Notebook tests run in a separate kernel process and their job opts into `--with-network`, so they can still fail on an upstream hiccup. The earlier kernel-injection machinery is gone with the retry module. I did not execute them locally — they run `!uv pip install --upgrade` cells that mutate the venv off the lock (#3581, #3583).

Separately: generated JSON-LD emits a bare relative `"core.context.jsonld"` in its `@context`, resolved against `@base https://w3id.org/linkml/tests/kitchen_sink/`. Nothing is published there, so it 404s and rdflib silently moves on. Preserved as-is here (served as a 404) to keep this change behaviour-neutral — it's a generator bug and deserves its own issue.

## Unrelated fix retained

`test_shapes_are_cached_per_schema`, added in #3893, asserted the SHACL plugin reuses a cached shape graph across calls. It doesn't: the key is `hash(str(schema))` and `process()` mutates the schema, so a second call keys differently. It passed only because of test ordering. Replaced with an assertion that's actually true, and documented the cache-key wart.

---

<details>
<summary><b>Original description</b> — the retry approach this replaced</summary>

> Root cause for #3421. Swapping individual hostnames treated instances; this treats the mechanism.
>
> ## The gap
>
> `tests/conftest.py` installs a session-wide autouse `requests_cache`. That's the right idea, but it only patches `requests`:
>
> ```
> requests.Session  -> patched      (_ConfiguredCachedSession)
> urllib.urlopen    -> NOT patched  (verified)
> ```
>
> Every flake that has cost us time arrives via **`urllib`** — `hbreader` reading schemas, and `rdflib.parser` — so the cache never sees them:
>
> | Failure | Path |
> | --- | --- |
> | `creature_view_direct_url` — `RemoteDisconnected` | `urllib` |
> | `test_metamodel[RDFGenerator]` — `503` | `urllib`, during YAML load |
> | biolink `503` (original report on #3421) | `urllib` |
>
> **Measured rate:** 5 failures in 108 `Slow Tests` jobs (4.6%). With four such jobs per PR, that's roughly one PR in six.
>
> ## Retry, not cache
>
> Deliberate choice. A cache would also hide genuine upstream breakage, and noticing that is what these tests are for. Retry keeps the signal: transient failures disappear, a genuinely broken upstream still fails.
>
> Only server-side statuses (`429, 500, 502, 503, 504`) and connection-level errors are retried. **A 404 fails on the first attempt** — there's a test pinning that specifically, because retrying a real failure into silence is the way this change could do harm.
>
> Also mounts a `urllib3.Retry` adapter on new `requests` Sessions: `requests_cache.clear()` runs at session start, so the first fetch of each URL in a run is live regardless of the cache. Verified this composes with `requests_cache` rather than displacing it — sessions keep `.cache` and gain retries.
>
> ## Tests
>
> Sixteen tests covering classification, retry-then-succeed, bounded retries on sustained failure, no-retry on 404, and — importantly — that the fixture is *actually installed* during a session. Without that last one the rest is theory.
>
> ## What this does not fix
>
> A sustained upstream outage still fails, after the retries. That's correct behaviour rather than a flake, but it means #3421 shouldn't be closed on the assumption that network failures can never redden a PR again.

**Why it was replaced.** @kevinschaper demonstrated the `urllib` patch never reached its callers: `hbreader` and `rdflib` do `from urllib.request import urlopen` at import time, so rebinding the module attribute is invisible to them. I reproduced it — injecting a 503 below `urlopen`, the PR as written failed with zero retries. His fix (patch `OpenerDirector.open`, the seam every caller funnels through) is what the offline guard uses to install itself.

He also caught that `test_retry_is_actually_installed_during_a_session` asserted the attribute had been reassigned, not that any consumer resolved through it — so it stayed green while the mechanism was inert. The replacement test exercises a real `hbreader` call for exactly that reason.

@sneakers-the-rat argued the whole direction was wrong: reduce network access rather than add retry infrastructure, mock the response directly for tests that explicitly test remote schemas, and treat "this is hard to mock" as a signal about the library rather than a reason for more test infra. Measuring the surface proved him right — it was 4 test functions, not the broad rewrite it looked like, and the retry module was more code than the fix.

</details>


tests/conftest.py
```diff
@@ -354,6 +361,90 @@ def pytest_assertrepr_compare(config, op, left, right):
...
+def _is_transient(exc: BaseException) -> bool:
+    """Return True if ``exc`` looks like a transient network failure.
+
+    Deliberately narrow. A 404 or a bad URL is a real failure and must keep
+    failing; only server-side and connection-level errors are retried, so a
+    genuinely broken upstream still surfaces rather than being papered over.
+    """
+    if isinstance(exc, urllib.error.HTTPError):
+        return exc.code in TRANSIENT_HTTP_STATUS
+    return isinstance(exc, urllib.error.URLError | http.client.HTTPException | ConnectionError | TimeoutError)
```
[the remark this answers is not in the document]
amc-corey-cox: Pushing back on this one — `isinstance` has accepted PEP 604 unions since Python 3.10, and the project requires `>=3.10`. Verified at both ends of the supported range:

```
python 3.14.4   URLError -> True   RemoteDisconnected -> True   ConnectionError -> True   ValueError -> False
python 3.10.20  URLError -> True
```

Three other things corroborate it: ruff's UP038 asked for this exact form; `test_transient_classification` already passes `URLError`, `RemoteDisconnected`, `ConnectionError` and `TimeoutError` through `_is_transient`; and the full suite is green. If the claim held, sixteen tests would be failing.

No change made. Resolving since the other three in this review were correct and acted on.

tests/conftest.py
```diff
@@ -354,6 +361,90 @@ def pytest_assertrepr_compare(config, op, left, right):
...
+    # go through requests_cache below, but it is cleared at session start, so the
+    # first fetch of each URL in a run is still live. Mount a retrying adapter on
+    # every new Session; this composes with requests_cache, which subclasses Session.
+    original_session_init = requests.sessions.Session.__init__
+
+    def session_init_with_retry(self, *args, **kwargs):
+        original_session_init(self, *args, **kwargs)
+        adapter = HTTPAdapter(
+            max_retries=Retry(
+                total=NETWORK_RETRY_ATTEMPTS,
```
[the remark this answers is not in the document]
amc-corey-cox: Good catch, fixed in e5557dd8b. `Retry(total=N)` counts retries rather than attempts, so `total=NETWORK_RETRY_ATTEMPTS` gave the requests path four attempts against the urlopen path's three.

Now `total=NETWORK_RETRY_ATTEMPTS - 1`, with the reason noted inline. Verified parity by exhausting a Retry:

```
urlopen path attempts : 3
requests path attempts: 3
```

tests/linkml/test_conftest_network_retry.py
```diff
@@ -0,0 +1,110 @@
...
+    inner, calls = _failing(urllib.error.HTTPError("u", 404, "Not Found", {}, None))
+    with pytest.raises(urllib.error.HTTPError):
+        with_urlopen_retry(inner)("https://example.invalid")
+    assert calls["n"] == 1
+
+
+def test_retry_is_actually_installed_during_a_session():
+    """The autouse fixture has patched urlopen; without this the rest is theory."""
+    import urllib.request
+
```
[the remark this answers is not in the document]
amc-corey-cox: Correct, fixed in e5557dd8b — the test now skips when `--without-cache` is set.

Worth noting the underlying mistake: I asserted a fixture was installed without accounting for the documented flag that disables it, i.e. an invariant that only holds in the default configuration. Same class of error as the ordering-dependent caching test also fixed in this PR.

tests/linkml/test_conftest_network_retry.py
```diff
@@ -0,0 +1,110 @@
...
+        "did not install, so transient failures will surface as test failures."
+    )
+
+
+def test_requests_sessions_get_a_retrying_adapter():
+    """New requests Sessions carry retries and remain cached."""
+    import requests
+
+    session = requests.Session()
+    assert session.get_adapter("https://example.invalid").max_retries.total == NETWORK_RETRY_ATTEMPTS
```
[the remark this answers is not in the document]
amc-corey-cox: Fixed in e5557dd8b, both parts. The test skips under `--without-cache`, and the assertion follows the `Retry.total` change from the other thread — it now expects `NETWORK_RETRY_ATTEMPTS - 1`, with a comment explaining why that equals three attempts.

Verified in both modes: 16 passed by default, 14 passed / 2 skipped with `--without-cache`.

sneakers-the-rat:
I could have sworn we added something that just full on blocks network access, maybe that was in the old linkml-runtime tests? i think that in general we should try and make the tests not access the network at all, except for a specific category of tests that are just tests for "does this potentially hardcoded URL resolve" (like, e.g. the linkml w3c purls). this seems fine, but sort of goes in the opposite direction of where i would naturally go - make it so the tests access the network less, rather than more.

tests/network_retry.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+
+    def wrapped(*args, **kwargs):
+        for attempt in range(1, NETWORK_RETRY_ATTEMPTS + 1):
+            try:
+                return inner(*args, **kwargs)
+            except Exception as exc:
+                if attempt == NETWORK_RETRY_ATTEMPTS or not _is_transient(exc):
+                    raise
+                time.sleep(NETWORK_RETRY_BACKOFF_SECONDS * attempt)
```
sneakers-the-rat: since we have concurrent tests, and the external server could respond at the same time across tests, we may want to add random jitter to this

kevinschaper:
# PR #3897 — Retry transient network failures instead of failing the suite

**Author:** amc-corey-cox · Root cause for #3421 · +364/-6 across 7 files
**Verdict: Request changes.** The diagnosis is right and the `requests` half works. The `urllib` half — which the PR body identifies as the path *every* cited flake takes — does not take effect. One-line fix.

> Checked empirically, not by reading: I ran the PR branch, inspected live module bindings under the installed fixture, and simulated a transient 503 on the exact path `creature_view_direct_url` uses.

## The urllib patch never reaches its callers

`install()` rebinds the module *attribute*:

```python
urllib.request.urlopen = with_urlopen_retry(original_urlopen)
```

But both libraries the PR names bind the function into their own namespace at import time:

```python
# hbreader
from urllib.request import Request, urlopen
# rdflib/_networking.py
from urllib.request import HTTPRedirectHandler, Request, urlopen
```

Both are imported while `tests/conftest.py` is being imported (it pulls in `linkml_runtime.linkml_model.meta` and `tests.linkml.utils.compare_rdf`), which happens long before any session-scoped fixture runs. They capture the original function, and the later rebinding is invisible to them. Measured on the PR branch, inside a test, with the fixture active:

```
urllib.request.urlopen      : wrapped
hbreader.urlopen            : urlopen        <-- not wrapped
rdflib._networking.urlopen  : urlopen        <-- not wrapped
hbreader.urlopen is urllib.request.urlopen : False
```

End-to-end confirmation. Injecting a single transient 503 *below* `urlopen` (at `AbstractHTTPHandler.do_open`, so nothing in the retry machinery is clobbered) and then doing `SchemaView(CREATURE_SCHEMA_RAW_URL)` — the `creature_view_direct_url` path:

| | result |
|---|---|
| PR as-is | `urllib.error.HTTPError: HTTP Error 503` — **fails, zero retries** |
| wrapping `OpenerDirector.open` instead | passes, 2 underlying opens |

By the PR's own table, all three cited failures (`creature_view_direct_url`, `test_metamodel[RDFGenerator]`, the biolink 503) arrive via `urllib`. As written, none of them are protected. The `requests` half is unaffected by this — patching `Session.__init__` is a class-method patch, so it works as claimed, and the notebook flake in `context_issue.ipynb` (which fetches with `requests`) is genuinely fixed.

## The fix

Wrap the shared seam instead of the module attribute. Every caller of `urlopen` — regardless of how they imported it — funnels through `OpenerDirector.open`:

```python
return opener.open(url, data, timeout)   # last line of urllib.request.urlopen
```

Patching the *class* method (not `_opener.open`) also covers `hbreader`'s `urlopen(req, context=ssl._create_unverified_context())` branch, which builds a fresh opener via `build_opener()` and so would miss an instance-level patch.

```python
original_open = urllib.request.OpenerDirector.open
urllib.request.OpenerDirector.open = with_urlopen_retry(original_open)
```

I applied exactly this and re-ran the 503 simulation: retried, passed.

## The test that should have caught this

```python
assert urllib.request.urlopen.__name__ == "wrapped", (
    "urllib.request.urlopen is not wrapped -- the retry_transient_network fixture did not install..."
)
```

This asserts the attribute was reassigned, not that any consumer resolves through it — so it passes green while the mechanism is inert. That is the same class of false confidence the PR body rightly warns about ("without that last one the rest is theory"). Replace it with something that exercises a real caller: inject a transient failure below `urlopen` and assert `SchemaView(url)` (or a bare `hbreader.hbread`) still succeeds.

## Summary

Fix the seam (`OpenerDirector.open`) and replace the installation assertion with one that exercises a real caller. With those two changes this does what the PR body says it does, and the rest of the design — retry over cache, narrow classification, reaching into the kernel process — is well-argued and well-tested.

sneakers-the-rat:
yeah especially after seeing which tests are flaky - ones where we are explicitly testing remote schema - i definitely think we should just patch the request to *not* use the network and instead mock the response directly. especially if we are having to reach deep into urllib to add retries. 

we could also mimic how cpython itself mocks urllib requests - https://github.com/python/cpython/blob/main/Lib/test/test_urllib.py

seems like compounding complexity since i think it's generally agreed that hbreader should go, having to add a ton of new test infra to handle hbreader seems like wrong direction. another way of saying that, maybe if it's this hard to mock network requests in our own lib, it's a sign we need to have greater control over network requests in the lib.

tests/network_retry.py
amc-corey-cox: Moot now — the retry module is gone. The measurement that came out of your pushback showed the whole network surface was four test functions, all fetching files that already exist on disk, so there's nothing left to back off from.

Thanks for pushing on this; the direction was right.

amc-corey-cox:
Rewrote this in the direction you both pointed. Summary of what changed and why.

@kevinschaper — you were right, and I reproduced it rather than taking it on faith. Injecting a 503 below `urlopen` and calling `hbreader.hbread()`: the PR as written failed with zero retries; wrapping `OpenerDirector.open` retried and passed. `hbreader` and `rdflib` both do `from urllib.request import urlopen` at import time, so the module-attribute rebind never reached them. Your point about the installation test was the sharper one — it asserted the attribute had been reassigned, not that any consumer resolved through it, so it stayed green while the mechanism was inert. The new test drives a real `hbreader` call for exactly that reason. The seam you identified is what the offline guard uses to install itself, so that part of your review survived the rewrite intact.

@sneakers-the-rat — before deciding, I instrumented `OpenerDirector.open` and `HTTPAdapter.send` and ran the actual CI commands. The result made your argument for me:

**The `Slow Tests` job — where every reported flake landed — never passes `--with-network`, and made 34 outbound requests.** These weren't network tests failing. It was a job that declares itself offline reaching the network anyway.

And the surface was four test functions, not the broad rewrite it looked like:

- 2 metamodel RDF tests fetching `@context` URLs, purely as a side effect of rdflib parsing generated JSON-LD
- 2 `SchemaView` tests loading schemas over https to exercise URL-based import resolution

Every URL they fetch is already on disk — the contexts are vendored in `linkml_runtime`, the schemas are in `tests/`. So it isn't really mocking; it's pointing at files we ship. The retry module was more code than the fix. Your instinct that "if it's this hard to mock, that's a signal" held up.

Two things that fell out of it:

**A blind spot.** Those GitHub URLs point at `refs/heads/main`, so a branch editing `creature_schema.yaml` was tested against *main's* copy of it. They now resolve to the working tree.

**`test_url_for_format` was hitting the network unmarked** — gated on `requests_cache` being importable rather than `@pytest.mark.network`. Now marked, which is why 42 tests move to skipped on the default runtime path. They still run in CI.

On your "I could have sworn we added something that blocks network access" — the marker half exists (`network` / `--with-network`), it just had nothing enforcing it. That's what the guard adds: unmapped access raises instead of silently going out.

Verified rather than assumed: with external sockets hard-blocked at `socket.connect`, the full slow suite passes `46 passed, 29 skipped, 3 xfailed` — identical to before, and 454s → 101s since the network was the bottleneck.

Two things I left alone deliberately, both noted in the description: notebook tests still reach the network (their job opts into `--with-network`), and generated JSON-LD emits an unresolvable relative `"core.context.jsonld"` that 404s upstream — preserved as a 404 here to keep this behaviour-neutral, but it looks like a real generator bug and I'd rather file it separately than fold it in.

Divergence checking is untouched — `test_vendored_files_match_upstream` and the `metamodel-compat.yaml` workflow still do that job. Worth stating explicitly, because if that system didn't already exist, serving these URLs from disk would have quietly destroyed the only drift signal we had.

Original description preserved at the bottom of the PR body.

tests/network_retry.py
amc-corey-cox: I went in a different direction that I think obsoletes this.

tests/linkml/test_validator/test_shacl_validation_plugin.py
```diff
@@ -72,13 +72,18 @@ def test_closed_does_not_reject_a_conforming_instance(validation_context):
         next(result_iter)
 
 
-def test_shapes_are_cached_per_schema(validation_context):
-    """Generated shape graphs are reused rather than regenerated per instance."""
+def test_generated_shapes_are_cached(validation_context):
+    """Generating shapes populates the cache, so they are not rebuilt from scratch.
```
sneakers-the-rat: this is a slightly less precise wording -  *per instance* is sort of an important detail

tests/linkml/test_offline_network.py
```diff
@@ -0,0 +1,121 @@
...
+    UnexpectedNetworkAccess,
+    resolve,
+)
+
+CREATURE_SCHEMA_RELPATH = "tests/linkml_runtime/test_utils/input/mcc/creature_schema.yaml"
+
+
+def test_model_urls_resolve_to_vendored_files():
+    """Published linkml-model URLs are served from the vendored copies."""
+    local = resolve(f"{LINKML_URL_BASE}meta.yaml")
```
sneakers-the-rat: i think for all of these we should use the functions in `linkml_files` - preserve the abstraction layer where we have a single way to e.g. "get the url of the metamodel". ensures that the tests don't drift from what (should be) used in the packages.

```python
URL_FOR(Source.META, Format.YAML)
```

(the allcaps were inherited, not my choice)

tests/linkml/test_offline_network.py
```diff
@@ -0,0 +1,121 @@
...
+    assert local is not None
+    assert local.is_file()
+    assert local.name == "meta.yaml"
+
+
+def test_model_url_map_covers_sources_without_a_source_member():
+    """units/datasets/validation ship vendored files but have no ``Source`` member."""
```
sneakers-the-rat: well then this is a sign that they should have a `Source` member, right? and the places that access them should use the accessing functions rather than however they are being accessed now, right?

sneakers-the-rat: could be done in another PR but yeah we _really really really really_ want to have a single source for all our contsants and not be hardcoding things in multiple places and in multiple ways.

tests/linkml/test_offline_network.py
```diff
@@ -0,0 +1,121 @@
...
+    patch of the module attribute would never reach it. This is the assertion that
+    the guard is genuinely installed -- checking that an attribute was reassigned
+    would pass even if no consumer resolved through it.
+    """
+    content = hbreader.hbread(f"{GITHUB_RAW_MAIN}{CREATURE_SCHEMA_RELPATH}")
+    # Compared as raw bytes: hbreader decodes the response without translating
+    # newlines, and ``.gitattributes`` sets ``text=auto``, so the working tree
+    # copy has CRLF on Windows. ``read_text`` would normalise it and never match.
+    expected = (REPO_ROOT / CREATURE_SCHEMA_RELPATH).read_bytes().decode("utf-8")
+    assert content == expected
```
sneakers-the-rat: pretty weak, doesn't directly test what we want to know, but hbreader doesn't make this easy to test with magicmock. another day, another bad consequence of hbreader

tests/linkml/test_offline_network.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def test_schemaview_loads_a_remote_url_offline():
+    """The end-to-end path the flaky tests take, with no network available."""
+    view = SchemaView(f"{GITHUB_RAW_MAIN}{CREATURE_SCHEMA_RELPATH}")
+    assert view.schema.name == "creature_schema"
+    assert "Creature" in view.all_classes()
```
sneakers-the-rat, line 190:
```suggestion
```

doesn't test anything

tests/linkml/test_offline_network.py
```diff
@@ -0,0 +1,121 @@
...
+    "url",
+    [
+        "https://example.invalid/schema.yaml",
+        "https://raw.githubusercontent.com/linkml/linkml/refs/heads/main/does/not/exist.yaml",
+    ],
+)
+def test_unmapped_url_raises_through_a_real_fetch(url):
+    """Unmapped access fails loudly instead of quietly hitting the network."""
+    with pytest.raises(UnexpectedNetworkAccess, match="not served from a local file"):
+        urllib.request.urlopen(url)  # noqa: S310
```
sneakers-the-rat: for the same reasons @kevinschaper flagged before, this should be urlopen and also `hbread` - we want to test the thing that actually does the network access. this reads as an LLM "make tests pass" dodge to me because another related test uses `hbreader` below

tests/linkml/test_offline_network.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+@pytest.mark.network
+def test_network_marked_tests_get_the_real_opener():
+    """Opting in with the marker hands back real network access.
+
+    Asserts the guard is absent rather than making a request, so this stays
+    offline-safe even though it only runs under ``--with-network``.
+    """
+    assert urllib.request.OpenerDirector.open.__name__ != "open_offline"
```
sneakers-the-rat: this is sort of close but also not really how i think this flag should work - for this library, tests should not ever use the network. the only thing we ever need live network for is a specific category of tests that ensure that some hardcoded values (like the URL accessors) correctly resolve. otherwise, we should mock all requests.

so we now have two different things that this flag means/does - now it allows real network access, but not having it uses the mocked local files. however not having `--with-network` also skips all the tests marked `network`. so if we now have a proper request mocking setup, then we can remove the part where we skip marked tests since they should not do any network access, and then the `--with-network` flag is a single-effect toggle that switches between mocked/live requests.

tests/conftest.py
```diff
@@ -354,6 +355,25 @@ def pytest_assertrepr_compare(config, op, left, right):
...
+    if request.node.get_closest_marker("network"):
+        yield
+        return
+
+    uninstall = install_offline_network()
+    try:
+        yield
+    finally:
+        uninstall()
+
```
sneakers-the-rat: yeah this is strange to me and related to https://github.com/linkml/linkml/pull/3897/changes#r3867844721

intuitively i would think *only* the tests marked `network` would need to have this patched, because otherwise the tests shouldn't use the network (arguably they should have network blocked at a socket access level to ensure that). but the mixed meaning of the flag makes this logic inverted.

tests/offline_network.py
```diff
@@ -0,0 +1,182 @@
...
+from email.utils import formatdate
+from functools import cache
+from io import BytesIO
+from pathlib import Path
+
+from linkml_runtime.linkml_model.linkml_files import LINKML_URL_BASE, LOCAL_PATH_FOR, Format, Source
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+
+GITHUB_RAW_MAIN = "https://raw.githubusercontent.com/linkml/linkml/refs/heads/main/"
```
sneakers-the-rat, line 53: this should go in `linkml_files` like

https://github.com/linkml/linkml/blob/680595df54a9daa633fd6f634c8092ea2d4eaf8c/packages/linkml_runtime/src/linkml_runtime/linkml_model/linkml_files.py#L12

tests/offline_network.py
```diff
@@ -0,0 +1,182 @@
...
+DEFAULT_CONTENT_TYPE = "text/plain"
+
+KNOWN_MISSING_URLS = frozenset(
+    {
+        # Generated JSON-LD carries a bare relative "core.context.jsonld" in its
+        # @context, resolved against @base https://w3id.org/linkml/tests/kitchen_sink/.
+        # Nothing is published there, so this 404s upstream too and rdflib moves on.
+        # Served as a 404 to keep behaviour identical; the generator emitting an
+        # unresolvable context reference is a separate bug.
+        "https://w3id.org/linkml/tests/kitchen_sink/core.context.jsonld",
```
sneakers-the-rat, line 114: lets make sure this has an issue to fix it

tests/offline_network.py
```diff
@@ -0,0 +1,182 @@
...
+    }
+)
+
+
+@cache
+def _model_url_map() -> dict[str, Path]:
```
sneakers-the-rat: this also seems like it should go in `linkml_files` - we already have enum -> url so it makes sense to have url -> enum.

tests/offline_network.py
```diff
@@ -0,0 +1,182 @@
...
+    -- which the enum product alone misses.
+    """
+    mapping: dict[str, Path] = {}
+    for fmt in Format:
+        probe = Path(LOCAL_PATH_FOR(Source.META, fmt))
+        extension = probe.name[len(f"{Source.META.value}.") :]
+        # A couple of formats share an extension (SQLDDL/SQLSCHEMA), so the flat
+        # URL space genuinely collides upstream. setdefault keeps this deterministic.
+        for path in sorted(probe.parent.glob(f"*.{extension}")):
+            mapping.setdefault(f"{LINKML_URL_BASE}{path.name}", path)
```
sneakers-the-rat: very odd. we already have the local path for the format. we also already have the path relative to the url base. so why are we reconstructing this naively here? 

`linkml_files` actually already [has something for this](https://github.com/linkml/linkml/blob/680595df54a9daa633fd6f634c8092ea2d4eaf8c/packages/linkml_runtime/src/linkml_runtime/linkml_model/linkml_files.py#L191).

```python
from linkml_runtime.linkml_model.linkml_files import Format, meta

# not very pretty because it's a class rather than a collection, but this could be smoothed if needed
mapping = {}
for fmt in Format.__members__.values():
    location = getattr(meta, fmt)
    mapping[location.url] = location.file
```

that's not *quite* right since there are some mismatches between format names and property names, but if we want this then we can construct the collection as a property on that rather than here.

observe the w3id htaccess file, which is the actual mapping from URLs -> paths: https://github.com/perma-id/w3id.org/blob/master/linkml/.htaccess

those `.sql` files aren't even in that, so they would be 404s, and this would create mappings for any other files in there that are not mapped by the w3id htaccess. if we have a single source of truth for these mappings (like `linkml_files`) rather than in ad-hoc test modules then we can do things like correctly model the htaccess file including its subtleties like which formats it maps.

tests/offline_network.py
```diff
@@ -0,0 +1,182 @@
...
+    parsed = urllib.parse.urlsplit(url)
+    return parsed.scheme in ALLOWED_SCHEMES or parsed.hostname in ALLOWED_HOSTS
+
+
+def _content_type_for(path: Path) -> str:
+    """Return the Content-Type to serve ``path`` as."""
+    for suffix, content_type in CONTENT_TYPE_BY_SUFFIX.items():
+        if path.name.endswith(suffix):
+            return content_type
+    return DEFAULT_CONTENT_TYPE
```
sneakers-the-rat:
```suggestion
    return CONTENT_TYPE_BY_SUFFIX.get(path.suffix, DEFAULT_CONTENT_TYPE)
```

tests/offline_network.py
```diff
@@ -0,0 +1,182 @@
...
+    """Raised when a test reaches for a URL that is not served from disk."""
+
+
+def _unexpected(url: str) -> UnexpectedNetworkAccess:
+    return UnexpectedNetworkAccess(
+        f"Test tried to fetch {url!r}, which is not served from a local file.\n"
+        f"Either add a mapping in {Path(__file__).name} (preferred, if the file is in the repo), "
+        f"or mark the test @pytest.mark.network to opt into real network access."
+    )
+
```
sneakers-the-rat: minor: could just be in the `__init__` method of the exception class

sneakers-the-rat (commented):
i like this a lot better, thanks. my main lingering thought here is that now the `network` marker and the `--with-network` flag have strange behavior.

Previously, we didn't have any sort of request mocking (or stubbing, the difference is small enough to me that i don't really care) so the `network` marker skipped tests that were known to access the network with the intention of making it so by default running the tests required no network access. in general this is how it should go, tests *shouldn't* need network access, that was a partial fix waiting for basically this to happen.

now they do something strange:
- all tests not marked `network` have network access stubbed to return local files if accessing a known URL, erroring otherwise. 
- tests marked `network` are skipped by default
- tests marked `network` *do not* have their requests stubbed and access a live network.

this means that tests that are explicitly about network access like the creature schema resolution tests are *not* marked `network`, which is confusing, and also means that tests marked `network` but don't need real network access (those should all use the stubbed responses too, ideally) are either never run or run with live network (rather than run with the stubbed responses).

So the marker is split across two orthogonal concerns now in a way that has some counterintuitive effects for both future test writers and test runners. 

The ideal fix for this, to me, is:
- all tests that are supposed to access the network are marked `network`, and by default those use the stubbed network (with exceptions for unknown URLs)
- all tests that are *not* supposed to access the network are *not* marked `network`, and an exception is thrown if *any* network requests are attempted. this is a global constraint on when the network can be used, if there is some future regression that causes unwanted network requests deep in the stack, we should be told.
- `network` no longer skips tests
- the `--with-network` flag then toggles whether the tests marked `network` use a real network or the stubbed network.

there is another fix that slices the concerns in the other direction, but this one i think is clearest to me and gives the strongest constraints about how network access is supposed to work.

one other thing: i'm not sure if `requests` passes through the patched codepath, that would be worth a test, because those would be the two ways i imagine network requests would be made.

and finally - we really should just rip out `hbreader`. it doesn't really do anything useful for us, and there is *way more* code that we have had to write to compensate for it than is actually in `hbreader` itself. we *absolutely should* differentiate in our code between filesystem and network access, and a package that mixes those together is a big negative on both code clarity and the complexity of any changes that touch i/o

amc-corey-cox:
@sneakers-the-rat - Thanks for the detailed and thoughtful review. You definitely know the stack and see deeper than I do and I appreciate all of your effort to make this better. It might take me a bit to work through this and get it right. I do think we likely need to address the hbreader problem you mention but I don't think that should happen here. I'll also make issues for the other items that feel out of scope.

tests/linkml/test_validator/test_shacl_validation_plugin.py
amc-corey-cox: Restored — now reads 'not rebuilt per instance validated'.

tests/linkml/test_offline_network.py
amc-corey-cox: Done — the tests now go through `URL_FOR`/`W3ID_FORMATS`.

tests/linkml/test_offline_network.py
amc-corey-cox: Agreed — tracked in #3954. Meanwhile the stub mirrors the htaccess, which serves by filename pattern, so these work without the enum.

tests/linkml/test_offline_network.py
amc-corey-cox: Tightened — asserts the schema name and a known class.

tests/linkml/test_offline_network.py
amc-corey-cox: Kept but rewritten to say what it pins: `file://` stays usable inside block mode. Without it, a regression in that exemption would surface as noise across the suite instead of here.

tests/linkml/test_offline_network.py
amc-corey-cox: Fair. Separate tests now drive urlopen, hbread, requests, and SchemaView through the guard.

tests/linkml/test_offline_network.py
amc-corey-cox: Implemented as proposed, with one addition: tests whose *assertion* is about the outside world (PURL resolution, upstream match) are a third category — `upstream` — run live in the weekly workflow, since stubbing them is vacuous. `network` no longer skips; `--with-network` only toggles stub↔live; unmarked tests hard-error. And you were right about requests — it bypassed the guard entirely; now patched at `HTTPAdapter.send`, with tests.

tests/conftest.py
amc-corey-cox: Fixed — unmarked tests are blocked at all three layers, including a `socket.connect` backstop.

tests/offline_network.py
amc-corey-cox: Left in tests deliberately: this is the URL of *this repo's* test fixtures, not a published linkml-model artifact, and `linkml_files` describes model release locations. Happy to move it if you see it differently.

tests/offline_network.py
amc-corey-cox: Filed: #3953.

tests/offline_network.py
amc-corey-cox: Agreed — same follow-up as the `Source`-member gap: #3954.

tests/offline_network.py
amc-corey-cox: Rewrote the map to mirror the htaccess directly — and you were right that the glob over-served: 14 URLs that 404 live (`meta.sql` included) were being stubbed. Moving the mapping into `linkml_files` is #3954.

tests/offline_network.py
amc-corey-cox: Taken.

tests/offline_network.py
amc-corey-cox: Taken.

.github/workflows/metamodel-compat.yaml
```diff
@@ -203,3 +203,32 @@ jobs:
...
+      # Notebook tests are excluded: their `!uv pip install` cells mutate the
+      # venv off the lock (#3581), which is why per-PR CI gives them a dedicated
+      # runner. Their stub-mode run in per-PR CI is the coverage that matters.
+      # (Cost: URLs only notebooks fetch -- the biocontext monarch context --
+      # are not live-checked anywhere; their stubs could drift silently.)
+      - name: Run network-marked tests against the live network
+        if: always()
+        run: |
+          uv run pytest tests/ --ignore=tests/linkml/test_notebooks \
+            -m "network and not kroki" --with-network -v
```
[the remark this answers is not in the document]
amc-corey-cox: Good catch — exactly right, and those are the URLs the whole PR exists for. Added --with-slow --with-biolink to the step.

sneakers-the-rat:
can we get a summary of the stuff that got vendored since last i looked? like do we need to actually vendor the shex vocab/is that actually used during generation? or something we could stub and just have the parts that are necessary? vendoring referenced code seems better to me than always fetching it from the network, but also stubbing so we are clear about what part of the vendored thing we are relying on would probably be better still: the thing to avoid in my mind is the same thing the network tests are for, preventing drift from some upstream thing that breaks in practice because our vendored code is stale

kevinschaper:
It does look like we could reduce how much of shex.jsonld we bundle, and monarch_context.jsonld even more so. I think I might just try to get this merged though for now.

kevinschaper (approved):

sneakers-the-rat:
cool cool, yeah i like where this ended up, thanks for the nice work, definitely a big improvement to test infra

PR #3901, opened 2026-08-17
Remove YARRRML generator codeowners

Ostrzyciel:
As mentioned [here](https://github.com/linkml/linkml/pull/3664#issuecomment-5318125680), we are no longer using or developing this generator, or the Python implementation of LinkML in general. We are focusing our efforts on [LinkML-Scala](https://github.com/NeverBlink-OSS/linkml-scala) instead.

## Summary

Fixes #

## Checklist

## AI Assistance
none

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -37,11 +37,6 @@
 /tests/linkml_runtime/test_loaders_dumpers/test_dumpers_pydantic.py  @sneakers-the-rat @kevinschaper
 /tests/linkml_runtime/test_loaders_dumpers/test_loaders_pydantic.py  @sneakers-the-rat @kevinschaper
 
-# yarrrml:
-/docs/generators/yarrrml.rst                          @Ostrzyciel @lapkinvladimir
-/packages/linkml/src/linkml/generators/yarrrmlgen.py  @Ostrzyciel @lapkinvladimir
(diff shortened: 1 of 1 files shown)


matentzn (approved):
As per SOP, this is your purview! Thanks for all your work anyways :)

Just as an FYI - it would be nice if we could keep the scope overlap between your ecosystem and this one a bit in check so that we dont split the attention of the community too much into two separate ecosystems (same generators maintained in different places, some people using the one, some the other, issues split onto different trackers, etc). I will mention that again at the workshop - I am sure we find a good solution.

Ostrzyciel:
> Just as an FYI - it would be nice if we could keep the scope overlap between your ecosystem and this one a bit in check so that we dont split the attention of the community too much into two separate ecosystems (same generators maintained in different places, some people using the one, some the other, issues split onto different trackers, etc).

Agreed on having this better organized. :) As for not duplicating generators, this is not really feasible. A SHACL generator for Python is useless for us, because we need a SHACL generator that runs in a browser and the JVM – Python does neither reliably, so it will never be an option for us. It is, however, possible to make native bindings for LinkML-Scala and use them in Python – this way we could make a Python library that wraps LinkML-Scala.

matentzn:
> It is, however, possible to make native bindings for LinkML-Scala and use them in Python – this way we could make a Python library that wraps LinkML-Scala

Yep thats the sort of things I'd look at / discuss in a meeting.

PR #3902, opened 2026-08-17
ci: warn weekly when the vendored model pin is behind the latest release

amc-corey-cox:
Closes the last gap in the vendored-model story: **nothing tells us we haven't picked up a new release.**

`UPSTREAM_SHA` only moves when someone runs `make update_model` by hand. That's deliberate — adopting a new model revision should be a decision, not something that lands on whoever runs CI next. But it means an rc can sit unadopted indefinitely while every existing check stays green:

| Check | Question it answers | Steady state |
| --- | --- | --- |
| hard drift (per-PR, blocking) | do the vendored files match the pinned SHA? | yes — the pin doesn't move |
| soft drift (weekly) | do they match upstream `main`? | almost always no — `main` moves constantly |
| **this (weekly)** | **is there a release we haven't adopted?** | **usually no — speaks only when actionable** |

Comparing against the newest release rather than `main` is the point. `main` diverges the moment anything merges upstream, so that signal is on most of the time and easy to tune out. A release means someone upstream decided something was ready to take.

Runs in the existing weekly `metamodel-compat` job and opens a `vendored-model-stale-pin` tracking issue, closed automatically once the pin catches up. It never blocks a PR.

## One resolver, not two

`update_model` already resolved "latest (pre)release → commit SHA" with two inlined API calls. Rather than repeating them in the workflow, that logic moves into a single Makefile target:

```
make -C packages/linkml_runtime -s print_latest_model_release
# -> v1.11.1-rc1 7657de64dc7098a55dc51dbc9ec500e66f2a1b4f
```

`update_model` and the weekly check now both call it, so they cannot disagree about what "latest" means. The extracted version also guards the empty-list case, which previously surfaced as a bare `IndexError`.

## Verification

Both branches exercised against the live API:

```
pin matches latest   -> exit 0, stale=false
pin altered          -> exit 0, stale=true, tag + commit populated for the issue body
make -n update_model -> resolves v1.11.1-rc1, clones that tag
```

Failure is fail-safe: if resolution returns nothing the step exits non-zero with an explicit message, and since neither `stale=true` nor `stale=false` is set, neither the open-issue nor the close-issue step fires. A broken API cannot silently open a spurious issue or close a real one.

## Note for the reviewer

The workflow now creates the `vendored-model-stale-pin` label itself via `gh label create --force`, so nothing needs doing by hand before the first weekly run — unlike `vendored-model-drift`, which had to be created manually when #3709 landed.

Worth considering separately: now that this covers the actionable case, the soft `main` comparison may be worth dropping. It reports drift in steady state, so its tracking issue will tend to sit permanently open and stop being read.

### DIFF
--- a/.github/workflows/metamodel-compat.yaml
+++ b/.github/workflows/metamodel-compat.yaml
@@ -203,3 +203,97 @@ jobs:
           fi
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      # Nothing updates UPSTREAM_SHA automatically -- `make update_model` is run by
+      # hand -- so a new linkml-model (pre)release can sit unadopted indefinitely
+      # with every check still green. This notices that. Comparing against the
+      # newest release rather than main is deliberate: main moves constantly, but a
+      # release means someone upstream decided something was ready to take.
+      #
+      # Resolution is delegated to `make print_latest_model_release`, the same
+      # target `make update_model` uses, so this check and the vendoring command
+      # cannot disagree about what "latest" means.
+      - name: Check whether the vendored pin is behind the latest release
+        id: pin_check
+        if: always()
+        run: |
+          PINNED=$(tr -d '[:space:]' < packages/linkml_runtime/src/linkml_runtime/linkml_model/UPSTREAM_SHA)
+          read -r TAG LATEST < <(make -C packages/linkml_runtime -s print_latest_model_release)
+
+          # Fail loudly rather than guessing: with no outputs set, neither the
+          # open-issue nor the close-issue step below will fire.
+          if [ -z "${PINNED}" ] || [ -z "${TAG}" ] || [ -z "${LATEST}" ]; then
+            echo "::error::Could not resolve the vendored pin or the latest release (pinned='${PINNED}' tag='${TAG}' latest='${LATEST}')"
+            exit 1
+          fi
+
+          echo "pinned=${PINNED}" >> "$GITHUB_OUTPUT"
+          echo "tag=${TAG}" >> "$GITHUB_OUTPUT"
+          echo "latest=${LATEST}" >> "$GITHUB_OUTPUT"
+          if [ "${PINNED}" = "${LATEST}" ]; then
+            echo "Vendored pin matches ${TAG}."
+            echo "stale=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "Vendored pin ${PINNED} is behind ${TAG} (${LATEST})."
+            echo "stale=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Check for existing stale-pin issue
+        if: always() && steps.pin_check.outputs.stale == 'true'
+        id: check_pin_issue
+        run: |
+          EXISTING_ISSUE=$(gh issue list --label "vendored-model-stale-pin" --state open --json number --jq '.[0].number // empty')
+          echo "existing_issue=${EXISTING_ISSUE}" >> $GITHUB_OUTPUT
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Create or update issue when the pin is stale
+        if: always() && steps.pin_check.outputs.stale == 'true'
+        run: |
+          ISSUE_TITLE="Vendored linkml-model pin is behind the latest release"
+          mkdir -p temp
+
+          {
+            cat <<EOF
+          ## Vendored linkml-model Pin Is Stale
+
+          A newer linkml-model (pre)release exists than the one the vendored copy is
+          pinned to. Nothing adopts it automatically.
+
+          | | |
+          | --- | --- |
+          | Pinned (\`UPSTREAM_SHA\`) | \`${{ steps.pin_check.outputs.pinned }}\` |
+          | Latest release | \`${{ steps.pin_check.outputs.tag }}\` |
+          | Its commit | \`${{ steps.pin_check.outputs.latest }}\` |
+
+          To adopt it, run \`make update_model\` in \`packages/linkml_runtime/\` and
+          commit the result. Expect generated-artifact snapshots to need regenerating
+          in the same change.
+
+          This does not block any PR. It stays open until the pin is updated.
+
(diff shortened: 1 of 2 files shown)


packages/linkml_runtime/Makefile
```diff
@@ -9,21 +9,45 @@
...
+
+BASE = "https://api.github.com/repos/linkml/linkml-model"
+
+releases = json.load(urllib.request.urlopen(BASE + "/releases?per_page=1"))
+if not releases:
+    sys.exit("No linkml-model releases returned. Refusing to guess a revision.")
+
+tag = releases[0]["tag_name"]
+sha = json.load(urllib.request.urlopen(BASE + "/commits/" + tag))["sha"]
+print(tag, sha)
```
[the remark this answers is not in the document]
amc-corey-cox: Both correct, fixed in d20133a4a.

**Drafts:** the resolver now fetches a small page and filters `draft == false`, keeping prereleases since rcs are exactly what we pin to. There are no drafts on linkml-model today so I couldn't reproduce the failure, but the field is there and drafts sort first by creation date, so a maintainer drafting a release would have had `update_model` clone a tag that may not exist yet.

**Timeout:** confirmed `urlopen`'s default is `socket._GLOBAL_DEFAULT_TIMEOUT`, i.e. none at all — a wedged endpoint would have hung the weekly job until the 6-hour runner limit rather than failing. Now 30s on both calls, via a small helper so neither call can be missed.

Verified the resolver still returns `v1.11.1-rc1 7657de64` and that `make -n update_model` picks up the same values.

.github/workflows/metamodel-compat.yaml
```diff
@@ -203,3 +203,97 @@ jobs:
...
+            echo "Vendored pin ${PINNED} is behind ${TAG} (${LATEST})."
+            echo "stale=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Check for existing stale-pin issue
+        if: always() && steps.pin_check.outputs.stale == 'true'
+        id: check_pin_issue
+        run: |
+          EXISTING_ISSUE=$(gh issue list --label "vendored-model-stale-pin" --state open --json number --jq '.[0].number // empty')
+          echo "existing_issue=${EXISTING_ISSUE}" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
amc-corey-cox: Right that this needed handling — fixed in d20133a4a with a `gh label create --force` step ahead of the issue steps. `--force` is documented as "Update the label color and description if label already exists", so it's a genuine no-op on re-runs. That also removes the "needs a label created by hand before the first run" caveat I'd put in the PR description, which is a better outcome than documenting the footgun.

One correction: `gh issue list --label` does *not* fail on a missing label — I tested it and it exits 0 with an empty result. Only `gh issue create --label` fails, which is what the new step covers. Worth noting because it means the check step itself was never at risk, only issue creation.

.github/workflows/metamodel-compat.yaml
```diff
@@ -203,3 +203,110 @@ jobs:
...
+        id: pin_check
+        if: always()
+        run: |
+          PINNED=$(tr -d '[:space:]' < packages/linkml_runtime/src/linkml_runtime/linkml_model/UPSTREAM_SHA)
+          read -r TAG LATEST < <(make -C packages/linkml_runtime -s print_latest_model_release)
+
+          # Fail loudly rather than guessing: with no outputs set, neither the
+          # open-issue nor the close-issue step below will fire.
+          if [ -z "${PINNED}" ] || [ -z "${TAG}" ] || [ -z "${LATEST}" ]; then
+            echo "::error::Could not resolve the vendored pin or the latest release (pinned='${PINNED}' tag='${TAG}' latest='${LATEST}')"
```
matentzn, line 226: Automated review analysis says this is dead code. GitHub runs `run:` under `bash --noprofile --norc -eo pipefail`, so when the resolver produces nothing, `read -r TAG LATEST < <(...)` returns 1 on EOF and `-e` kills the step before it reaches your guard. I simulated it — step exits 1, no message. The behaviour you actually care about is intact (no outputs set, so neither the open nor the close step fires, and I agree that's the property that matters), but "exits non-zero with an explicit message" in the description is only half true, and the message is the half that tells whoever opens the run what went wrong. `|| true` on the read, or capture into a variable first, gets you the diagnostic back.

.github/workflows/metamodel-compat.yaml
```diff
@@ -203,3 +203,110 @@ jobs:
...
+            exit 1
+          fi
+
+          echo "pinned=${PINNED}" >> "$GITHUB_OUTPUT"
+          echo "tag=${TAG}" >> "$GITHUB_OUTPUT"
+          echo "latest=${LATEST}" >> "$GITHUB_OUTPUT"
+          if [ "${PINNED}" = "${LATEST}" ]; then
+            echo "Vendored pin matches ${TAG}."
+            echo "stale=false" >> "$GITHUB_OUTPUT"
```
matentzn, line 235: Super unlikely nit: if anyone ever re-vendors from `main` by hand the weekly job will confidently file "pin is behind the latest release" while it's ahead.

.github/workflows/metamodel-compat.yaml
```diff
@@ -203,3 +203,110 @@ jobs:
...
+          commit the result. Expect generated-artifact snapshots to need regenerating
+          in the same change.
+
+          This does not block any PR. It stays open until the pin is updated.
+
+          **Workflow run:** ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+          EOF
+          } > temp/stale_pin_issue.md
+
+          if [ -n "${{ steps.check_pin_issue.outputs.existing_issue }}" ]; then
```
matentzn, line 292: NIT: When it is stale it re-comments the whole body every week. An rc that sits unadopted for two months is eight identical comments. I know this matches the existing `vendored-model-drift` step so I'm not going to make you fix a pattern you inherited — but it connects to your closing question. I don't think the answer is dropping the soft `main` comparison; it answers a genuinely different question ("do we need to re-vendor before we release" vs "did upstream cut something"). I think the answer is that both tracking issues should only speak when what they'd say has changed, and stay quiet otherwise. That'd fix the rot you're worried about without losing the signal. Separate issue, happy to be argued out of it.

PR #3926, opened 2026-08-27
Use SchemaView in JSON-LD generator

### LINKED ISSUE #3689: Adapt JSON-LD generator to use SchemaView

Silvanoc:
> [!Warning]
> This PR builds upon #3796 and #3928 and will remain as `draft` until those PRs have been merged.

## Summary

Fixes #3689

Replace the use of SchemaLoader (to-be-deprecated) with its replacement SchemaView.

The goals of this PR are not only updating the generator to replace SchemaLoader, but also getting closer (through the introduction of SchemaView) to the possibility of addressing the [namespacing topic](https://github.com/orgs/linkml/discussions/1739) through [SchemaView recursivity](https://github.com/orgs/linkml/discussions/1739#discussioncomment-8599940) as proposed by @sneakers-the-rat .

## How was this tested?

This is a refactoring that shouldn't require additional tests, but run with existing ones. Unfortunately it involves some small modifications that require snapshots updates.

## Areas of uncertainty

None

## Checklist

## AI Assistance

### DIFF
--- a/packages/linkml/src/linkml/generators/jsonldgen.py
+++ b/packages/linkml/src/linkml/generators/jsonldgen.py
@@ -84,13 +84,6 @@ class JSONLDGenerator(Generator):
         self.schema.classes.update(self.schemaview.all_classes())
         self.schema.subsets.update(self.schemaview.all_subsets())
         self.schema.enums.update(self.schemaview.all_enums())
-        # SchemaView's inject_metadata sets from_schema on class attributes (embedded
-        # SlotDefinitions), but SchemaLoader does not. SchemaView's induced_slot may also
-        # set domain_of on attributes as a side effect. Clear both to match SchemaLoader.
-        for cls in self.schema.classes.values():
-            for attr in cls.attributes.values():
-                attr.from_schema = None
-                attr.domain_of = []
         # Materialise reverse inverses, mirroring SchemaLoader (schemaloader.py:304-317):
         # for every slot declaring ``inverse``, set the inverse relationship on the target
         # slot when it does not already declare one. SchemaView does not do this, so the
@@ -772,12 +765,6 @@ class JSONLDGenerator(Generator):
             context_kwargs = {**default_context_kwargs, **context_kwargs}
 
         self._add_type(self.schema)
-        # SchemaView's induced_slot may set domain_of on class attributes as a side effect
-        # during inlined inference. Clear it before serialization — SchemaLoader never sets
-        # domain_of on embedded class attributes.
-        for cls in self.schema.classes.values():
-            for attr in cls.attributes.values():
-                attr.domain_of = []
         base_prefix = self.default_prefix()
 
         # `context` can be a `str`, a `list[str]` or a `tuple[str]`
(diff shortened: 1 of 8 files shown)


packages/linkml/src/linkml/generators/jsonldgen.py
```diff
@@ -64,8 +65,555 @@ class JSONLDGenerator(Generator):
...
 
     def __post_init__(self) -> None:
+        if self.uses_schemaloader:
+            raise ValueError(
+                f"{type(self).__name__} is a SchemaView-only generator and does not support the "
+                "SchemaLoader path; 'uses_schemaloader' must remain False."
+            )
         self.original_schema = deepcopy(self.schema)
         super().__post_init__()
+        # super().__post_init__() takes the SchemaView path (uses_schemaloader is False)
```
sneakers-the-rat, line 75: so what is all this stuff in the post-init method? or, why does it go here? i am reading the comments and not exactly sure why this generator needs so much specialization outside of the schemaview class itself. In my mind, the division of labor is like 

- schemaview is a means of materializing the schema and interacting with the materialized schema
- a generator is a way of translating the materialized schema into a different modality

so a generator should not be manipulating the schema much, if at all, and just be using the schemaview to construct the output domain. that's an important division of labor so that the schemaview can be developed independently, and then any bugfix/behavior change propagates to the generators which have an abstract view on the schema via schemaview (the lack of information hiding in schemaview, as well as the ability to use schemaview to mutate the schema is imo a design flaw)

It looks like here we are using `SchemaView` to update the `SchemaDefinition` object, maybe to preserve how the rest of the generator expects to use the `schema` attr, but in my mind that would be the target of the schemaview retrofit, to replace the usage of the raw `SchemaDefinition` with a `SchemaView`. This is a general problem with linkml, where there is an ambiguity between the concrete definitions and the induced definitions, which share the same classes and only differentiated by how they are accessed - and so in the absence of proper type distinction, the schemaview serves as the materialized schema while `SchemaDefinition` is more or less just an in-memory representation of the yaml.

in any case this shouldn't all be in `__post_init__` and at least should be split up into small, intelligible methods, but yeah i would think that we would want to not do most of this and replace the rest of the generators calls to `self.schema` with `self.schemaview`

Silvanoc: TL;DR: I fully agree with you, the code is not meant to be reviewed until "Ready to review".

Sorry, I didn't mean to pull you for a review. I just wanted to give context in the PR description. This PR is still very early draft not meant to be reviewed yet. Even applying patch #3793 manually, I still find some differences between local and GitHub testing, that's why I create PRs in so early phases.

I some times let AI propose a solution (like in this case) that I then review and only mark as "Ready for review" once I'm satisfied with the result or after I've implemented it myself. In this case I fully agree that the AI generated code feels wrong: too much in jsonldgen.py that probably belongs into SchemaView. For example, during that analysis I've realized that the AI implementation is working around bug #3927. So I'm fixing that bug first and that way throwing more AI crap away. I rewrite the git history, so no rubbish commits are left behind.

PR #3941, opened 2026-09-01
Add the RDFS generator

### LINKED ISSUE #3939: Add an RDFS generator

Add a generator for RDF Schema, as a much lighter alternative to the OWL generator. This should cover the classes, property definitions, subclass/subproperty relations and ranges.

**Motivation:** OWL is a very large and complex standard, making working with it challenging. Still, in many cases, a concise open-world description of a LinkML schema is useful, for example for light vocabularies, or for pairing it with SHACL shapes.

**Usage examples:** We currently use the RDFS generator to generate the [RDF Tensor vocabulary](https://neverblink-oss.github.io/rdf-tensor/ontology/functions/) from LinkML. It does not need OWL, so RDFS is a good fit. A second use case is pairing the RDFS generator with the SHACL one – [SHACL validators do respect RDFS subclassing](https://www.w3.org/TR/shacl/#terminology), so exporting both SHACL and RDFS is pretty common. We use RDFS extensively in our customers' enterprise projects (often paired with SHACL).

**Implementation:** An RDFS generator is already present in [LinkML-Scala](https://github.com/NeverBlink-OSS/linkml-scala). As discussed with @matentzn last week, we want to try upstreaming one generator to LinkML-Py using LinkML-Scala's bindings. This generator is relatively simple, so we start with it – we can later expand this to more generators.

**Additional context**

Like I wrote above, this issue is a bit of a pilot for testing how the upstreaming process works. If successful, we can also upstream the Scala and Frictionless generators, which are currently not present in LinkML-Py. We initially planned to upstream the Frictionless generator first, but RDFS turned out to be a bit simpler.

The ultimate goal is to lower the maintenance burden on the LinkML-Py core team. In the future, if the compatibility is proven to be good enough, we could use this process to also take over the maintenance of some less critical generators like GraphQL, which are also present in LinkML-Scala. Of course, this is to be discussed with the community and depends on how this pilot will work.

I will submit a PR for this issue shortly.

Ostrzyciel:
## Summary

Fixes #3939

Please see the issue for the context and the motivation. This is important to understand what is going on here.

This is a wrapper around the [LinkML-Scala](https://github.com/NeverBlink-OSS/linkml-scala) Python bindings – [`neverblink-linkml`](https://pypi.org/project/neverblink-linkml/) on PyPI.

Design notes:

- The API mimics the Python generators as close as possible. There are some differences, though. You cannot pass an already parsed schema to the generator, because LinkML-Scala has its own importer and SchemaView implementations.
- I created the `GeneratorBase` ABC as a neutral interface for generators that does not depend on anything specific from LinkML-Py, other than generation options. This is important, because it lets us skip creating SchemaView in Python and in Scala, which would be very wasteful.
- I took care to keep the `Generator` ABC exactly as it was. For example, there are tests asserting that the field order is unchanged. This is a safe, backwards-compatible change.
- Not all import options are supported yet, we have an upstream issue for this: https://github.com/NeverBlink-OSS/linkml-scala/issues/210 I don't think this is a blocker, we can iteratively improve this as we go.
- The tests for RDFS generator are minimal, because we already test it extensively in LinkML-Scala.
- LinkML-Scala's SchemaViews are held in native memory. Even if users forget to use the context manager, they will be freed correctly.
- This adds a dependency on `neverblink-linkml`, which is a natively-compiled package, about 2–3 MB. It has zero transitive dependencies. If we later upstream other generators from LinkML-Scala, they will require no further dependencies.
- `neverblink-linkml` has native wheels **and** an sdist, so it can be built from source during installation. The 8 wheels cover all popular platforms. In case users have something exotic (e.g., RISC-V), they can use the sdist. We've tested it in QEMU and it does work, for example on RISC-V and NetBSD.
  - conda-forge is not supported, though. We could do this, but the conda-forge package for linkml has about as many downloads in a year as the PyPI one does in a *day*, so it doesn't seem worthwhile at the moment.
- The help texts clearly instruct users to file bugs in LinkML-Scala. If users still come here with bugs, we can work out a process to redirect them somehow. We'd be happy to help with that.

**EDIT 2026-09-11:** I've updated the PR to neverblink-linkml version 0.16.0, which introduces sdist and more wheels, thanks to a different build process based on Scala Native. The binaries are also 2–3x smaller now.

## How was this tested?

- There is decent test coverage of the new code.
- There are tests asserting that the Generator ABC is intact and stable.
- I did test it manually with some large schemas. It works fine.

## Areas of uncertainty

I think that what I did to the `Generator` here is the minimal possible edit. But, if you see any possibility to make this change even smaller, I'd be happy to do that.

## Checklist

## AI Assistance


Ostrzyciel:
@niegrzybkowski requesting a review :)

niegrzybkowski (approved):
Looks good to me

matentzn (changes requested):
This looks great - requesting only changes pro forma because this needs a larger discussion I will have to lead. 

@Ostrzyciel THANKS a ton!

Ostrzyciel:
I was asked off-GitHub in relation to this PR whether this would work for people who install LinkML from conda-forge: https://anaconda.org/channels/conda-forge/packages/linkml/overview

In short no, it would not. These people will get a message on runtime when they try to use the RDFS generator that it's not supported on their platform.

I checked the conda-forge requirements, and while we could adapt to this in some way, it would require quite a lot of work. The `linkml` package in question has been downloaded only 3.4k times over the last six months ([stats](https://anaconda.org/channels/conda-forge/packages/linkml/overview)). For context, `linkml` on PyPI gets 2x more downloads [***per day***](https://pypistats.org/packages/linkml), so this is a very niche distribution.

I'd be happy to add support for conda-forge if someone requests this specifically, but I don't think we need to do this right now.

Ostrzyciel:
Hey, just a heads-up that we've found a solution for making sdist :) this also allows LinkML-Scala in Python to run on basically any platform (OS or CPU) and makes the binary size even smaller. I will update the PR in a few days with this new release, this should help with compatibility concerns.

Ostrzyciel:
@matentzn I've updated the PR. We now have sdist, more wheels, and binary is also smaller. I hope this will help address any concerns :)

We also have 3 further generators that we can upstream in quick succession if this gets merged: Apache Ossie, Scala, and Frictionless.

Ostrzyciel:
Following a discussion with the core team yesterday, we've decided to first merge the Ossie generator, which does not require additional discussions around the shape of the generated RDFS, so we can progress with it faster. See #4029

I will keep this PR open. After we merge the Ossie PR, I will rebase it to remove the parts that are identical between the PRs.

PR #3943, opened 2026-09-01
pydantic: Correctly handle a titled enum value in ifabsent.

### LINKED ISSUE #3942: Pydantic generator produces invalid code for enum-typed slots with `ifabsent`
**Describe the bug**
Given a slot (i) whose range is set to an enumeration where (ii) enumeration values have a `title` attribute and (iii) the slot has a `ifabsent` attribute referring to one of the enumerated values, then the Pydantic generator will produce code that cannot be loaded by Python (loading results in an `AttributeError`).

(This looks similar to #2961, but is worse, because this does not merely result in a Mypy error.)

Tip of the _main_ branch as of September 1st, 2026.

Consider the following schema (preamble omitted _brevitatis causa_):

```yaml
enums:
  LengthUnit:
    permissible_values:
      mm:
        title: millimeter
      cm:
        title: centimeter
      nm:
        title: nanometer

classes:
  MyClass:
    attributes:
      length:
        range: integer
      length_unit:
        range: LengthUnit
        ifabsent: LengthUnit(mm)
```

Observe how all enum values in _LengthUnit` have a `title` attribute.

Generating Python (data classes) code from that schema works fine:

```sh
$ linkml generate python test.yaml > test.py
$ python -mtest
[no error]
```

But doing the same with Pydantic code yields an error:

```sh
$ linkml generate pydantic test.yaml > test.py
$ python -m test
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/damien/Tests/linkml/pydantic-enum-title/test.py", line 83, in <module>
    class MyClass(ConfiguredBaseModel):
  File "/Users/damien/Tests/linkml/pydantic-enum-title/test.py", line 87, in MyClass
    length_unit: Optional[LengthUnit] = Field(default=LengthUnit.mm, json_schema_extra = { "linkml_meta": {'domain_of': ['MyClass'], 'ifabsent': 'LengthUnit(mm)'} })
                                                      ^^^^^^^^^^^^^
  File "/Users/damien/.local/share/uv/python/cpython-3.11.14-macos-aarch64-none/lib/python3.11/enum.py", line 786, in __getattr__
    raise AttributeError(name) from None
AttributeError: mm
```

Looking at the generated Pydantic code (comments added and “meta” stuff removed):

```python
class LengthUnit(str, Enum):
    # The symbol name is derived from the title, not from the value;
    # this is supposed to be used as LengthUnit.millimeter,
    # not as LengthUnit.mm!
    millimeter = "mm"
    centimeter = "cm"
    nanometer = "nm"

class MyClass(ConfiguredBaseModel):
    length: Optional[int] = Field(default=None)
    # The default value is trying to refer to the `mm` enum value
    # by its value rather than by its symbol
    length_unit: Optional[LengthUnit] = Field(default=LengthUnit.mm)
```

gouttegd:
## Summary

Fixes #3942

This commit amends the _PydanticIfAbsentProcessor_ to make sure it does not produce invalid (unloadable) code when a `ifabsent` attribute refers to an enum value that has a `title`.

When an enum value has a title, the symbol for the value is derived from the title rather than from the value’s `text`, so trying to resolve the default value with `{enum_name}.{permissible_value_name}` will result in an attribute error when the generated module is loaded because `permissible_value_name` is always derived from the value’s `text`, regardless of whether the permissible value has a `title` or not.

To fix, we make the generator reference the default value as `{enum_name}("permissible_value_name")` instead – that is, calling the enum constructor with the text of the default value.

## How was this tested?

Added an explicit test case with a schema that (i) defines an enum with “titled values” (permissible values that have a `title` attribute) and (ii) defines a class with an attribute with a range set to the enum and a `ifabsent` referencing one of the permissible values. The test generates the Pydantic code for that schema and attempts to compile it – which would currently fail without the fix in this PR.

## Areas of uncertainty

Not really.

Of note, I would personally have preferred to make the _PydanticIfAbsentProcessor_ reference the default value as `EnumName.ValueSymbol` (where `ValueSymbol` is derived either from the permissible value’s `text` or from its `title`, depending on whether it has a title or not), rather than as `EnumName("value")`. But this would have required, either

* duplicating the logic of [`oocodegen.generate_enum_label`](https://github.com/linkml/linkml/blob/bb14aef63f721e741e7a50e28a11ee0738e340be/packages/linkml/src/linkml/generators/oocodegen.py#L194) – not keen to do that as it introduces a risk of the duplicated logic not being kept in sync with the original;
* somehow find a way to call `oocodegen.generate_enum_label` from the _PydanticIfAbsentProcessor_, which does not seem easily doable because the processor has seemingly no handle to the code generator that called it.

So in the end, relying on the enum constructor to do the right thing (`EnumName("value")`) seems by far the easiest and least intrusive fix.

## Checklist

## AI Assistance

None.

### DIFF
--- a/packages/linkml/src/linkml/generators/pydanticgen/pydantic_ifabsent_processor.py
+++ b/packages/linkml/src/linkml/generators/pydanticgen/pydantic_ifabsent_processor.py
@@ -10,4 +10,4 @@ class PydanticIfAbsentProcessor(PythonIfAbsentProcessor):
     def map_enum_default_value(
         self, enum_name: EnumDefinitionName, permissible_value_name: str, slot: SlotDefinition, cls: ClassDefinition
     ):
-        return f"{enum_name}.{permissible_value_name}"
+        return f'{enum_name}("{permissible_value_name}")'
--- a/tests/linkml/test_issues/test_linkml_issue_3942.py
+++ b/tests/linkml/test_issues/test_linkml_issue_3942.py
@@ -1,3 +1,5 @@
+import pytest
+
 from linkml.generators.pydanticgen import PydanticGenerator
 from linkml_runtime.linkml_model import (
     ClassDefinition,
@@ -7,6 +9,8 @@ from linkml_runtime.linkml_model import (
     SlotDefinition,
 )
 from linkml_runtime.utils.compile_python import compile_python
+
+pytestmark = [pytest.mark.pydanticgen]
 
 
 def test_pydantic_generator_correctly_reference_default_enum_values():
--- a/tests/linkml/test_issues/test_linkml_issue_3942.py
+++ b/tests/linkml/test_issues/test_linkml_issue_3942.py
@@ -1,0 +1,35 @@
+from linkml.generators.pydanticgen import PydanticGenerator
+from linkml_runtime.linkml_model import (
+    ClassDefinition,
+    EnumDefinition,
+    PermissibleValue,
+    SchemaDefinition,
+    SlotDefinition,
+)
+from linkml_runtime.utils.compile_python import compile_python
+
+
+def test_pydantic_generator_correctly_reference_default_enum_values():
+    """ifabsent: MyEnum(MyValue) should work when MyValue has a title."""
+    schema = SchemaDefinition(
+        id="ifabsent_referencing_titled_enum_value",
+        name="ifabsent_referencing_titled_enum_value",
+        enums=[
+            EnumDefinition(
+                name="LengthUnit",
+                permissible_values=[
+                    PermissibleValue(text="mm", title="millimeter"),
+                    PermissibleValue(text="cm", title="centimeter"),
+                ],
+            ),
+        ],
+        classes=[
+            ClassDefinition(
+                name="MyClass",
+                attributes=[SlotDefinition(name="length_unit", range="LengthUnit", ifabsent="LengthUnit(mm)")],
+            ),
+        ],
+    )
+    gen = PydanticGenerator(schema=schema)
+    code = gen.serialize()
+    compile_python(code)
--- a/tests/linkml/test_issues/test_linkml_issue_3942.py
+++ b/tests/linkml/test_issues/test_linkml_issue_3942.py
@@ -36,4 +36,5 @@ def test_pydantic_generator_correctly_reference_default_enum_values():
     )
     gen = PydanticGenerator(schema=schema)
     code = gen.serialize()
-    compile_python(code)
+    mod = compile_python(code)
+    assert mod.MyClass.model_fields["length_unit"].default is mod.LengthUnit.millimeter


packages/linkml/src/linkml/generators/pydanticgen/pydantic_ifabsent_processor.py
```diff
@@ -10,4 +10,4 @@ class PydanticIfAbsentProcessor(PythonIfAbsentProcessor):
     def map_enum_default_value(
         self, enum_name: EnumDefinitionName, permissible_value_name: str, slot: SlotDefinition, cls: ClassDefinition
     ):
-        return f"{enum_name}.{permissible_value_name}"
+        return f'{enum_name}("{permissible_value_name}")'
```
sneakers-the-rat, line 13: nice. i haven't accessed python enums by value like this before, but [lo and behold](https://docs.python.org/3/library/enum.html#enum.EnumType.__call__) - `assert LengthUnit("mm") is LengthUnit.millimeter` - learned something new today, thanks for showing me this

tests/linkml/test_issues/test_linkml_issue_3942.py
sneakers-the-rat, line 1: this style is endemic in the package already, but my strong preference is that tests are organized in meaningful units that correspond to the code under test - among other reasons, so that it is possible to "run the pydantic tests" and "see what tests pydantic enums." organizing tests by issue largely turns them into a junk drawer, because the test becomes a de facto regression test against "the thing that is in issue xyz" as opposed to a positive test that affirms some desired behavior. 

So i will make the same comment i always make when people add new "issue xyz" tests - this is not a hard requirement because similar tests do already exist and it wouldn't be fair to make some special requirement for this pr/you/the pydantic generator, but my strong preference would be to have this be within the pydanticgen tests and be named/structured something like

```python
def test_ifabsent_enums_with_title():
    """
    PermissibleValues with titles are instantiated from the enum value when used as defaults

    References:
        https://github.com/linkml/linkml/issues/3942
    """
    # ...
```

failing that, could we at least add `pytest.mark.pydanticgen` to it?

tests/linkml/test_issues/test_linkml_issue_3942.py
```diff
@@ -0,0 +1,35 @@
...
+        classes=[
+            ClassDefinition(
+                name="MyClass",
+                attributes=[SlotDefinition(name="length_unit", range="LengthUnit", ifabsent="LengthUnit(mm)")],
+            ),
+        ],
+    )
+    gen = PydanticGenerator(schema=schema)
+    code = gen.serialize()
+    compile_python(code)
```
sneakers-the-rat: this satisfies as a regression test, but while we're here we should actually test the thing we want to remain true in the future (that the default is the member of the enum accessed by the title)

```suggestion
    gen = PydanticGenerator(schema=schema)
    code = gen.serialize()
    mod = compile_python(code)

    assert mod.MyClass.model_fields["length_unit"].default is mod.LengthUnit.millimeter
```

sneakers-the-rat (approved):
great! looks good :)

I went and refreshed my memory and took a look at the vanilla pythongen to see why this didn't need to get fixed there, and it's because the pythongen uses the special linkml-style enums instead of the stdlib enums and instantiates with a string rather the enum member - so just confirming to other reviewers that we don't need to do that/the pydantic-specific patch to handling enums is warranted.

only minor things in my comments - reorganizing the test to be within the pydanticgen tests is optional but strongly preferred (by me, i dont' know how other maintainers feel about new issue-numbered tests, which is why i'm saying optional) - if that isn't done, would definitely want to add the pydanticgen marker so that it's run whenever running pydantic tests. also would strongly prefer adding something like the suggestion to explicitly protect the behavior we do want vs. preventing behavior we don't want.

> I would personally have preferred to make the PydanticIfAbsentProcessor reference the default value as EnumName.ValueSymbol

yeah, same, and i agree with your reasons why not (the ifabsent processor should not sprout a reference to its parent for information hiding reasons and we should not duplicate code) and like where you ended up. i think that all the ad-hoc string manipulation in the library is one of its biggest weaknesses, and having that be `oocodegen.generate_enum_label` rather than a more generalized "string as symbol" method is a good example of why (@Silvanoc and i [tried to get started on this](https://github.com/linkml/linkml/blob/main/packages/linkml/src/linkml/generators/common/naming.py) awhile ago in #2303 but we haven't made it that far)

thanks again!

tests/linkml/test_issues/test_linkml_issue_3942.py
gouttegd: We’ve had this discussion before, as I recall. :)

Same answer as before: I have no personal preference in the matter, but it so happens that creating a `test_linkml_issue_NNNN.py` file is what the LinkML project [explicitly recommends](https://linkml.io/linkml/maintainers/contributing.html#directions-for-creating-a-unit-test) when adding a test for a bug fix:

> While creating a test for a bug, you should make a file called `test_linkml_issue_NNN.py`.

(So it’s not really surprising that this “style” is “endemic”.)

> if that isn't done, would definitely want to add the pydanticgen marker so that it's run whenever running pydantic tests.

I didn’t know that there was a `pydanticgen` marker, but using it would indeed makes perfect sense, I’ll do that know, thanks!

tests/linkml/test_issues/test_linkml_issue_3942.py
gouttegd: Makes sense, thanks!

tests/linkml/test_issues/test_linkml_issue_3942.py
sneakers-the-rat: Oops, sorry, didn't mean if that came across as aggro to repeat something I already said. Can never remember, its just mental boilerplate for me.

gouttegd: No worries. :)

sneakers-the-rat (approved):
Cool cool I'm good with this

kevinschaper (approved):

Issue #3923, opened 2026-08-25
Prepare support for Python 3.15


Silvanoc:
**What is your feature request?**

[Python v3.15rc1](https://www.python.org/downloads/release/python-3150rc1/) has been released.

We should start testing LinkML against this and future release candidates to be v3.15 compatible form day 1.

Probably preparing a PR running the tests against Python 3.15 RCs is the best way to identify the issues early on.

**Additional context**

Until now LinkML has been acting reactively to Python version releases. This issue tries to address release of Python v3.15 pro-actively to reduce future pressure once the release is done.

dalito:
Oh. Now I see this is assigned to @melonora. #3945 is just to get us started early.

melonora:
Yeah I need to check all the shex libraries:) it is on my radar

dalito:
Interestingly, linkml-runtime is fully clean on 3.15 (rc2). All failures are in the linkml suite, 

It was important to fix caching to avoid 45 min built time on every run of 3.15.

Regarding #3945, I suggest:
- [ ] review & merge #3945
- [ ] in repo settings: switch 3.13 bound gates to 3.14 (or write gates in version-independent way)
- [ ] Update ci to no longer run gated runs on 3.13 (this reduces CI time as #3945 runs them on 3.13 and 3.14 to allow migration of the gates)

PR #3945, opened 2026-09-01
ci: test against Python 3.15 release candidates

dalito:
### Summary

Adds 3.15 to the `test` matrix, first only on Linux, following the suggestion in #3923 to run the test suite against release candidates. The job reports failures without blocking a PR.

Supporting changes were needed:

- **uv supplies the 3.15 interpreter.** actions/python-versions built rc2 on 2026-09-02, but its `versions-manifest.json`, which setup-python actually reads, still lists 3.15.0-rc.1 as newest. python-build-standalone published rc2 on 2026-09-01, so setup-uv now sets `UV_PYTHON` for pre-release versions and setup-python is skipped for them; stable versions are unchanged. This needs uv 0.12.x. The `UV_VERSION` pin was moved from 0.11.21 to 0.12.9 in every workflow.

 - **Each job gets its own uv cache entry.** setup-uv derives the Python component of its cache key from `uv python find`, which runs before setup-python and therefore reports the runner's system Python, so all six matrix jobs shared one key. The bigger issue is that `quality-checks` caches by default and populates nothing, but it won the cache reservation race.  This stopped every other job from saving. So the 3.15 job, which has to build many dependencies from source, spent ~40 minutes in `uv sync` on every run. `cache-suffix` now separates the matrix jobs and `quality-checks` no longer caches.

- Windows tested on 3.13 and 3.14 although this makes no sense. I changed Windows tests to run on the oldest and newest stable release (3.10 and 3.14) in the 2nd commit.

- Despite the two test suites of linkml and linkml-runtime being independent, a failure in the linkml suite prevented the linkml-runtime suite being run. This was change to always run both which is important here to see the issues in both packages. The job still fails if either suite fails.

### How was this tested?

See CI output here.

### Checklist


dalito:
It may well be that we need to drop 3.10 at the same time since some dependencies add 3.15 and drop 3.10 at the same time, for example [rpds-py](https://github.com/crate-py/rpds/releases/tag/v2026.5.1).

Silvanoc:
> It may well be that we need to drop 3.10 at the same time since some dependencies add 3.15 and drop 3.10 at the same time, for example [rpds-py](https://github.com/crate-py/rpds/releases/tag/v2026.5.1).

That's gonna be a challenge. It makes sense to start deprecating 3.10, since it'll reach EOL October this year.

But still my feeling is that there are more people running outdated systems (who will want to keep 3.10 some time after its EOL) than those running bleeding edge (who will want to use 3.15 right from its release).

dalito:
Keeping older versions seems to be fine. With an updated uv.lock Python 3.10 and 3.15 can both be used with the same config.

See last run https://github.com/linkml/linkml/actions/runs/33575184388/job/100077639262?pr=3945

dalito:
@matentzn - Can you check if there is something configured in the settings which required Python 3.13 jobs to pass? I think there is and so I kept 3.13. - However it would save CI time if we could use 3.14 which is already tested as "latest".

Nice surprise: linkml-runtime is fully clean on 3.15. All failures are in the linkml suite,

dalito:
Run https://github.com/linkml/linkml/actions/runs/33624750370/job/100229958000?pr=3945 shows that dependencies built for 3.15 are now installed from cache, thanks to the fixed caching. This reduces the time for the 3.15 tests by ~40 min. 

Testing with 3.15 is now as fast as testing with the other Python versions.

dalito:
This PR is ready now. 

Once merged, there should be no negative effect on other PRs (at least that was the plan). It just adds the 3.15 test run. The ability to merge should not depend on green 3.15 tests. - But please review carefully.

matentzn:
Great effort @dalito @Silvanoc;

I am not that certain about whether we should keep (even non-blocking) red tests around.. It would be a quite confusing signal for folks opening PR to see the RED, even if it is not blocking. Why not just wait until everything upstream is fixed and merge then?

dalito:
> I am not that certain about whether we should keep (even non-blocking) red tests around.. It would be a quite confusing signal for folks opening PR to see the RED, even if it is not blocking. Why not just wait until everything upstream is fixed and merge then?

Waiting and frequently rebasing this PR is also fine. Whoever is interested in the status of Python 3.15 support can look at this PR's runs.

@matentzn - The PR fixes some other issues, the caching and the use of Python 3.13 (instead of 3.14) as default build env and gate. Do you want to tackle those in another issue/PR?

Currently the `uv.lock` file is in conflict with the one on `main`, which makes rebasing more work. Would be great to avoid that.

PR #3963, opened 2026-09-06
fix(openapigen): handle linkml examples slot mapping to example field

### LINKED ISSUE #3962: gen-openapi: slot `examples` crashes generation (openapi 3.0 has no `examples`)
**Describe the bug**

A LinkML slot with `examples:` cannot be generated to OpenAPI at all. 

The OpenAPI 3.0.3 Schema Object has no `examples` field - only singular `example` (a bare value, [[§4.7.16.1][schema-ex]](https://spec.openapis.org/oas/v3.0.3.html#fixed-fields-19)): _A free-form property to include an example of an instance for this schema. To represent examples that cannot be naturally represented in JSON or YAML, a string value can be used to contain the example with escaping where necessary._

OpenAPI does have a plural `examples`, but in different context and its a `Map` (`string ->  Example Object`) and mutually exclusive of above `example` field, and where both are used `examples` wins.
Parameter Type: https://spec.openapis.org/oas/v3.0.3.html#fixed-fields-9
Media Type: https://spec.openapis.org/oas/v3.0.3.html#fixed-fields-11

`_fix_openapi_spec` has no rule for either case,

main branch

```yaml
# /tmp/s.yaml
id: https://example.org/s
name: s
default_range: string
classes:
  Foo:
    attributes:
      name:
        examples:
          - value: sample
```
```yaml
# /tmp/t.yaml
openapi: 3.0.3
info: {title: t, version: '1.0.0'}
paths:
  /foo:
    get:
      responses:
        '200': {description: ok, content: {application/json: {schema: {$ref: '#/components/schemas/Foo'}}}}
components:
  schemas:
    Foo: {type: object, x-linkml-schema: https://example.org/s, x-linkml-source: Foo}
```
```
python -m linkml.generators.openapigen /tmp/s.yaml -t /tmp/t.yaml 

openapi_spec_validator.validation.exceptions.OpenAPIValidationError: {'type': 'object', 'additionalProperties':True, 'description': '', 'properties': {'name': {'examples': ['sample']}}} is not valid under any of the given schemas

Failed validating 'oneOf' in schema['properties']['components']['properties']['schemas']['patternProperties']['^[a-zA-Z0-9\\.\\-_]+$']:
    {'oneOf': [{'$ref': '#/definitions/Schema'},
               {'$ref': '#/definitions/Reference'}]}

On instance['components']['schemas']['Foo']:
    {'type': 'object',
     'additionalProperties': True,
     'description': '',
     'properties': {'name': {'examples': ['sample']}}}
```

noelmcloughlin:
## Summary

Fixes #3962

In `_fix_openapi_spec`, convert `examples` (list) -> `example`(first element) on the Schema Object. Its lossy becuase openapi 3.0.3 `example` field is singular, but what else can we do.

Also this [json-ld issue comment](https://github.com/json-ld/json-ld.org/issues/612#issuecomment-1431076708) mentions "_... following the spirit of the OAS recommendations to deprecate the use of examples within API specifications and add them as schema annotations ..._"  - which suggests examples in (open)API schema are promoted in general.

## How was this tested?

Added tests

## Areas of uncertainty

- To solve (?)  lossy conversion we could pre-merge linkml examples into a singular 'thing' for over engineering?
- The  `examples` field in Media Type / Parameter Objects is out of scope is my understanding.

## Checklist

## AI Assistance

### DIFF
--- a/packages/linkml/src/linkml/generators/openapigen.py
+++ b/packages/linkml/src/linkml/generators/openapigen.py
@@ -185,6 +185,8 @@ class OpenApiGenerator(Generator):
 
         - ``const`` becomes ``enum`` with a single value (OpenAPI 3.0 doesn't support ``const``)
         - ``type`` as a list (e.g. nullable ``["string", "null"]``) becomes ``anyOf``
+        - ``examples`` (a list) becomes ``example`` (its first element); OpenAPI 3.0 has
+          no plural ``examples`` keyword on the Schema Object, only singular ``example``
         - ``$ref`` paths are rewritten from ``#/$defs/`` to ``#/components/schemas/``
         """
         fixed_element = None
@@ -195,6 +197,13 @@ class OpenApiGenerator(Generator):
                     fixed_element["enum"] = [value]
                 elif key == "type" and isinstance(value, list):
                     fixed_element["anyOf"] = [{"type": item} for item in value if item != "null"]
+                elif key == "examples" and isinstance(value, list):
+                    if value:
+                        # lossy by necessity: OpenAPI 3.0 allows only one example.
+                        # assigned rather than recursed into, since an example is data,
+                        # not schema -- recursing could rewrite a `const`/`type` key
+                        # that happens to appear inside the example value itself
+                        fixed_element["example"] = value[0]
                 else:
                     if isinstance(value, dict | list):
                         value = self._fix_openapi_spec(value)
--- a/tests/linkml/test_generators/input/openapi/schema_examples.yaml
+++ b/tests/linkml/test_generators/input/openapi/schema_examples.yaml
@@ -1,0 +1,16 @@
+# LinkML schema with a slot declaring multiple examples
+id: https://w3id.org/linkml/tests/examples
+name: examples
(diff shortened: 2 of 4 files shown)


Silvanoc (approved):
In general it looks good to me, at least as a first approach. Therefore I'm approving it.

I'm only wondering (after reading that "... following the spirit of the OAS recommendations to deprecate the use of examples within API specifications and add them as schema annotations ..." ) if it's the best way to add it. Probably we should investigate the "annotations" way in the future.

Silvanoc:
In the spirit of the [self-contained-area experiment](https://github.com/linkml/linkml/blob/ae9cfb742e46203fbb1a47091a6582aee95fbf67/.github/CODEOWNERS#L42) I'm merging. @matentzn that's the idea, right?

noelmcloughlin:
Yes, this is fine as 1st pass.

> add them as schema annotations

It was not obvious to me which schema the remark referred to (JSON or JSON-LD or some API schema). 

Is "annotation" part of OA or OAD template vocabulary?

PR #3982, opened 2026-09-10
fix(jsonschema): emit CURIE $ref targets under --use-curies

### LINKED ISSUE #3981: openapi: generation with `--use-curies` creates invalid `$ref`s
**Describe the bug**

When generating JSON Schema with `gen-json-schema --use-curies`, the `$defs`
entries for classes are keyed by their CURIE (e.g. `ex:Item`), but the internal
`$ref` targets that point at those classes are still emitted with the
camelCased element name (e.g. `#/$defs/Item`). The `$ref` targets therefore do
not match any key in `$defs`, producing a schema with dangling references that
fails validation/`$ref` resolution.

Every inlined class-range slot is affected (single-valued, multivalued inlined
lists, and inlined-as-dict `additionalProperties`), because all of them build
their references through `JsonSchema.ref_for()` / `add_lax_def()`, which
camelCase the class name regardless of `--use-curies`. Enum ranges are not
affected (enum `$defs` are intentionally not CURIE-keyed).

The `--not-use-curies` (default) path is unaffected.

Current `main` commit: c10d16a3d6dbf0c89819093fed2465b267b6ce38

Minimal schema (`repro.yaml`):

```yaml
id: https://example.org/test
name: test
prefixes:
  ex: https://example.org/test/
default_prefix: ex
default_range: string
imports:
  - linkml:types

classes:
  Container:
    class_uri: ex:Container
    tree_root: true
    attributes:
      has_item:
        slot_uri: ex:has_item
        range: Item
        inlined: true
        multivalued: true

  Item:
    class_uri: ex:Item
    attributes:
      label:
        slot_uri: ex:label
```

Command:

```console
$ linkml generate json-schema --use-curies repro.yaml
```

**Actual behavior**

`$defs` is keyed by CURIE, but the `$ref` points at the camelCased name:

```jsonc
{
  "$defs": {
    "ex:Container": { /* ... */ },
    "ex:Item": { /* ... */ }
  },
  // inside ex:Container.properties.has_item:
  "items": { "$ref": "#/$defs/Item" }   // <-- dangling: no "Item" key in $defs
}
```

Extracting the keys makes the mismatch obvious:

```console
$ gen-json-schema --use-curies repro.yaml | jq -r '.["$defs"] | keys[]'
ex:Container
ex:Item

$ gen-json-schema --use-curies repro.yaml | grep -oE '#/\$defs/[A-Za-z:]+'
#/$defs/Item
#/$defs/Item
```

**Expected behavior**

Under `--use-curies`, `$ref` targets for class ranges should use the same CURIE
keys that `$defs` uses:

```console
$ gen-json-schema --use-curies repro.yaml | grep -oE '#/\$defs/[A-Za-z:]+'
#/$defs/ex:Item
#/$defs/ex:Item
```

**Root cause**

`$defs` for classes are stored under the CURIE via
`JsonSchema.add_def(self._curie(cls), ..., is_curie=True)` (in `handle_class`),
but the reference side is name-based:

- `get_type_info_for_slot_subschema()` sets `reference = slot.range`
  (the element *name*), not a CURIE.
- `JsonSchema.ref_for()` and `JsonSchema.add_lax_def()` then apply
  `camelcase(name)` (unless `PRESERVE_NAMES`), yielding `#/$defs/Item`.

There is no code path that converts the reference to a CURIE when
`use_curies` is enabled, so the two sides diverge.

**Suggested fix**

Convert class-range references to CURIEs when `use_curies` is set, and make the
name-transforming helpers (`ref_for`, `add_lax_def`) leave already-CURIE names
untouched (analogous to the existing `PRESERVE_NAMES` handling). A `USE_CURIES`
class variable on `JsonSchema` mirrors the existing `PRESERVE_NAMES` mechanism.

Silvanoc:
## Summary

Fixes #3981

Under `--use-curies`, class `$defs` are keyed by CURIE (e.g. `ex:Item`), but
`$ref` targets were still emitted with the camelCased element name (e.g.
`#/$defs/Item`), producing dangling references that fail `$ref` resolution.
Every inlined class-range slot was affected (single-valued, multivalued
inlined lists, and inlined-as-dict `additionalProperties`).

This PR:

- Converts class-range references to CURIEs in
  `get_type_info_for_slot_subschema()` when `use_curies` is set.
- Adds a `USE_CURIES` class variable on `JsonSchema` (mirroring the existing
  `PRESERVE_NAMES` mechanism) so `ref_for()` and `add_lax_def()` leave
  already-CURIE names untouched instead of camelcasing them.

The default (`--not-use-curies`) path is unchanged.

As usual on my bug-fixing PRs:

1. Commit that can be used to reproduce the bug and avoid future regressions.
2. Commit that fixes the bug.

## How was this tested?

- Added a regression test (`test_use_curies_refs_resolve_to_defs`) using the
  minimal schema from #3981. It walks every `$ref` in the generated schema and
  asserts each `#/$defs/` target resolves to an existing `$defs` key. The test
  fails on `main` (dangling `#/$defs/Item`) and passes with this fix.
- Existing `test_use_uris` still passes.

## Areas of uncertainty

- Enum ranges are intentionally not CURIE-keyed; this PR leaves them untouched.
  Worth a reviewer sanity-check that no other reference path (descendants
  lists, identifier-slot ranges) needs the same conversion.

## Checklist

## AI Assistance

### DIFF
--- a/packages/linkml/src/linkml/generators/jsonschemagen.py
+++ b/packages/linkml/src/linkml/generators/jsonschemagen.py
@@ -218,6 +218,7 @@ def _slot_examples_for_json_schema(
 class JsonSchema(dict):
     OPTIONAL_IDENTIFIER_SUFFIX = "__identifier_optional"
     PRESERVE_NAMES: bool = False
+    USE_CURIES: bool = False
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -247,7 +248,7 @@ class JsonSchema(dict):
             names = [names]
 
         for name in names:
-            canonical_name = name if self.PRESERVE_NAMES else camelcase(name)
+            canonical_name = name if self.PRESERVE_NAMES or self.USE_CURIES else camelcase(name)
 
             if "$defs" not in self or canonical_name not in self["$defs"]:
                 self._lax_forward_refs[canonical_name] = identifier_name
@@ -319,7 +320,7 @@ class JsonSchema(dict):
     @classmethod
     def ref_for(cls, class_name: str | list[str], identifier_optional: bool = False, required: bool = True):
         def _ref(class_name):
-            def_name = class_name if cls.PRESERVE_NAMES else camelcase(class_name)
+            def_name = class_name if cls.PRESERVE_NAMES or cls.USE_CURIES else camelcase(class_name)
             def_suffix = cls.OPTIONAL_IDENTIFIER_SUFFIX if identifier_optional else ""
             return JsonSchema({"$ref": f"#/$defs/{def_name}{def_suffix}"})
 
@@ -476,6 +477,9 @@ class JsonSchemaGenerator(Generator, LifecycleMixin):
 
         # Set the class variable for JsonSchema to use
         JsonSchema.PRESERVE_NAMES = self.preserve_names
+        # Under --use-curies, class $defs are keyed by CURIE and $ref targets are
+        # pre-converted to CURIEs, so they must not be camelcased again.
+        JsonSchema.USE_CURIES = self.use_curies
 
         if self.top_class:
             if self.schemaview.get_class(self.top_class) is None:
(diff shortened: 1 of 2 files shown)


Silvanoc:
@matentzn I've pulled this PR into the milestone [v1.12](https://github.com/linkml/linkml/milestone/15) because otherwise we will be releasing a broken feature.

kevinschaper (approved):
Looks good!

PR #3985, opened 2026-09-11
feat(rdf): add opt-in diff-stable blank-node labels

jdsika:
# Add opt-in diff-stable blank-node labels

## Background

This is the follow-up to linkml/linkml#3295, which was closed with a clear
instruction rather than a rejection:

> Regardless of formal correctness of approach, the implementation should live
> elsewhere, where others can take advantage of it! Have you considered either
> rdflib or some kind of independent rdflib sidecar library?
> — @cmungall, linkml/linkml#3295

> For part 1, I'd like to encourage you to identify another RDF library where
> this can live [...] this seems like content we should consume from an
> upstream not manage ourselves.
> — @amc-corey-cox, linkml/linkml#3295

> This should be implemented somewhere upstream of linkml, after which we'll
> happily consider a PR to integrate it.
> — @cmungall, closing linkml/linkml#3295

So the algorithm was extracted, published and hardened outside linkml as
[`diffable-rdf`](https://pypi.org/project/diffable-rdf/), and this PR is the
thin integration that was invited. linkml gains a flag and a dependency; it
does not gain an algorithm to maintain.

It is also shaped by the constraint attached to linkml/linkml#3407:

> I would be amenable to a minimal diff serialization form as an opt-in feature
> clearly labeled as such, as a separate change, if it is lossless and etc.
> — @sneakers-the-rat, linkml/linkml#3407

Opt-in, separate, and lossless is exactly what this is, and the losslessness
claim is asserted by a test rather than argued.

## Problem

RDFC-1.0 canonicalization (merged in linkml/linkml#3407) makes RDF output
**deterministic**: two isomorphic graphs always serialize to identical bytes.
That solves reproducibility, but it does not make output **diffable**.

RDFC-1.0 assigns blank-node labels `c14n0, c14n1, …` from a single *global*
ordering over the whole graph. So a blank node's name depends on every other
blank node in the file. Adding one class near the top of a schema shifts that
ordering and renumbers every blank node after it.

The practical effect: a one-line semantic change produces a whole-file diff.
linkml/linkml#3481 demonstrates it on a real schema — one slot added, roughly
1,100 changed lines. For generated OWL and SHACL that are checked into version
control, this makes review effectively impossible, because you cannot see what
actually changed when everything changed. linkml/linkml#3702 is the open issue
for this specific property.

## What this changes

A new `diff_stable` argument on `canonicalize_rdf_graph()`, and a matching
`--diff-stable / --no-diff-stable` flag on the four generators that emit RDF
(`owlgen`, `rdfgen`, `shaclgen`, `shexgen`).

When enabled, blank-node labels are derived from each node's **own
neighbourhood** using Weisfeiler-Lehman refinement instead of from a global
ordering. A blank node's label then depends only on the subgraph around it, so
editing one part of a schema leaves the labels in unrelated parts untouched.

This runs *after* RDFC-1.0, not instead of it. Determinism is unchanged and
still comes from RDFC-1.0; this only changes which label each blank node ends
up with.

## Effect

Adding a single class to a real schema, counting changed lines in the output:

| generator  | default | `--diff-stable` | reduction |
| ---------- | ------: | --------------: | --------: |
| `owlgen`   |    2091 |              17 |     123× |
| `shexgen`  |     796 |              50 |      16× |
| `shaclgen` |     291 |              13 |      22× |
| `rdfgen`   |     115 |              25 |     4.6× |

In every case the output was verified to be **isomorphic** to the default
output and **byte-identical across repeated runs**.

## Design notes

**Off by default.** Enabling it relabels every blank node in existing output,
which is a one-time churn event for anyone with generated files in git. That
should be an explicit choice, not something a version bump does to you.

**Determinism is not traded away.** Diff stability is the weaker property;
determinism is the stronger one. The output is still fully deterministic with
the flag on — there is a test asserting exactly this.

**It refuses rather than pretending.** `canonicalize_rdf_graph()` already falls
back to rdflib when pyoxigraph cannot parse a graph, and on that path there are
no canonical quads to relabel, so `diff_stable` cannot be honoured. It now says
so with a warning instead of returning unchanged output. This is reachable
today: `gen-shacl --include-annotations` emits a literal in predicate position
for any annotation tag without a `:`, which forces the fallback.

**No new transitive dependencies.** The refinement lives in `diffable-rdf`, a
small pure-Python library. Its only two dependencies — `rdflib` and
`pyoxigraph` — are already `linkml-runtime` dependencies at *higher* version
floors, so the dependency tree does not grow.

## Testing

Six new tests in `tests/linkml_runtime/test_utils/test_rdf_canonicalize.py`:

- `test_diff_stable_is_opt_in` — the default output is byte-identical to
  before, so nothing changes for existing users.
- `test_diff_stable_preserves_semantics` — relabelled output is isomorphic to
  the default output.
- `test_diff_stable_is_deterministic` — repeated runs produce one distinct
  result.
- `test_diff_stable_confines_an_insertion_to_the_lines_it_touches` — the
  behavioural claim: inserting a subject must churn far fewer lines than the
  baseline. Asserts on the *ratio*, not an absolute count, so it is not brittle
  across rdflib versions.
- `test_diff_stable_reaches_every_rdf_generator[owlgen|rdfgen|shaclgen|shexgen]`
  — parametrized over all four generators. Asserts the *observable consequence*
  (no `c14nN` labels survive in the output), so it fails if a generator accepts
  the flag but forgets to thread it into `canonicalize_rdf_graph()`.
- `test_diff_stable_warns_instead_of_silently_no_opping_on_the_fallback` — the
  case above, where the request cannot be honoured and must be reported.

Each new assertion was validated by **reverting the change and confirming the
test fails** — including unwiring each generator individually, which is what
caught an earlier version of the generator test passing vacuously. The
fallback test came out of the same exercise: the generator test could not have
caught that bug, because the fallback emits `_:cb0` labels rather than the
`c14nN` labels it looks for.

Full suite: green on CI across Ubuntu and Windows for Python 3.10–3.14,
including `Validate Dependencies`.

Rebased onto current `main`, which resolved a `pyproject.toml` conflict with
the dependabot floor bumps for `pyoxigraph` and `pydantic` — main's higher
floors are kept.

## Relation to the open PRs

linkml/linkml#3704 and linkml/linkml#3754 (@amc-corey-cox) target the same
property with an in-tree implementation. I am not proposing this instead of
that work, and the two are not mutually exclusive — #3754 covers dumpers, which
this PR explicitly does not.

The reason to prefer an external implementation for the algorithm itself is the
one @cmungall and @amc-corey-cox gave on #3295: it is general-purpose RDF
tooling and other projects want it too. It also happens to answer the three
review objections @matentzn raised on #3704 — recursion depth at large N, the
24-hex-prefix collision grouping, and reliance on `sorted` stability — because
WL refinement is iterative rather than recursive, compares full labels, and
does not depend on sort stability for correctness. Those are properties of the
algorithm, not of who maintains it, so whichever route is chosen they should be
covered by tests either way.

Happy to fold this into either PR, or to close it, if that is the shorter path.

## Notes for reviewers

**`uv.lock`.** The lock entry for `diffable-rdf` is included. The repo sets
`exclude-newer = "7 days"` in `[tool.uv]`, so a *full* re-resolution will not
pick up a release younger than that; `uv lock --check` and `uv sync --frozen`
both pass against the committed lock, which is what CI runs — the
`Validate Dependencies` job is green on this PR.

**A separate finding.** `rdf_canonicalize.py` and the extracted library are the
same code, and while integrating I compared them property by property. Nine
correctness gaps in linkml's copy showed up, two of which are silent data
corruption — a `@base` ending in `#` rewrites every IRI in a way that still
parses, and a shared `rdf:List` tail is written twice, so nine triples in
become eleven out. One gap runs the other way: the library drops `@base` on its
fallback path and linkml does not.

That is out of scope here and is not folded into this PR. It is written up as
linkml/linkml#3986 (mirrored on the fork as ASCS-eV/linkml#25), which adds a
conformance test file asserting each property against both implementations with
the failing side marked `xfail(strict=True)`. It changes no behaviour. It is
also why the library is not yet used for the default path: it needs `base_iri`
support first.

## Deliberately out of scope

- **Dumpers** (`linkml_runtime.dumpers.rdflib_dumper`) — same argument applies,
  but instance data is a separate surface with separate compatibility concerns.
  linkml/linkml#3754 already covers this.
- **`shexgen` default format** — `--diff-stable` only has an effect when
  `--format rdf` is used; the default ShExC output is text, not RDF, so there
  are no blank-node labels to stabilize.
- **Making it the default** — see "Off by default" above.

## Dependency

Pins `diffable-rdf>=0.4.0`. 0.4.0 fixes the last correctness property that
linkml held and the library did not — `@base` surviving the rdflib fallback —
which is what unblocks linkml/linkml#3987 from deleting the in-tree
canonicalizer entirely. See ASCS-eV/diffable-rdf#60 and the
[v0.4.0 release](https://github.com/ASCS-eV/diffable-rdf/releases/tag/v0.4.0).

The library is Apache-2.0, has no dependencies beyond `rdflib` and
`pyoxigraph` (both already linkml dependencies), and is published to PyPI with
a Trusted Publisher via OIDC.

## Related

Fulfils: linkml/linkml#3295 · Addresses: linkml/linkml#3702, linkml/linkml#3212,
linkml/linkml#696 · Demonstrated by: linkml/linkml#3481 · Tracking:
linkml/linkml#3721
Builds on: linkml/linkml#3407, linkml/linkml#3703, linkml/linkml#3696,
linkml/linkml#3518, linkml/linkml#3524

### DIFF
--- a/packages/linkml/src/linkml/generators/owlgen.py
+++ b/packages/linkml/src/linkml/generators/owlgen.py
@@ -122,6 +122,22 @@ class OwlSchemaGenerator(Generator):
     """Suffix to add to the schema name to create the ontology URI, e.g. .owl.ttl"""
 
     # ObjectVars
+    diff_stable: bool = False
+    """Label blank nodes so that unrelated edits leave them untouched.
+
+    Output is already deterministic: RDFC-1.0 guarantees that isomorphic
+    graphs serialize identically. It does not guarantee that *similar*
+    graphs serialize *similarly* — blank nodes are numbered ``c14nN`` in a
+    global order, so adding one class can renumber every blank node after
+    it and rewrite most of the file.
+
+    When ``True``, blank-node labels are instead derived from each node's
+    own neighbourhood via Weisfeiler-Lehman refinement, so an edit relabels
+    only the blank nodes it actually touches. The output stays
+    deterministic and isomorphic either way; only the choice of label
+    changes. Off by default because enabling it relabels existing output.
+    """
+
     metadata_profile: MetadataProfile | None = None
     """Deprecated - use metadata_profiles."""
 
@@ -353,7 +369,7 @@ class OwlSchemaGenerator(Generator):
         """
         self.as_graph()
         fmt = "turtle" if self.format in ["owl", "ttl"] else self.format
-        return canonicalize_rdf_graph(self.graph, output_format=fmt)
+        return canonicalize_rdf_graph(self.graph, output_format=fmt, diff_stable=self.diff_stable)
 
     def add_metadata(self, e: Definition | PermissibleValue, uri: URIRef) -> None:
         """
@@ -1842,6 +1858,16 @@ class OwlSchemaGenerator(Generator):
         "(e.g. en, de, zh-Hans).  When set, rdfs:label, rdfs:comment, "
         "skos:definition and other text annotations are emitted with the "
         "specified language tag.  Element-level in_language overrides this."
+    ),
+)
+@click.option(
+    "--diff-stable/--no-diff-stable",
+    default=False,
+    show_default=True,
+    help=(
+        "Derive blank-node labels from each node's own neighbourhood so that "
+        "unrelated edits leave them unchanged. Output is deterministic either "
+        "way; this makes successive versions of a file diff cleanly."
     ),
 )
 @click.version_option(__version__, "-V", "--version")
--- a/packages/linkml/src/linkml/generators/rdfgen.py
+++ b/packages/linkml/src/linkml/generators/rdfgen.py
@@ -78,6 +78,22 @@ class RDFGenerator(Generator):
     uses_schemaloader = True
 
     # ObjectVars
+    diff_stable: bool = False
+    """Label blank nodes so that unrelated edits leave them untouched.
+
+    Output is already deterministic: RDFC-1.0 guarantees that isomorphic
+    graphs serialize identically. It does not guarantee that *similar*
+    graphs serialize *similarly* — blank nodes are numbered ``c14nN`` in a
+    global order, so adding one class can renumber every blank node after
+    it and rewrite most of the file.
+
+    When ``True``, blank-node labels are instead derived from each node's
+    own neighbourhood via Weisfeiler-Lehman refinement, so an edit relabels
+    only the blank nodes it actually touches. The output stays
+    deterministic and isomorphic either way; only the choice of label
+    changes. Off by default because enabling it relabels existing output.
+    """
+
     emit_metadata: bool = False
     context: list[str] = None
     original_schema: SchemaDefinition = None
@@ -89,7 +105,7 @@ class RDFGenerator(Generator):
 
     def _data(self, g: Graph) -> str:
         fmt = "turtle" if self.format == "ttl" else self.format
-        return canonicalize_rdf_graph(g, output_format=fmt)
+        return canonicalize_rdf_graph(g, output_format=fmt, diff_stable=self.diff_stable)
 
     def end_schema(self, output: str | None = None, context: str = None, **_) -> str:
         gen = JSONLDGenerator(
@@ -136,6 +152,16 @@ class RDFGenerator(Generator):
     show_default=True,
     multiple=True,
     help="JSONLD context file (default: vendored meta.context.jsonld)",
+)
+@click.option(
+    "--diff-stable/--no-diff-stable",
+    default=False,
+    show_default=True,
+    help=(
+        "Derive blank-node labels from each node's own neighbourhood so that "
+        "unrelated edits leave them unchanged. Output is deterministic either "
+        "way; this makes successive versions of a file diff cleanly."
+    ),
 )
 @click.version_option(__version__, "-V", "--version")
 def cli(yamlfile, **kwargs):
--- a/packages/linkml_runtime/pyproject.toml
+++ b/packages/linkml_runtime/pyproject.toml
@@ -48,6 +48,7 @@ dependencies = [
     "prefixmaps >=0.1.4",
     "curies>=0.14.6",
     "pyoxigraph>=0.5.11",
+    "diffable-rdf>=0.2.0",
     "pydantic>=2.13.5,<3.0.0",
     "isodate >=0.7.2, <1.0.0; python_version < '3.11'",
 ]
--- a/packages/linkml_runtime/pyproject.toml
+++ b/packages/linkml_runtime/pyproject.toml
@@ -48,7 +48,7 @@ dependencies = [
     "prefixmaps >=0.1.4",
     "curies>=0.14.6",
     "pyoxigraph>=0.5.11",
-    "diffable-rdf>=0.3.0",
+    "diffable-rdf>=0.4.0",
     "pydantic>=2.13.5,<3.0.0",
     "isodate >=0.7.2, <1.0.0; python_version < '3.11'",
 ]
--- a/packages/linkml_runtime/pyproject.toml
+++ b/packages/linkml_runtime/pyproject.toml
@@ -48,7 +48,7 @@ dependencies = [
     "prefixmaps >=0.1.4",
     "curies>=0.14.6",
     "pyoxigraph>=0.5.11",
-    "diffable-rdf>=0.2.0",
+    "diffable-rdf>=0.3.0",
     "pydantic>=2.13.5,<3.0.0",
     "isodate >=0.7.2, <1.0.0; python_version < '3.11'",
 ]
--- a/packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
+++ b/packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
@@ -42,6 +42,7 @@ import warnings
 
 import pyoxigraph as ox
 import rdflib
+from diffable_rdf import wl_relabel_quads
 from rdflib.compare import to_canonical_graph
 
 
@@ -287,6 +288,7 @@ def _is_safe_prefix_iri(iri: str) -> bool:
 def canonicalize_rdf_graph(
     graph: rdflib.Graph,
     output_format: str = "turtle",
+    diff_stable: bool = False,
 ) -> str:
     """Serialize an rdflib Graph deterministically using RDFC-1.0 canonicalization.
 
@@ -300,6 +302,12 @@ def canonicalize_rdf_graph(
 
     :param graph: The rdflib Graph to serialize.
     :param output_format: Target serialization format (e.g. ``"turtle"``, ``"nt"``).
+    :param diff_stable: Derive blank-node labels from each node's own
+        neighbourhood instead of RDFC-1.0's global ``c14nN`` numbering, so that
+        editing one part of a schema does not renumber unrelated blank nodes.
+        Output is deterministic and isomorphic either way; only the choice of
+        label changes. Off by default because enabling it relabels existing
+        output.
     :return: Deterministic string serialization of the graph.
     """
     ox_format = _FORMAT_MAP.get(output_format.lower())
@@ -339,13 +347,26 @@ def canonicalize_rdf_graph(
     # 3. Canonicalize blank node labels with RDFC-1.0.
     dataset.canonicalize(ox.CanonicalizationAlgorithm.RDFC_1_0)
 
+    quads = list(dataset)
+
+    # 3b. Optionally re-label blank nodes with locality-sensitive hashes.
+    # RDFC-1.0 guarantees that identical graphs get identical labels, but it
+    # does not guarantee that *similar* graphs get similar labels: the labels
+    # are assigned by a global ordering, so inserting one blank node can
+    # renumber every label after it and turn a one-line semantic change into a
+    # whole-file diff. Weisfeiler-Lehman labels are derived only from each
+    # node's local neighbourhood, so unrelated regions keep their labels.
+    # Output stays deterministic and isomorphic either way; this only changes
+    # which label each blank node receives.
+    if diff_stable:
+        quads = wl_relabel_quads(quads)
+
     # 4. Sort triples for deterministic ordering.
     # RDFC-1.0 stabilizes blank-node labels but pyoxigraph's Dataset
     # iteration order is not sorted and varies across processes (verified
     # empirically against pyoxigraph 0.5.8). The explicit string-key sort
     # is load-bearing for byte-identical output across runs; see
     # tests/linkml_runtime/test_utils/test_rdf_canonicalize.py::test_sort_is_load_bearing.
-    quads = list(dataset)
     sorted_triples = sorted(
         (ox.Triple(q.subject, q.predicate, q.object) for q in quads),
         key=lambda t: (str(t.subject), str(t.predicate), str(t.object)),
--- a/packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
+++ b/packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
@@ -307,7 +307,8 @@ def canonicalize_rdf_graph(
         editing one part of a schema does not renumber unrelated blank nodes.
         Output is deterministic and isomorphic either way; only the choice of
         label changes. Off by default because enabling it relabels existing
-        output.
+        output. Has no effect on the rdflib fallback path (non-standard RDF),
+        which warns rather than silently ignoring the request.
     :return: Deterministic string serialization of the graph.
     """
     ox_format = _FORMAT_MAP.get(output_format.lower())
@@ -338,6 +339,23 @@ def canonicalize_rdf_graph(
             RDFCanonicalizationWarning,
             stacklevel=2,
         )
+        if diff_stable:
+            # Weisfeiler-Lehman refinement consumes canonical pyoxigraph quads,
+            # and this path exists precisely because pyoxigraph refused the
+            # graph, so there are none to refine. The fallback is deterministic
+            # but not diff-stable: say so rather than returning output that
+            # silently ignores the argument. ``shaclgen --include-annotations``
+            # reaches this path, because an annotation tag without a ``:``
+            # becomes a literal predicate.
+            warnings.warn(
+                "diff_stable=True was requested but this graph took the rdflib fallback, "
+                "which cannot apply Weisfeiler-Lehman blank-node labels. Output is "
+                "deterministic but NOT diff-stable: an unrelated edit may still renumber "
+                "blank nodes. Make the offending terms standard RDF (absolute IRIs, IRI "
+                "predicates) to get diff-stable labels.",
+                RDFCanonicalizationWarning,
+                stacklevel=2,
+            )
         return _deterministic_fallback_serialize(graph, output_format)
 
     dataset = ox.Dataset()
--- a/tests/linkml_runtime/test_utils/test_rdf_canonicalize.py
+++ b/tests/linkml_runtime/test_utils/test_rdf_canonicalize.py
@@ -642,3 +642,28 @@ def test_diff_stable_reaches_every_rdf_generator(tmp_path, generator, kwargs):
         f"{generator.__name__} still emits RDFC-1.0 blank-node labels with diff_stable=True; "
         "the flag is probably not threaded into canonicalize_rdf_graph()"
     )
+
+
+def test_diff_stable_warns_instead_of_silently_no_opping_on_the_fallback():
+    """A request the fallback cannot honour must be reported, not ignored.
+
+    ``wl_relabel_quads`` consumes canonical pyoxigraph quads, and the rdflib
+    fallback exists precisely because pyoxigraph refused the graph. Returning
+    the same bytes for ``diff_stable=True`` and ``diff_stable=False`` without
+    saying so lets a caller believe the output is diff-stable when it is not.
+    """
+    graph = _make_graph_with_bnodes()
+    # A relative IRI is non-standard RDF, so pyoxigraph rejects the graph and
+    # canonicalize_rdf_graph degrades to rdflib -- the same path that
+    # ``shaclgen --include-annotations`` takes via its literal predicates.
+    graph.add((URIRef("testing"), URIRef("http://example.com/p"), Literal("v")))
+
+    with pytest.warns(RDFCanonicalizationWarning, match="NOT diff-stable"):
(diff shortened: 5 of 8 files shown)


sneakers-the-rat:
Great, will give full review next week. On a quick skim - i like this a lot better than something we take on directly, thanks for that! having it as an optional dependency (use is opt-in) would be great.

also, if this is a coupled package that is mostly for linkml, as a courtesy it would be nice to add some canary tests from linkml to run in that package's CI, e.g. i have a workflow in numpydantic that ensures i dont break anything linkml is known to depend on - https://github.com/p2p-ld/numpydantic/blob/main/.github/workflows/tests-linkml.yml 
an optional request, but something i think is a "sleep easier at night" thing for taking on a one-off dependency that doesnt have the same kind of stability guarantees as a matured/broadly used dep.

jdsika:
@sneakers-the-rat re: the canary-tests courtesy request — done.

- Pushed a small follow-up commit here (`test(rdf): mark diff-stable tests for diffable-rdf canary CI`) that registers a `diffable_rdf` pytest marker and applies it to the six `--diff-stable` tests in `test_rdf_canonicalize.py`, so an external job can select exactly those with `pytest -m diffable_rdf` without pulling in the rest of the suite.
- Opened [ASCS-eV/diffable-rdf#61](https://github.com/ASCS-eV/diffable-rdf/pull/61), which adds `.github/workflows/tests-linkml.yml` to the diffable-rdf repo, modeled directly on your numpydantic workflow: it checks out linkml, swaps the pinned `diffable-rdf` release for an editable install of the current diffable-rdf checkout, and runs the marked tests against it.

That workflow is already green on real CI across ubuntu/windows × Python 3.10/3.13, checked out against this PR's branch (since `linkml/linkml@main` doesn't have the integration yet). Once this PR merges, `ASCS-eV/diffable-rdf#61`'s `LINKML_REPO`/`LINKML_REF` just need to flip to `linkml/linkml`/`main`.

<del>Let me know if you'd like anything about either change adjusted before your full review, and whether it's okay to fold both into this PR / the next release.</del> I merged in my repo and I think it is unrelated to a pypi release as it is simply a workflow addon.

packages/linkml/src/linkml/generators/owlgen.py
```diff
@@ -122,6 +122,22 @@ class OwlSchemaGenerator(Generator):
...
+    graphs serialize *similarly* — blank nodes are numbered ``c14nN`` in a
+    global order, so adding one class can renumber every blank node after
+    it and rewrite most of the file.
+
+    When ``True``, blank-node labels are instead derived from each node's
+    own neighbourhood via Weisfeiler-Lehman refinement, so an edit relabels
+    only the blank nodes it actually touches. The output stays
+    deterministic and isomorphic either way; only the choice of label
+    changes. Off by default because enabling it relabels existing output.
+    """
```
sneakers-the-rat, line 139: thanks for this documentation. I think one of the main things on my "nice to have" list is a bit of narrative documentation in one of the pages re: "maintaining a schema" since this would otherwise get buried and only seen by those who are watching the issues/reading the api docs: thinking about Chris's "how to use RDF in version control" document/presentation, which is super useful contextual knowledge. we *sort of* have a page where this would fit here - https://linkml.io/linkml/howtos/collaborative-development.html - since keeping diffs tidy is part of a collaborative process. Just having a mention of "minimize diffs, for example by using the `diff_stable` option..." in one of the best practices tables there would be great to have.

packages/linkml/src/linkml/generators/rdfgen.py
```diff
@@ -78,6 +78,22 @@ class RDFGenerator(Generator):
     uses_schemaloader = True
 
     # ObjectVars
+    diff_stable: bool = False
```
sneakers-the-rat, line 81: i'm sort of amazed that there isn't a grouping class for all rdf-like generators for params like this (not a request, just remarking that it's a code smell that you have to duplicate the docstring like this due to existing structure of the code)

packages/linkml/src/linkml/generators/rdfgen.py
```diff
@@ -137,6 +153,16 @@ def end_schema(self, output: str | None = None, context: str = None, **_) -> str
...
     help="JSONLD context file (default: vendored meta.context.jsonld)",
 )
+@click.option(
+    "--diff-stable/--no-diff-stable",
+    default=False,
+    show_default=True,
+    help=(
+        "Derive blank-node labels from each node's own neighbourhood so that "
+        "unrelated edits leave them unchanged. Output is deterministic either "
+        "way; this makes successive versions of a file diff cleanly."
```
sneakers-the-rat, line 163: this would be a good place to have a dedicated docs url to be able to link "see more: {url}" to keep the cli docs short but have the path to learning more info clear.

packages/linkml_runtime/pyproject.toml
```diff
@@ -48,6 +48,7 @@ dependencies = [
     "prefixmaps >=0.1.4",
     "curies>=0.14.6",
     "pyoxigraph>=0.5.11",
+    "diffable-rdf>=0.4.0",
```
sneakers-the-rat, line 51: on reviewing again, i really think this makes sense as an optional dependency: this gives it a "true opt in" feel, and gives a point of documentation where we can describe the relationship to the package so that you also can feel free to extend it as you like. it's an opt-in feature for a specific subset of generators and we already have precedent for that being what to do with framework-specific deps. And is has exactly one import and call site, so the work of making it optional would be minimal.

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
```diff
@@ -330,6 +339,23 @@ def canonicalize_rdf_graph(
...
+            # becomes a literal predicate.
+            warnings.warn(
+                "diff_stable=True was requested but this graph took the rdflib fallback, "
+                "which cannot apply Weisfeiler-Lehman blank-node labels. Output is "
+                "deterministic but NOT diff-stable: an unrelated edit may still renumber "
+                "blank nodes. Make the offending terms standard RDF (absolute IRIs, IRI "
+                "predicates) to get diff-stable labels.",
+                RDFCanonicalizationWarning,
+                stacklevel=2,
+            )
```
sneakers-the-rat, line 358: i understand why this is here, but it's also why i didn't really like this fallback in the first place - i can imagine with two or three more things that touch this function we might end up with a pretty messy tree of code paths in a place that feels like it should be more of a pipeline. no change requested again, i am just sort of noting this early return block as a point of future fragility

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
```diff
@@ -330,6 +339,23 @@ def canonicalize_rdf_graph(
...
             stacklevel=2,
         )
+        if diff_stable:
+            # Weisfeiler-Lehman refinement consumes canonical pyoxigraph quads,
+            # and this path exists precisely because pyoxigraph refused the
+            # graph, so there are none to refine. The fallback is deterministic
+            # but not diff-stable: say so rather than returning output that
+            # silently ignores the argument. ``shaclgen --include-annotations``
+            # reaches this path, because an annotation tag without a ``:``
+            # becomes a literal predicate.
```
sneakers-the-rat, line 349:
```suggestion
            # diff_stable uses oxigraph and can't be applied
```

if we even need a comment here since the warning explains things. LLM comments are a practical hazard that makes the code less readable (and also forces future LLMs to consume tokens needlessly, if one cares about that kind of thing).

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
```diff
@@ -339,13 +365,26 @@ def canonicalize_rdf_graph(
...
+
+    # 3b. Optionally re-label blank nodes with locality-sensitive hashes.
+    # RDFC-1.0 guarantees that identical graphs get identical labels, but it
+    # does not guarantee that *similar* graphs get similar labels: the labels
+    # are assigned by a global ordering, so inserting one blank node can
+    # renumber every label after it and turn a one-line semantic change into a
+    # whole-file diff. Weisfeiler-Lehman labels are derived only from each
+    # node's local neighbourhood, so unrelated regions keep their labels.
+    # Output stays deterministic and isomorphic either way; this only changes
+    # which label each blank node receives.
```
sneakers-the-rat, line 378: similarly, as numbering comments doesn't generally age well and becomes counterinformative, a point 3b that doesn't require a comment - this belongs in the docs not the comments, a "see: {url} " if anything

tests/linkml_runtime/test_utils/test_rdf_canonicalize.py
```diff
@@ -507,3 +507,169 @@ def run(seed: str) -> str:
...
+    diff = difflib.unified_diff(before.splitlines(), after.splitlines(), n=0, lineterm="")
+    return sum(1 for line in diff if line[:1] in "+-" and not line.startswith(("+++", "---")))
+
+
+@pytest.mark.diffable_rdf
+def test_diff_stable_is_opt_in():
+    """The default must keep producing exactly the output it produced before."""
+    graph = _make_graph_with_bnodes()
+    assert canonicalize_rdf_graph(graph) == canonicalize_rdf_graph(graph, diff_stable=False)
+
```
sneakers-the-rat, line 537: tests the value of a default parameter - if we wanted to test this then we could do it more cheaply with `inspect.signature`

tests/linkml_runtime/test_utils/test_rdf_canonicalize.py
```diff
@@ -507,3 +507,169 @@ def run(seed: str) -> str:
...
+    """
+    before, after = _shapes_graph(20), _shapes_graph(20, extra=True)
+
+    baseline = _changed_line_count(canonicalize_rdf_graph(before), canonicalize_rdf_graph(after))
+    stable = _changed_line_count(
+        canonicalize_rdf_graph(before, diff_stable=True),
+        canonicalize_rdf_graph(after, diff_stable=True),
+    )
+
+    assert stable < baseline / 4, f"expected diff-stable output to churn far less; got {stable} vs baseline {baseline}"
```
sneakers-the-rat, line 578: a test that would be great to have for this would be a hand-constructed graph that has exactly the minimal representation of the behavior that this mode changes and test for exact matches of the two modes - "that one one creates fewer diff lines" is an important behavior of the code, but it would be nice to have a direct test of the thing that this algorithm is supposed to do differently to the serialization than rdfc.

tests/linkml_runtime/test_utils/test_rdf_canonicalize.py
```diff
@@ -507,3 +507,169 @@ def run(seed: str) -> str:
...
+    return [
+        pytest.param(OwlSchemaGenerator, {}, id="owlgen"),
+        # rdfgen and shexgen resolve JSON-LD contexts (linkml types, shex.jsonld);
+        # the `network` marker serves those from local stubs. See tests/conftest.py.
+        pytest.param(RDFGenerator, {}, id="rdfgen", marks=pytest.mark.network),
+        pytest.param(ShaclGenerator, {}, id="shaclgen"),
+        # ShExGenerator only emits RDF in this format; its default is ShExC text,
+        # where blank-node labelling does not apply.
+        pytest.param(ShExGenerator, {"format": "rdf"}, id="shexgen", marks=pytest.mark.network),
+    ]
```
sneakers-the-rat, line 621: strange closure. if we wanted this to be reusable it should be a parameterized fixture or a tuple. we don't gain anything from having the imports in a closure here. since this is only used once, just move to `pytest.mark.parameterize` and move the imports to the top

tests/linkml_runtime/test_utils/test_rdf_canonicalize.py
```diff
@@ -507,3 +507,169 @@ def run(seed: str) -> str:
...
+    plain = generator(str(schema), **kwargs).serialize()
+    stable = generator(str(schema), diff_stable=True, **kwargs).serialize()
+
+    # Guards the test itself: if the fixture stopped producing blank nodes the
+    # assertion below would hold vacuously.
+    assert re.search(r"c14n\d+", plain), f"{generator.__name__} output has no blank nodes to relabel"
+    assert not re.search(r"c14n\d+", stable), (
+        f"{generator.__name__} still emits RDFC-1.0 blank-node labels with diff_stable=True; "
+        "the flag is probably not threaded into canonicalize_rdf_graph()"
+    )
```
sneakers-the-rat, line 649: also strange - if we are using magic strings to test, we might as well test affirmatively what we expect to see, rather than what we *don't* expect to see and having to generate twice.

tests/linkml_runtime/test_utils/test_rdf_canonicalize.py
sneakers-the-rat: yeah after reading the rest of the tests, having the minimal case that gives us a positive target of "what the output should look like exactly" (as in, assertions against a designed schema, not a blanket string equality assertion) is the test i was looking for and did not find

packages/linkml_runtime/src/linkml_runtime/utils/rdf_canonicalize.py
sneakers-the-rat, line 1: i think that there should be a graph isomorphism test with a complex naturalistic schema, but that might work better in the downstream package tests than here - another test i was looking for and did not find was "is the transformation here correct across the fuckery that is possible in RDF" (which will surely be work-in-forever-progress, part of the reason to separate the package from linkml, and i certainly don't expect to be fully solved in this PR) but yeah i don't think the linkml tests are the right place for it

sneakers-the-rat (commented):
hell yeah. so i am pretty much good with this and really appreciate the work put in to splitting it off and coming back with the reduced PR. Since what we need here from a linkml-centric perspective is "some insurance or means of detection if upstream package accidentally goes off the rails" (which i'm sure it won't!) is some basic correctness tests, most of my comments and review focus on tests. I don't necessarily think we should "chuck the metamodel and biolink into it and see what comes out the other side" as on a macro scale that poses other problems, but some more elaborated and designed test cases would be welcome: an exact demonstration of the kind of schema that would suck to maintain with just RDFC with exact expectations about what should be different with this PR.

the other thing that i think would be great would be moving some of the comments and docstring chatter into the docs - this stems out of a real problem with maintaining schemas in version control, and is directly relevant to anyone who is doing that, so to the degree we can maintain a "how to do RDF with git and stay sane" survival guide we should do that (though i don't want to shoulder you with that responsibility lol). Just a few lines, a place that can be linked to from the docstrings, cli output, and in future issues when someone is like "how do i make my schema work good in a repo" - since you went to the trouble of carrying this through, feel free to carve out a space on the website where you get to soapbox about how good and cool it is <3

thanks again!

jdsika:
Thanks for the careful review — all of it addressed, and the naturalistic case
went where you said it belonged.

`diffable-rdf` is optional now: `linkml-runtime[diff-stable]` and
`linkml[diff-stable]`, imported lazily, with an install hint when the extra is
missing. Both comment blocks are gone.

On the tests: `test_diff_stable_is_opt_in` uses `inspect.signature`; the closure
is replaced by `parametrize` with the imports at the top; the generator test
asserts what it expects to see and generates once instead of twice. The minimal
case you were looking for is `test_diff_stable_insertion_labels` — a two-triple
hand-built graph plus one independent insertion, with exact quad sets asserted
in both modes. RDFC relabels the untouched node `c14n0` -> `c14n1`; diff-stable
leaves it alone and the only difference is the inserted pair.

Docs: the guidance moved into a new "Keep generated RDF diffs small" section
under Principles in the collaborative development guide, anchored as
`rdf-in-version-control`, which the docstrings and `--diff-stable` help now link
to. That page has no tables, so I put it next to "Version Control Your Schema"
rather than inventing one.

The complex naturalistic schema belongs upstream, as you said, so it went into
diffable-rdf instead of here (ASCS-eV/diffable-rdf#63): a complete OWL/SHACL document with shared list
tail cells, a blank-node cycle, OWL restrictions, SHACL shapes and mixed
literals, checked for isomorphism, idempotence, and blank-node-label and
insertion-order independence across every public rendering. Adding one
unrelated class rewrites zero existing lines; the same edit moves 40 lines
under RDFC-1.0 labels alone, and the test asserts that baseline too, so a
fixture that stopped demonstrating the problem can't pass quietly.

I'll leave the runtime release target to you. The existing `>=1.10.0` minimum
still admits a runtime without this feature, so that bound needs to follow
whichever release carries it.

.github/tests/test_diff_stable_install.py
sneakers-the-rat, line 1: I find this and the accompanying change to the testing CI action to be extremely strange. So these are tests for handling the optional dependency, that we can import linkml and that we get a proper warning if we don't have the optional dep installed. But its a very cumbersome way to do that, and is the only thing of its kind tested like this.

I also had written a test like this for the black optional dep in pydanticgen, which takes place within the testing framework rather than split between CI setup and tests, and the mechanism is already in place to make it generic across deps: https://github.com/linkml/linkml/blob/b4c3261fddcad14f0221ba980f6e84f9d88eb873/tests/conftest.py#L446

I just think we don't really need this, and if we do want to have explicit tests for optional dep handling, this is a lot of one-off complexity that can't really be re-used, so it should probably be made general so other optional deps can do something similar.

.github/tests/test_diff_stable_install.py
```diff
@@ -0,0 +1,147 @@
...
+"""
+
+
+def _assert_installed(package: str) -> None:
+    """Require imports from this environment's installed package directory."""
+    module = importlib.import_module(package)
+    assert Path(module.__file__).resolve().is_relative_to(Path(sysconfig.get_path("purelib")).resolve())
+
+
+def _assert_runtime_requirements() -> None:
```
sneakers-the-rat, line 35: This basically tests if `uv` correctly creates wheel metadata from pyproject.toml, if im reading it right. Not testing anything about linkml

sneakers-the-rat (commented):
Appreciate making the dep optional, and also thanks for adding the tests upstream. Last round of changes came with a set of tests that I think are a bit unnecessary, but I don't want to goaltend this PR so I will let someone else make final evaluations.

I am assuming from the preserved linebreaks and em-dashes that the above response to review was copy-pasted LLM output, and it seems to have interpreted some of my comments strangely (e.g. it interpreted my comment about `inspect.signature` testing the default values in a function signature as a request to actually do that, rather than a comment implying we don't need to test default values in a function signature, which is understandable just very literal.). So I suppose a note for the LLM then, if that's the case: since the minimum testing we need is in place, less is more for this PR.

PR #3986, opened 2026-09-11
test(rdf): record the canonicalizer gaps against the extracted library

jdsika:
# Record the RDF canonicalizer gaps against the extracted library

> **Stacked on linkml/linkml#3985** (mirrored on the fork as ASCS-eV/linkml#24).
> The first two commits belong to that PR; this one adds a single test file. It
> has to be stacked, because the test compares against `diffable-rdf` and that
> dependency arrives in #3985.

**This changes no behaviour.** It adds one test file and nothing else.

## Why

`linkml_runtime/utils/rdf_canonicalize.py` and `diffable_rdf`'s canonicalizer
are the same code. The module was extracted into a standalone library at the
maintainers' request (linkml/linkml#3295), and the two copies have since
diverged.

While integrating the library I compared them property by property. The
divergence turned out to be worth writing down, so this adds a conformance file
that asserts each property against **both** implementations and marks whichever
side fails it `xfail(strict=True)`. Strict, so the file is a ratchet in both
directions: when a fix lands on either side its case starts passing and the
suite fails until the mark is removed.

Nothing is asserted about the library that is not also asserted about linkml,
so the library is not taken on trust.

## What it finds

Nine gaps in linkml's copy. Two are silent data corruption, which is the part
worth a second look — the output parses cleanly and says something the input
never said, so nothing reports it:

- **A base IRI ending in `#` rewrites every term.** Relativizing
  `http://ex.org/d#a` against base `http://ex.org/d#` gives `<#a>`, correct per
  RFC 3986, but rdflib's parser resolves a fragment reference by concatenation
  and reads it back as `http://ex.org/d##a`.
- **A shared `rdf:List` tail is written twice.** rdflib's Turtle writer renders
  `( … )` collection syntax per list, so a tail referenced from two lists
  becomes two separate blank nodes. Nine triples in, eleven out.

The rest fail loudly or produce unusable output:

- N-Triples output is written for a graph pyoxigraph already refused, so it
  fails on line 1. The fallback is reached *because* a term is non-standard,
  and for N-Triples that term is always one N-Triples cannot write either.
- Fallback sorting uses `str.splitlines()`, which breaks on U+2028, U+2029,
  U+0085, U+000B, U+000C and U+001C–1E. N-Triples permits those raw inside a
  quoted literal, so one statement becomes two lines, the halves sort
  independently, and the document no longer parses.
- Degraded RDF/XML is ordered by rdflib's graph traversal, degraded Turtle with
  unbound namespaces gets `ns1`/`ns2` names allocated in traversal order, and
  `json-ld` is absent from the format map so it falls through to rdflib. All
  three vary across processes — tested with four `PYTHONHASHSEED` values.
- Trailing newlines differ by format: RDF/XML gives none, Turtle gives one.
- A `Dataset` is a `Graph` subclass, so it passes the type check and is
  flattened, dropping the graph names.

## One gap ran the other way — now fixed

> [!NOTE]
> ~~`diffable-rdf` drops `@base` on its own rdflib fallback, and linkml does
> not. That is a bug in the library, it is ours to fix, and it is the reason
> this PR does not simply replace linkml's implementation with the library's.~~
>
> **Fixed in diffable-rdf 0.4.0** (ASCS-eV/diffable-rdf#60), pinned by
> linkml/linkml#3985. The case is now unmarked and passes on both sides.

`diffable-rdf` dropped `@base` on its own rdflib fallback where linkml did not.
`rdflib_dumper.dumps(..., prefix_map={"@base": ...})` hits exactly that path,
because the metamodel's bare `status: testing` on a `uriorcurie` slot
serializes to a relative `<testing>` that forces the fallback.

Asserting it here from both directions is what got it fixed where it belonged.
The naive fix is wrong, which is why it is worth a sentence: rdflib's
`relativize()` is a string prefix strip rather than RFC 3986 §5.2.2 component
resolution, so carrying a base of `http://ex.org/vocab#` rewrites
`http://ex.org/vocab#Thing` *and* every unrelated IRI that merely starts with
those characters. Rejecting bases by shape does not work either —
`http://ex.org/a/b` and `http://ex.org/d?q=1` corrupt too. The library now
carries the base, re-parses, and keeps it only if every absolute IRI of the
source survives. RFC 3986 specifies resolution and never relativization, so
verification is the only sound test.

## Why not just delegate

> [!NOTE]
> ~~Delegation becomes viable once the library carries `base_iri`. It also
> wants a `diff_stable` parameter, so the flag in linkml/linkml#3985 stops
> running on the unfixed local pipeline. Both are on the library's side; I'll
> raise them there.~~
>
> Both landed in 0.4.0. **Delegation is now linkml/linkml#3987**, stacked on
> this PR: it deletes the in-tree copy, closes all nine gaps at once, and
> leaves all four RDF generators byte-identical.

I tried delegating first. Replacing the default path with
`diffable_rdf.canonicalize_rdf_graph` fixes all nine gaps and leaves all four
RDF generators byte-identical — but it regressed `test_loaders_no_namespace`,
because of the `@base` gap above. So this PR states the facts and leaves the
code alone, which keeps the nine gaps reviewable as evidence before the change
that fixes them.

If you would rather fix them in place, or accept them as known, the evidence is
the same either way — and the file keeps running against both entry points
afterwards, so a future divergence fails the suite instead of going unnoticed.

## Related

Follows from: linkml/linkml#3295 · Module added by: linkml/linkml#3407 ·
Hardened by: linkml/linkml#3524, linkml/linkml#3703 · Known fallback issue:
linkml/linkml#3803 · Nondeterminism leaks: linkml/linkml#3516

### DIFF
--- a/packages/linkml/src/linkml/generators/owlgen.py
+++ b/packages/linkml/src/linkml/generators/owlgen.py
@@ -122,6 +122,22 @@ class OwlSchemaGenerator(Generator):
     """Suffix to add to the schema name to create the ontology URI, e.g. .owl.ttl"""
 
     # ObjectVars
+    diff_stable: bool = False
+    """Label blank nodes so that unrelated edits leave them untouched.
+
+    Output is already deterministic: RDFC-1.0 guarantees that isomorphic
+    graphs serialize identically. It does not guarantee that *similar*
+    graphs serialize *similarly* — blank nodes are numbered ``c14nN`` in a
+    global order, so adding one class can renumber every blank node after
+    it and rewrite most of the file.
+
+    When ``True``, blank-node labels are instead derived from each node's
+    own neighbourhood via Weisfeiler-Lehman refinement, so an edit relabels
+    only the blank nodes it actually touches. The output stays
+    deterministic and isomorphic either way; only the choice of label
+    changes. Off by default because enabling it relabels existing output.
+    """
+
     metadata_profile: MetadataProfile | None = None
     """Deprecated - use metadata_profiles."""
 
@@ -353,7 +369,7 @@ class OwlSchemaGenerator(Generator):
         """
         self.as_graph()
         fmt = "turtle" if self.format in ["owl", "ttl"] else self.format
-        return canonicalize_rdf_graph(self.graph, output_format=fmt)
+        return canonicalize_rdf_graph(self.graph, output_format=fmt, diff_stable=self.diff_stable)
 
     def add_metadata(self, e: Definition | PermissibleValue, uri: URIRef) -> None:
         """
@@ -1842,6 +1858,16 @@ class OwlSchemaGenerator(Generator):
         "(e.g. en, de, zh-Hans).  When set, rdfs:label, rdfs:comment, "
         "skos:definition and other text annotations are emitted with the "
         "specified language tag.  Element-level in_language overrides this."
+    ),
+)
+@click.option(
+    "--diff-stable/--no-diff-stable",
+    default=False,
+    show_default=True,
+    help=(
+        "Derive blank-node labels from each node's own neighbourhood so that "
+        "unrelated edits leave them unchanged. Output is deterministic either "
+        "way; this makes successive versions of a file diff cleanly."
     ),
 )
 @click.version_option(__version__, "-V", "--version")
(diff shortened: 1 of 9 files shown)


jdsika:
Follow-up opened: linkml/linkml#3987 deletes the in-tree canonicalizer and calls the library. Every xfail in this PR's evidence file goes away with it.

PR #4000, opened 2026-09-17
fix(schemaview): get_uri resolves elements of relatively-imported schemas

### LINKED ISSUE #3878: gen-shacl: KeyError on imported schema when an attribute name is reused across classes with per-class slot_uri
## Summary

`gen-shacl` raises `KeyError: '<schema name>'` on any modular schema (imported via a directory-prefixed relative path, e.g. `../core/core`) where **two or more classes reuse the same attribute name, each with its own `slot_uri` override**, and at least one of those classes has a subclass (`is_a`) somewhere in the import graph.

`linkml lint`, `gen-owl`, `gen-pydantic`, and `gen-sqlddl` all handle the same schema fine — this is specific to `gen-shacl`'s `sh:closed`/`sh:ignoredProperties` computation.

linkml 1.11.1 (`pip install linkml`), Python 3.13, macOS.

## Minimal reproduction

Two files, `nucleo/core.yaml`:

```yaml
id: https://example.org/core
name: core
prefixes:
  core: https://example.org/core/
default_prefix: core
default_range: string
imports:
  - linkml:types

classes:
  Transaccion:
    attributes:
      id_transaccion:
        identifier: true
      estado:
        slot_uri: core:transaccion_estado
  Compromiso:
    attributes:
      id_compromiso:
        identifier: true
      estado:
        slot_uri: core:compromiso_estado
```

and `dominios/domain.yaml`:

```yaml
id: https://example.org/domain
name: domain
prefixes:
  core: https://example.org/core/
  dom: https://example.org/domain/
default_prefix: dom
default_range: string
imports:
  - linkml:types
  - ../nucleo/core

classes:
  Pedido:
    is_a: Transaccion
    attributes:
      importe:
        range: float
```

```
$ gen-shacl dominios/domain.yaml
KeyError: 'core'
```

(with `--stacktrace`, or via the Python API directly, the traceback bottoms out in `SchemaView.get_uri`):

```
File ".../linkml/generators/shaclgen.py", line 466, in _build_ignored_properties
    collect_child_properties(c.name, child_properties)
File ".../linkml/generators/shaclgen.py", line 459, in collect_child_properties
    URIRef(self.schemaview.get_uri(prop, expand=True))
File ".../linkml_runtime/utils/schemaview.py", line 1400, in get_uri
    schema = self.schema_map[self.in_schema(e.name)]
KeyError: 'core'
```

## Isolating the two required ingredients

I narrowed it down with a few variants of the same two files:

- **One class, `estado` with a `slot_uri` override, no second class reusing the name** → works fine.
- **Two classes, both declaring `estado`, neither with an explicit `slot_uri`** (falls back to the default-prefix minting path) → works fine.
- **Two classes, both declaring `estado`, each with its own `slot_uri`** (`core:transaccion_estado` / `core:compromiso_estado` as above) → `KeyError`.
- Importing `core.yaml` from the *same* directory as `domain.yaml` (`imports: [core]` instead of `imports: [../nucleo/core]`) → no crash, even with the two-classes-same-attribute-name setup above.

So it needs **both**: (a) an attribute name reused across ≥2 classes with distinct `slot_uri`s, and (b) the schema being imported via a relative path string that differs from the imported schema's own `name:` field (i.e. any subdirectory-prefixed import — a common layout for modular schemas: a `core`/`nucleo` directory imported from a `domains` directory).

## Likely root cause

`SchemaView` resolves "which schema does this bare attribute name belong to" through two different keyspaces that happen to coincide only when the import string equals the target schema's declared name:

- `schema_map` (consulted by `get_uri`'s fallback branch, `self.schema_map[self.in_schema(e.name)]`) is keyed by the **string used in `imports:`** — here, `"../nucleo/core"`.
- `element_by_schema_map()` (backing `in_schema()`) is keyed by the **imported schema's own `name:`** — here, `"core"`.

`shaclgen.py`'s `_build_ignored_properties`/`collect_child_properties` calls `schemaview.get_uri(prop, expand=True)` with a bare slot-name **string** (not a `SlotDefinition`, and not scoped to a class) while walking a class's subclasses to compute `sh:ignoredProperties`. For an attribute name that's unambiguous (declared on exactly one class, or with no `slot_uri` override so it falls through the default-prefix branch instead) this bare lookup happens to resolve fine either way. For an attribute name reused across multiple classes each with a distinct `slot_uri`, `get_element("estado")` apparently returns some merged/last-write-wins element whose `.from_schema` is unset, forcing the `self.schema_map[self.in_schema(e.name)]` fallback — and that's where the two keyspaces (import-string-keyed vs. schema-name-keyed) diverge and the lookup misses.

I haven't traced far enough into `get_element`/`class_slots` to say exactly why the per-class `slot_uri` override is what tips it into the ambiguous path rather than the plain literal minting path — happy to dig further if useful, but the repro above should be enough to chase it down directly.

## Impact

Any non-trivial modular schema with a shared upper module (a coordination/event/base-entity pattern that multiple domains import and extend via `is_a`, each domain reusing an attribute name like `estado`/`status` with its own per-class URI) is one such collision away from `gen-shacl` becoming unusable, while every other generator (`gen-owl`, `gen-pydantic`, `gen-sqlddl`, `linkml lint`) keeps working on the exact same schema. Happy to test a fix against the real (larger) schema this originated from if one lands.

noelmcloughlin:
## Summary

Fixes #3878 and unblocks https://github.com/IBM/ai-atlas-nexus/issues/257

- gen-shacl KeyError on `../dir/module` imports with reused attribute + slot_uri (#3878 scenario)
- gen-shacl KeyError on `./module` imports with reused attribute, no slot_uri
- gen-shacl on the `ai-atlas-nexus` ontology: `KeyError: 'ai_aiuc'`
- SchemaView.get_uri() for any element with no from_schema in a relatively-imported schema
- Bare KeyError -> a ValueError on a genuine miss

## Checklist

## AI Assistance


packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
```diff
@@ -1439,7 +1439,17 @@ def get_uri(
                     msg = f"Cannot find {e.from_schema} in schema_map"
                     raise ValueError(msg)
             else:
-                schema = self.schema_map[self.in_schema(e.name)]
+                # schema_map is keyed by the import as written in the importing schema, which
+                # is not always the imported schema's name: `imports: [./child]` is keyed
+                # './child' while in_schema() reports 'child'. Fall back to matching on the
+                # name, as the from_schema branch above does on the id.
+                schema_name = self.in_schema(e.name)
+                schema = self.schema_map.get(schema_name)
```
Silvanoc: Is there a reason for breaking the pre-existing one-liner into these two lines? They do exactly the same, right?

IMO following code shows clearly, that the change is about:
1. Adding a fallback in case of mismatch between the import and the schema name.
2. Error reporting

But the most common case (import == schema name) remains untouched.

Perhaps the comment would need a small reformulation.

```suggestion
                schema = self.schema_map[self.in_schema(e.name)]
                # schema_map is keyed by the import as written in the importing schema, which
                # is not always the imported schema's name: `imports: [./child]` is keyed
                # './child' while in_schema() reports 'child'. Fall back to matching on the
                # name, as the from_schema branch above does on the id.
```

Silvanoc (approved):
Only comment that I have is probably a nitpick, therefore approving the PR.

packages/linkml_runtime/src/linkml_runtime/utils/schemaview.py
noelmcloughlin: My bad - the comment was not clear enough, and fooled me too - `get()` was key but hard to spot.

`    test_shacl_modular_schema_with_reused_attribute_name - KeyError: 'core'`

Test failed because`schema_map`'s key **('../nucleo/core')** is different schema name **('core')** - so I made the comment is more clear now I think
```
-             schema = self.schema_map[self.in_schema(e.name)]
+             # Two classes reusing an attribute name give a bare placeholder slot with no
+             # from_schema, so the schema has to be looked up by name instead.
+             schema_name = self.in_schema(e.name)  # the schema's own name, e.g. 'core'
+             # .get(), not [...]: keys are imports as written, so a relative import misses
+             # here -- a miss is expected and must fall through to the name match below.
+             schema = self.schema_map.get(schema_name)
```

**Maybe** that type of test is also needed in kitchen sink so other generators flex this use case?

noelmcloughlin: The comment I added to schemaview was clearer.  

Thanks for the review @Silvanoc, appreciated.

kevinschaper (approved):
looks good! thanks @noelmcloughlin, and thanks for the review @Silvanoc!

PR #4004, opened 2026-09-18
build: exclude vendored linkml-model files from reformatting hooks

amc-corey-cox:
The files under `linkml_runtime/linkml_model/` are byte-for-byte copies from linkml-model, but this repo's tooling rewrites them: `ruff-format` reformats six `.py` files, `end-of-file-fixer` strips trailing blank lines, and `trailing-whitespace` edits `protobuf/meta.proto`.

Two consequences:

1. The vendored copy silently diverges from upstream, which is why `test_vendored_files_match_upstream` can only cover the YAML and JSONLD files.
2. `make update_model` records `VENDORED_MANIFEST` *before* running pre-commit, so the rewrite leaves the manifest describing content that no longer matches disk — and `test_vendored_files_match_manifest` fails on the vendoring PR itself, looking like the files were hand-edited.

The second hasn't fired because nobody has re-vendored since #3897 added the manifest. I hit it doing a dry run of vendoring the upcoming linkml-model rc.

Excluding the directory is better than reordering the Makefile: reordering keeps the manifest consistent but leaves the vendored copy permanently divergent from upstream, and linting vendored code yields findings that can only be fixed upstream anyway. There's precedent in both directions already — `tests/offline_stubs/` is excluded from codespell for exactly this reason, and the vendored model schemas were already excluded from `trailing-whitespace`; this generalises that to the whole vendored tree.

The ruff exclusion goes in `packages/linkml_runtime/pyproject.toml`, not the root — that package carries its own `[tool.ruff]`, so it's the config that actually governs these files.

## Verification

Re-vendored from the linkml-model rc branch and ran `pre-commit run --all-files`: every hook passes and the vendored tree is **byte-identical to upstream** (0 differences), which it has never been. Full suite at baseline.

Aside, not changed here: the root `per-file-ignores` entry for `linkml_model/*.py` looks dead for the same config-resolution reason, since the nested config wins.


kevinschaper (approved):
makes sense to me

PR #4009, opened 2026-09-20
Javagen: Allow template-specific custom type maps

### LINKED ISSUE #4008: Javagen: Allow template-specific type mapping
The Java generator should allow every “template variant” (including custom variants, outside of those bundled with the generator) to define its own mappings between a LinkML type and a Java type.

**Additional context**

The Java generator maintains its own map between LinkML types and Java types, hardcoded in the `javagen.py` module:

```python
TYPEMAP = {
    "xsd:string": "String",
    "xsd:integer": "Integer",
    "xsd:float": "Float",
    "xsd:double": "Double",
    "xsd:boolean": "Boolean",
    "xsd:dateTime": "ZonedDateTime",
    "xsd:date": "LocalDate",
    "xsd:time": "LocalTime",
    "xsd:anyURI": "URI",
    "xsd:decimal": "BigDecimal",
}
```

However, having only such a static map is not satisfying, for at least two reasons:

**A** A given template variant may want to render a particular LinkML type into another Java type than the type from the default map.
**B** A static map leaves no room for dealing with additional custom types.

Currently, the only way to address this is to subclass the `JavaGenerator` class and override the `map_type` method (or, as a last recourse, to play heavily with Jinja conditionals in a custom template).

I am envisioning a solution that reuses the same logic as the one already used for the selection of template variant: Whenever we are generating code using template variant _VAR_, we look up in the templates directory for a file named `_types-VAR.map`. If that file exists, it is assumed to be a simple map between type URIs (first column) and Java types (second column), as in for example:

```
# Render xsd:base64Binary as an array of bytes
http://www.w3.org/2001/XMLSchema#base64Binary byte[]
```

The `map_type` method can then look up in that table _before_ looking up in the hardcoded `TYPEMAP`.

gouttegd:
## Summary

Fixes #4008

This PR allows each “template variant” used with the Java generator to define its own type map (to map a LinkML type into a Java type), to override the hardcoded static map found within the `javagen.py` module.

For a given template variant named _VAR_, the custom type map is expected to be found in the template directory as a file named `_types-VAR.map`, and that file is expected to contain one mapping per line (lines starting with `#` are ignored).

For example, the custom map for the `org.incenp.linkml` template variant is `_types-org.incenp.linkml.map`, and it currently contains the following mappings:

```
# The LinkML-Java runtime expects uriorcurie slots to be rendered as strings
https://w3id.org/linkml/Uriorcurie String
# Both xsd:base64Binary and xsd:hexBinary are rendered as BinaryBlob, which is a wrapper in LinkML-Java for an array of bytes
http://www.w3.org/2001/XMLSchema#base64Binary BinaryBlob
http://www.w3.org/2001/XMLSchema#hexBinary BinaryBlob
```

## How was this tested?

New tests added to the existing test suite for the Java generator. The updated generator was also tested on the LinkML-Java project, where it generates exactly the expected code.

## Areas of uncertainty

None.

## Checklist

## AI Assistance

None.


packages/linkml/src/linkml/generators/javagen/_types-org.incenp.linkml.map
```diff
@@ -0,0 +1,3 @@
+https://w3id.org/linkml/Uriorcurie String
```
noelmcloughlin, line 1: I was curious about the specificity, or appropriateness, of the `org.incenp..` namespace but having read the markdown documentation and **LinkML Core Java Runtime** website (my first time to see it). So now I know its a legitimate controlled namespace for "_one of the three_" template variants, and we are extending the feature.

packages/linkml/src/linkml/generators/javagen.py
```diff
@@ -160,14 +160,14 @@ class TemplateCache:
     """
 
     def __init__(self):
-        self.template_files: dict[str, Path] = {}
+        self.files: dict[str, Path] = {}
```
noelmcloughlin, line 163: I understand the variable name change captures the new type map extension `.map` which is not a template so the name change makes sense, just to normalize things.

noelmcloughlin (approved):
I see no issues, LGTM
Thanks @gouttegd

packages/linkml/src/linkml/generators/javagen/_types-org.incenp.linkml.map
gouttegd: Yes, `org.incenp.linkml` is the namespace for my LinkML-Java runtime. I use this namespace because I happen to own the `incenp.org` domain name. :)

(At some point in the future, the LinkML core devs may ask themselves whether they are willing to “adopt” my LinkML-Java runtime as _the_ LinkML-Java runtime – that is, the one LinkML runtime for Java, the same way as `linkml-runtime` is the one LinkML runtime for Python. If so, then LinkML-Java could be transferred under a `io.linkml` namespace.)

